PR #68, opened 2020-12-29
Enhance tracker

dreamer-coding-555:
I believe these enhanced features would make the upstream wrap file request part of the transaction a little more enjoyable.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-wrap.md
+++ b/.github/ISSUE_TEMPLATE/new-wrap.md
@@ -1,0 +1,8 @@
+---
+name: New wrap file request
+about: Request for a new Meson wrap file
+title: 'new wrap: '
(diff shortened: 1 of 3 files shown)


xclaesse:
Thanks for your contribution, sorry it hasn't been noticed. The wrapdb system changed as you can see in the new README.md, so I think the proposed templates would need to be updated too.

michaelbrockus:
I see, so how would I go about making this enhancement work? Would like to hear a suggestion before trying to change things around.

eli-schwartz:
The "request new wrap" template is likely obsolete since people can just submit a PR directly and don't need to have a repo created first. So that can be deleted.

The other thing here just seems to be a vacation autoresponder or something? What are the advantages of using it vs. just responding in person?

PR #99, opened 2021-06-05
new wrap: httplib

Tachi107:
Not sure if WrapDB v2 is already up, but in the meantime I'll leave this PR here. 

I've also opened a PR upstream (yhirose/cpp-httplib#955) to add basic Meson support, and if that gets merged having this wrap in WrapDB wouldn't be that useful... But in the event that the PR gets rejected this could be merged instead.

And also, this new way of creating wraps seems much better :)


Tachi107:
Turns out I'm not that good at sorting things alphabetically :/
(having to write sorted JSON by hand is the only thing that I don't like about v2)

subprojects/packagefiles/httplib/meson.build
```diff
@@ -0,0 +1,10 @@
+project(
+  'httplib',
+  'cpp',
+  version: '0.8.9',
+  license: 'MIT'
+)
+
+httplib_dep = declare_dependency(include_directories: include_directories('.'))
```
xclaesse, line 8: This is a header-only library, nothing to compile?

releases.json
```diff
@@ -389,6 +389,13 @@
       "1.9-1"
     ]
   },
+  "httplib": {
+    "dependency_names": [],
```
xclaesse, line 393: It provides the `httplib` dependency, you should also add it in the wrap file `[provide]` section. See https://mesonbuild.com/Wrap-dependency-system-manual.html#provide_section.

xclaesse:
>  (having to write sorted JSON by hand is the only thing that I don't like about v2)

I could relax that requirement, there is no technical reason to enforce it, but I thought it would be easier to navigate that file if it's ordered.

xclaesse:
> But in the event that the PR gets rejected this could be merged instead.

Looks like it has been rejected already.



> And also, this new way of creating wraps seems much better :)

Thanks, you're the first one to submit a wrap into the new system, you're our beta tester ;D

subprojects/packagefiles/httplib/meson.build
xclaesse: Relpying to myself: Yes it's header only.

releases.json
```diff
@@ -389,6 +389,13 @@
       "1.9-1"
     ]
   },
+  "httplib": {
```
xclaesse, line 392: Shouldn't it be "cpp-httplib" instead?

Tachi107: I named it this way because the CMake target is called just `httplib`, but in https://github.com/yhirose/cpp-httplib/pull/470#issuecomment-630562373 I saw that the target has been named by a third party contributor... I asked the upstream owner, and I'm waiting for a response.

Tachi107:
Unfortunately for this PR, upstream has accepted to merge meson support as long as I maintain it (https://github.com/yhirose/cpp-httplib/pull/955#issuecomment-855280013).
Should I close this? Or should this be merged anyway (without the patch in `packagefiles`)?

> I could relax that requirement, there is no technical reason to enforce it, but I thought it would be easier to navigate that file if it's ordered.

A sorted fine is better IMO.

jpakkane:
> Or should this be merged anyway

Yes, because it means that people can install it with `meson wrap install httplib`. There are other projects that build with Meson in WrapDB already. Their patch files are empty.

releases.json
Tachi107: Isn't the `httplib` dependency implicitly provided? _A wrap file named `foo.wrap` implicitly provides the dependency name `foo` even when the `[provide]` section is missing._

xclaesse: Ah right, good point.

xclaesse:
@Tachi107 one reason for going with this new wrapdb system is to make easier to add wraps for projects that already have native Meson build system. In that case you can drop `packagefiles/` part and remove `patch_directory` from your wrap file. That way the upstream tarball will be used and no patch will be applied on top.

xclaesse:
Note that we could still merge this (one you fix the name to cpp-httplib) because I guess it's going to take a while before upstream makes a release tarball with the Meson build system merged, right?

Tachi107:
The latest release tag still doesn't support Meson though. For the time being leaving the patch should be fine (right?).

Tachi107:
> The latest release tag still doesn't support Meson though. For the time being leaving the patch should be fine (right?).

Didn't see your comment as I was writing this one, so yes everything is fine.

releases.json
eli-schwartz: The thing is that the provide section is checked for by the older `mesonwrap review` tooling, and

> A wrap file named foo.wrap implicitly provides the dependency name foo even when the [provide] section is missing

only counts for meson.override_dependency which, however, the wrapdb generally avoided using in preference for:

> Since that method was introduced in version 0.54.0, as a transitional aid for projects that do not yet make use of it the variable name can be provided in the wrap file with entries in the format foo-1.0 = foo_dep.


This wrap uses meson.override_dependency... and it doesn't even specify a minimum meson_version. So if people click through to https://mesonbuild.com/Wrapdb-projects.html and download the linked foo.wrap file because their meson version isn't old enough to `meson wrap install` from the v2 wrapdb... then they'll get something that just errors out. And even if they use the current version of meson to install the wrap, if they support older versions of meson then people using that older version support will hit the error.

e.g. see how the upstream PR uses `if meson.version().version_compare('>=0.54.0')`, which once this wrap is updated to pull the upstream tag will require using the [provide] section anyway.

xclaesse: Hm, we never defined the minimum meson we require in wraps. Personally I don't mind just depending on latest release. But it's true that it should be set in `project()`.

eli-schwartz: Maybe you didn't.

Tachi107: @eli-schwartz you're right, I forgot to define the minimum required version..
I also thought that using override_dependency was fine since this wrap is only in WrapDB v2, that isn't completely published yet

xclaesse: > The thing is that the provide section is checked for by the older `mesonwrap review` tooling

I don't think it's changed by the review tool, I copied all checks AFAIK. What should it be checking?

I personally think enough time passed, we should use `meson.override_dependency()` in our wraps. Hardcoding variable names is an horror that should stop ASAP.

eli-schwartz: And yet... people still do use things other than the very latest meson. In fact, people still use things older than just over a year ago. The stable versions of both Debian and Ubuntu don't provide anything that new.

We're already "hardcoding" the names of `dependency_names = foo` even for newer meson, and "hardcoding" variable names in meson.build in order to pass them to meson.override_dependency() -- I don't comprehend the meaning of the sentence "Hardcoding variable names is an horror that should stop ASAP".

I'm a strong believer in being lenient in permitted versions on the grounds we would like people to actually use the wrapdb, and I still don't see a single downside...

eli-schwartz: > I don't think it's changed by the review tool, I copied all checks AFAIK. What should it be checking?

Well, like I said, one of the review tool checks is literally `upstream.wrap has [provide] section:`. It's pedantically required by the v1 tooling, and may or may not have dependency_names vs `foo = foo_dep` but it ***must*** exist.

xclaesse: Oh you're right, I did not notice that `print_status()` in reviewtool.py is fatal by default. Ok, let's make that fatal in v2 too then.

PR #101, opened 2021-06-06
cpp-httplib: Do not require Meson >= 0.54

xclaesse:


xclaesse:
CC @eli-schwartz

eli-schwartz:
The thing is, the upstream merged meson.build adds a version_compare guard before running override_dependency(): https://github.com/yhirose/cpp-httplib/blob/master/meson.build

So once we update this wrap and drop the custom packagefiles, we will need to use [provide] anyway, and there will be no version check. Given that, I'd prefer to just drop/version guard the override_dependency here and provide a `foo = foo_dep` provide...

xclaesse:
@eli-schwartz Ok, updated from that upstream version then.

eli-schwartz (approved):

PR #102, opened 2021-06-06
ci: Add more sanity checks

xclaesse:


xclaesse:
@eli-schwartz Let's make provide section mandatory from now on.

eli-schwartz:
Yeah, this looks good.

Note: the new print() call interacts funnily with the use of unittests, which is doing dot reporting with prints interspersed. There might be room to prettify the stdout reporting, perhaps by making each wrap name construct a test class instance to run the tests. We could then rely on unittest reporting the names.

Just a shower thought.

PR #105, opened 2021-06-09
rdkafka: Version 1.7.0

jcaesar:
Also fixes a lot of whitespacing issues adds building tests and examples. (Running the tests requires a running Kafka daemon, hence it's deliberately not done.)

(Original PR: https://github.com/mesonbuild/rdkafka/pull/5)


subprojects/packagefiles/zstd/meson.build
```diff
@@ -17,7 +17,7 @@ project('zstd',
     'buildtype=release'
   ],
   version: '1.4.5',
-  meson_version: '>=0.47.0')
+  meson_version: '>=0.55.0')
```
eli-schwartz, line 20: This seems quite restrictive?

xclaesse:
I like how the CI picked lz4 and zstd from wrapdb. :D

subprojects/packagefiles/zstd/meson.build
xclaesse: From a quick look I don't see what requires 0.55.0. I'm not against it if there is a reason with no obvious workaround.

xclaesse:
Other than the meson version question, looks all good to me. Thanks!

subprojects/packagefiles/zstd/meson.build
eli-schwartz: There's no reason whatsoever. Maybe it's the first version where meson started erroring out with fatal warnings and telling you "btw -Dwarning_level has existed for ages, please stop using hardcoded -Wextra already". Which is absolutely not a reason to bump the minimum meson version, it's fully backwards compatible...

And in general, wraps need to be conservative and refrain from requiring the latest version of meson which is not available on commonly used distros, unless there is a good reason (e.g. it is hard to do correctly without new features).

jpakkane:
> I like how the CI picked lz4 and zstd from wrapdb. 

Things working as designed. What will they think of next?

subprojects/packagefiles/zstd/meson.build
jcaesar: I'm getting
```
zstd| WARNING: Project targeting '>=0.47.0' but tried to use feature introduced in '0.53.0': Fallback without variable name.
```
and once I lift it to 0.53
```
zstd| WARNING: Project targeting '>=0.53.0' but tried to use feature introduced in '0.55.0': patch_directory.
```

I don't actually understand the first error message, since zstd isn't using fallbacks. So if there is a neat workaround, please do tell. Though... `patch_directory` seems necessary for wrapdb? So it's a lost cause anyway?

eli-schwartz: @xclaesse the fact that CI uses patch_directory and the released wraps do not, seems like a hard dealbreaker. We can't use --fatal-meson-warnings...

xclaesse: > zstd| WARNING: Project targeting '>=0.53.0' but tried to use feature introduced in '0.55.0': patch_directory.

`patch_directory` is only used by wrapdb project itself for the CI, it's not required for wrapped subprojects. I think that should be fixed by https://github.com/mesonbuild/wrapdb/pull/107.

> zstd| WARNING: Project targeting '>=0.47.0' but tried to use feature introduced in '0.53.0': Fallback without variable name.

Not sure why this one happens, which dependency does it try to get that has no variable name?

eli-schwartz: Wait hmm, so warnings about syntax used in the .wrap file are based on the superproject minimum... But if there isn't a minimum it uses the subproject minimum?

Confusing. :D But if it works...

eli-schwartz: I think maybe the "without variable name" is referring to dependency() with automatic wrap promotion? Is that also a case of a subproject minimum being hoisted to the superproject if the superproject doesn't specify one?

xclaesse: > Wait hmm, so warnings about syntax used in the .wrap file are based on the superproject minimum

Hmm, In this case it uses the minimum of the project that first checkout zstd subproject, which is rdkafka and it has no meson_version defined, so it should work... I'm confused too now, #107 should have no effect (but still good to define a minimum version).

I *think* if a wrap requires other wraps, it needs >=0.55.0 (or not define meson_version) to pass the CI.

It's a bit unfortunate, would be sad to drop `--fatal-meson-warnings` in the CI just for this :/

jcaesar: How about I disable the zstd dependency on rdkafka for the CI for now so this can be sorted out separately?

xclaesse: If you change change meson_version back to '>=0.47.0' and rebase, I think it should work. At least it works for me locally.

jcaesar: - [x] green ci
- [x] not touching anything not belonging to rdkafka

I think this is ready for merge.

xclaesse:
Thanks!

PR #110, opened 2021-07-15
Add libcjson 1.7.14.

schmidtw:


subprojects/packagefiles/cjson/meson.build
```diff
@@ -0,0 +1,20 @@
+project('cjson', 'c',
+        version: '1.7.14',
+        meson_version: '>=0.47.0',
+        license: ['MIT'])
+
+headers = ['cJSON.h', 'cJSON_Utils.h']
```
eli-schwartz, line 6: You can use [`files()`](https://mesonbuild.com/Reference-manual.html#files) here to avoid string concatenation later on as '../../'+h

schmidtw: Good suggestion.  Thank you!

subprojects/packagefiles/cjson/include-workaround-meson/cjson/meson.build
```diff
@@ -0,0 +1,9 @@
+# This is a workaround for creating a cjson/ folder with the headers in it in
+# the builddir space.
+#
+# See https://github.com/mesonbuild/meson/issues/2546 for where the technique comes from
+foreach h: headers
+  configure_file(copy: true,
```
tristan957, line 6: Should these configure_file() outputs not also be included as sources with the declare_dependency()

eli-schwartz: What would that accomplish? Headers don't need to be added as source files to executable() or library() either.

tristan957: Thought you needed that if you configured a header file that was meant to be public?

eli-schwartz: Why? You just need the include directory, and to guarantee the file is built first. configure_file() is built at setup time, always.

In the specific (non-default) case of headers generated using custom_target or generator, this can be useful to set up an order-only dependency.

tristan957: Sounds good. Thanks

eli-schwartz:
Thank you!

Issue #111, opened 2021-07-16
libjpeg-turbo fails to compile on 32bit MSVC


benjamin-otte:
Running the GTK compile with the MSVC 2019 32bit toolchain (apparently I was opening the wrong shortcut and got that one) failed with:

```
[1986/3743] Linking target subprojects/libjpeg-turbo-2.1.0/jpeg-8.2.2.dll
FAILED: subprojects/libjpeg-turbo-2.1.0/jpeg-8.2.2.dll subprojects/libjpeg-turbo
-2.1.0/jpeg-8.2.2.pdb
"link" @subprojects/libjpeg-turbo-2.1.0/jpeg-8.2.2.dll.rsp
   Creating library subprojects\libjpeg-turbo-2.1.0\jpeg.lib and object subproje
cts\libjpeg-turbo-2.1.0\jpeg.exp
jcphuff.c.obj : error LNK2019: unresolved external symbol ___builtin_ctzl refere
nced in function _count_zeroes
subprojects\libjpeg-turbo-2.1.0\jpeg-8.2.2.dll : fatal error LNK1120: 1 unresolv
ed externals
[1987/3743] Linking target subprojects...pixbuf/gdk-pixbuf/pixops/timescale.exe
ninja: build stopped: subcommand failed.
```

After trying to debug it a bit, it seems to be that there's a not just the check for `___builtin_ctzl` but also one for `sizeof(long)==sizeof(size_t)` and that 2nd check fails on 64bit but not on 32bit, so `___builtin_ctzl` gets disabled on the CI boxes because of that.

eli-schwartz:
It should be checking that the __builtin_ctzl is functional: https://github.com/mesonbuild/wrapdb/blob/0421165f76846b438227dae634a756a3a8153678/subprojects/packagefiles/libjpeg-turbo/meson.build#L68-L69

Apparently on MSVC this does compile (both in the meson check, and in jcphuff.c.obj), but does not link? Is this somehow available on MSVC, but needs an additional library?

tp-m:
I can confirm this fails to link on MSVC for x86. The only reason it "works" on x64 is that the check is bypassed entirely because of the size checks.

I think the best / correct thing to do here is to just do `cc.has_function('__builtin_ctzl')`? (Which works correctly).

I don't know why it compiles though, I'm not aware of these builtins being available on MSVC, they've got their own builtins for that (`_BitScan*`) as far as I know.

xclaesse:
[...]

PR #112, opened 2021-07-16
libjpeg-turbo: Fix compiler checks

xclaesse:
This is a translation from CMake's check_c_source_compiles() that
compiles and links the code.

Fixes: #111


subprojects/glib.wrap
```diff
@@ -1,9 +1,6 @@
-[wrap-file]
-directory = glib-2.68.1
-
-source_url = https://download.gnome.org/sources/glib/2.68/glib-2.68.1.tar.xz
-source_filename = glib-2.68.1.tar.xz
-source_hash = 241654b96bd36b88aaa12814efc4843b578e55d47440103727959ac346944333
+[wrap-git]
+url = https://gitlab.gnome.org/gnome/glib.git
+revision = main
```
tp-m, line 3: Is this meant to be part of this PR?

xclaesse: Oops, sorry, leftover from some unrelated tests I did. `git add -p` followed by `git commit -a`....

Dropped that now.

eli-schwartz:
This should just use has_function, following the intent of the check, rather than taking an overly literal translation of the cmake code.

xclaesse:
> This should just use has_function, following the intent of the check, rather than taking an overly literal translation of the cmake code.

Ok, changed that, as long as someone can test it for me... :D

jpakkane:
Since no-one volunteered to test, let's merge.

PR #115, opened 2021-07-19
Add netstring-c

nazar-pc:


subprojects/packagefiles/netstring-c/meson.build
```diff
@@ -0,0 +1,32 @@
...
+  'netstring-c',
+  'c',
+  version: '0.0.0+92104d3',
+  license : 'Unlicense',
+  meson_version: '>= 0.55',
+)
+
+# libm (glibc)
+cc = meson.get_compiler('c')
+libm_dep = cc.find_library('m')
```
dcbaker, line 11: Shouldn't this be `required : false`?

nazar-pc: Seems to be required by the library: https://github.com/PeterScott/netstring-c/blob/92104d314ad885fd444b70bffd12a2963ef37a2f/makefile#L2

dcbaker: That actually looks like an upstream bug as well. Most libc implementations don't have a separate libm, it's a quirk of glibc (which they plan to remove in the future i think, but with music to keep things working for people hard coding it). By hard coding it, we prevent this from building with a different libc, even on Linux, since i think only glibc and musl provide a libm

releases.json
```diff
@@ -606,6 +606,14 @@
       "1.2.3-1"
     ]
   },
+  "netstring-c": {
+    "dependency_names": [
+      "netstring-c"
+    ],
+    "versions": [
+      "0.0.0+92104d3"
```
xclaesse, line 614: must end with `-1`

subprojects/netstring-c.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = netstring-c-92104d314ad885fd444b70bffd12a2963ef37a2f
+source_url = https://github.com/PeterScott/netstring-c/archive/92104d314ad885fd444b70bffd12a2963ef37a2f.tar.gz
```
xclaesse, line 3: Hmm, upstream has no release tags? Last code change seems to be 11 years ago, so I guess the commit sha1 is just as fine...

subprojects/packagefiles/netstring-c/meson.build
nazar-pc: My experience with C is very limited and what you're saying makes sense, adjusted accordingly.

subprojects/netstring-c.wrap
nazar-pc: Yeah, I also checked that author doesn't use GitHub lately, so it might be challenging to convince them to create a tag and meaningful version.

subprojects/packagefiles/netstring-c/LICENSE.build
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2020 The Meson development team
```
xclaesse, line 1: My clock says we are in 2021 already :D

subprojects/netstring-c.wrap
jpakkane: Using a serialisation framework to be used in network traffic that has not been updated since 2015 seems like a massive security hole waiting to happen.

nazar-pc: It is a very, very simple serialization format and there is really nothing to be updated in it, works just fine since 2015 :sunglasses: 

Source code is 100 lines, half of which are comments.

jpakkane:
Please don't use merges, they mess up everything. All MRs should be kept up to date by rebasing instead.

nazar-pc:
Sure, rebased and squashed. Also had to add exception for version check since there is no meaningful version in this case unfortunately.

jpakkane:
I suspect that people are going to use this as an excuse to add more Git-based wraps, which we as a policy don't do. However since this project:

- is clearly dead
- has only a small number of users
- has not made any releases ever

Let's add it for now. Note that this does _not_ mean that other such projects will get accepted. In fact they probably won't. In any case that would need to be decided on a case-by-case basis.

nazar-pc:
Makes sense and thank you!

subprojects/packagefiles/netstring-c/meson.build
```diff
@@ -0,0 +1,35 @@
+project(
+  'netstring-c',
+  'c',
+  version: '0.0.0',
+  license : 'Unlicense',
+  meson_version: '>= 0.55',
```
eli-schwartz, line 6: This was wrong, it only needed 0.40 -- you can test this by setting the version to `>=0.00` and checking which versions it complains about.

nazar-pc: It was a somewhat random choice to be honest, didn't now about the trick with `>=0.00`, will use it next time, thanks!

eli-schwartz: No problem -- fixed and pushed.

PR #117, opened 2021-07-24
Add libfuse2

mensinda:
Builds the libraries and the util executables, but not the tests, docs and examples.

This is basically a copy-paste from the current fuse3 `meson.build` with some adjustments for fuse2.


eli-schwartz:
TBH I'm confused by this wrap, because surely people should simply use libfuse3... is this wrap being actively used somewhere?

mensinda:
Yes, it is used imdirectly here https://github.com/mensinda/libRuntime

libfuse2 is needed, since fuse3 is not present on older systems, wich is a problem for AppImmages

On 28 July 2021 07:59:06 CEST, Eli Schwartz ***@***.***> wrote:
>TBH I'm confused by this wrap, because surely people should simply use libfuse3... is this wrap being actively used somewhere?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/mesonbuild/wrapdb/pull/117#issuecomment-888033449
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #118, opened 2021-07-25
Don't define min/max macros on MSVC, it will cause compile errors for abseil-cpp

nazar-pc:
Creating PR against `abseilvs` branch since that contains other fixes required for abseil-cpp to compile anyway


jpakkane:
I added this directly to that branch, so closing. Thanks.

Issue #121, opened 2021-08-01
Make it possible to run CI's scripts in forks and locally


nazar-pc:
The way CI is currently configured makes it basically impossible to check things before submitting PR and then annoying everyone watching the repo with test commits.

The first issue is trilom/file-changes-action action that works in a very strange way in general and I would argue is basically broken: https://github.com/trilom/file-changes-action/issues/116

The second issue related to that is that it is not possible to run `sanity_checks.py` locally due to the fact that it relies in `files.json` in the home folder.

From what I understand it is basically necessary to find out what files have changed as of commit on which the script is running relatively to:
* `master` if not on `master` branch
* relatively to older `master` revision right before current commit otherwise

If there was a script that does this just by using `git diff` or something, that would make it much nicer to work with.

eli-schwartz:
Ah, so *that*s why I got nonsense errors when I pushed a new PR branch to this repo.

xclaesse:
Maybe instead of checking changed files to know which wraps needs to be sanity checked, I could just rely on the new release tags. As long as the PR is not making new releases, who cares if it's not conformant?

PR #122, opened 2021-08-02
sanity-checks: Make it run locally

xclaesse:
trilom/file-changes-action action has issues and it prevents running
sanity checks locally. Instead of relying on the changed files to know
which wraps are modified by a PR, rely on the added release tags.

Fixes: #121.

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -16,8 +16,6 @@ jobs:
           sudo apt-get -y install build-essential python3-pip ninja-build
           pip install meson
 
-      - uses: trilom/file-changes-action@v1.2.3
-
       - name: Sanity Checks
         run: |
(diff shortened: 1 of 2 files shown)


xclaesse:
CC @eli-schwartz @nazar-pc

eli-schwartz:
So if I'm reading this correctly, the idea is to test each subproject wrap, iff its latest declared version is not yet tagged?

My original thought would have been to run git diff between the latest tag and the current HEAD, but this looks pretty good too.

xclaesse:
> So if I'm reading this correctly, the idea is to test each subproject wrap, iff its latest declared version is not yet tagged?

Yes exactly, because that's when a new release is going to be made after merge. That's simpler than trying to play with git diff, because with multi commit PRs it's not always trivial to now exactly the diff that's going to be merged.

PR #123, opened 2021-08-03
adding Unit system v0.5.0

noah1510:
This adds the unit-system library to the wrapdb.
At the moment only version 0.5.0 will be added.
Newer version will follow once they release.

Should I change the revision in the wrap file to the tag of the specific release or can that track the stable branch?


subprojects/unit-system.wrap
```diff
@@ -0,0 +1,6 @@
+[wrap-git]
```
eli-schwartz, line 1: wrap-git is not permitted in the wrapdb, please use the `${repo}/archive/${gitref}.tar.gz` tarball

subprojects/unit-system.wrap
```diff
@@ -0,0 +1,6 @@
...
+url = https://github.com/noah1510/unit-system.git
+revision = v0.5.0
+
+[provides]
+dependency_names = ['unit-system-full','unit-system-generic','unit-system-base','unit-system-common']
```
eli-schwartz, line 6: The wrapdb documentation says this section is called `[provide]`, no "s", and that ini values are a comma-separated (and unquoted) list.

As is, this does not currently work at all... it isn't generally ini syntax, it looks toml inspired? But the previous section isn't quoted, anyway. :p

An easy way to check is to add `dependency('unit-system-base')` to the main meson.build and see if the subproject gets required.

eli-schwartz (changes requested):

eli-schwartz:
Also note that these 3 commits should really be squashed into one commit since they represent one logical change and fixups to the original commit.

eli-schwartz:
Thank you! The changes look good and CI now passes. Can you do a final squash of the commits (general how-to: `git rebase -i origin/master` and change "pick" to "fixup" for each commit besides the first)?

Then I will hit the big green button. :)

noah1510:
This should have squashed the commits.

eli-schwartz:
Perfect!

PR #124, opened 2021-08-03
Add jbigkit

dgcampea:


subprojects/packagefiles/jbigkit/pbmtools/meson.build
```diff
@@ -0,0 +1,19 @@
+jbgtopbm = executable('jbgtopbm',
+    'jbgtopbm.c',
+    include_directories: '../libjbig',
+    link_with: libjbig)
```
eli-schwartz, line 4: If you use `dependencies: libjbig_dep` it will handle both link_with and include_directories for you.

subprojects/packagefiles/jbigkit/meson.build
```diff
@@ -0,0 +1,5 @@
+project('jbigkit', 'c', version: '2.1')
```
eli-schwartz, line 1: No minimum version of meson specified...

subprojects/packagefiles/jbigkit/libjbig/meson.build
```diff
@@ -0,0 +1,14 @@
+sources = files('jbig.c', 'jbig_ar.c')
+sources85 = files('jbig85.c', 'jbig_ar.c')
+
+libjbig = library('jbig', sources)
+libjbig_dep = declare_dependency(
+    include_directories: include_directories('.'),
+    link_with: libjbig)
+meson.override_dependency('libjbig', libjbig_dep)
```
eli-schwartz, line 8: ... yet override_dependency is used, which is a relatively new API.

If, instead of `dependency_names = libjbig`, you used `libjbig = libjbig_dep`, then the meson.build file becomes backwards compatible with lots of older versions of meson, including one or two versions that support this alternative wrap-file syntax but not the dependency_names format.

releases.json
```diff
@@ -416,6 +416,11 @@
       "2.11-1"
     ]
   },
+  "jbigkit": {
+    "version": [
```
eli-schwartz, line 420: Should also include the json dependency_names section...

eli-schwartz (changes requested):
Upstream also includes a testsuite which should get run.

eli-schwartz:
Also note that these 2 commits should really be squashed into one commit since they represent one logical change and fixups to the original commit.

dgcampea:
> Upstream also includes a testsuite which should get run.

Their testsuite is not a typical executable that gets run, it's a series of command sequences in the makefile that convert files back and forth. I've got no good idea how to port those to meson without it turning into a pile of hacks.

dgcampea:
Squashed and added minimum meson version.

eli-schwartz:
Thanks.

However I'd actually prefer to use a backwards-compatible provide section if it means we can lower the minimum version of meson required.

dgcampea:
> However I'd actually prefer to use a backwards-compatible provide section if it means we can lower the minimum version of meson required.

What should the minimal version be for the backward-compat syntax?

eli-schwartz:
A neat trick is to use `meson_version: '>=0.0.0'` and see which versioned features it warns you about (this will surely trigger every FeatureNew warning possible).

xclaesse:
`meson.override_dependency()` has been added in Meson `0.54.0`, more than a year ago. Distros are not relying on wrapdb, and anyone else can easily pip install latest meson. IMHO it makes sense to stop using deprecated fallback mechanism, wrapdb could also serve as example of best meson practice. OTOH it's not a montain I'm willing to die on.

dgcampea:
Personally I'd prefer to use the new syntax and afaik, this dependency is for a somewhat obscure and old image format and I doubt there are many users anxious to integrate it in their projects (well, this PR exists because I ripped it out from one of my personal projects for some ancient machine that underwent a meson port a while ago)

I don't mind changing it to the older syntax though, just sharing my 2¢ on it.

dgcampea:
If this library is intended to be linked against other programs (not a build-time dependency), is it necessary to add `install: true` to `library()`?

eli-schwartz:
If the parent project isn't built using `-Djbigkit:default_library=static` then yes, the subproject needs to install its shared libraries.

As far as syntax goes, I'd prefer the backwards compatible one... its only downside is that if you use explicit `subproject()` instead of dependency fallbacks and copy the project directly but don't include the wrap file, dependencies won't be overridden. The existing syntax wasn't even deprecated.

dgcampea:
Added `install: true` and changed syntax

subprojects/packagefiles/jbigkit/libjbig/meson.build
```diff
@@ -0,0 +1,16 @@
+sources = files('jbig.c', 'jbig_ar.c')
+sources85 = files('jbig85.c', 'jbig_ar.c')
+
+libjbig = library('jbig', sources)
+libjbig_dep = declare_dependency(
+	include_directories: include_directories('.'),
+	link_with: libjbig,
+	install: true)
```
eli-schwartz, line 8:
```
jbigkit| subprojects/jbigkit-2.1/libjbig/meson.build:5: WARNING: Passed invalid keyword argument "install".
jbigkit| WARNING: This will become a hard error in the future.
```

This kwarg would need to be added to library(), not declare_dependency

eli-schwartz (changes requested):
If you're going to mark the libraries as installable, you probably need to also install the headers that let people compile their code against it...

releases.json
```diff
@@ -423,6 +423,21 @@
...
```
eli-schwartz, line 427: CI failure points out that this key should be "versions", not "version"

subprojects/packagefiles/jbigkit/libjbig/meson.build
```diff
@@ -0,0 +1,18 @@
+sources = files('jbig.c', 'jbig_ar.c')
+sources85 = files('jbig85.c', 'jbig_ar.c')
+
+libjbig = library('jbig', sources,
+	install: true)
```
eli-schwartz, line 5: CI rules forbid tabs, please exchange for spaces here and elsewhere.

eli-schwartz (changes requested):
Trivial rebase on master to get the CI running, found hopefully the last problems?

eli-schwartz:
Thanks!

PR #125, opened 2021-08-04
updating unit-system to version 0.5.2

noah1510:
I know the initial pull request to add 0.5.0 was approved just now.
 However in parallel to getting that one merged there have been two
 point releases. These contain some compatibility improvements and
 some fixes for the build system.

 Since 0.5.2 is a hotfix for 0.5.1, only 0.5.2 needs to be available
 in the wrapdb.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -987,6 +987,7 @@
       "unit-system-common"
     ],
     "versions": [
+      "0.5.2-1",
       "0.5.0-1"
     ]
   },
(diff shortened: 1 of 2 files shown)


eli-schwartz:
> Since 0.5.2 is a hotfix for 0.5.1, only 0.5.2 needs to be available in the wrapdb.

We generally avoid adding old versions because new stuff is better than old stuff (but we do not delete existing releases of course). The exception is if someone specifically says they need and will use the old version and cannot use the new version because $REASONS.

PR #127, opened 2021-08-05
add liburing-2.0 wrap

fischerling:
The meson build files for liburing-2.0 are taken from https://github.com/fischerling/liburing/tree/2.0-meson
and are based upon https://github.com/axboe/liburing/pull/297.

I tested the wrap using the method described in the README and meson version 0.59:

```
$meson setup builddir -Dwraps=liburing

The Meson build system
Version: 0.59.0
Source dir: /home/muhq/code/wrapdb
Build dir: /home/muhq/code/wrapdb/builddir
Build type: native build
Project name: wrapdb
Project version: undefined
Host machine cpu: x86_64

Executing subproject liburing

liburing| Project name: liburing
liburing| Project version: 0.7
liburing| C compiler for the host machine: ccache cc (gcc 11.1.0 "cc (GCC) 11.1.0")
liburing| C linker for the host machine: cc ld.bfd 2.36.1
liburing| C++ compiler for the host machine: ccache c++ (gcc 11.1.0 "c++ (GCC) 11.1.0")
liburing| C++ linker for the host machine: c++ ld.bfd 2.36.1
liburing| Run-time dependency threads found: YES
liburing| Compiler for C supports arguments -Wstringop-overflow: YES
liburing| Compiler for C supports arguments -Warray-bounds: YES
liburing| Checking for type "__kernel_rwf_t" : YES
liburing| Checking whether type "struct __kernel_timespec" has members "tv_sec", "tv_nsec" : YES
liburing| Checking if "open_how" compiles: NO
liburing| Checking if "statx" compiles: YES
liburing| Checking if "C++" compiles: YES
liburing| Checking for type "ucontext_t" : YES
liburing| Configuring config-host.h using configuration
liburing| Configuring compat.h using configuration
liburing| Build targets in project: 2
liburing| Subproject liburing finished.

Program meson found: YES (/sbin/meson)
Build targets in project: 3

liburing 0.7

  Directories
    bindir  : bin
    libdir  : lib
    datadir : share

  Configuration
    examples: NO
    tests   : NO

wrapdb undefined

  Subprojects
    liburing: YES

Found ninja-1.10.2 at /sbin/ninja

$ meson compile -C builddir
ninja: Entering directory `builddir'
[6/6] Linking target subprojects/liburing-liburing-2.0/src/liburing.so.2.0.0
```


subprojects/packagefiles/liburing/meson.build
```diff
@@ -0,0 +1,115 @@
+project('liburing', ['c','cpp'],
+        version: run_command('awk', '/Version:/ { print $2 }', 'liburing.spec').stdout().strip(),
```
jpakkane, line 2: This is not acceptable, calling Unix tools is prohibited for things in WrapDB as they are not portable. The "proper" solution would be to use a `VERSION` file, for which Meson has builtin support.

fischerling: Even for a convenience library wrapping the Linux exclusive io_uring syscall?
But I can understand the concern and am willing to use a different solution if needed.

eli-schwartz: A very simple solution would be to statically code it, and update it, together with the version number declared in subprojects/liburing.wrap

fischerling: > A very simple solution would be to statically co

This is what I would do. But this means having two points of truth and the call to awk is what the upstream make-based buildsystem does.

eli-schwartz: Well, apparently the redhat spec file says the version is 0.7 and the git tag says the version is 2.0, so color me very confused.

eli-schwartz:
@xclaesse the CI should NOT pass here, because releases.json never got updated. I think maybe this is because comparing tags to releases.json doesn't actually check when new wraps are added without the latter?

subprojects/packagefiles/liburing/meson.build
fischerling: I will ask upstream what version should be used where and why ;)

fischerling: https://github.com/axboe/liburing/issues/405

eli-schwartz:
ping?

fischerling:
> ping?

Upstream has not answered my question yet.

I would merge it as it is using 0.7 as project version and 2.0 as the library
version.
As far as I know this allows resolving dependencies as expected in meson because the
library version overrides the project version in the declared dependency object.

jpakkane:
Which version number do they end up writing in the .pc file? I could not immediately tell all their configure code is custom and handwritten.

fischerling:
> Which version number do they end up writing in the .pc file? I could not immediately tell all their configure code is custom and handwritten.

Using the liburing-2.0 git tag they will write 0.7 into the pkgconfig file which is wrong and was fixed by [1815337f7dbb](https://github.com/axboe/liburing/commit/1815337f7dbb530ae3e3bc1175d8da214cf76eb5).
After seeing this commit I think it is correct to use 2.0 everywhere and drop the project and library version separation

subprojects/packagefiles/liburing/examples/meson.build
```diff
@@ -0,0 +1,21 @@
+executable('io_uring-cp',
+           'io_uring-cp.c',
+           include_directories : inc,
+           link_with : liburing)
```
eli-schwartz, line 4: dependencies: uring

Since the declare_dependency combines both the include directory and the link_with into one convenient virtual interface.

subprojects/packagefiles/liburing/meson.build
```diff
@@ -0,0 +1,113 @@
+project('liburing', ['c','cpp'],
+        version: 2.0,
```
eli-schwartz, line 2: This can't be a float, needs to be a string or meson immediately errors out. Did this work for you at some point?

fischerling: No. Seams I have pushed a outdated version. Sorry for that!

subprojects/packagefiles/liburing/examples/meson.build
fischerling: Changed. Thanks!

eli-schwartz:
Needs to be rebased on master, sorry. :D

fischerling:
> Needs to be rebased on master, sorry. :D

No problem ;)

fischerling:
I expect the CI to still fail because the C++ code included in the meson.build file to check certain compiler features include tabs.
Which is detected by sanity_checks.py and apparently not allowed in meson code.
Should I replace the tabs with spaces? In the C++ code?
This is probably way easier than modifying sanity_checks.py to allow tabs in strings.

fischerling:
> I expect the CI to still fail because the C++ code included in the meson.build file to check certain compiler features include tabs.
> Which is detected by sanity_checks.py and apparently not allowed in meson code.
> Should I replace the tabs with spaces? In the C++ code?
> This is probably way easier than modifying sanity_checks.py to allow tabs in strings.

I have removeed the tabs from the C code strings.

eli-schwartz:
The alternative might involve adding an AST parser, so yeah, that's a lot easier no matter how meh the error is in this highly specific case.

fischerling:
> The alternative might involve adding an AST parser, so yeah, that's a lot easier no matter how meh the error is in this highly specific case.

Seems like I missed some tabs. Sorry for all those incremental fixes and force pushes

eli-schwartz:
And as I've merged another PR ahead of yours, releases.json is clashing again... sorry to make you rebase yet again. :D

eli-schwartz:
subprojects/packagefiles/liburing/src/include/liburing/compat.h.in seems unfortunate, actually pretty curious why upstream doesn't use a template with a simple sed to replace e.g. `#if @HAS__KERNEL_TIMESPEC@` with `#if 1`

jpakkane:
The "typical" way of doing these is that instead of an `foo.h.in`, you have a `foo.h.meson`. This way it won't clash with the existing configuration template file and you can add any needed changes there. Maybe we should loosen the check and also permit `.h.in` files?

fischerling:
> The "typical" way of doing these is that instead of an `foo.h.in`, you have a `foo.h.meson`. This way it won't clash with the existing configuration template file and you can add any needed changes there. Maybe we should loosen the check and also permit `.h.in` files?

Upstream has no template file they generate the header directly in their custom configure script.
But I have no problem with changing it from `compat.h.in` to `compat.h.meson` if that is the typical naming scheme.
Just let me know ;)

jpakkane:
That is the current scheme, yes. Changing that should make this pass CI.

fischerling:
> That is the current scheme, yes. Changing that should make this pass CI.



> That is the current scheme, yes. Changing that should make this pass CI.

Done :)

subprojects/packagefiles/liburing/meson.build
fischerling: Upstream changed their version in the spec file after their liburing-2.0 git tag with [1815337f7dbb](https://github.com/axboe/liburing/commit/1815337f7dbb530ae3e3bc1175d8da214cf76eb5). And since I used the git tag as base for the meson changes we had this version mismatch.

PR #130, opened 2021-08-12
Add libgpiod

amboar:
The libgpiod library encapsulates the ioctl calls and data structures for accessing Linux's GPIO chardevs behind a straightforward API.

### DIFF
--- a/subprojects/packagefiles/libgpiod/bindings/cxx/examples/meson.build
+++ b/subprojects/packagefiles/libgpiod/bindings/cxx/examples/meson.build
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+executable('gpiodetectcxx', 'gpiodetectcxx.cpp', dependencies: gpiodcxx_dep)
+executable('gpiofindcxx', 'gpiofindcxx.cpp', dependencies: gpiodcxx_dep)
+executable('gpiogetcxx', 'gpiogetcxx.cpp', dependencies: gpiodcxx_dep)
+executable('gpioinfocxx', 'gpioinfocxx.cpp', dependencies: gpiodcxx_dep)
+executable('gpiomoncxx', 'gpiomoncxx.cpp', dependencies: gpiodcxx_dep)
+executable('gpiosetcxx', 'gpiosetcxx.cpp', dependencies: gpiodcxx_dep)
--- a/subprojects/packagefiles/libgpiod/bindings/cxx/meson.build
+++ b/subprojects/packagefiles/libgpiod/bindings/cxx/meson.build
@@ -1,0 +1,35 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+add_languages('cpp')
+
+# See version commentary in lib/meson.build
+#
+# From upstream:
+# > ABI_CXX_VERSION=2.2.1
+abi_cxx_version = '1.1.2'
+
+gpiodcxx_sources = files('chip.cpp', 'iter.cpp', 'line.cpp', 'line_bulk.cpp')
+gpiodcxx_includes = include_directories('.')
+gpiodcxx = library('gpiodcxx', gpiodcxx_sources,
+                   include_directories: [ gpiodcxx_includes ],
+                   gnu_symbol_visibility: 'hidden',
+                   version: abi_cxx_version,
+                   dependencies: gpiod_dep,
+                   install: true)
+pkg.generate(gpiodcxx,
+             name: 'libgpiodcxx',
+             description: 'C++ bindings for libgpiod')
+
+gpiodcxx_dep = declare_dependency(link_with: gpiodcxx,
+                                  include_directories: gpiodcxx_includes,
+                                  dependencies: gpiod_dep,
+                                  version: abi_cxx_version)
+
+install_headers('gpiod.hpp')
+
+subdir('examples')
+
+if get_option('tests')
+    subdir('tests')
+endif
--- a/subprojects/packagefiles/libgpiod/bindings/cxx/tests/meson.build
+++ b/subprojects/packagefiles/libgpiod/bindings/cxx/tests/meson.build
@@ -1,0 +1,20 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+catch2_dep = dependency('catch2')
+
+gpiod_cxx_test = executable('gpiod-cxx-test',
+                            'gpiod-cxx-test-main.cpp',
+                            'gpiod-cxx-test.cpp',
+                            'gpio-mockup.cpp',
+                            'gpio-mockup.hpp',
+                            'tests-chip.cpp',
+                            'tests-event.cpp',
+                            'tests-event.cpp',
+                            'tests-iter.cpp',
+                            'tests-line.cpp',
+                            dependencies: [ gpiodcxx_dep,
+                                            gpiomockup_dep,
+                                            catch2_dep,
+                                            threads_dep ],
+                            install: true)
--- a/subprojects/packagefiles/libgpiod/bindings/python/meson.build
+++ b/subprojects/packagefiles/libgpiod/bindings/python/meson.build
@@ -1,0 +1,14 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+python = import('python')
+python3 = python.find_installation('python3')
+python3_dep = python3.dependency()
+
+gpiodpy = python3.extension_module('gpiod', 'gpiodmodule.c',
+                                   dependencies: [ gpiod_dep, python3_dep ],
+                                   install: true)
+
+if get_option('tests')
+    subdir('tests')
+endif
--- a/subprojects/packagefiles/libgpiod/lib/meson.build
+++ b/subprojects/packagefiles/libgpiod/lib/meson.build
@@ -1,0 +1,41 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+# Upstream's ABI_VERSION is defined in terms of libtool versioning.
+#
+# From upstream configure.ac:
+# > ABI_VERSION=4.2.2
+#
+# From upstream lib/Makefile.am:
+# > libgpiod_la_LDFLAGS = -version-info $(subst .,:,$(ABI_VERSION))
+#
+# So libtool is invoked with:
+# > -version-info 4:2:2
+#
+# Libtool documentation defines the argument for -version-info as[1]:
+# > This flag accepts an argument of the form 'current[:revision[:age]]'
+#
+# In accordance with the libtool script, targetting Linux:
+# > version=$(($current - $age)).${age}.${revision}
+#
+# Solving for upstream ABI_VERSION:
+# version=2.2.2
+#
+# [1] https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
+abi_version = '2.2.2'
+
+gpiod_sources = [ 'core.c', 'ctxless.c', 'helpers.c', 'iter.c', 'misc.c' ]
+gpiod = library('gpiod', gpiod_sources,
+                include_directories: gpiod_includes,
+                gnu_symbol_visibility: 'hidden',
+                c_args: [ '-include', '../config.h' ],
+                version: abi_version,
+                install: true)
+
+pkg.generate(gpiod,
+             name: 'libgpiod',
+             description: 'Library and tools for the Linux GPIO chardev')
+
+gpiod_dep = declare_dependency(link_with: gpiod,
+                               include_directories: gpiod_includes,
+                               version: abi_version)
--- a/subprojects/packagefiles/libgpiod/man/meson.build
+++ b/subprojects/packagefiles/libgpiod/man/meson.build
@@ -1,0 +1,23 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+if get_option('docs')
+    help2man_prog = find_program('help2man')
+    tools = [ 'gpiodetect', 'gpioinfo', 'gpioget', 'gpioset', 'gpiofind', 'gpiomon' ]
+    mandir = join_paths(get_option('prefix'), get_option('mandir'))
+    man1 = join_paths(mandir, 'man1')
+    manpages = []
+    foreach t : tools
+        custom_target('@0@.1'.format(t),
+                      input: get_variable(t),
+                      output: '@BASENAME@.1',
+                      depend_files: 'template',
+                      command: [ help2man_prog,
+                                 '@INPUT@',
+                                 '--include=@CURRENT_SOURCE_DIR@/template',
+                                 '--output=@OUTPUT@',
+                                 '--no-info' ],
+                      install: true,
+                      install_dir: man1)
+    endforeach
+endif
--- a/subprojects/packagefiles/libgpiod/meson.build
+++ b/subprojects/packagefiles/libgpiod/meson.build
@@ -1,0 +1,112 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+project('libgpiod', 'c',
+        default_options: [ 'buildtype=debugoptimized',
+                           'warning_level=2',
+                           'c_std=gnu99',
+                           'cpp_std=gnu++11' ],
+        version: '1.6.3',
+        license: 'LGPL-2.1-or-later',
+        meson_version: '>=0.48.0')
+
+add_project_arguments('-D_GNU_SOURCE', language: 'c')
+
+pkg = import('pkgconfig')
+
+cc = meson.get_compiler('c')
+
+err_not_found_lib = '@0@ not found (needed to build the library)'
+
+lib_header_deps = [ 'getopt.h',
+                    'dirent.h',
+                    'sys/poll.h',
+                    'sys/sysmacros.h',
+                    'linux/gpio.h',
+                    'linux/version.h' ]
+foreach h : lib_header_deps
+    if not cc.has_header(h)
+        error(err_not_found_lib.format(h))
+    endif
+endforeach
+
+lib_function_deps = [ [ 'ioctl', '#include <sys/ioctl.h>' ],
+                      [ 'asprintf', '#include <stdio.h>' ],
+                      [ 'scandir', '#include <dirent.h>' ],
+                      [ 'alphasort', '#include <dirent.h>' ],
+                      [ 'ppoll', '#include <poll.h>' ],
+                      [ 'realpath', '#include <stdlib.h>' ] ]
+foreach fd : lib_function_deps
+    if not cc.has_function(fd[0], prefix: fd[1], args: '-D_GNU_SOURCE')
+        error(err_not_found_lib.format(fd[0]))
+    endif
+endforeach
+
+linux_headers_4_8_0 = '''#include <linux/version.h>
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 8, 0)
+#error
+#endif'''
+if not cc.compiles(linux_headers_4_8_0, name: 'linux headers v4.8.0 check')
+    error('libgpiod needs linux headers version >= 4.8.0')
+endif
+
+linux_headers_5_5_0 = '''#include <linux/version.h>
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 5, 0)
+#error
+#endif'''
+kernel_pre_5_5 = not cc.compiles(linux_headers_5_5_0, name: 'linux headers v5.5.0 check')
+
+gpiod_config = configuration_data()
+gpiod_config.set_quoted('GPIOD_VERSION_STR', meson.project_version())
+gpiod_config.set('KERNEL_PRE_5_5', kernel_pre_5_5)
+configure_file(configuration: gpiod_config, output: 'config.h')
+gpiod_includes = include_directories('include')
+
+subdir('include')
+subdir('lib')
+
+if get_option('tools')
+    err_not_found_tools = '@0@ not found (needed to build the tools)'
+
+    # Check headers
+    if not cc.has_header('sys/signalfd.h')
+        error(err_not_found_tools.format('sys/signalfd.h header'))
+    endif
+
+    tool_function_deps = [ [ 'basename', '#include <libgen.h>' ],
+                           [ 'daemon', '#include <unistd.h>' ],
+                           [ 'signalfd', '#include <sys/signalfd.h>' ],
+                           [ 'setlinebuf', '#include <stdio.h>' ] ]
+    foreach fd : tool_function_deps
+        if not cc.has_function(fd[0], prefix: fd[1], args: '-D_GNU_SOURCE')
+            error(err_not_found_tools.format(fd[0]))
+        endif
+    endforeach
+
+    subdir('tools')
+    subdir('man')
+endif
+
+if get_option('tests')
+    err_not_found_tests = '@0@ not found (need to build the tests)'
+
+    if not cc.has_function('qsort', prefix: '#include <stdlib.h>', args: '-D_GNU_SOURCE')
+        error(err_not_found_tests.format('qsort'))
+    endif
+
+    threads_dep = dependency('threads')
+
+    subdir('tests')
+endif
+
+subdir('bindings')
+
+if get_option('docs')
+    doxygen_prog = find_program('doxygen')
+
+    if doxygen_prog.found()
+        # Discussion:
+        # https://github.com/mesonbuild/wrapdb/pull/130#pullrequestreview-729526961
+        warning('The supplied Doxyfile uses a fixed output path, skipping generation of library documentation and manpages')
+    endif
+endif
--- a/releases.json
+++ b/releases.json
@@ -448,6 +448,15 @@
       "0.6.22-1"
     ]
   },
+  "libgpiod": {
+    "versions": [
+      "1.6.3-1"
+    ],
+    "dependency_names": [
+      "libgpiod",
+      "libgpiodcxx"
+    ]
+  },
   "libjpeg-turbo": {
     "dependency_names": [
       "libjpeg"
--- a/subprojects/libgpiod.wrap
+++ b/subprojects/libgpiod.wrap
@@ -1,0 +1,12 @@
+[wrap-file]
+directory = libgpiod-1.6.3
+
+source_url = https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/snapshot/libgpiod-1.6.3.tar.gz
+source_filename = libgpiod-1.6.3.tar.gz
+source_hash = eb446070be1444fd7d32d32bbca53c2f3bbb0a21193db86198cf6050b7a28441
+
+patch_directory = libgpiod
+
+[provide]
+libgpiod = gpiod_dep
+libgpiodcxx = gpiodcxx_dep
--- a/subprojects/packagefiles/libgpiod/LICENSE.build
+++ b/subprojects/packagefiles/libgpiod/LICENSE.build
@@ -1,0 +1,502 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+[This is the first released version of the Lesser GPL.  It also counts
+ as the successor of the GNU Library Public License, version 2, hence
+ the version number 2.1.]
+
+                            Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+Licenses are intended to guarantee your freedom to share and change
+free software--to make sure the software is free for all its users.
+
+  This license, the Lesser General Public License, applies to some
+specially designated software packages--typically libraries--of the
+Free Software Foundation and other authors who decide to use it.  You
+can use it too, but we suggest you first think carefully about whether
+this license or the ordinary General Public License is the better
+strategy to use in any particular case, based on the explanations below.
+
+  When we speak of free software, we are referring to freedom of use,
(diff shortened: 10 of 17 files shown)


subprojects/packagefiles/libgpiod/bindings/cxx/tests/meson.build
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+catch2_dep = dependency('catch2', required: true)
```
eli-schwartz, line 4: required: true is the default.

subprojects/packagefiles/libgpiod/bindings/python/meson.build
```diff
@@ -0,0 +1,19 @@
...
+
+python3_dep = dependency('python3', required: true)
+python3_prog = find_program('python3')
+python3_config_prog = find_program('python3-config')
+if not python3_config_prog.found()
+    error('python3-config not found - needed for python bindings')
+endif
+
+# TODO: Set up PYTHON_CPPFLAGS and PYTHON_LIBS
+gpiodpy = shared_module('gpiod', 'gpiodmodule.c',
```
eli-schwartz, line 12: I think you probably want https://mesonbuild.com/Python-module.html#extension_module

subprojects/packagefiles/libgpiod/man/meson.build
```diff
@@ -0,0 +1,28 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+if get_option('manpages').enabled()
+    warning('Meson lacks help2man support, unable to generate manpages')
```
eli-schwartz, line 5: What do you mean by lacking support?

It is admittedly true that:
- you should be using custom_target, not generator
- install_man kind of sucks at the moment, you need to manually build the install_dir path without its help, e.g. https://gitlab.archlinux.org/pacman/pacman/-/blob/fc7986485cad9c93df7e59979a7e950fffdc4271/doc/meson.build#L45-47

subprojects/packagefiles/libgpiod/meson.build
```diff
@@ -0,0 +1,123 @@
...
+if with_tests
+    err_not_found_tests = '@0@ not found (need to build the tests)'
+
+    if not cc.has_function('qsort', prefix: '#include <stdlib.h>', args: '-D_GNU_SOURCE')
+        error(err_not_found_tests.format('qsort'))
+    endif
+
+    bats_prog = find_program('bats', required: false)
+    if not bats_prog.found()
+        warning('bats not found - gpio-tools tests cannot be run')
```
eli-schwartz, line 102: Shouldn't the tests be aborted, then?

In fact you can easily do this: https://mesonbuild.com/Reference-manual.html#subdir

```meson
subdir('tests', if_found: bats_prog)
```

subprojects/packagefiles/libgpiod/readme.txt
```diff
@@ -0,0 +1,3 @@
+This repository contains a Meson build definition for project libgpiod.
```
eli-schwartz, line 1: This file is no longer needed and doesn't make much sense inside the subprojects setup, IMO.

eli-schwartz (changes requested):

subprojects/packagefiles/libgpiod/bindings/cxx/tests/meson.build
amboar: Ack

subprojects/packagefiles/libgpiod/bindings/python/meson.build
amboar: Ah neat. Ack.

subprojects/packagefiles/libgpiod/man/meson.build
amboar: Hmm, I tried a few different approaches and eventually found an open issue about supporting `help2man` via `run_target()`. I stopped experimenting at that point.

I can try with `custom_target()`.

subprojects/packagefiles/libgpiod/meson.build
amboar: `if_found` looks promising, let me rework it.

subprojects/packagefiles/libgpiod/readme.txt
amboar: Right, I put it there because I was running `mesonwrap review` against the separate repo I had going. This was before it was pointed out to me that the process had changed.

subprojects/packagefiles/libgpiod/man/meson.build
eli-schwartz: run_target is basically the equivalent of a Makefile .PHONY target which runs arbitrary commands such as a "dist" target that makes a tarball, then scp's it to a server.

custom_target is meson's generic rule for creating a file input->output target with a manually coded compile command.

executable or library or shared_module is meson's super-smart macro rule for detecting the best compile command per platform, and using that.

subprojects/packagefiles/libgpiod/meson.build
amboar: So this came from transliterating the upstream autotools build system; they don't actually run the tests that depend on bats, they just install the test scripts. I'm just going to drop the warning and install the scripts if we have `-Dtests=enabled` and `-Dtools=enabled`.

eli-schwartz: Oh... I'm actually slightly curious why one would wish to install the tests anyways. Are they more useful when installed on end user systems than when getting run at the same time the project is built? What do they test, anyway?

eli-schwartz: Speaking of enabled, a meson feature option enabled/disabled/auto is supposed to be different from a true/false boolean option, specifically because it has the third "auto" state. And also because a feature option has a special-cased meaning when used as a `required: get_option('some-feature')` kwarg. meson will then:
- do a clever log message for the disabled case, saying that lookup was "skipped, because feature blah is disabled",
- for the auto case it is treated as `required: false`
- for the enabled case it is treated as `required: true`

If you don't check all 3 cases, then I'd generally recommend using a boolean instead, just for the sake of clarity.

subprojects/packagefiles/libgpiod/man/meson.build
amboar: Alright, I got manpages for the tools working, but the library API manpages generated with doxygen seem a little more difficult with the upstream `Doxyfile.in`. For now I'm building them but not installing them by default.

subprojects/packagefiles/libgpiod/man/meson.build
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+if get_option('manpages').enabled()
+    help2man_prog = find_program('help2man')
+    tools = [ gpiodetect, gpioinfo, gpioget, gpioset, gpiofind, gpiomon ]
+    mandir = join_paths(get_option('prefix'), get_option('mandir'))
+    mandirn = mandir + '1'
```
eli-schwartz, line 8: This adds a "1" to "/usr/share/man", what you actually want is the install dir "/usr/share/man/man1"

eli-schwartz (changes requested):

subprojects/packagefiles/libgpiod/meson.build
amboar: > Oh... I'm actually slightly curious why one would wish to install the tests anyways. Are they more useful when installed on end user systems than when getting run at the same time the project is built? What do they test, anyway?

The tests exercise the kernel's GPIO chardev interface, which requires inserting and removing kernel modules among other kinds of gunk. Things that are probably best not run as automated tests and run at the user or maintainer's leisure I guess?

amboar: I've switched to using booleans for options as it didn't seem like I needed to exploit the tri-state nature of features.

xclaesse:
I should add in README/Doc that new wraps should specify if meson port has been proposed upstream first.

@amboar this is a non trivial meson port, has it been proposed upstream? If no, please do. If yes, please link to the discussion :)

subprojects/packagefiles/libgpiod/lib/meson.build
```diff
@@ -0,0 +1,19 @@
...
+                install: true)
+
+pkg.generate(gpiod,
+             name: 'libgpiod',
+             description: 'Library and tools for the Linux GPIO chardev')
+
+gpiod_dep = declare_dependency(link_with: gpiod,
+                               include_directories: gpiod_includes,
+                               version: abi_gpiod_version)
+meson.override_dependency('libgpiod', gpiod_dep)
```
eli-schwartz, line 19: This requires meson 0.54, preventing use with older meson versions, e.g. it will not work with the meson provided by debian stable or Ubuntu 20.04 LTS.

Instead, you can use
```ini
[provide]
libgpiod = gpiod_dep
```

Which will do the same thing as dependency_names but without meson.override_dependency

Later calls to meson.override_find_program don't have a backwards-compatible declaration form, but they are also probably not needed. Are any of them actually going to be used in other projects depending on libgpiod via `find_program('gpioinfo')` or whatnot?

By dropping the use of meson.override_*, you can get down to a minimum meson version of 0.48.0, which can be declared in `project(..., meson_version: '>=0.48.0')`

eli-schwartz (changes requested):

amboar:
> @amboar this is a non trivial meson port, has it been proposed upstream? If no, please do. If yes, please link to the discussion :)

I've started the discussion upstream: https://lore.kernel.org/linux-gpio/cc4926af-95bb-4178-a760-d55821dfb626@www.fastmail.com/

subprojects/packagefiles/libgpiod/man/meson.build
eli-schwartz: In my project, we somehow got it to work by instructing doxygen to emit man pages to a "man3" directory, and listing that directory name as the doxygen output.

meson manages to do the right thing if a directory name is the custom_target output, it copies the entire directory tree to the install dir.

amboar: Yeah I saw that, but I'm not imaginative enough to make that work here without patching the upstream source. Upstream sets the output directory to 'doc' as a literal (as opposed to via `@var@`) in `Doxyfile.in`, and then we get `doc/man3`. `custom_target` barfs about not allowing path components if I do `output: [ 'doc/man3' ]`.

subprojects/packagefiles/libgpiod/man/meson.build
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+if get_option('manpages')
+    help2man_prog = find_program('help2man')
```
eli-schwartz, line 5: A classic case where you might use the tri-state nature of features is here:

```meson
help2man_prog = find_program('help2man', required: get_option('manpages'))
if help2man_prog.found()
    ...
endif
```

It will then let you autodetect when manpages can be built, or force them on or off.

eli-schwartz: (This is an observational comment, not a requested change. I don't have a strong opinion here.)

subprojects/packagefiles/libgpiod/man/meson.build
eli-schwartz: You could create a doc/meson.build which tells it to output to ../ and then it creates ../doc/. I think this should work...

subprojects/packagefiles/libgpiod/man/meson.build
amboar: > A classic case where you might use the tri-state nature of features is here:

Oh of course. That's trivial enough. I'll rework it.

eli-schwartz: The same applies to doxygen, although...

If you want to avoid the "auto" case building one and not the other (I don't know if it matters, people using "auto" may well be perfectly content to build as much as they can) then you'd need to hoist their find_programs up and set a `have_manpages = foo.found() and bar.found()` boolean.

In the enabled case, if one of help2man/doxygen is found and not the other, then configuring fails anyway, and in the disabled case, neither one is even looked for. This is just a fun thought about the auto case alone.

subprojects/packagefiles/libgpiod/lib/meson.build
eli-schwartz: Thanks for relaxing the 90.54 requirement. Personally, I'd still like to explicitly list the minimum requirement though:

```meson
project('libgpiod', 'c', 'cpp',
        ...,
        meson_version: '>=0.48.0')
```

It serves as useful documentation of expectations, and raises warnings if someone tries to update it to require functions from newer versions, so that doesn't happen by accident.

subprojects/packagefiles/libgpiod/meson.build
```diff
@@ -0,0 +1,126 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+project('libgpiod', 'c', 'cpp',
```
eli-schwartz, line 4: Hmm, I think you do not actually need a C++ compiler, unless the cxx bindings option is turned on.

So you could remove this from the project options and instead use https://mesonbuild.com/Reference-manual.html#add_languages

eli-schwartz (changes requested):

subprojects/packagefiles/libgpiod/meson.build
amboar: I also forgot to add in the `meson_version` constraint.

amboar: Fixed it to use `add_languages('cpp')` in the cxx binding `meson.build`.

amboar:
@eli-schwartz regarding the docs, I'm going to leave `manpages` as a boolean option and keep the current behaviour of not installing the doxygen output. The `Doxyfile.in` is too antisocial in its current form; I tried various hacks with the `doc` subdirectory but couldn't get it working (output goes into `build/doc` rather than `build/subprojects/libgpiod-1.6.3/doc`, and then there's the "path element" issue with the `output` kwarg for `custom_target()`.

subprojects/packagefiles/libgpiod/bindings/cxx/meson.build
```diff
@@ -0,0 +1,27 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+if add_languages('cpp', native: false)
+    abi_cxx_version = '2.2.1'
```
eli-schwartz, line 5: add_languages() defaults to required: true, there's no need to put this in an if block since the build will abort if the C++ compiler is not found.

Also, if someone specifies -Dbindings=cxx then failure to configure the cxx bindings *should* be fatal. :p

subprojects/packagefiles/libgpiod/man/meson.build
```diff
@@ -0,0 +1,23 @@
...
+    tools = [ gpiodetect, gpioinfo, gpioget, gpioset, gpiofind, gpiomon ]
+    mandir = join_paths(get_option('prefix'), get_option('mandir'))
+    man1 = join_paths(mandir, 'man1')
+    manpages = []
+    foreach t : tools
+        custom_target('@0@.1'.format(t.name()),
```
eli-schwartz, line 11: For the record, you don't need t.name() here, you could go the other direction and use:

```meson
tools = [ 'gpiodetect', 'gpioinfo', 'gpioget', 'gpioset', 'gpiofind', 'gpiomon' ]
foreach t: tools
    toolprog = get_variable(t)
    custom_target(...)
```

eli-schwartz (changes requested):
On the doxygen issue, I see that Doxyfile.in specifies `OUTPUT_DIRECTORY       = doc` so yeah, it will unfortunately build in whichever is the working directory of doxygen, and that does not play nicely with subprojects at all.

Given this, I don't believe you can enable the doxygen target at all, whether it gets installed or not. But if upstream makes that variable configurable, then this is solvable without patching Doxyfile.in

subprojects/packagefiles/libgpiod/bindings/cxx/meson.build
amboar: Yep, fair. I'll fix that.

subprojects/packagefiles/libgpiod/man/meson.build
amboar: Hah, I wasn't aware of `get_variable()`. This works for me.

amboar:
> Given this, I don't believe you can enable the doxygen target at all, whether it gets installed or not.

Good point. I'll skip doxygen.

amboar:
Okay, in the latest push:

* Reduce meson version requirement to `>=0.48.0`
* Only require a C++ compiler for `-Dbindings=cxx`
* Compare and correct shared object versioning against autotools
* Ensure all artefacts installed by upstream configure options are installed by meson configure options, including test artefacts (see discussion about test environment requirements on the kernel above)
* Disable and warn for library documentation build due to incompatible Doxyfile from upstream

eli-schwartz:
```
$ meson setup --wipe builddir . -Dwraps=libgpiod -Dlibgpiod:bindings=cxx,python
[...]
$ ninja -C builddir
[5/27] Compiling C++ object bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o
FAILED: bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o 
ccache c++ -Ibindings/cxx/examples/gpiodetectcxx.p -Ibindings/cxx/examples -I../bindings/cxx/examples -Ibindings/cxx -I../bindings/cxx -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -std=gnu++11 -O2 -g -MD -MQ bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o -MF bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o.d -o bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o -c ../bindings/cxx/examples/gpiodetectcxx.cpp
In file included from ../bindings/cxx/examples/gpiodetectcxx.cpp:10:
../bindings/cxx/gpiod.hpp:13:10: fatal error: gpiod.h: No such file or directory
   13 | #include <gpiod.h>
      |          ^~~~~~~~~
compilation terminated.
```


I think you should be making much heavier use of the declare_dependencies() you have set up.

subprojects/packagefiles/libgpiod/bindings/cxx/meson.build
```diff
@@ -0,0 +1,34 @@
...
+                   gnu_symbol_visibility: 'hidden',
+                   version: abi_cxx_version,
+                   link_with: gpiod,
+                   install: true)
+pkg.generate(gpiodcxx,
+             name: 'libgpiodcxx',
+             description: 'C++ bindings for libgpiod')
+
+gpiodcxx_dep = declare_dependency(link_with: gpiodcxx,
+                                  include_directories: gpiod_includes,
```
eli-schwartz, line 25: Should this also add the include_directories: gpiodcxx_includes ?

subprojects/packagefiles/libgpiod/bindings/cxx/meson.build
```diff
@@ -0,0 +1,34 @@
...
+# See version commentary in lib/meson.build
+#
+# From upstream:
+# > ABI_CXX_VERSION=2.2.1
+abi_cxx_version = '1.1.2'
+
+gpiodcxx_sources = files('chip.cpp', 'iter.cpp', 'line.cpp', 'line_bulk.cpp')
+gpiodcxx_includes = include_directories('.')
+gpiodcxx = library('gpiodcxx', gpiodcxx_sources,
+                   include_directories: [ '.', gpiod_includes ],
```
eli-schwartz, line 15: '.' -> gpiodcxx_includes

subprojects/packagefiles/libgpiod/bindings/cxx/examples/meson.build
```diff
@@ -0,0 +1,26 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+executable('gpiodetectcxx', 'gpiodetectcxx.cpp',
+           include_directories: '..',
+           link_with: gpiodcxx)
```
eli-schwartz, line 6: '.' -> gpiodcxx_includes

... or just dependencies: gpiodcxx_dep which is supposed to expose everything all together for you.

eli-schwartz (changes requested):

amboar:
> ```
> $ meson setup --wipe builddir . -Dwraps=libgpiod -Dlibgpiod:bindings=cxx,python
> [...]
> $ ninja -C builddir
> [5/27] Compiling C++ object bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o
> FAILED: bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o 
> ccache c++ -Ibindings/cxx/examples/gpiodetectcxx.p -Ibindings/cxx/examples -I../bindings/cxx/examples -Ibindings/cxx -I../bindings/cxx -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -std=gnu++11 -O2 -g -MD -MQ bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o -MF bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o.d -o bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o -c ../bindings/cxx/examples/gpiodetectcxx.cpp
> In file included from ../bindings/cxx/examples/gpiodetectcxx.cpp:10:
> ../bindings/cxx/gpiod.hpp:13:10: fatal error: gpiod.h: No such file or directory
>    13 | #include <gpiod.h>
>       |          ^~~~~~~~~
> compilation terminated.
> ```

Hmm, I wonder why I didn't see this in my testing. 

> 
> I think you should be making much heavier use of the declare_dependencies() you have set up.

Yep.

eli-schwartz:
The cxx bindings are disabled by default, so it felt like a useful thing to poke at to see if I can break it.

... look, my IRC nick isn't "elibrokeit" for nothing.

amboar:
> > ```
> > $ meson setup --wipe builddir . -Dwraps=libgpiod -Dlibgpiod:bindings=cxx,python
> > [...]
> > $ ninja -C builddir
> > [5/27] Compiling C++ object bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o
> > FAILED: bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o 
> > ccache c++ -Ibindings/cxx/examples/gpiodetectcxx.p -Ibindings/cxx/examples -I../bindings/cxx/examples -Ibindings/cxx -I../bindings/cxx -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -std=gnu++11 -O2 -g -MD -MQ bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o -MF bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o.d -o bindings/cxx/examples/gpiodetectcxx.p/gpiodetectcxx.cpp.o -c ../bindings/cxx/examples/gpiodetectcxx.cpp
> > In file included from ../bindings/cxx/examples/gpiodetectcxx.cpp:10:
> > ../bindings/cxx/gpiod.hpp:13:10: fatal error: gpiod.h: No such file or directory
> >    13 | #include <gpiod.h>
> >       |          ^~~~~~~~~
> > compilation terminated.
> > ```
> 
> Hmm, I wonder why I didn't see this in my testing.
> 

That would be because I had `libgpiod-dev` installed on my system :facepalm: 

FWIW I was testing with `meson setup -Dtools=true -Dtests=true -Dbindings=cxx,python -Dmanpages=true build` in the `subprojects/libgpiod-1.6.3` subdirectory, so was enabling the C++ bindings.

eli-schwartz:
Yeah, I guess that would do it.

...

On the upstream note, looks like this will be a wrapdb exclusive for now?

amboar:
> On the upstream note, looks like this will be a wrapdb exclusive for now?

Yep, looking that way.

amboar:
Okay, this time around:

* Removed `libgpiod-dev` from my system, retested and fixed the broken includes
* Refactored cumbersome use of `link_with:` / `include_directories:` kwargs by instead using `depends:` with existing results from `declare_dependency()`
* Removed `with_tests` in favour of just inlining `get_option('tests')`
* Renamed the `manpages` option to `docs` for forwards compatibility with generating other formats (e.g. html) whenever that becomes possible with the upstream Doxyfile

subprojects/packagefiles/libgpiod/bindings/cxx/tests/meson.build
```diff
@@ -0,0 +1,21 @@
...
+                            'gpiod-cxx-test-main.cpp',
+                            'gpiod-cxx-test.cpp',
+                            'gpio-mockup.cpp',
+                            'gpio-mockup.hpp',
+                            'tests-chip.cpp',
+                            'tests-event.cpp',
+                            'tests-event.cpp',
+                            'tests-iter.cpp',
+                            'tests-line.cpp',
+                            include_directories: '.',
```
eli-schwartz, line 16: > implicit_include_directories (since 0.42.0): a boolean telling whether Meson adds the current source and build directories to the include path, defaults to true.

It's unneeded in this case.

subprojects/packagefiles/libgpiod/bindings/cxx/tests/meson.build
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright 2021 IBM Corp.
+
+catch2_dep = dependency('catch2')
```
eli-schwartz, line 4: Randomly noticed this:

```
FAILED: subprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test.p/gpiod-cxx-test-main.cpp.o 
ccache c++ -Isubprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test.p -Isubprojects/libgpiod-1.6.3/bindings/cxx/tests -I../subprojects/libgpiod-1.6.3/bindings/cxx/tests -Isubprojects/libgpiod-1.6.3/bindings/cxx -I../subprojects/libgpiod-1.6.3/bindings/cxx -Isubprojects/libgpiod-1.6.3/include -I../subprojects/libgpiod-1.6.3/include -Isubprojects/libgpiod-1.6.3/tests/mockup -I../subprojects/libgpiod-1.6.3/tests/mockup -I../subprojects/Catch2-2.13.3/single_include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -g -pthread -MD -MQ subprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test.p/gpiod-cxx-test-main.cpp.o -MF subprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test.p/gpiod-cxx-test-main.cpp.o.d -o subprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test.p/gpiod-cxx-test-main.cpp.o -c ../subprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test-main.cpp
In file included from /usr/include/signal.h:328,
                 from ../subprojects/Catch2-2.13.3/single_include/catch2/catch.hpp:8034,
                 from ../subprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test-main.cpp:9:
../subprojects/Catch2-2.13.3/single_include/catch2/catch.hpp:10822:58: error: call to non-‘constexpr’ function ‘long int sysconf(int)’
10822 |     static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
      |                                                          ^~~~~~~~~~~
In file included from /usr/include/bits/sigstksz.h:24,
                 from /usr/include/signal.h:328,
                 from ../subprojects/Catch2-2.13.3/single_include/catch2/catch.hpp:8034,
                 from ../subprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test-main.cpp:9:
/usr/include/unistd.h:641:17: note: ‘long int sysconf(int)’ declared here
  641 | extern long int sysconf (int __name) __THROW;
      |                 ^~~~~~~
In file included from ../subprojects/libgpiod-1.6.3/bindings/cxx/tests/gpiod-cxx-test-main.cpp:9:
../subprojects/Catch2-2.13.3/single_include/catch2/catch.hpp:10881:45: error: size of array ‘altStackMem’ is not an integral constant-expression
10881 |     char FatalConditionHandler::altStackMem[sigStackSize] = {};
      |                                             ^~~~~~~~~~~~
```

It works fine if I acquire 2.13.6 from my distro... or if I update catch2.wrap from 2.13.3, weird.

eli-schwartz: Meh, according to google, MINSIGSTKSZ is a glibc change, conveniently my system runs glibc-git.

eli-schwartz: In fact, also if I rebase to grab #132

amboar: I'll add `version: '>=2.13.7'`

subprojects/packagefiles/libgpiod/bindings/cxx/tests/meson.build
amboar: Ack.

subprojects/packagefiles/libgpiod/bindings/cxx/tests/meson.build
eli-schwartz: I wouldn't. libgpiod doesn't depend on any specific version of catch2, only glibc depends on a specific version of catch2.

(Also, it is fixed in .5, not .7)

eli-schwartz: Using directly via the wrapdb, people will automatically get .7, but using distro dependencies, whichever version comes from the distro should be updated enough to be compatible with the distro glibc, which might be less than .7 or even .5

amboar:
Fixed the include path issue in the c++ binding tests.

PR #132, opened 2021-08-13
catch2: update to 2.13.7

nazar-pc:


nazar-pc:
Interestingly, this didn't get published on the website, looks like 2 CI pipelines were running concurrently and that caused some race condition :thinking:

PR #133, opened 2021-08-19
Set missing define SDL_THREAD_PTHREAD_RECURSIVE_MUTEX in SDL meson.build

SkaillZ:
Fixes an issue where SDL gets stuck when initializing audio devices on macOS. I've also enabled the define on Linux since it's also enabled in the original CMake build.


eli-schwartz:
Rather than hardocoding some systems where it is assumed to work, shouldn't this check if a test program can be compiled with the feature? e.g. upstream https://github.com/libsdl-org/SDL/blob/f5794f9eebaac3e70d4fc849641f30bc36913367/configure.ac#L3006-L3012

Also this needs an update to releases.json

SkaillZ:
@eli-schwartz Done.

eli-schwartz:
Thanks!

Excitingly, it fails CI because libxi isn't installed. But it actually works fine.

@xclaesse Perhaps there should be a general facility for a tested wrap to install its own dependencies in sanity_checks, if running under github actions.

xclaesse:
@eli-schwartz we could add that, makes sense. Initially my hope was that all deps would come themself from wraps. That's a nice trick of how I made wrapdb project, wraps can fallback to other wraps.

eli-schwartz:
Looks like they're all `required: false` globally, instead of `required: true` but only checked for on the OS that needs them.

Many dependencies can be acquired via wraps, but not all. I'm not even sure it makes sense to try to build xorg stuff as a subproject wrap, specifically (it is usually provided by the host system even when shipping distro-agnostic precompiled binaries).

xclaesse:
I agree, X stuff makes little sense as wraps. Seems a valid use-case for adding deps on the CI runner.

PR #134, opened 2021-08-21
Add uthash

zweimach:
This new repo for all wrap files is a great improvement.

I'm not sure if wrapdb accepts header-only libraries, but here we go.


jpakkane:
> I'm not sure if wrapdb accepts header-only libraries

Of course we do. Thanks.

PR #144, opened 2021-08-26
Update cJSON to 1.7.15

tristan957:


tristan957:
I wasn't too sure on what the protocol for tests was, but they are there and they pass so yay!

eli-schwartz:
Cool, thanks!

eli-schwartz:
I'm actually immensely puzzled by upstream, since the tests appear to mostly be in tests/ but there is also one test at https://github.com/DaveGamble/cJSON/blob/master/test.c ?

tristan957:
lol. cJSON upstream is something else...

PR #150, opened 2021-08-28
openblas-dll: create wrap

danielcjacobs:
Create new wrap for openblas-dll (windows only)

Version: 0.3.7 (will also add most recent version)

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -336,7 +336,14 @@
     ]
   },
   "gtest": {
+    "dependency_names": [
+      "gtest",
+      "gtest_main",
+      "gmock",
+      "gmock_main"
+    ],
     "versions": [
+      "1.11.0-1",
       "1.10.0-1",
       "1.8.1-1",
       "1.8.0-5",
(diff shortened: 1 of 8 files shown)


danielcjacobs:
I know I didn't need to, but I merged mesondb master into my branch before creating the PR, and now it's acting as if I all the recent changes came from this PR...shouldn't it detect that the only differences between this PR and the current master is the openblass-dll changes?

tp-m:
This looks like a binary wrap for a specific platform + architecture - is this something we want in wrapdb?

jpakkane:
WrapDB is for source code only.

danielcjacobs:
So I can't host this wrap here?

xclaesse:
You can, but you would have to build from source instead of using prebuilt binaries. Upstream is using plain Makefile, you can "translate" them to meson.build. Maybe upstream could be interested to use a real build system instead of their hand crafted mess?

xclaesse:
Sorry, they have cmake build system actually. You could try Meson's cmake module: https://mesonbuild.com/CMake-module.html.

PR #152, opened 2021-08-30
Add wraps for xtl, xtensor, and xtensor-python

danielcjacobs:
Add wraps for xtensor and xtl from the [xtensor-stack](https://github.com/xtensor-stack)

NOTE: Shows gtest commit even though this has already been merged into `wrapdb`.  Somewhat new to the whole forking process, so not sure how I should've changed my process.

### DIFF
--- a/subprojects/packagefiles/xtensor/meson.build
+++ b/subprojects/packagefiles/xtensor/meson.build
@@ -1,0 +1,43 @@
+project('xtensor', 'cpp',
+        version:'0.21.5',
+        license:'BSD-3-Clause',
+        default_options : ['warning_level=3', 'cpp_std=c++14'])
+
+xtl_dep = dependency('xtl',
+    version:['>=0.6.12', '<1.0.0'])
+
+json_dep = dependency('nlohmann_json',
+    version:'>=3.1.1')
+
+xsimd_dep = dependency('xsimd',
+    required: get_option('xsimd'),
+    version: ['>=7.4.6', '<8.0.0'])
+
+if xsimd_dep.found()
+    xtensor_cpp_arg = ['-Duse_xsimd']
+else
+    xtensor_cpp_arg = []
+endif
+
+xtensor_dep = declare_dependency(
+    include_directories:include_directories('include', is_system: true),
+    dependencies: [xtl_dep, json_dep, xsimd_dep],
+    compile_args: xtensor_cpp_arg)
+
+extra_cflags = []
+
+# Workaround from issue: https://github.com/mesonbuild/meson/issues/2550
+if meson.is_subproject()
+    install_dir = get_option('subproj_includedir')
+    extra_cflags += '-I' + get_option('prefix') / install_dir
+else
+    install_dir = get_option('includedir')
+endif
+
+install_subdir('include/xtensor', install_dir: install_dir)
+
+pkgconfig = import('pkgconfig')
+pkgconfig.generate(name: meson.project_name(),
+    version: meson.project_version(),
+    description: 'xtensor is a C++ library meant for numerical analysis with multi-dimensional array expressions.',
+    extra_cflags: extra_cflags)
--- a/subprojects/packagefiles/xtensor/meson_options.txt
+++ b/subprojects/packagefiles/xtensor/meson_options.txt
@@ -1,0 +1,6 @@
+option('xsimd', type: 'feature', value: 'auto', description: 'simd acceleration for xtensor')
+option('subproj_includedir',
+    type: 'string',
+    value: 'include',
+    description: 'Header file directory when built as subproject',
+    yield: true)
--- a/releases.json
+++ b/releases.json
@@ -1183,6 +1183,22 @@
       "0.3.0-1"
     ]
   },
+  "xtensor": {
+    "dependency_names": [
+      "xtensor"
+    ],
+    "versions": [
+      "0.21.5-1"
+    ]
+  },
+  "xtl": {
+    "dependency_names": [
+      "xtl"
+    ],
+    "versions": [
+      "0.6.12-1"
+    ]
+  },
   "xxhash": {
     "dependency_names": [
       "libxxhash"
--- a/subprojects/packagefiles/xtl/meson.build
+++ b/subprojects/packagefiles/xtl/meson.build
@@ -1,0 +1,26 @@
+project('xtl', 'cpp', version:'0.6.12', license:'BSD-3-Clause', default_options : ['warning_level=3', 'cpp_std=c++14'])
+
+json_dep = dependency('nlohmann_json',
+    version:'>=3.1.1')
+
+xtl_dep = declare_dependency(
+    include_directories:include_directories('include', is_system: true),
+    dependencies: [json_dep])
+
+extra_cflags = []
+
+# Workaround from issue: https://github.com/mesonbuild/meson/issues/2550
+if meson.is_subproject()
+    install_dir = get_option('subproj_includedir')
+    extra_cflags += '-I' + get_option('prefix') / install_dir
+else
+    install_dir = get_option('includedir')
+endif
+
+install_subdir('include/xtl', install_dir: install_dir)
+
+pkgconfig = import('pkgconfig')
+pkgconfig.generate(name: meson.project_name(),
+    version: meson.project_version(),
+    description: 'Basic tools (containers, algorithms) used by other quantstack packages.',
+    extra_cflags: extra_cflags)
--- a/subprojects/packagefiles/xtl/meson_options.txt
+++ b/subprojects/packagefiles/xtl/meson_options.txt
@@ -1,0 +1,6 @@
+
+option('subproj_includedir',
+    type: 'string',
+    value: 'include',
+    description: 'Header file directory when built as subproject',
+    yield: true)
--- a/subprojects/xtensor.wrap
+++ b/subprojects/xtensor.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = xtensor-0.21.5
+source_url = https://github.com/xtensor-stack/xtensor/archive/0.21.5.tar.gz
+source_filename = xtensor-0.21.5.tar.gz
+source_hash = 30cb896b6686683ddaefb12c98bf1109fdfe666136dd027aba1a1e9aa825c241
+patch_directory = xtensor
+
+[provide]
+xtensor = xtensor_dep
--- a/subprojects/xtl.wrap
+++ b/subprojects/xtl.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = xtl-0.6.12
+source_url = https://github.com/xtensor-stack/xtl/archive/0.6.12.tar.gz
+source_filename = xtl-0.6.12.tar.gz
+source_hash = 25a16958195f939e6fb7c8feccad89b3e450a18124be6aee119c69a1ee2b308c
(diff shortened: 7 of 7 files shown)


dcbaker:
The way you deal with that is rebasing on the master branch from the mesonbuild/wrapdb.

danielcjacobs:
Do I have permission to do that, or will someone else have to do that after this is merged?

danielcjacobs:
Do I have permission to do that, or does someone else have to do it after this gets merged?

danielcjacobs:
Any advice on how to handle the numpy error?

eli-schwartz:
Any reason for using pip instead of apt? It should be called something like "python3-numpy" in apt-get.

danielcjacobs:
Not really, when testing on fedora `dnf install numpy` worked and seemed to do the same thing as `pip install numpy`.  I tried apt "numpy" and it didn't work, and pip was an easy alternative.   Afaik there's not really a difference, but I can switch it to apt if that's desired

subprojects/packagefiles/xtensor/meson.build
```diff
@@ -0,0 +1,42 @@
...
+endif
+
+xtensor_dep = declare_dependency(
+    include_directories:include_directories('include', is_system: true),
+    dependencies: [xtl_dep, json_dep, xsimd_dep],
+    compile_args: xtensor_cpp_arg)
+
+extra_cflags = []
+
+# Workaround from mesonbuild/meson github issue 2550
```
xclaesse, line 28: It's easier to write the full URL I can click.

subprojects/packagefiles/xtensor/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option('XTENSOR_USE_XSIMD', type: 'boolean', value: false, description: 'simd acceleration for xtensor')
```
xclaesse, line 1: we don't usually use capslock when writing options, unlike cmake.

subprojects/packagefiles/xtensor-python/meson.build
```diff
@@ -0,0 +1,47 @@
...
+
+    extra_cflags += '-I' + npy_incstr
+
+# Workaround from mesonbuild/meson github issue 2550
+if meson.is_subproject()
+    install_dir = get_option('subproj_includedir')
+    extra_cflags += '-I' + get_option('prefix') / install_dir
+else
+    install_dir = get_option('includedir')
+endif
```
xclaesse, line 38: Why is the install_dir different when it's a subproject?

subprojects/packagefiles/xtensor/meson_options.txt
eli-schwartz: Also please remove the project name from the option.

When used as a subproject, the current name of the option is
```
-D xtensor:XTENSOR_USE_XSIMD=false
```

subprojects/packagefiles/xtensor-python/meson.build
danielcjacobs: bc of [issue 2550](https://github.com/mesonbuild/meson/issues/2550)

danielcjacobs: sorry, a better explanation would be that it is nice to have control over where a subproject gets installed. The resolution to issue `2550` was to allow the user to decide whether or not to install a subproject, but not to decide where to install it.

tristan957: @danielcjacobs whether or not to install a subproject in handled by the `--skip-subprojects` option on `meson install`

danielcjacobs: Yes, but that doesn't allow you to control _where_ to install, just _whether_ to install

.github/workflows/sanity_checks.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
       - name: Install packages
         run: |
-          sudo apt-get -y install build-essential python3-pip ninja-build nasm
+          sudo apt-get -y install build-essential python3-pip python3-numpy ninja-build nasm
```
xclaesse, line 16: We can know define package per wrap in ci_config.json

subprojects/packagefiles/xtensor-python/LICENSE.build
```diff
@@ -0,0 +1,19 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
xclaesse, line 19: This file is not needed anymore, it will be added automatically into release tarballs created by CI.

xclaesse:
Please rebase on master, we have Windows CI now.

danielcjacobs:
Sanity checks passed locally but are failing now

eli-schwartz:
It is due to fatal.warnings, plus:
```
xtensor-python| WARNING: Broken python installation detected. Python files installed by Meson might not be found by python interpreter.
```

But this has been fixed in meson master...

danielcjacobs:
I rebased onto master :/

xclaesse:
> But this has been fixed in meson master...

I'm not sure it has been fixed in master, on Windows we still have the issue that python modules needs to be installed outside of Meson's prefix. If your prefix is `C:\` it will install your python modules in `C:\lib\python3\site-package` but they should go to `C:\Python39\lib\site-packages` (or something like that).

However, that fatal warning is still a good thing, it means you must add a project option to set your python install dir. Ideally we should add a meson built-in option for that.

eli-schwartz:
No, I mean that meson 0.60 will include a quirk for debian which lets it detect the debian-specific path which is unique to debian, and which avoids the warning for "your python is weird and we dunno what its paths are".

xclaesse:
> I rebased onto master :/

Meson master, not wrapdb master, there is no way currently to run wrapdb CI with meson master.

xclaesse:
> No, I mean that meson 0.60 will include a quirk for debian which lets it detect the debian-specific path which is unique to debian, and which avoids the warning for "your python is weird and we dunno what its paths are".

Yes, that will probably make Ubuntu CI passing, but the Windows CI will still fail with the same warning.

eli-schwartz:
Yes but at least it will get some testing.

subprojects/packagefiles/xtensor-python/meson.build
xclaesse: Still don't understand why whether it's a subproject makes a difference to *where* to install.

subprojects/packagefiles/xtensor-python/meson.build
```diff
@@ -0,0 +1,47 @@
...
+
+extra_cflags = []
+
+xtensor_python_dep = declare_dependency(
+    include_directories: [include_directories('include', is_system: true),
+        include_directories(npy_incstr, is_system: true)],
+    dependencies:[xtensor_dep, pybind_dep])
+
+    extra_cflags += '-I' + npy_incstr
```
xclaesse, line 30: Indentation is off.

xclaesse:
> However, that fatal warning is still a good thing, it means you must add a project option to set your python install dir. Ideally we should add a meson built-in option for that.

(diverging) Hm, actually there is no way to avoid that warning, it gets printed as soon as you call find_installation() even if you don't install any python module. Hmm, that's not acceptable situation I think.

subprojects/packagefiles/xtensor/meson.build
```diff
@@ -0,0 +1,42 @@
...
+
+json_dep = dependency('nlohmann_json',
+    version:'>=3.1.1')
+
+if get_option('use_xsimd')
+    xsimd_dep = dependency('xsimd',
+        version: ['>=7.4.6', '<8.0.0'])
+    xtensor_cpp_arg = ['-Duse_xsimd']
+else
+    xsimd_dep = dependency('', required: false) # aka "not found"
```
eli-schwartz, line 17: We could really just use a feature option in meson_options.txt and then have dependency('xsimd') natively handle the =disabled case by skipping. Then key off of xsimd_dep.found()

danielcjacobs: should be good

eli-schwartz: That's not what I suggested at all.

danielcjacobs: I don't really know what you were suggesting then. Seemed like what I changed it to is a simpler way of carrying out the same behavior as before

eli-schwartz: > We could really just use a feature option in meson_options.txt

https://mesonbuild.com/Build-options.html#features

eli-schwartz: Note that xsimd_dep would then always be added to the relevant targets, but if not found, or disabled, then it would be a no-op (adds no additional compile or link arguments).

danielcjacobs: ohh..._now_ it should be good if I understand correctly

eli-schwartz: Thank you! Yes, that does look much better now. Very beautiful. ❤️

subprojects/packagefiles/xtensor-python/meson.build
```diff
@@ -0,0 +1,45 @@
...
+if (npy_incstr == '')
+    warning('Unable to find numpy (Looking for it using ' + python.path() +
+        '), disabling python bindings.')
+endif
+
+extra_cflags = ['-I' + npy_incstr]
+
+xtensor_python_dep = declare_dependency(
+    include_directories: [include_directories('include', is_system: true),
+        include_directories(npy_incstr, is_system: true)],
```
xclaesse, line 27: nitpicking: indentation seems off here.

danielcjacobs: I don't think it is...? Just happens to look that way because of the repeated `include_directories`

xclaesse: Ah sorry, it's an item in the list not the next kwarg indeed. I would align both `include_directories()` but I don't think we have any strong coding style, so I guess we can leave it.

jpakkane:
:crossed_swords: conflicts.

danielcjacobs:
Fixed the merge conflict...which of course required another commit. Not sure if and how I need to rebase to eliminate that commit.

eli-schwartz:
In order to solve the python meson fatal warning, I can think of two solutions:
- wait until meson is updated and no longer has the problem on debian
- remove the xtensor-python wrap from this PR and watch the CI succeed now. Then add xtensor-python in a new PR. This would make it easier to at least verify two out of the three wraps work properly. The third wrap could be accepted on faith / after manual tests run locally.

PR #153, opened 2021-08-31
Add typing to Python tools

tristan957:
Could add mypy too. The add_to_db() function is weird.

### DIFF
--- a/tools/import-wraps.py
+++ b/tools/import-wraps.py
@@ -21,6 +21,7 @@ import configparser
 import requests
 import json
 import io
+import typing as T
 
 from utils import Version
 
@@ -30,8 +31,8 @@ def get_wrap_list():
     stdout = subprocess.check_output(['meson', 'wrap', 'list'])
     return stdout.decode().splitlines()
 
-def get_wrap_info(wrap):
-    versions = []
+def get_wrap_info(wrap: str) -> T.List[T.Tuple[str, str]]:
+    versions: T.List[T.Tuple[str, str]] = []
     try:
         stdout = subprocess.check_output(['meson', 'wrap', 'info', wrap])
         for line in stdout.decode().splitlines()[1:]:
@@ -42,7 +43,7 @@ def get_wrap_info(wrap):
        pass
     return versions
 
-def rewrite_wrap(wrap):
+def rewrite_wrap(wrap: str):
     # Add 'patch_directory' key info wrap file
     filename = f'subprojects/{wrap}.wrap'
     config = configparser.ConfigParser(interpolation=None)
@@ -52,7 +53,7 @@ def rewrite_wrap(wrap):
     with open(filename, 'w') as f:
         config.write(f)
 
-def fetch_git(wrap, branch):
+def fetch_git(wrap: str, branch: str):
     # Fetch history from legacy repository and move files to new location
     os.makedirs(f'subprojects/packagefiles/{wrap}', exist_ok=True)
     subprocess.check_call(['git', 'fetch', f'https://github.com/mesonbuild/{wrap}', branch])
@@ -73,10 +74,10 @@ def fetch_git(wrap, branch):
         subprocess.check_call(['git', 'rm', 'LICENSE.build'])
     subprocess.check_call(['git', 'commit', '-m', f'Move {wrap} files'])
 
-def create_release(tag, token):
+def create_release(tag: str, token: str):
     api = f'https://api.github.com/repos/{upload_repo}/releases'
     headers = { 'Authorization': f'token {token}' }
-    json = {
+    json: T.Dict[str, str] = {
         'tag_name': tag,
         'name': tag,
     }
@@ -87,7 +88,7 @@ def create_release(tag, token):
     response.raise_for_status()
     return response.json()['upload_url'].replace(u'{?name,label}','')
 
-def upload(upload_url, content, mimetype, name, token):
+def upload(upload_url: str, content: T.AnyStr, mimetype: str, name: str, token: str):
     headers = {
         'Authorization': f'token {token}',
         'Content-Type': mimetype,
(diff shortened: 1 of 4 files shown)


eli-schwartz:
This PR mixes typing changes with other changes and makes it hard to reason about.

tools/import-wraps.py
```diff
@@ -137,22 +138,20 @@ def get_provide(wrap):
...
-def add_to_db(wrap, versions, releases):
-    releases.setdefault(wrap, {})
-    releases[wrap].setdefault('versions', [])
-    releases[wrap].setdefault('dependency_names', [])
-    releases[wrap].setdefault('program_names', [])
-    versions = [Version(f'{version}-{revision}') for version, revision in versions]
-    versions = sorted(versions, reverse=True)
-    versions = [v._s for v in versions]
+def add_to_db(wrap: str, versions: T.List[T.Tuple[str, int]], releases: T.Dict[str, configparser.SectionProxy]):
+    releases.setdefault(wrap, {}) # type: ignore
```
eli-schwartz, line 142: `releases: T.Dict[str, configparser.SectionProxy]`

But then ***every single usage site*** is breaking that rule and getting ignored. So I don't quite understand the point of it all. What do we gain from this typing anyway?

tristan957: What types fits it better? I don't understand this function as I called out in the description.

tristan957:
The only changes I made besides typing to more easily coerce type checkers into getting it right without a cast. Feel free to take it or leave it. Untyped Python just like undocumented JavaScript is the worst thing in the world to read.

tools/import-wraps.py
```diff
@@ -137,22 +138,20 @@ def get_provide(wrap):
...
+    releases[wrap].setdefault('versions', []) # type: ignore
+    releases[wrap].setdefault('dependency_names', []) # type: ignore
+    releases[wrap].setdefault('program_names', []) # type: ignore
+    version_objs = sorted([Version(f'{version}-{revision}') for version, revision in versions], reverse=True)
```
tristan957, line 146: This change was made because overwriting a parameter is a no-no in the type checking world.

tools/utils.py
```diff
@@ -19,7 +19,7 @@
 # a helper class which implements the same version ordering as RPM
 class Version:
     def __init__(self, s: str) -> None:
-        self._s = s
```
tristan957, line 22: This change was because complaints about accessing a restricted member.

tools/import-wraps.py
```diff
@@ -30,19 +31,19 @@ def get_wrap_list():
...
-    versions = []
+def get_wrap_info(wrap: str):
+    versions: T.List[T.Tuple[str, int]] = []
     try:
         stdout = subprocess.check_output(['meson', 'wrap', 'info', wrap])
         for line in stdout.decode().splitlines()[1:]:
             line = line.strip()
             version, revision = line.split()
-            versions.append((version, revision))
+            versions.append((version, int(revision)))
```
tristan957, line 41: This was made because Meson itself uses revisions as ints, so it made sense to be consistent.

eli-schwartz: Meson doesn't. When it parsed the v1 API, it did, but this was in order to compare it with the wrapdb API itself, which returned revision in a separate field as a standalone int.

In the v2 API porting work, there was a bug that it continued to cast the local version's revision to an int, in order to compare it to the remote version which was no longer automatically an int due to a side effect of the wrapdb field layout.

That got fixed in https://github.com/mesonbuild/meson/commit/38db4602d8f80c3d2aec80ed527e3eed3c41b08c by only using revisions as strings everywhere.

In this script, the revision is only ever used for building f-strings, and making it an int feels a bit inconsistent. both with its intended use and with current git versions of meson.

tools/import-wraps.py
eli-schwartz: It mixes two changes :p one change renames the temporary variable to not be the same as the parameter, and one change moves around code so that it is sorted() in the same line rather than on a second line.

The old version built a list comprehension, assigned it, then overwrote itself by sorting itself (and you can do `versions.sort(reverse=True)`). Just pass a generator expression directly to `sorted()` (you're still using an inline list) and call it "rewriting for readability and efficiency" (mypy does not care if you assign `version_objs = [...]; version_objs = sorted(version_objs, reverse=True)`). In a separate commit, rename the actual variable for mypy purposes.

tools/import-wraps.py
eli-schwartz: I don't know what it is either, but the obvious answer seems to be a `T.Any` rather than `configparser.SectionProxy` and then ignoring it every single time.

Moreso given that nowhere is configparser used for the releases variable.

eli-schwartz: Okay, so the general layout of releases is:

```
releases = {
    'foo': {
        versions: ['2.0, '1.0'],
        'program_names': ['foo-config'],
        'dependency_names': ['libfoo']
    }
}
```

So intuitively, this would be:
```
T.Dict[str, T.Dict[str, T.List[str]]]
```

And indeed, that works after removing all ignores.

tools/utils.py
eli-schwartz: This is pylint W0212: Access to a protected member _s of a client class (protected-access)

mypy doesn't care.

pylint has 45 other complaints about tools/import-wraps.py, and two complaints about W0212 protected-access for utils.py

tristan957: Reverted the change since it is unclear if pylint compliance is wanted.

tools/import-wraps.py
tristan957: Time to go change my Meson PR....

eli-schwartz (commented):
Thanks for the updates! Looks great.

I notice that this is doing `from typing import X, Y, Z` while meson itself uses `import typing as T; T.X`. Any particular reason?

tristan957:
I just decided to do it this way. I can change it to be consistent however.

eli-schwartz:
Okay then, please do make that change. :)

tristan957:
Would you also like me to introduce mypy at all in the pipeline?

eli-schwartz:
I don't care one way or the other, do it if you want. :)

tristan957:
If I add mypy, I will do it in another PR

PR #154, opened 2021-08-31
Use the RawConfigParser to not have to escape URLs

### LINKED ISSUE #151: ConfigParser not parsing `%` in openblas sourceforge url
I'm creating a wrap for openblas-0.3.17, and I've set the source_url to `https://sourceforge.net/projects/openblas/files/v0.3.17/OpenBLAS%%200.3.17%%20version.zip/`, but python requires a `%` escape character to parse this string in `sanity_checks.py`:
```
  File "/home/user/wrapdb/tools/sanity_checks.py", line 127, in test_releases
    self.check_source_url(name, wrap_section, ver)
  File "/home/user/wrapdb/tools/sanity_checks.py", line 147, in check_source_url
    source_url = wrap_section['source_url']
  File "/usr/lib64/python3.9/configparser.py", line 1258, in __getitem__
    return self._parser.get(self._name, key)
  File "/usr/lib64/python3.9/configparser.py", line 799, in get
    return self._interpolation.before_get(self, section, option, value,
  File "/usr/lib64/python3.9/configparser.py", line 395, in before_get
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
  File "/usr/lib64/python3.9/configparser.py", line 442, in _interpolate_some
    raise InterpolationSyntaxError(
configparser.InterpolationSyntaxError: '%' must be followed by '%' or '(', found: '%200.3.17%20version.zip/'
```

But even though the url is checked in python, it is accessed in meson, so then it doesn't get the correct address:
```
meson.build:8:2: ERROR: could not get https://sourceforge.net/projects/openblas/files/v0.3.17/OpenBLAS%%200.3.17%%20version.zip/
```

How can I get around this?  Solution could just be finding a different url that works to download it, but I'm not sure how to locate one.

tristan957:
URL encoded strings have percents in them. Percents are interpolated by
ConfigParser, so use RawConfigParser to avoid the interpolation
completely.

Fixes #151

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -69,7 +69,7 @@ class TestReleases(unittest.TestCase):
                 extra_checks = latest_tag not in self.tags
 
                 # Make sure we can load wrap file
-                config = configparser.ConfigParser()
+                config = configparser.RawConfigParser()
                 config.read(f'subprojects/{name}.wrap')
 
                 # Basic checks


xclaesse:
Nitpicking: The doc recommends using `ConfigParser(interpolation=None)` instead, which is what meson's wrap.py does. But it doesn't change anything in our case I think. LGTM.

tristan957:
Oh. I don't mind submitting a follow up PR.

xclaesse:
> Oh. I don't mind submitting a follow up PR.

Not worth it, but go ahead if you want :D

PR #159, opened 2021-09-09
Create Mathjax wrap

danielcjacobs:


eli-schwartz:
I don't quite understand this... this is a javascript project.

There is no pkg-config dependency to provide, and declare_dependency cannot provide anything useful -- include_directories is going to add something to the C compiler header include path but mathjax does not have any C headers to be detected.

Can you provide an example of how you expect this to be used?

danielcjacobs:
I had essentially nothing in my meson.build, but I was required to add a provide section, so I did.

Daniel

danielcjacobs:
We use it to build documentation. Here's how it's used in a project `meson.build`
mathjax_directory = subproject('mathjax').get_variable('mathjax_directory')
```
if get_option('show_sphinx_warnings')
    run_target('docs', command: ['build_documentation.py', mathjax_directory, meson.project_build_root(), '--show-warnings'])
else
    run_target('docs', command: ['build_documentation.py', mathjax_directory, meson.project_build_root()])
endif
```

This subproject clearly doesn't provide any dependencies, but every wrap is required to have a provide section...

danielcjacobs:
> I had essentially nothing in my meson.build, but I was required to add a provide section, so I did. Daniel
> […](#)
> On Sep 9, 2021, at 5:37 PM, Eli Schwartz ***@***.***> wrote: ﻿ I don't quite understand this... this is a javascript project. There is no pkg-config dependency to provide, and declare_dependency cannot provide anything useful -- include_directories is going to add something to the C compiler header include path but mathjax does not have any C headers to be detected. Can you provide an example of how you expect this to be used? — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or unsubscribe. Triage notifications on the go with GitHub Mobile for iOS or Android.

What would be the resolution to this?

xclaesse:
Please rebase on master, we have Windows CI now.

eli-schwartz:
Well, e.g. what would be the advantage of this over just adding mathjax as a git submodule? It's not a dependency, it may only be used via direct subproject(), and its only purpose is to:
- clone the repository / download the repository as a tarball
- add the repository root as a custom variable resource directory to a custom program

It feels a bit odd and inconsistent.

PR #160, opened 2021-09-13
libtiff: fix build with MSVC

tp-m:
Would fail to link with

LINK : fatal error LNK1181: cannot open input file 'subprojects\tiff-4.1.0\libtiff\tiff4.lib'

because no symbols were exported.


xclaesse:
You need to edit releases.json to add the release version.

tp-m:
Ah, thanks!

I wonder whether that couldn't be checked as part of the sanity checks on the CI?

tp-m:
Fixed, hopefully.

xclaesse:
Thanks.

PR #163, opened 2021-09-19
Add wrap for libccp4c

taw10:
This adds a wrap for the C-language part of the [CCP4](https://www.ccp4.ac.uk/) core library.  The source code also includes a set of Fortran bindings, which are currently ignored but could be added as an optional extra if there were demand.

The license for this library is LGPL3, with liability disclaimer modified for UK law.  I've licensed my meson.build file as BSD to match the other wraps, but I'd also be happy to use LGPL3 instead with a view to eventually upstreaming the Meson file.  Note that LGPL applies only to this core library - the rest of the main CCP4 suite (which is huge), based on this library, has a proprietary licence.

The code is old (20 years, some parts older still) and there are a number of compiler warnings.  Following wrap guidelines, I haven't fixed them.  However, I did take the liberty of suppressing them with compiler options, to stop them from drowning warnings from the main project.

The only download source for the libraries is unfortunately an unreliable FTP server, so I've provided a more reliable fallback.  I've been in touch with the upstream authors to ask about hosting it on their [Github account](https://github.com/ccp4) and also fixing the compiler warnings.

### DIFF
--- a/subprojects/packagefiles/libccp4c/meson.build
+++ b/subprojects/packagefiles/libccp4c/meson.build
@@ -1,0 +1,108 @@
+# Meson file for libccp4c (CCP4 core libraries, C part only)
+project('libccp4c', ['c'],
+        version: '6.5.1',
+        meson_version: '>=0.49.0',
+        license: 'LGPL3')
+
+
+cc = meson.get_compiler('c')
+mdep = cc.find_library('m', required: false)
+
+if cc.get_id() == 'msvc'
+    error('libccp4c cannot be compiled using MSVC')
+endif
+
+# Note that the source code assumes that datadir='share'.
+# If it is not, programs using the library will not be able
+# to find 'environ.def'.
+datadir = get_option('datadir') / 'ccp4'
+
+
+add_project_arguments('-DPACKAGE_ROOT="' + get_option('prefix') + '"',
+                      language: 'c')
+add_project_arguments(cc.get_supported_arguments(['-Wno-uninitialized',
+                                                  '-Wno-unused-function',
+                                                  '-Wno-stringop-truncation',
+                                                  '-Wno-stringop-overflow',
+                                                  '-Wno-format-overflow',
+                                                  '-Wno-misleading-indentation',
+                                                  '-Wno-pointer-compare']),
+                      language: 'c')
+
+
+# C library
+libccp4c = library('ccp4c', ['ccp4/ccp4_array.c',
+                             'ccp4/cmap_accessor.c',
+                             'ccp4/cmap_open.c',
+                             'ccp4/csymlib.c',
+                             'ccp4/pack_c.c',
+                             'ccp4/ccp4_general.c',
+                             'ccp4/cmap_close.c',
+                             'ccp4/cmap_skew.c',
+                             'ccp4/cvecmat.c',
+                             'ccp4/ccp4_parser.c',
+                             'ccp4/cmap_data.c',
+                             'ccp4/cmap_stats.c',
+                             'ccp4/library_err.c',
+                             'ccp4/ccp4_program.c',
+                             'ccp4/cmap_header.c',
+                             'ccp4/cmap_symop.c',
+                             'ccp4/library_file.c',
+                             'ccp4/ccp4_unitcell.c',
+                             'ccp4/cmap_labels.c',
+                             'ccp4/cmtzlib.c',
+                             'ccp4/library_utils.c'],
+                   dependencies: [mdep],
+                   install: true)
+
+# CCP4 headers are included with prefix: <ccp4/header.h>
+incdir = include_directories('.')
+
+install_headers(['ccp4/ccp4_file_err.h',
+                 'ccp4/ccp4_program.h',
+                 'ccp4/ccp4_unitcell.h',
+                 'ccp4/cmap_errno.h',
+                 'ccp4/cmap_stats.h',
+                 'ccp4/csymlib.h',
+                 'ccp4/library_file.h',
+                 'ccp4/ccp4_fortran.h',
+                 'ccp4/ccp4_spg.h',
+                 'ccp4/ccp4_utils.h',
+                 'ccp4/cmap_header.h',
+                 'ccp4/cmaplib.h',
+                 'ccp4/cvecmat.h',
+                 'ccp4/mtzdata.h',
+                 'ccp4/ccp4_array.h',
+                 'ccp4/ccp4_general.h',
+                 'ccp4/ccp4_sysdep.h',
+                 'ccp4/ccp4_vars.h',
+                 'ccp4/cmap_labels.h',
+                 'ccp4/cmaplib_f.h',
+                 'ccp4/overview.h',
+                 'ccp4/ccp4_errno.h',
+                 'ccp4/ccp4_parser.h',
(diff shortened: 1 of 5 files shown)


subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -0,0 +1,104 @@
+# Meson file for libccp4c (CCP4 core libraries, C part only)
+project('libccp4c', ['c'],
+        version: '6.5.1',
+        meson_version: '>=0.49.0',
+        license: 'LGPL3')
+
+
+cc = meson.get_compiler('c')
+mdep = cc.find_library('m', required: true)
```
jpakkane, line 9: This should be `required: false` to make it portable. Most platforms (in fact everything except Linux) have math functions in `libc` and don't have a standalone `m` library at all.

Also, `true` is the default value so this would be superfluous in any case.

subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -0,0 +1,104 @@
...
+
+
+# Note that the source code assumes that datadir='share'.
```
eli-schwartz, line 12: You can assert this I suppose... how does it handle different prefixes?

subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -0,0 +1,104 @@
...
+
+add_project_arguments('-DPACKAGE_ROOT="' + get_option('prefix') + '"',
+                      language: 'c')
+add_project_arguments('-Wno-uninitialized', language: 'c')
+add_project_arguments('-Wno-unused-function', language: 'c')
+add_project_arguments('-Wno-stringop-truncation', language: 'c')
+add_project_arguments('-Wno-stringop-overflow', language: 'c')
+add_project_arguments('-Wno-format-overflow', language: 'c')
+add_project_arguments('-Wno-misleading-indentation', language: 'c')
+add_project_arguments('-Wno-pointer-compare', language: 'c')
```
xclaesse, line 26: Better only add warning flags your compiler actually understands: `add_project_arguments(cc.get_supported_arguments([...]), language: 'c')`

subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -0,0 +1,104 @@
...
+        license: 'LGPL3')
+
+
+cc = meson.get_compiler('c')
+mdep = cc.find_library('m', required: true)
+
+
+# Note that the source code assumes that datadir='share'.
+# If it is not, programs using the library will not be able
+# to find 'environ.def'.
```
xclaesse, line 14: You don't seems to install `environ.def`.

eli-schwartz:
Please rebase on master -- we've added a Windows CI check.

subprojects/packagefiles/libccp4c/meson.build
taw10: Ok, fixed!

subprojects/packagefiles/libccp4c/meson.build
taw10: Nice hint - fixed!

taw10:
> Please rebase on master -- we've added a Windows CI check.

Ok, rebased.  I don't have the ability to test on Windows here, so I'm expecting some trouble from the CI.  Let's see...

subprojects/packagefiles/libccp4c/meson.build
taw10: It literally does the following:
```
      env_file = ccp4_utils_joinfilenames(ccp4_utils_joinfilenames(
                  ccp4_utils_joinfilenames(pkgroot,"share"),"ccp4"),"environ.def");
```
Where `pkgroot` is set like this:
```
#ifdef PACKAGE_ROOT
  const char *pkgroot=PACKAGE_ROOT;
#else
  const char *pkgroot=get_parent_directory(argv[0]);
#endif
```
And `PACKAGE_ROOT` is set by meson.build.  However, this can be bypassed by setting an environment variable (`CINCL`).  This environment variable would be set when using the "full" CCP4 installation, but I wouldn't expect it to be the case when using the libraries via the wrap.

However, the environ.def (also default.def) file isn't part of these libraries, but rather part of the full CCP4 installation which is under a proprietary license.  It's just a list of default values for various program arguments, so it's a bit mean to not include it in the LGPL part, but that's the way it is.  Therefore, the question is kind of theoretical because the installed file will almost certainly never be seen when the wrap is used.  Adding an assertion seems like it would prevent the top-level project from customising the paths, without really adding any security because the project is almost certainly not using these files.  If it were, it would be part of the full CCP4 project and the environment variable would be set.

There is a related problem for some of the data files that are installed by the package, e.g. `syminfo.lib`.  The code will only be able to find them if an environment variable is set to the right location, even though the real location could easily be set by the build system (there is no particular reason to want to override the file).  This is something I _really_ want to fix via an upstream PR, but cannot yet do.

It's awful, I agree.  I don't see any way to fix it without changes upstream.

subprojects/packagefiles/libccp4c/meson.build
taw10: I've hopefully addressed this with my comment on the review by @eli-schwartz - see above.

eli-schwartz:
Sorry -- merging another PR broke the CI so I rebased it myself.

eli-schwartz:
Alright, yeah, this works on Linux but fails on Windows:

```
D:\a\wrapdb\wrapdb\subprojects\libccp4-6.5.1\ccp4\ccp4_sysdep.h(244): fatal error C1189: #error:  "Can't determine machine number format"
```

subprojects/packagefiles/libccp4c/meson.build
eli-schwartz: Ah, okay -- thanks for explaining.

taw10:
Hmmm, this would probably be possible to fix, but only with upstream involvement.  Building CCP4 (the whole project, not just this library) on Windows is reported to work  with MinGW or Cygwin, but not VS (see http://devtools.fg.oisin.rc-harwell.ac.uk/#windows).

Is this a deal-breaker?  If so, I'll close the PR for now, see what I can do upstream, and come back later!

eli-schwartz:
This is not a dealbreaker, but I would recommend adding an explicit

```
if cc.get_id() == 'msvc'
    error('libccp4c cannot be compiled using MSVC')
endif
```

This will make it obvious to people that they shouldn't bother trying to figure out if they have done something wrong to make it not work, and that yes, it is actually a known and expected issue.

jpakkane:
You need to edit `ci_config.json` so that it does not try to build it on Windows.

eli-schwartz:
According to README.md, the correct way to do this is to mark the wrap as `linux_only`.

The undocumented method you used instead is commented in the json file as "mark the wrap for automatic failure on Windows as known-broken".

The nuance here is that `skip_*` is used solely by the mass regression tests, and indicates that the wrap is ***known to fail and needs fixing*** and is always run in CI as normal, but, it is skipped by the mass regression tests because we cannot test for regressions in things that are already broken.

The actual documented functionality, on the other hand, marks it as "not intended to compile on Windows because it wasn't authored upstream to allow for that possibility".

taw10:
Got it, sorry for the noise!

PR #164, opened 2021-09-21
Fixes to vorbis, flac, and sfml

MisterDA:
stb is a header-only dependency. Projects vendor the header in their
sources. The library isn't (most of the time) installed system-wide.
SFML doesn't use the stb system headers, and this is documented:

> The stb_image_* header in the extlibs directory are used regardless
> of this option.

https://www.sfml-dev.org/tutorials/2.5/compile-with-cmake.php (look
for SFML_USE_SYSTEM_DEPS).


MisterDA:
> You need to bump version in releases.json

Done, updated the package revision.

subprojects/packagefiles/sfml/src/SFML/Graphics/meson.build
```diff
@@ -40,10 +40,12 @@ graphics_sources += files(
     'RenderTextureImplDefault.cpp',
 )
 
+stb = include_directories('../../../extlibs/headers/stb_image')
```
xclaesse, line 43: Wait, that look for headers outside of the subproject itself? That's not allowed, I'm curious if the CI will tell.

The correct way for making this work is the project that vendor headers should do:
```meson
incdir = include_directories('extlibs/headers/stb_image')
stb_dep = declare_dependency(include_directories: incdir)
meson.override_dependency('stb', stb_dep)
```

That way when sfml subproject lookup for `stb` dependency it will get stb_dep defined by the main project instead of looking on the system.

MisterDA: > Wait, that look for headers outside of the subproject itself? That's not allowed, 

No, course not. This goes up of three levels, so to the root of SFML sources. That's allowed.

> I'm curious if the CI will tell.

The CI fails on OpenAL... but it's working fine on my system without STB installed. 

> The correct way...

Mine's correct and simply works. I have no strong opinion in favour or disfavour of one or another, so if overriding the dependency is the recommended way, I'm with that.
I do think that removing the STB dependency declaration from the root meson file helps by constraining the declaration to the module that's using it. 

> That way when sfml subproject lookup for stb dependency it will get stb_dep defined by the main project instead of looking on the system.

There is no "main" project here, `stb_dep` is defined in `sfml/meson.build`, in the meson overlay for SFML itself, not in the root project. But that's just nitpicking on my side.

xclaesse: Sorry, my mistake, those headers are indeed inside sfml, I got the number of `..` levels wrong :)

xclaesse:
This fix looks good, but we should find a way to get it pass the CI ideally. Maybe just install some packages on the CI job?

xclaesse:
@MisterDA If you need to install some packages on the CI runner to get it build, with that PR you'll be able to define the list: https://github.com/mesonbuild/wrapdb/pull/166

subprojects/packagefiles/sfml/src/SFML/Graphics/meson.build
eli-schwartz: Well, you can also just declare `stb_inc = include_directories('extlibs/headers/stb_image')` in the top-level meson.build and reuse that variable in `src/SFML/Graphics/meson.build`, which has the advantage of not adding literal `-I../../../` to the compiler command line AND is clearly less confusing for people reading this PR.

eli-schwartz:
Fascinating, that stb is used by a wrap here. I've seen this project before, very unfortunate. e.g. https://github.com/nothings/stb/issues/359

I almost wonder if, instead, we should add a stb.wrap so projects can download stb as a subproject. But, I have *no* clue how versioning would work, and I also have no clue how API-stable this is because apparently every header has its own version and no compatibility description; the assumption seems to be you code your project against whichever version you download and vendor that version like SFML does.

Meanwhile at least one other project I've seen is using wrap-git to download the stb HEAD revision, which I assume means they think it is somewhat stable.

MisterDA:
> I almost wonder if, instead, we should add a stb.wrap so projects can download stb as a subproject. 

Also when you vendor files, its possible to have patches on the vendored files. One reason we should not blindly replace them. One other reason would be when the project asks not to.
but providing a stb wrap for the general case could be useful.

I think the PR's good now :) thanks to all for the help!

eli-schwartz:
The openal CI failure seems weird, why is the wrap which provides it claiming it is only supported on Windows?

MisterDA:
Rebased and declared the `libopenal-dev` package, could you re-run the CI?

eli-schwartz:
It would be convenient if the openal wrap was updated to support all the platforms it is supposed to per upstream, but I guess not a real blocker for this PR...

ci_config.json
```diff
@@ -5,4 +5,9 @@
       "WITH_ZSTD=disabled"
     ]
   }
+  "sfml": {
```
eli-schwartz, line 8: Oops, typo.

xclaesse: I hate that JSON does not allow coma on last item of the list.

eli-schwartz: It is pretty bizarre, yes...

I guess there is always ini. :)

eli-schwartz:
```
E: Could not open lock file /var/lib/dpkg/lock-frontend - open (13: Permission denied)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
```

Uh, apparently the sanity checks script is a bit incomplete...

MisterDA:
The CI's missing sudo or something?

eli-schwartz:
Fixed in master, rebased. It will still fail, but not because of sudo.

MisterDA:
```
FAILED: subprojects/libvorbis-1.3.5/vq/latticebuild.p/latticebuild.c.o 
cc -Isubprojects/libvorbis-1.3.5/vq/latticebuild.p -Isubprojects/libvorbis-1.3.5/vq -I../subprojects/libvorbis-1.3.5/vq -I../subprojects/libvorbis-1.3.5/include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -g -MD -MQ subprojects/libvorbis-1.3.5/vq/latticebuild.p/latticebuild.c.o -MF subprojects/libvorbis-1.3.5/vq/latticebuild.p/latticebuild.c.o.d -o subprojects/libvorbis-1.3.5/vq/latticebuild.p/latticebuild.c.o -c ../subprojects/libvorbis-1.3.5/vq/latticebuild.c
In file included from ../subprojects/libvorbis-1.3.5/vq/bookutil.h:24,
                 from ../subprojects/libvorbis-1.3.5/vq/latticebuild.c:23:
../subprojects/libvorbis-1.3.5/vq/localcodebook.h:21:10: fatal error: ogg/ogg.h: No such file or directory
   21 | #include <ogg/ogg.h>
      |          ^~~~~~~~~~~
compilation terminated.
```
Why isn't `-I../subprojects/libogg-1.3.2/include` in this command line, even when I apply this patch?

```diff
diff --git a/releases.json b/releases.json
index d901dc2..cfd3c86 100644
--- a/releases.json
+++ b/releases.json
@@ -1117,6 +1117,7 @@
       "vorbisenc"
     ],
     "versions": [
+      "1.3.5-8",
       "1.3.5-7",
       "1.3.5-6",
       "1.3.5-5",
diff --git a/subprojects/packagefiles/vorbis/vq/meson.build b/subprojects/packagefiles/vorbis/vq/meson.build
index 67e1812..c325b0f 100644
--- a/subprojects/packagefiles/vorbis/vq/meson.build
+++ b/subprojects/packagefiles/vorbis/vq/meson.build
@@ -2,19 +2,19 @@ latticebuild = executable('latticebuild',
   'latticebuild.c',
   'vqgen.c',
   'bookutil.c',
-  dependencies : vorbis_dep,
+  dependencies : [ogg_dep, vorbis_dep],
 )

 latticetune = executable('latticetune',
   'latticetune.c',
   'vqgen.c',
   'bookutil.c',
-  dependencies : vorbis_dep,
+  dependencies : [ogg_dep, vorbis_dep],
 )

 huffbuild = executable('huffbuild',
   'huffbuild.c',
   'vqgen.c',
   'bookutil.c',
-  dependencies : vorbis_dep,
+  dependencies : [ogg_dep, vorbis_dep],
 )
```

eli-schwartz:
Better question, why is flac or vorbis running when you're trying to test ***imgui***?

MisterDA:
> Better question, why is flac or vorbis running when you're trying to test _**imgui**_?

I'm not trying to test imgui, I'm trying to test sfml and it has an audio module.
There, I think I've fixed everything. The CI is passing on my fork.

eli-schwartz:
Ah, I see. :p 

The 3 commits fixing flac's config.h can probably be rolled into one, though it's not super important.

eli-schwartz:
Can you please rebase anyways, because we added a Windows CI check?

MisterDA:
The Windows CI successfully passes (on my fork), I think this is good to go.

eli-schwartz:
Thanks!

PR #165, opened 2021-09-22
Add VisualStudio2019 CI

xclaesse:


xclaesse:
I have no idea how it works to setup azure CI, I just copied the file from Meson. Anyone knows?

eli-schwartz:
I think you need to have an Azure account and manually link things... you may even need to be an org owner, I don't know.

Anyway I believe github actions supports Windows CI, so using Azure is not required.

xclaesse:
> Anyway I believe github actions supports Windows CI, so using Azure is not required.

ohh, did not know that, it works indeed! Why are we using azure for meson then?

PR updated and it works now!

eli-schwartz:
We are running like 40 test runners, azure gives us 8 more simultaneous slots.

Also because GitHub actions is relatively new and azure has the first-mover advantage due to inertia.

PR #166, opened 2021-09-22
Add script to update packagefiles

xclaesse:
This helps developping patch overlays, meson files can be created
    directly in the right position in source tree, and this script will copy
    them into packagefiles/ directory.

### DIFF
--- a/tools/update-packagefiles.py
+++ b/tools/update-packagefiles.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+
+# Copyright 2021 Xavier Claessens <xclaesse@gmail.com>
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+import configparser
+import tarfile
+import zipfile
+import shutil
+from pathlib import Path
+
+def read_wrap(filename):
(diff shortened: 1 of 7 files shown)


xclaesse:
Added a commit to add `debian_packages` list to configure the CI, that should be useful in the future.

tools/update-packagefiles.py
```diff
@@ -0,0 +1,52 @@
...
+
+if __name__ == '__main__':
+    for f in Path('subprojects').glob('*.wrap'):
+        wrap_section = read_wrap(f)
+        patch_directory = wrap_section.get('patch_directory')
+        if not patch_directory:
+            continue
+        archive_path = Path('subprojects', 'packagecache', wrap_section['source_filename'])
+        if not archive_path.exists():
+            continue
```
eli-schwartz, line 36: Hmm, should this also skip on the fast path if the packagecache archive exists but is not extracted to `directory`?

Sure, directory_files will be empty and so will `directory_files - archive_files` but we could also skip decompressing and reading the tarfile.

tools/update-packagefiles.py
```diff
@@ -0,0 +1,52 @@
...
+        archive = tarfile.open(archive_path)
```
eli-schwartz, line 37: Also tarfile.open does not work for zip files, which is why in meson itself we handle unpacking via shutil.unpack_archive() -- which sadly does not support getting file lists.

I recently noticed this and wondered why we use zip files at all :D since they are rather larger on average, and at least since github archives supports both, we should prefer the nicer format.

eli-schwartz: #167 because why not...

However, this won't help for upstream projects which host their own downloads and which use zip.

tools/update-packagefiles.py
xclaesse: Good point, fixed.

tools/update-packagefiles.py
xclaesse: Added support for zipfile now.

eli-schwartz:
Nice! Especially the ability to add CI dependencies to a wrap!

PR #168, opened 2021-09-23
Make possible to run create_release.py locally for debugging

xclaesse:


eli-schwartz:
Makes sense and seems useful.

PR #169, opened 2021-09-25
add wrap for frozen

neheb:

### DIFF
--- a/subprojects/packagefiles/frozen/LICENSE.build
+++ b/subprojects/packagefiles/frozen/LICENSE.build
@@ -1,0 +1,17 @@
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


eli-schwartz:
Had to double-check the dependency name because upstream does not provide pkg-config, only cmake... dependency('frozen', method: 'cmake') does work though.

Would be nice for upstream to add a simple pkg-config file too. :)

eli-schwartz:
Hmm, the tests/ directory should probably be added as test binaries for the meson port too...

neheb:
Why?

eli-schwartz:
This allows people to test that the header library works correctly for them while building their project. :) wraps do generally have the testsuites set up, sometimes guarded by a "tests" option in meson_options.txt though.

subprojects/packagefiles/frozen/LICENSE.build
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
```
xclaesse, line 4: This should not be the license of the project, but for build file. I think we want to keep the same license for all wraps.

dcbaker: I'm the past @jpakkane said they need to be the same as the project they're for, i think that's in the docs too

xclaesse: Hum, could be worth clarifying this, I thought the intention was to keep them all under a permissive license. What if we have a wrap for a gpl project?

I was actually considering to set a single license file at the root of wraps, applying to the whole repository. Then release script can copy it in all patch tarballs automatically and we can stop duplicating it in packagefiles.

jpakkane: The original reason for that was that we want to make it as easy as possible for projects to grab the wrapdb definitions and add them upstream.

neheb: Replaced.

PR #170, opened 2021-09-28
Delete duplicated LICENSE.build files

xclaesse:
Add COPYING file and copy it in release tarballs by default
    
The license is voluntarily permissive to make it easy to upstream the build system in any project regardless of their respective license.

There are 2 wraps that does not match the same license, I guess you just copied project's license, any objection on dropping the file to have everything in wrapdb under the same terms? (This PR does not relicense them yet)
./subprojects/packagefiles/lzo/LICENSE.build (Apache by @eli-schwartz)
./subprojects/packagefiles/libgpiod/LICENSE.build (LGPL by @amboar)

Note that wrapdb project itself was missing a license, but tools/* scripts have Apache header. It's probably not an issue, but maybe we should have the Apache license (same as Meson) in root COPYING file, and keep a separate LICENSE.build file at the root to copy it into wrap release tarballs? Not sure what's best practice here?

Another issue, I saw some of the LICENSE.build file contains Copyright for individual contributors and this PR drops that information. I'm not a lawyer, but if feels like license and copyright are 2 different things, but I'm not sure how we should preserve that information? I guess I should preserve LICENSE.build files that don't have `The Meson development team` copyright?


tristan957:
Maybe you could add a COPYRIGHT file with year, name, and email to each of the subprojects where you have individual copyright?

amboar:
> There are 2 wraps that does not match the same license, I guess you just copied project's license, any objection on dropping the file to have everything in wrapdb under the same terms? (This PR does not relicense them yet)
./subprojects/packagefiles/lzo/LICENSE.build (Apache by @eli-schwartz)
./subprojects/packagefiles/libgpiod/LICENSE.build (LGPL by @amboar)

@xclaesse For the libgpiod wrap I just copied its license as you guessed. The meson build scripts were pretty much transliterated from the upstream autotools build scripts. As the contributor of the libgpiod meson build scripts I'm happy for them to be relicensed so long as you feel that's reasonable with respect to the upstream project license and how the meson build scripts were derived.

fzwoch:
> To avoid any issue, I left files that have specific copyright assignment: subprojects/packagefiles/liburing/LICENSE.build (@fzwoch)

I think you meant `Florian Fischer` here. At least I don't know what `liburing` actually is.. ;-)

xclaesse:
> I think you meant `Florian Fischer` here. At least I don't know what `liburing` actually is.. ;-)

You're right, sorry for the noise.

xclaesse:
Ok, let's start with this.

PR #173, opened 2021-09-29
Update spng to v0.7.0

randy408:


xclaesse:
Wow, was not expecting GitHub Action Windows image to contain pkg-config...
```
spng| Found pkg-config: C:\Strawberry\perl\bin\pkg-config.BAT (0.26)
spng| Run-time dependency zlib found: YES 1.2.11
```

Wondering why it's not working though. It has the pc file but not headers? Maybe workaround this by adding force_fallback_for=zlib option in ci_config.json?

eli-schwartz:
Please squash this into one commit when doing "fixup" style commits, and in general use `git commit --amend && git push -f myfork pr-branch` when it makes sense.

...

hmm, the ci_config.json change had no effect, because it is getting passed as `-Dspng:force_fallback_for=zlib`.

This needs to be a non-subproject option though. :D

However, `-D:force_fallback_for=zlib` does work as a top-level option (apparently?) and passes the sanity_checks.py test `':' not in option`.

xclaesse:
Should fix sanity_checks script to not add subproject prefix any more, then.

eli-schwartz:
Well no, you do need the prefix if the option is in fact one from meson_options.txt rather than a built-in option.

I'm not sure how the sanity checks shod distinguish that difference. Maybe it's either hack it with a leading empty `:` or add two json keys, one for built-in options and one for project options.

xclaesse:
> Well no, you do need the prefix if the option is in fact one from meson_options.txt rather than a built-in option.

We can add the prefix manually in `ci_config.json`, e.g. `rdkafka:WITH_SASL=disabled`.

eli-schwartz:
Fair enough!

xclaesse:
@randy408 If you rebase now, it should pass CI... hopefully...

randy408:
This branch is messed up and I have no idea why, started a new branch and PR #175

eli-schwartz:
You don't need to start a.new branch!

Just `git push --force` to completely overwrite the PR with the new version.

eli-schwartz:
Also you did not rebase, you did `git merge` or `git pull`?

randy408:
> You don't need to start a new branch!

I didn't want to but this branch has listed changes that are already in `master` so it shouldn't show them(?) I just assumed it's broken somehow and not worth the trouble.

eli-schwartz:
That is almost certainly because you merged instead of rebasing...

PR #176, opened 2021-09-30
add wrap for pugixml

eli-schwartz:
This includes some options from the future -- individual defines which are documented in https://pugixml.org/docs/manual.html#install.building.config but are added to CMakeLists.txt via cmake's impossible options system in git master. So I figured why not add them upfront? :D


subprojects/packagefiles/pugixml/meson.build
```diff
@@ -0,0 +1,61 @@
...
+
+foreach opt: ['wchar_mode', 'compact', 'no_xpath', 'no_stl', 'no_exceptions']
+  if get_option(opt)
+    args += '-DPUGIXML_' + opt.to_upper()
+  endif
+endforeach
+
+pugixml_lib = library('pugixml', sources,
+  cpp_args: args,
+  version: meson.project_version(),
```
xclaesse, line 29: Nitpicking: meson doesn't set that by default?

subprojects/packagefiles/pugixml/meson.build
```diff
@@ -0,0 +1,61 @@
...
+  cmake.configure_package_config_file(
+    input: 'scripts/pugixml-config.cmake.in',
+    name: meson.project_name(),
+    configuration: configuration_data(),
+  )
+endif
+
+subdir('tests')
+pugixml_check = executable('pugixml-check', test_sources, dependencies: pugixml_dep)
+test('pugixml-check', pugixml_check, workdir: meson.current_source_dir())
```
xclaesse, line 61: Any reason for not moving that into test/ subdir?

xclaesse (commented):
Have you tried to submit upstream?

subprojects/packagefiles/pugixml/meson.build
eli-schwartz: It actually wants that subdir, and it is easier than playing around with meson versions and global_source_root and suchlike.

subprojects/packagefiles/pugixml/meson.build
eli-schwartz: No it does not, this is a soname ABI. A pkg-config version is set by default, though.

eli-schwartz:
I have not (yet?) tried to submit this upstream, I suppose I could check if they are interested...

subprojects/packagefiles/pugixml/meson.build
xclaesse: It's more complicated than `workdir: meson.project_source_root()` ?

xclaesse:
> I have not (yet?) tried to submit this upstream, I suppose I could check if they are interested...

I think it's best practice to always engage with upstream, at least so they are informed the port exists and they can point other people asking to wrapdb. But we don't have to wait for their reply to merge this PR of course.

subprojects/packagefiles/pugixml/meson.build
eli-schwartz: That would raise the minimum version of meson! Which I don't like to do in general without good reason, and especially since I actually did offer this upstream now.

And I really have no issue with programs being built in the root build dir, so it never crossed my mind to try to make it work like that anyway.

eli-schwartz:
I actually did end up doing that last night. :)

subprojects/packagefiles/pugixml/meson.build
xclaesse: ok fair enough, was a small nitpicking anyway.

xclaesse:
For the record, upstream PR: https://github.com/zeux/pugixml/pull/443.

neheb:
This can be used in gerbera. Cool.

PR #177, opened 2021-09-30
update tinyxml2 to 9.0.0

dcbaker:
And use the upstream meson definitions instead of rolling our own.

I wrote upstream definition files to use with another project that never really took off, but since they have upstream definitions we should use those instead of our own hand rolled ones.


eli-schwartz:
Nice, upstream build definitions are a definite win!

eli-schwartz:
I think you forgot to update releases.json though :)

eli-schwartz:
It needs to be in descending order.

dcbaker:
And here I expected it to be in ascending order :) hopefully it's all sorted now

PR #178, opened 2021-10-04
Add wrap for LodePNG

tuorqai:
Wrap for [lvandeve/lodepng](https://github.com/lvandeve/lodepng).

LodePNG doesn't use any build system. It's intended to be dropped into source tree. The library consists of only two files: source (`lodepng.cpp`) and header (`lodepng.h`). Also this means that there is no need to build any static/shared library.

It supports both C and C++, but there is a caveat: to use it in C project, you have to rename `lodepng.cpp` to `lodepng.c`.

LodePNG is written in C89 with a thin C++ wrapper on top of it, but technically it's still C++. So attempt to use `lodepng.cpp` in C project leads to linkage errors. To address this problem specifically for this wrap, I might've tried to patch lodepng.h and add `extern "C"` here and there... or just duplicate `lodepng.cpp` and provide two separate dependencies for C and C++. I decided to go with the latter option.

This wrap ignores all testing utilities from upstream.

The upstream repository doesn't have any releases/tags, but the library is kinda versioned.

### DIFF
--- a/subprojects/packagefiles/lodepng/meson.build
+++ b/subprojects/packagefiles/lodepng/meson.build
@@ -1,0 +1,18 @@
+project('lodepng', ['c', 'cpp'],
+    version: '20210627',
+    license: 'zlib')
+
+# C++ version
+lodepng_dep = declare_dependency(
+    include_directories: include_directories('.'),
+    sources: files('lodepng.cpp'))
+
+# C89 version
+copy = find_program('cp', 'copy', required: true)
+run_command(copy, 'lodepng.cpp', 'lodepng.c')
+lodepng_c_dep = declare_dependency(
+    include_directories: include_directories('.'),
+    sources: files('lodepng.c'))
+
+meson.override_dependency('lodepng', lodepng_dep)
+meson.override_dependency('lodepng_c', lodepng_c_dep)
--- a/releases.json
+++ b/releases.json
@@ -627,6 +627,15 @@
       "3.0.1-1"
     ]
   },
+  "lodepng": {
+    "dependency_names": [
+      "lodepng",
+      "lodepng_c"
+    ],
+    "versions": [
+      "20210627-1"
+    ]
+  },
   "lua": {
     "dependency_names": [
       "lua-5.4"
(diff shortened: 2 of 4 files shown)


eli-schwartz:
I'm not super excited to add a project that has no versions and doesn't export any form of dependency whatsoever but wants to provide a `files()` source. I'm especially not thrilled that to provide these unique dependency names you need to encode the programming language you're using into the call to `dependency(...)`

subprojects/packagefiles/lodepng/meson.build
```diff
@@ -0,0 +1,18 @@
...
+    license: 'zlib')
+
+# C++ version
+lodepng_dep = declare_dependency(
+    include_directories: include_directories('.'),
+    sources: files('lodepng.cpp'))
+
+# C89 version
+copy = find_program('cp', 'copy', required: true)
+run_command(copy, 'lodepng.cpp', 'lodepng.c')
```
eli-schwartz, line 12: This is writing to the source tree...

xclaesse: Use configure_file(..., copy: true) to copy and/or rename a file into the build directory instead.

xclaesse:
Why not building at least a static_library() and exposing that as a dependency?

tuorqai:
OK, I will try do it another way. Probably I have to patch `lodepng.h` (to avoid duplication at all) and make a static library.

subprojects/packagefiles/lodepng/meson.build
tuorqai: > Use configure_file(..., copy: true) to copy and/or rename a file into the build directory instead.

Thank you.

tuorqai:
> Why not building at least a static_library() and exposing that as a dependency?

I wasn't sure that this is necessary.

jpakkane:
I'd really like us to have a wider policy decision on what to do with projects that do not do any releases. I'm strictly of the opinion that they should not be put in WrapDB at all

eli-schwartz:
That is my opinion too... It already became a big discussion in the one submission for a dead project that would presumably never need to be updated, and we ended up allowing it. This one, however, is not in that situation as it does, in fact, update.

PR #179, opened 2021-10-05
Add hse to wrapdb

tristan957:
HSE aims to be parallel-installable with itself, but a project should only use one version. HSE provides the following:

* hse2 - CLI
* hse_dep - Dependency object for correctly linking/compiling with HSE

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -389,6 +389,17 @@
       "2.4.1-1"
     ]
   },
+  "hse": {
+    "dependency_names": [
+      "hse-2"
+    ],
+    "program_names": [
+      "hse2"
+    ],
+    "versions": [
+      "2.0.0-1"
+    ]
+  },
   "htslib": {
     "dependency_names": [
       "htslib"
--- a/subprojects/hse.wrap
+++ b/subprojects/hse.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = hse-r2.0.0
+source_url = https://github.com/hse-project/hse/archive/refs/tags/r2.0.0.tar.gz
+source_filename = r2.0.0.tar.gz
+source_hash = 6f9d5a2c08441bf35eff736f9ec1e9a251e841684d834e6f6207d5563e0696fc
+
+[provide]
+program_names = hse2
+dependency_names = hse-2


tristan957:
Not sure what to do about: https://github.com/mesonbuild/wrapdb/pull/179/checks?check_run_id=3805066460#step:5:25

eli-schwartz:
This is the same problem as in the xtensor PR. Meson in git master no longer reports that (fatal) warning on Debian. Windows would still be broken,  but there is discussion on fixing it so that may be fixable by adding an option `-Dpython.install_platlib` or similar. Which would then be handled by the sanity check script directly.

tristan957:
Good to know that I haven't done anything wrong!

xclaesse:
Maybe in the meantime we can add in ci_config.json a new field to skip --meson-fatal-warnings?

eli-schwartz:
That being said, I realize now I should obviously have reviewed the upstream meson.build, mea culpa.

I'm declaring a review of the upstream files to be a merge blocker, BTW.

tristan957:
https://github.com/hse-project/hse/blob/master/subprojects/packagefiles/lz4/lib/meson.build#L6

It is a compile-time define used by lz4 itself. We support compiling against system installed lz4, but performance will suffer. Our lz4 symbols are hidden so there shouldn't be any conflict there if I understand symbol visibility correctly. For instance, we already have this problem with lz4 and xxhash where lz4 includes a copy of xxhash. No symbol collision to speak of.

xclaesse:
@tristan957 `'-DLZ4LIB_VISIBILITY=__attribute__((visibility("hidden")))',` that should be dealt with gnu_symbol_visibility, and that seems to be an important fix that must go in wrapdb's lz4.

tristan957:
After doing both, I get 0 PLT calls whereas previously I had one. Not bad. Will update HSE accordingly.

eli-schwartz:
Also shouldn't there be an upstream option to inline the code rather than having downstreams fork the build?

tristan957:
FWIW lz4 seems to fail to link if you add the visibility define to a shared library. I added some protection for my downstream wrap in that case.

eli-schwartz:
@tristan957 right, because shared libraries exist to expose visible symbols but if you go ahead and hide them because you want them to be inline...

tristan957:
It was more a comment for @xclaesse upstream PR. I don't think he is protecting against it. Should've made the comment there.

tristan957:
PR will be put on hold until our 2.1.0 release later this year.

tristan957:
Once this PR is merged and released on wrapdb (https://github.com/lz4/lz4/pull/1064). I am happy to drop my custom lz4 in favor of the wrapdb build. Then that should unblock this PR maybe.

tristan957:
Update for myself. lz4 PR was merged, HdrHistogram_c PRs were merged (yay) and released. At this point HSE is waiting only on an lz4 release, which is taking a while...

Some of the upstream HSE build has changed since this PR was originally created. We are currently working on 3.0, so I will update the PR when that release comes out assuming that lz4 gets a new release.

neheb:
This should be rebased against master. lz4 problem can be avoided for Ubuntu and macOS CI by supplying debian_ and brew_packages.

tristan957:
@neheb HSE needs access to a v1.9.4 of lz4 which doesn't exist since the Meson build that I updated for it hasn't be included in any release.

PR #180, opened 2021-10-05
Add wrap for fdk_aac

thiblahute:


subprojects/packagefiles/fdk-aac/meson.build
```diff
@@ -0,0 +1,382 @@
...
+fdk_aac_dep = declare_dependency(
+  link_with : fdk_aac,
+  dependencies : [m_dep],
+  include_directories : fdk_aac_inc
+)
+
+pkgc = import('pkgconfig')
+pkgc.generate(name: 'fdk-aac',
+    version: meson.project_version(),
+    description: 'AAC codec library'
```
xclaesse, line 373: You forgot to pass the library argument.

xclaesse: nitpicking: meson.project_version() is the default already.

subprojects/packagefiles/fdk-aac/meson.build
```diff
@@ -0,0 +1,382 @@
...
+  'libPCMutils/src/version.h'
+)
+
+cc = meson.get_compiler('c')
+m_dep = cc.find_library('m', required : false)
+
+fdk_aac = shared_library(
+  'fdk_aac',
+  fdk_aac_src,
+  build_by_default : true,
```
xclaesse, line 357: `true` is the default already.

subprojects/packagefiles/fdk-aac/meson.build
```diff
@@ -0,0 +1,382 @@
...
+    description: 'AAC codec library'
+)
+
+if not get_option('build-programs').disabled()
+  executable(
+    'aac-enc',
+    files('aac-enc.c', 'wavreader.c'),
+    dependencies: fdk_aac_dep
+  )
+endif
```
xclaesse, line 382: nitpicking: newline at the end of file :D

releases.json
```diff
@@ -198,6 +198,14 @@
       "0.7.5-1"
     ]
   },
+  "fdk-aac": {
+    "dependency_names": [
+      "fdk-aac"
+    ],
+    "versions": [
+      "2.0.2",
```
xclaesse, line 206: json is really annoying at not allowing the coma here.

thiblahute: OOps,

thiblahute:
> @thiblahute Has this been proposed upstream?

Yes: https://github.com/mstorsjo/fdk-aac/pull/139

subprojects/packagefiles/fdk-aac/meson.build
```diff
@@ -0,0 +1,403 @@
...
+  extra_args += ['-DSUPPRESS_BUILD_DATE_INFO']
+endif
+
+fdk_aac = library(
+  'fdk_aac',
+  fdk_aac_src,
+  dependencies: [m_dep],
+  include_directories : fdk_aac_inc,
+  install : true,
+  cpp_args : ['-fno-exceptions', '-fno-rtti', '-Dfdk_aac_EXPORTS'] + extra_args,
```
xclaesse, line 380: For msvc you probably need to use cc.get_supported_arguments()

xclaesse: cl : Command line warning D9002 : ignoring unknown option '-fno-exceptions'
cl : Command line warning D9002 : ignoring unknown option '-fno-rtti'

thiblahute: Thanks! Fixed.

releases.json
```diff
@@ -198,6 +198,14 @@
       "0.7.5-1"
     ]
   },
+  "fdk-aac": {
+    "dependency_names": [
+      "fdk-aac"
+    ],
+    "versions": [
+      "2.0.2"
```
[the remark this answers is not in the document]
thiblahute: Fixed already yes :-)

subprojects/packagefiles/fdk-aac/meson.build
```diff
@@ -0,0 +1,403 @@
...
+install_headers(files(fdk_aac_headers), subdir : 'fdk-aac')
+
+# Make using fdk-aac as a subproject behave the same as when used installed.
+subdir('fdk-aac')
+
+cc = meson.get_compiler('c')
+m_dep = cc.find_library('m', required : false)
+
+extra_args = ['-Dfdk_aac_EXPORTS']
+if get_option('remove-date').enabled()
```
eli-schwartz, line 370: What is the point of using .enabled() instead of a boolean, when auto and disabled are identical?

subprojects/packagefiles/fdk-aac/meson.build
```diff
@@ -0,0 +1,403 @@
...
+
+pkgc = import('pkgconfig')
+pkgc.generate(
+  fdk_aac,
+  name: 'fdk-aac',
+  description: 'AAC codec library'
+)
+
+if not get_option('build-programs').disabled()
+  executable(
```
eli-schwartz, line 398: And here. enabled and auto are identical, why not just use a boolean?

eli-schwartz:
From cmake:

```cmake
  if(WIN32)
    target_sources(aac-enc PRIVATE win32/getopt.h)
    target_include_directories(aac-enc PRIVATE win32)
  endif()
```

Windows CI fails due to:

```
[167/175] Compiling C object subprojects/fdk-aac-2.0.2/aac-enc.exe.p/aac-enc.c.obj
FAILED: subprojects/fdk-aac-2.0.2/aac-enc.exe.p/aac-enc.c.obj 
"cl" "-Isubprojects\fdk-aac-2.0.2\aac-enc.exe.p" "-Isubprojects\fdk-aac-2.0.2" "-I..\subprojects\fdk-aac-2.0.2" "-I..\subprojects\fdk-aac-2.0.2\libAACdec\include" "-I..\subprojects\fdk-aac-2.0.2\libAACenc\include" "-I..\subprojects\fdk-aac-2.0.2\libSYS\include" "-I..\subprojects\fdk-aac-2.0.2\libArithCoding\include" "-I..\subprojects\fdk-aac-2.0.2\libDRCdec\include" "-I..\subprojects\fdk-aac-2.0.2\libSACdec\include" "-I..\subprojects\fdk-aac-2.0.2\libSACenc\include" "-I..\subprojects\fdk-aac-2.0.2\libSBRdec\include" "-I..\subprojects\fdk-aac-2.0.2\libSBRenc\include" "-I..\subprojects\fdk-aac-2.0.2\libMpegTPDec\include" "-I..\subprojects\fdk-aac-2.0.2\libMpegTPEnc\include" "-I..\subprojects\fdk-aac-2.0.2\libFDK\include" "-I..\subprojects\fdk-aac-2.0.2\libPCMutils\include" "/MDd" "/nologo" "/showIncludes" "/W2" "/Od" "/Zi" "/Fdsubprojects\fdk-aac-2.0.2\aac-enc.exe.p\aac-enc.c.pdb" /Fosubprojects/fdk-aac-2.0.2/aac-enc.exe.p/aac-enc.c.obj "/c" ../subprojects/fdk-aac-2.0.2/aac-enc.c
../subprojects/fdk-aac-2.0.2/aac-enc.c(23): fatal error C1083: Cannot open include file: 'getopt.h': No such file or directory
```

subprojects/packagefiles/fdk-aac/meson.build
xclaesse: There is one difference, you can control the value with -Dauto_feautures. Dunno if that was the intention of @thiblahute. As he's a gstreamer dev, I think he just used feature because we use that everywhere in gst even for simple booleans.

thiblahute: `--auto-features=disabled` is very handy...

xclaesse:
`LINK : fatal error LNK1181: cannot open input file 'subprojects\fdk-aac-2.0.2\fdk_aac.lib'`

Most of the time that means your dll does not export any symbol. Check if cmake has a .def file, of some macro defined to dllexport.

subprojects/packagefiles/fdk-aac/meson.build
eli-schwartz: This is not a feature. It makes no sense to disable it.

Opinionated statement: Gst is wrong for using this pattern. Please submit a ticket to meson for adding --default-bools-to-false.

Objective statement that cannot be argued with: this neither enables nor disables any part of the build, so even misusing --auto-features serves no purpose here. This controls whether to replace the `__DATE__` and `__TIME__` macros with an empty string rather than the default compiler behavior which is to   ignore both date and time and use the value of $SOURCE_DATE_EPOCH instead.

Your CI is not benefiting from this.

eli-schwartz: Also it seems that cmake and autotools both don't offer this option, but Android.bp always defines it unconditionally... huh.

xclaesse: Tbh I would just drop that weird option for now. Indeed it does not seems to fit a feature, unlike build_programs that do make sense as a feature.

thiblahute: Agreed, it is not a feature, I overthought this one. It allows making my build reproducible so I still need it, will make it a bool

thiblahute: Well, it is building on debian, and iiuc the code requires to support it, which is not the case here.

I guess it is good enough to add an option no?

eli-schwartz: It's fine to add an option.

Anyway, some compilers automatically support SOURCE_DATE_EPOCH (gcc does), but not all. So it might or might not be a general solution for you, but if it is, you don't need the code to support it.

You just need the code to use `__DATE__` rather than idk, the build system adding the compiler flag `-DBUILD_DATE="$(/usr/bin/date)"`.

Yes, there are people doing the latter. (!!!)

subprojects/packagefiles/fdk-aac/meson.build
```diff
@@ -0,0 +1,408 @@
...
+fdk_aac_headers = [
+  'libAACdec/include/aacdecoder_lib.h',
+  'libAACenc/include/aacenc_lib.h',
+  'libSYS/include/FDK_audio.h',
+  'libSYS/include/genericStds.h',
+  'libSYS/include/machine_type.h',
+  'libSYS/include/syslib_channelMapDescr.h',
+]
+
+if build_machine.system() == 'windows'
```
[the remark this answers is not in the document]
thiblahute: I think the example I used in `gstreamer` was wrong :-)

Thanks!

xclaesse:
Looks all good to me now, and CI passed \o/

PR #184, opened 2021-10-06
ci: Add manual job that build all wraps

xclaesse:
This builds all wraps using meson master and report which passed/failed at the end. It can be triggered manually from github's actions page in wrapdb project. The goal is to blacklist those that are known to fail, and then run this when doing meson RC releases.


meson.build
```diff
@@ -7,7 +7,3 @@ wraps = get_option('wraps')
 foreach w : wraps
   subproject(w)
 endforeach
-
-run_target('update',
-  command: ['meson', 'subprojects', 'update', '--sourcedir', meson.current_source_dir(), '--reset'] + wraps,
-)
```
eli-schwartz, line 13: Why are we deleting this?

xclaesse: Otherwise it fails because in the CI job "meson" is not in PATH, and I thought nobody would be using that anyway. IIRC we have a PR somewhere to get meson cmd_array that could fix this.

tools/sanity_checks.py
```diff
@@ -143,11 +145,26 @@ def test_releases(self):
                             self.check_source_url(name, wrap_section, ver)
                     if i == 0 and t not in self.tags:
                         with self.subTest(step='check_new_release'):
-                            self.check_new_release(name, info, wrap_section)
+                            self.check_new_release(name)
```
eli-schwartz, line 148: In commit c8c74737cf6b82f6b200acc2d27b8d021ec5161c this stopped being used, indeed.

Can this cleanup be done in a separate commit for review purposes?

tools/sanity_checks.py
```diff
@@ -180,9 +197,9 @@ def check_new_release(self, name: str, info: T.Dict[str, T.Union[T.List[str], st
             else:
                 s = ', '.join(debian_packages)
                 print(f'The following packages could be required: {s}')
-        subprocess.check_call(['meson', 'setup', '_build'] + options)
-        subprocess.check_call(['meson', 'compile', '-C', '_build'])
-        subprocess.check_call(['meson', 'test', '-C', '_build'])
+        subprocess.check_call(['meson', 'setup', builddir] + options)
+        subprocess.check_call(['meson', 'compile', '-C', builddir])
+        subprocess.check_call(['meson', 'test', '-C', builddir])
```
eli-schwartz, line 202: This has worked okay for me locally in the past since it passes --wipe instead of using a temporary directory. Curious why this was needed.

xclaesse: Note that it still reuse "_build" when you run locally.

tools/sanity_checks.py
xclaesse: That was indeed unrelated cleanup.

PR #186, opened 2021-10-06
WIP: openssl: Use our own generator python script

xclaesse:
This makes the openssl wrap self contained and does not rely on any
external scripts like node.js.

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -336,7 +336,7 @@ class TestReleases(unittest.TestCase):
                         print('cannot verify in wrapdb due to missing dependency')
                         return
             raise Exception(f'Wrap {name} failed to configure due to bugs in the wrap, rather than due to being unsupported')
-        subprocess.check_call(['meson', 'compile', '-C', builddir])
+        subprocess.check_call(['meson', 'compile', '-C', builddir], env=meson_env)
         try:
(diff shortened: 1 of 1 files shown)


subprojects/packagefiles/openssl/meson.build
```diff
@@ -357,40 +180,32 @@ if get_option('default_library') == 'static'
   ]
 endif
 
-foreach define : defines + openssl_defines
-  c_args += ['-D' + define]
+foreach def : defines
+  c_args += '-D' + def
```
nazar-pc, line 184: `c_args` is an array here:
```suggestion
  c_args += ['-D' + def]
```

xclaesse:
Sorry, should have mentioned this is not ready for review, opened the MR just to test on CI.

subprojects/packagefiles/openssl/meson.build
xclaesse: Meson language is special here, `c_args += '-D' + def` does add string to array without having to created a temporary 1-item array.

nazar-pc:
> I don't think we have a way to test that until it gets merged...

Can you test it in your fork?

xclaesse:
> > I don't think we have a way to test that until it gets merged...
> 
> Can you test it in your fork?

If I merge in the master branch of my fork it triggers that CI indeed, nice. Unsurprisingly, it's not working yet...

PR #187, opened 2021-10-08
Fix using zlib.wrap with Clang on Windows

andrzejc:
Clang is using Microsoft STL from Visual Studio on Windows, so without
this change it fails to compile the subproject with lots of issues like:

FAILED: subprojects/zlib-1.2.11/libz.a.p/gzread.c.obj
../subprojects/zlib-1.2.11/gzread.c:35:15: error: 'read' is deprecated:
The POSIX name for this item is deprecated. Instead, use the ISO C and
C++ conformant name: _read. See online help for details.
[-Werror,-Wdeprecated-declarations]

This breaks the build if parent project uses werror flag like in the example above.


subprojects/packagefiles/zlib/meson.build
```diff
@@ -4,7 +4,7 @@ cc = meson.get_compiler('c')
 
 link_args = []
 compile_args = []
-if cc.get_id() == 'msvc'
+if cc.get_id() == 'msvc' or (cc.get_id().contains('clang') and host_machine.system() == 'windows')
```
dcbaker, line 7: This still isn't right, there are some other compilers that will be hit here, really:
```meson
if cc.get_compiler_argument_syntax() == 'msvc' or (host_machine.system() == 'windows'  and cc.get_id() == 'clang')
```

andrzejc: Could you please elaborate? I think even in the case this is actually mingw64 Clang, not the "official" Clang that is piggy-backing on MSVC headers defining -D_CRT_SECURE_NO_DEPRECATE and -D_CRT_NONSTDC_NO_DEPRECATE won't hurt, because those defines are not used in the zlib code, right?

dcbaker: `cc.get_compiler_argument_syntax() == 'msvc'` catche all of the compilers that pretend to be msvc, which would be `msvc`, `clang-cl`, and `intel-cl`.

andrzejc: sorry, I haven't noticed this change to `if cc.get_compiler_argument_syntax() == 'msvc'` in your comment on first sight and misinterpreted it; it's been a long day

jpakkane:
⚔️ conflicts. But with a rebase, this should be good to go.

neheb:
This was fixed differently with get_argument_syntax. Closing.

PR #188, opened 2021-10-08
Fix using openssl.wrap with Clang on Windows

andrzejc:
Clang was experiencing the same build issue (failure to find ws2_32 and
other Windows SDK libraries) as MSVC for which the workaround was
already present. Additionally, Clang was producing some minor warnings which
break the build of superprojects with werror flag enabled.


subprojects/packagefiles/openssl/meson.build
```diff
@@ -333,7 +340,7 @@ endif
 
 foreach library: openssl_libraries
   # MSVC fails with "ERROR: C static library 'ws2_32' not found" when `static` is specified at all
-  if compiler.get_id() == 'msvc'
+  if compiler.get_id() == 'msvc' or (compiler.get_id().contains('clang') and host_machine.system() == 'windows')
```
nazar-pc, line 343: This check is explicitly for MSVC. I didn't see any error because of this with either GCC 8.1 or Clang 12.0.1.

Here is the run with all 3 compilers on Windows that include OpenSSL as part of the build process: https://github.com/nazar-pc/mediasoup/runs/3836364833

nazar-pc: Hm, if that is really the case (I'm very far from native Windows development) then probably the whole `if/else` can be reduced to just `if` branch contents.

andrzejc: As I thought, this is caused by the fact that mingw64-Clang is finding import libraries that are part of mingw64 distribution, see: https://packages.msys2.org/package/mingw-w64-x86_64-crt-git?repo=mingw64 , there's libws2_32.a in the list which is found as a 'static' option but actually this is an import library for ws2_32.dll.

PR #193, opened 2021-10-11
update liburing wrap to 2.1

fischerling:
Updates the wrap to offer liburing 2.1.

Should I change the LICENSE.build file (like #170) ?
Problem is I don't hold the full Copyright on the build files a lot of the build code is from @eszlari.


subprojects/packagefiles/liburing/meson.build
```diff
@@ -71,7 +71,16 @@ int main(int argc, char **argv)
 '''
 has_cxx = cpp.compiles(code, name: 'C++')
 
-has_ucontext = cc.has_type('ucontext_t', prefix: '#include <ucontext.h>')
+code = '''#include <ucontext.h>
```
eli-schwartz, line 74: I think the goal of this change is because the upstream configure now checks for makecontext vs. makecontext_e2k -- is it possible to just check if the header provides the former symbol instead of reverting to a compile test?

xclaesse:
It fails on Windows: cl : Command line error D8021 : invalid numeric argument '/Wno-unused-parameter'

I think it needs to use cc.get_supported_arguments().

eli-schwartz:
It's linux-only software anyway, it is a wrapper over a Linux kernel feature.

subprojects/packagefiles/liburing/meson.build
fischerling: I think you are right. It seams that checking if `makecontext` is available is sufficient to match upstream's behavior.

I probably have just seen the diff and not the commit message of https://github.com/axboe/liburing/commit/b5f2347b53d4336da355aeafbb81ee748f8bc600.

What do you think about:

```diff
diff --git a/subprojects/packagefiles/liburing/meson.build b/subprojects/packagefiles/liburing/meson.build
index 17544b1..81b954c 100644
--- a/subprojects/packagefiles/liburing/meson.build
+++ b/subprojects/packagefiles/liburing/meson.build
@@ -71,16 +71,8 @@ int main(int argc, char **argv)
 '''
 has_cxx = cpp.compiles(code, name: 'C++')

-code = '''#include <ucontext.h>
-int main(int argc, char **argv)
-{
-  ucontext_t ctx;
-  getcontext(&ctx);
-  makecontext(&ctx, 0, 0);
-  return 0;
-}
-'''
-has_ucontext = cc.compiles(code, name: 'ucontext')
+has_ucontext = cc.has_type('ucontext_t', prefix: '#include <ucontext.h>')
+  and cc.has_function('makecontext', prefix: '#include <ucontext.h>')

 conf_data = configuration_data()
 conf_data.set('CONFIG_HAVE_KERNEL_RWF_T', has__kernel_rwf_t)
```

xclaesse:
Adding support for linux-only wraps: https://github.com/mesonbuild/wrapdb/pull/196.

xclaesse:
@fischerling after the PR above gets merged, you'll have to rebase, and add in ci_config.json:
```
  "liburing": {
    "linux_only": true
  }
```

subprojects/packagefiles/liburing/meson.build
eli-schwartz: This seems like it should do the right thing, yeah.

fischerling:
> Windows CI will now pass if you remove liburing from "skip_windows".

Should we document the fields in `ci_config.json` somewhere so new contributors can know how to configure the CI for their wraps?

PR #197, opened 2021-10-13
lame: Add [provides] section

thiblahute:
See https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/1068#note_1095243

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -38,6 +38,9 @@ PER_PROJECT_PERMITTED_FILES = {
         'meson.build.tmpl',
         'README.md',
     ],
+    'lame': [
+        'lame.h',
+    ],
 }
 NO_TABS_FILES = ['meson.build', 'meson_options.txt']
 PERMITTED_KEYS = {'versions', 'dependency_names', 'program_names'}
--- a/releases.json
+++ b/releases.json
@@ -488,7 +488,12 @@
     ]
   },
   "lame": {
+    "dependency_names": [
+      "mp3lame",
+      "mpglib"
+    ],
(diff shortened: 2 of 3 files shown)


thiblahute:
CI should be fixed, not ideal as we have our own `.h` file, but it works.

eli-schwartz:
Not only is it not ideal, the original rationale for it was mistaken.

The correct solution there was always to use vs_module_defs. Additionally it is possible to support symbol visibility on GCC by reimplementing libtool's version map: https://github.com/mesonbuild/meson/issues/3047

I hacked it. :)

eli-schwartz:
Weird: https://github.com/mesonbuild/wrapdb/runs/3878392317#step:6:232

The vs module defs file doesn't seem to work?

thiblahute:
> The vs module defs file doesn't seem to work?

Looks like it.

Using symbol visibility prototypes makes it simpler to me tbh.

tools/sanity_checks.py
```diff
@@ -38,6 +38,9 @@
         'meson.build.tmpl',
         'README.md',
     ],
+    'lame': [
+        'include/lame.h',
```
xclaesse, line 42: I think it's only the file name, not the path.

eli-schwartz:
> Using symbol visibility prototypes makes it simpler to me tbh.

Not forking the upstream header makes it simpler to me!

xclaesse:
> While I don't disagree, it is orthogonal to that PR :-)

while I don't disagree, we are trying to take the opportunity of wrap updates to improve them, otherwise nobody will... :P

PR #198, opened 2021-10-14
openjp2: Fix win32 build

dabrain34:
Tried to use 'add_project_arguments' after a build target has been
declared.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -14,7 +14,7 @@
     "fuse", "gdbm", "glib", "google-benchmark", "google-snappy", "google-woff2",
     "hinnant-date", "htslib", "icu", "imgui-sfml", "inih", "jbigkit", "json-c",
     "jsoncpp", "libexif", "libgpiod", "libjpeg-turbo", "liblbfgs", "liblzma",
-    "libopenjp2", "libpng", "libtomcrypt", "libuv", "libwebsockets",
+    "libpng", "libtomcrypt", "libuv", "libwebsockets",
     "libxml2", "libxmlpp", "lua", "lzo2", "mocklibc", "mpdecimal", "netstring-c",
     "nng", "openh264", "openssl", "pcre2", "protobuf", "protobuf-c", "pugixml",
     "quazip", "rdkafka", "reflex", "sdl2", "sdl2_image", "sdl2_mixer",


tp-m:
Ignore my comment please, I misread what you meant :)

subprojects/packagefiles/libopenjp2/meson.build
```diff
@@ -20,22 +20,17 @@ openjpeg_version_build = '1'
...
-
 # clock_gettime might require -rt, or it might not. find out
-if not cc.has_function('clock_gettime', prefix : '#define _GNU_SOURCE\n#include <time.h>')
-  # XXX: untested
-  rt_dep = cc.find_library('rt')
-else
-  rt_dep = []
+librt = cc.find_library('rt', required : false)
+if librt.found() and cc.has_header_symbol('time.h', 'clock_gettime')
+  add_project_arguments('-DHAVE_LIBRT=1', language : 'c')
```
xclaesse, line 29: Are you sure `HAVE_LIBRT` is used? The only place I saw for `clock_gettime()` is there:

https://github.com/uclouvain/openjpeg/blob/15615751b840d5b85a46292cf8fe393dab8e1a8b/src/bin/jp2/opj_decompress.c#L992

So it's only on linux, I think `librt = cc.find_library('rt', required : false)` should be enough here, no?

xclaesse:
So in thirdparty they are all wraps already, except lcms2. I think we should make its own wrap. Sorry, one bug uncovers another... but that would be a massive cleanup.

xclaesse:
https://github.com/mesonbuild/wrapdb/pull/200

xclaesse:
@dabrain34 Fixed some of the issues there: https://github.com/xclaesse/wrapdb/commits/pr-198, you can cherry-pick those commits into this PR. But CI still fails on Windows, there are more issues.

xclaesse:
@dabrain34 no, it makes no sense to maintain those third-party copies, please remove them.

xclaesse:
@dabrain34 you could add the wraps in subprojects/ but that's not required for wrapdb, they will be picked automatically by the CI.

xclaesse:
`meson wrap install zlib` will add the .wrap needed. You could add those wraps in subprojects/ if you want.

eli-schwartz:
There is no need to add recursive wraps to a wrapdb subproject. Users who are adding openjpeg to their project via `meson wrap install` are perfectly capable of adding its optional dependencies too (assuming they need those optional dependencies which are only used by some CLI tools that are highly unlikely to be needed by other projects simply linking to the openjpeg library).

It would be appropriate to include *.wrap files to an upstream PR adding meson to the upstream project.

Please do not add wraps to `subprojects/packagefiles/libopenjp2/subprojects/` or I will just have to add a review comment requesting (as a merge blocker!) that you remove them again.

dabrain34:
Thanks for your feedback guys

PR #200, opened 2021-10-14
Add lcms2 to wrapdb

xclaesse:
This is extracted from libopenjp2's thirdparty directory.

### DIFF
--- a/subprojects/packagefiles/lcms2/meson.build
+++ b/subprojects/packagefiles/lcms2/meson.build
@@ -1,0 +1,76 @@
+project('Little-CMS', 'c',
+  version: '2.12',
+  meson_version: '>=0.55.0',
+)
+
+library_version = '2.0.12'
+
+cc = meson.get_compiler('c')
+
+lcms2_srcs = [
+  'src/cmsalpha.c',
+  'src/cmscgats.c',
+  'src/cmserr.c',
+  'src/cmsgmt.c',
+  'src/cmsintrp.c',
+  'src/cmsio1.c',
+  'src/cmsmd5.c',
+  'src/cmsnamed.c',
+  'src/cmspack.c',
+  'src/cmsplugin.c',
+  'src/cmssamp.c',
+  'src/cmstypes.c',
+  'src/cmswtpnt.c',
+  'src/cmscam02.c',
+  'src/cmscnvrt.c',
+  'src/cmsgamma.c',
+  'src/cmshalf.c',
+  'src/cmsio0.c',
+  'src/cmslut.c',
+  'src/cmsmtrx.c',
+  'src/cmsopt.c',
+  'src/cmspcs.c',
+  'src/cmsps2.c',
+  'src/cmssm.c',
+  'src/cmsvirt.c',
+  'src/cmsxform.c',
+]
+
+install_headers([
+  'include/lcms2.h',
+  'include/lcms2_plugin.h'
+])
+
+inc_dirs = include_directories('include')
+
+deps = [
+  cc.find_library('m', required: false),
+  dependency('threads'),
+  dependency('zlib'),
+  dependency('libtiff-4'),
+]
+
+cargs = ['-DHasTHREADS=1', '-DHasZLIB=1', '-DHasTIFF=1']
+
+if cc.has_argument('-fvisibility')
+  cargs +='-DHAVE_FUNC_ATTRIBUTE_VISIBILITY=1'
+elif cc.get_id() == 'msvc'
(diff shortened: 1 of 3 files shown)


xclaesse:
We already had a meson port there: https://github.com/mesonbuild/wrapdb/tree/master/subprojects/packagefiles/libopenjp2/thirdparty/liblcms2. Better move it as a proper wrap so openjpeg can fallback to it.

xclaesse:
Submitted upstream too, just in case: https://github.com/mm2/Little-CMS/pull/284.

eli-schwartz:
Aha, there we go -- linux CI failed due to missing libm. :D

subprojects/packagefiles/lcms2/meson.build
```diff
@@ -0,0 +1,65 @@
...
+deps = [
+  cc.find_library('m', required: false),
+  dependency('threads'),
+  dependency('zlib'),
+  dependency('libtiff-4'),
+]
+
+cargs = ['-DHasTHREADS=1', '-DHasZLIB=1', '-DHasTIFF=1']
+
+if cc.has_argument('-fvisibility')
```
eli-schwartz, line 46: Would it make more sense to use `cc.has_function_attribute('visibility')` here?

xclaesse: Pretty much the same thing, I copy that from another project...

xclaesse:
> Missing install targets, missing options and deps from configure.ac (it looks up thread dep, libm, optional jpeg support, sse2 option, zlib dep, tiff dep, GNU symbol visibility but only if `-DHAVE_FUNC_ATTRIBUTE_VISIBILITY`, a couple random symbols, WORDS_BIGENDIAN), the "fastfloat" plugin, testsuite in `testbed/`, various CLI programs and more.

Added the needed/easy parts, I'm not really interested in doing a full 1:1 port. Can always be extended later.

subprojects/packagefiles/lcms2/meson.build
```diff
@@ -0,0 +1,65 @@
...
+  cargs +='-DHAVE_FUNC_ATTRIBUTE_VISIBILITY=1'
+elif cc.get_id() == 'msvc'
+  cargs += '-DCMS_DLL_BUILD=1'
+endif
+
+liblcms2_lib = library('lcms2', lcms2_srcs,
```
eli-schwartz, line 52: Also the autotools build creates a versioned library "2.0.12" (I believe due to misunderstanding libtool, but everyone misunderstands that anyway).

xclaesse: done.

eli-schwartz: nooooooooo the version has the 0 in the middle :p

(Yes I know it makes no sense and goes completely against the actual project version.)

Also: version, not soversion, the soversion is the part encoded in the DT_NEEDED ("Library soname") while version is the filename and can set the soversion automatically from the first component of the version.

subprojects/packagefiles/lcms2/meson.build
```diff
@@ -0,0 +1,74 @@
+project('Little-CMS', 'c',
+  version: '2.12'
+)
+
+library_version = '2.0.12'
+
+cc = meson.get_compiler('c')
+
+lcms2_srcs = [ 'src/cmsalpha.c'
+        , 'src/cmscgats.c'
```
eli-schwartz, line 10: I have just noticed that here in this one place there are commas at the beginning of each line. Which now that I look again, is a general style in the libopenjp2 wrap... madness.

Can this be moved to proper formatting, and include commas after the last element too (as that is the real rationale for making them at the beginning, so that adding new lines doesn't require editing an existing line to add the comma -- but this isn't json. :D)

subprojects/packagefiles/lcms2/meson.build
```diff
@@ -0,0 +1,74 @@
+project('Little-CMS', 'c',
+  version: '2.12'
```
eli-schwartz, line 2: This could use `meson_version: '>=0.48.0'`

xclaesse: actually 0.55 because of patch_directory.

eli-schwartz:
0.55 is wrong, probably caused by https://github.com/mesonbuild/meson/issues/9118

eli-schwartz:
As in, please I'd rather not use meson_version at all than do that.

xclaesse:
yep, exactly that, does not matter much tbh.

PR #203, opened 2021-10-17
ogg,vorbis: Install library files

seungha-yang:
Add missing install statement


tp-m:
You also need to update `releases.json` to add new revisions for each wrap.

And while you're at it, please fix the version number in vorbis's `meson.build` `project()`, should be 1.3.7 (as per https://github.com/mesonbuild/wrapdb/pull/201#issuecomment-945027960)

tp-m:
Also, what about the header files and pkg-config files?

I don't see any `install_headers()` calls etc.

subprojects/packagefiles/ogg/src/meson.build
```diff
@@ -4,6 +4,7 @@ libogg = library('ogg',
   sources,
   include_directories : incdir,
   vs_module_defs: '../win32/ogg.def',
+  install: true,
```
eli-schwartz, line 7: While we are at it, the official build system installs this with a soname "0" and a full version 0.8.5

subprojects/packagefiles/vorbis/lib/meson.build
```diff
@@ -31,37 +31,59 @@ if host_machine.system() == 'windows'
 endif
 
 math_dep = meson.get_compiler('c').find_library('m', required: false)
+pkgconfig = import('pkgconfig')
 
 vorbis_lib = library('vorbis', vorbis_sources,
   include_directories : [incdir],
   dependencies : [ogg_dep, math_dep],
   vs_module_defs: '../win32/vorbis.def',
+  install: true,
```
eli-schwartz, line 40: Upstream installs as libvorbis.so.0.4.9

subprojects/packagefiles/vorbis/lib/meson.build
```diff
@@ -31,37 +31,59 @@ if host_machine.system() == 'windows'
...
+  description: 'vorbis is the primary Ogg Vorbis library',
+  version: meson.project_version()
+)
+
 vorbisfile_lib = library('vorbisfile', vorbisfile_sources,
   include_directories : [incdir],
   link_with: vorbis_lib,
   dependencies : [ogg_dep, math_dep],
   vs_module_defs: '../win32/vorbisfile.def',
+  install: true,
```
eli-schwartz, line 58: libvorbisfile.so.3.3.8

subprojects/packagefiles/vorbis/lib/meson.build
```diff
@@ -31,37 +31,59 @@ if host_machine.system() == 'windows'
...
+  description: 'vorbisfile is a library that provides a convenient high-level API for decoding and basic manipulation of all Vorbis I audio streams',
+  version: meson.project_version()
+)
+
 vorbisenc_lib = library('vorbisenc', vorbisenc_sources,
   include_directories : [incdir],
   link_with: vorbis_lib,
   dependencies : [ogg_dep, math_dep],
   vs_module_defs: '../win32/vorbisenc.def',
+  install: true,
```
eli-schwartz, line 75: libvorbisenc.so.2.0.12

seungha-yang:
rebased

PR #204, opened 2021-10-17
checks: Ensure version match between releases.json and meson.build

thiblahute:
And fix all mistmatches


subprojects/packagefiles/abseil-cpp/meson.build
```diff
@@ -1,5 +1,5 @@
 project('abseil-cpp', 'cpp',
-  version : '20210324.1',
+  version : '20210324.2',
```
eli-schwartz, line 2: Haha, after complaining about vorbis  forgetting to bump the version in meson.build, the *only* other wrap with the problem is... my fault? :D

thiblahute: :-)

tools/sanity_checks.py
```diff
@@ -72,6 +72,21 @@ def test_releases_json(self):
...
                 self.assertIn(k, PERMITTED_KEYS)
 
+    def get_patch_path(self, wrap_section):
+        patch_directory = wrap_section.get('patch_directory')
+        if patch_directory:
+            return Path('subprojects', 'packagefiles', patch_directory)
+
+        return None
+
+    def check_meson_version(self, name: str, version: str,patch_path: str, builddir: str = '_build'):
```
eli-schwartz, line 82:
```suggestion
    def check_meson_version(self, name: str, version: str, patch_path: str, builddir: str = '_build'):
```

tools/sanity_checks.py
```diff
@@ -72,6 +72,21 @@ def test_releases_json(self):
...
+        with self.subTest(step="check_meson_version"):
+            with open(Path(builddir) / "meson-info/intro-projectinfo.json") as project_info_file:
+                project_info = json.load(project_info_file)
+                subproject, = [subproj for subproj in project_info["subprojects"] if subproj["name"] == name]
+                if subproject['version'] != 'undefined' and not patch_path:
```
eli-schwartz, line 87: So we only run the check if the patch_path does not exist -- i.e. meson.build comes from upstream? Oops.

eli-schwartz (changes requested):

tools/sanity_checks.py
```diff
@@ -72,6 +72,21 @@ def test_releases_json(self):
...
+                    self.assertEqual(subproject['version'], version, f"for '{name}'")
```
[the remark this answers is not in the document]
thiblahute: Makes sense, removed it.

eli-schwartz:
Thanks!

PR #205, opened 2021-10-19
libpng: Add MSVC assembly support for arm64

gaborkertesz-linaro:
MSVC supports arm64 on Windows on Arm, starting with Visual
Studio 2019.
MSVC assembler requires explicit C preprocessor call if
C preprocessor language is used in the assembly source file.
So arm/filter_neon.S is preprocessed then passed to assembler,
if MSVC is used.


gaborkertesz:
Can I help in anything to move forward with this change?

eli-schwartz:
Thanks for reminding me!

Seems like this is missing the revision bump in releases.json, without which this won't release a new version of the wrap.

gaborkertesz-linaro:
Is my understanding aoout json versioning is correct, where the "1.6.37" is the libpng version and the "-3" is the version of the supporting wrap?
I was a bit confused, because the last digits of 1.6.17 versions are not contiguous.

gaborkertesz-linaro:
Do you mean this https://github.com/mesonbuild/wrapdb/blob/master/ci_config.json#L17 ?

PR #206, opened 2021-10-19
lcms2: Remove unused dependencies

xclaesse:
Turns out zlib and tiff are only needed for utilities that we are not
building.

Also fix the visibility attribute check that was failing with:
cc: error: unrecognized command line option '-fvisibility'; did you mean '-fvisibility='?


eli-schwartz:
Nice!

PR #208, opened 2021-10-19
libuv: update to 1.42.0

nazar-pc:
Changes:
* update of libuv to 1.42.0 with corresponding changes imported from CMakeLists.txt (ignoring s390 platform, which meson.build didn't support anyway)
* various fixes for Windows, now CI compiles, but fails tests with timeouts, but I don't care enough to debug why exactly and didn't find timeout configuration quickly to try to increase it)
* remove enforced warnings


xclaesse:
If that's the problem, we could also add `--timeout-multiplier 2` in wrapdb because CI runners are often slower than real machines.

xclaesse:
> 145.18 sec

Ok, definitely need to increase timeout in your meson.build.

eli-schwartz:
Well, shouldn't we instead ask why doesn't it succeed?

nazar-pc:
> You could set build_tests=false in build_options.

It works on Linux though, so still valuable I think

eli-schwartz:
I was in the process of doing this... 431cf40f528333ec2ce78139c3af8759bbe05945

Mine is a bit nicer I think.

nazar-pc:
https://github.com/libuv/libuv/pull/3212/files was merged just 5 days ago upstream and is not part of the official release yet.

nazar-pc:
Yeah, apparently it is broken on MSVC 2019, I think this is as far as this PR will go until the next libuv release

nazar-pc:
Well, it is already in `ci_config.json`, should it still have `build_tests=false`? I'd leave it as is and just remove from `build_tests=false` next time update happens.

PR #211, opened 2021-10-20
Update glib to 2.70.0

xclaesse:


tools/sanity_checks.py
```diff
@@ -182,14 +182,11 @@ def test_build_all(self):
...
-
-            config = configparser.ConfigParser(interpolation=None)
-            config.read(f'subprojects/{name}.wrap')
-            wrap_section = config['wrap-file']
             try:
                 with tempfile.TemporaryDirectory() as d:
-                    self.check_new_release(name, d)
+                    # Disable --fatal-meson-warnings because Meson could have added
+                    # warnings/deprecations since last time this wrap has been updated.
+                    self.check_new_release(name, d, fatal_warnings=False)
```
eli-schwartz, line 189: This makes me nervous precisely because after a new meson release that causes something to become deprecated, is when I would like to know about it and update the build system to be compliant.

And for downstream wraps specifically, we control it and can release new wrap revisions independent of a stable project release, so... we really should require that no matter what (or mandate via policy that it is only permitted in the event that a subproject dependency cannot be fixed).

xclaesse: I think there are 2 separate and valid use-cases: Regression tests for Meson where introducing warnings is fine, and wrapdb maintainance where we want to fix warnings. I updated this PR to make fatal_warning a input option for the CI trigger.

ci_config.json
```diff
@@ -21,6 +21,14 @@
     "sdl2_net", "sdl2_ttf", "termbox", "unit-system", "xxhash", "zstd"
   ],
 
+  "glib": {
+    "_comment": ["- Disable fatal warnings: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2304",
+                 "- Disable unit tests because upstream does not run them on Windows and some are flicky on Linux."],
```
eli-schwartz, line 26:
```suggestion
                 "- Disable unit tests because upstream does not run them on Windows and some are flaky on Linux."],
```

PR #213, opened 2021-10-23
Update openh264 to v2.3.1

xclaesse:
It has an upstream meson build system. Unfortunately it has many meson
warnings, but they are all being fixed upstream for the next release:
https://github.com/cisco/openh264/pull/3351


eli-schwartz:
I've been specifically checking this periodically and holding off on updating this precisely because I was waiting for them to merge a fix.

jpakkane:
So what's the current status?

jpakkane:
:crossed_swords: conflicts.

eli-schwartz:
Fun. Upstream released an ABI breakage without bumping sover, which is a bit of an awkward situation... there's a ticket open, asking for an upstream patch release to fix that.

xclaesse:
Hmm, trying many tricks, but I can't make meson find nasm.exe installed by choco. I would have expected that starting meson in a new powershell session would have the latest PATH, but maybe not.

xclaesse:
oh great, been wasting my time trying to update env, it's a 10 years old nasm bug that nobody cares about... https://bugzilla.nasm.us/show_bug.cgi?id=3392224.

xclaesse:
> Fun. Upstream released an ABI breakage without bumping sover, which is a bit of an awkward situation... there's a ticket open, asking for an upstream patch release to fix that.

There seems to have a PR to bump version: https://github.com/cisco/openh264/pull/3565.

It's probably not a big deal for a wrap anyway, the point is to use openh264 as fallback when not installed on the system.

eli-schwartz:
Yeah but it's still less than 100% elegant to have what amounts to a pulled release. :)

I think we could probably hold off until the bumped version with minimal waiting.

xclaesse:
@eli-schwartz 2.3.1 has been released with the fix.

neheb:
@xclaesse as an alternative, how about using https://github.com/ilammy/setup-nasm ?

seems weird to rely on choco.

xclaesse:
@neheb a bit late, meson and wrapdb both search nasm in it's default install path. Been a while.

A reason for using choco is to install nasm only for wraps that need it. Using a gh action would do for every wraps.

PR #215, opened 2021-10-25
openjp2: cleanup obsolete components

dabrain34:
improve bin dependency checking

png, libtiff, lcms2 dep checking
can be conditional.
Remove zlib dependency as nobody is using it.

---------------------------------------------------------------

Remove unmaintained JPWL, JP3D and MJ2

Since https://github.com/uclouvain/openjpeg/pull/1350
the components JPWL, JP3d and MJ2 will not be
maintained in 2.5.0.

As win32 build was facing an issue with mj2 build and lrintf inline,
(see  https://github.com/uclouvain/openjpeg/pull/1339), better
to remove these useles components who will be removed in v2.5.0.


jpakkane:
LGMT. Anyone else?

eli-schwartz:
I had previously mentioned that enabling additional functionality (png, tiff, lcms2) for the command line programs using `required: false` is [automagic dependencies](https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies) and I'm not entirely sure that is a good idea compared to using a feature option.

neheb:
Needs rebasing.

eli-schwartz:
What about adding feature options for the optional dependencies?

jpakkane:
Still looks good. Anyone else have comments?

subprojects/packagefiles/libopenjp2/meson_options.txt
```diff
@@ -12,3 +12,6 @@ option('build_java', type : 'boolean', value : false, description : 'Build the o
 option('build_testing', type : 'boolean', value : false, description : 'Build the tests.')
 option('build_pkgconfig_files', type : 'boolean', value : true, description : 'Build and install pkg-config files')
 option('with_astyle', type : 'boolean', value : false, description : 'Build with astyle')
+option('bin_png', type : 'boolean', value : false, description : 'Build binaries with libpng')
+option('bin_tiff', type : 'boolean', value : false, description : 'Build binaries with libtiff')
+option('bin_lcms2', type : 'boolean', value : false, description : 'Build binaries with liblcms2')
```
eli-schwartz, line 16: Since these are used as `required: get_option('XXX')` for dependency lookups, setting `-Dbin_png=false` will actually build binaries with libpng, it just won't error out if it cannot.

It should be a feature option instead. The alternative is to check
```meson
if get_option('bin_png')
    png_dep = dependency('libpng')
else
    # always an empty not-found dependency
    png_dep = dependency('', required: false)
endif
```

eli-schwartz (changes requested):

jpakkane:
Please ack, I don't want to merge things marked as blocked.

subprojects/packagefiles/libopenjp2/meson_options.txt
eli-schwartz: Thanks. This fixes the automagic dependencies, so I retract my change request.

I will caution that I think feature options are nicer and more modern / DRY, and I'd generally prefer to use them in cases like this.

eli-schwartz:
I personally don't love the booleans instead of feature options, but that's a personal preference.

Feel free to merge it if you like, I myself would rather wait to make up my mind over the weekend but that's not a merge blocker anymore.

jpakkane:
Incremental improvements can always be done later. Thanks.

xclaesse:
This PR is missing a new entry in releases.json, so it did not generate a new release.

PR #216, opened 2021-10-25
libpng: Add aarch64 argument for win-arm64

gaborkertesz-linaro:
In order to get the same Aarch64 specific behavior for libpng
C source files, the corresponding __aarch64__ argument is added.
Even though MSVC defines _M_ARM64, this change fixes the
inconsistency without libpng host repository change.
Github issue is referred in the source code.


eli-schwartz:
Sorry for the delay, merged now.

PR #209, opened 2021-10-20
fluidsynth: update to 2.2.3

kcgen:

PR #217, opened 2021-10-26
Fuidsynth: Upgrade to 2.2.3 and fix Windows builds

shermp:
This PR is a replacement for #209 

As well as upgrading Fluidsynth to 2.2.3, changes were made to allow the wrap to build on Windows, including both MSVC and MSYS2 environments.


shermp:
Just a gentle nudge to ask what the status of the glib wrap is? I saw @eli-schwartz started work on it.

eli-schwartz:
Meson 0.60.1 will be released as soon as the final CI runner completes, and once that is out, I should be able to merge the glib update.

shermp:
> Meson 0.60.1 will be released as soon as the final CI runner completes, and once that is out, I should be able to merge the glib update.

Thanks for the update.

eli-schwartz:
Glib is updated, please rebase :D

subprojects/packagefiles/fluidsynth/meson.build
```diff
@@ -72,6 +94,21 @@ gthread_dep = dependency('gthread-2.0', static : get_option('try-static-deps'))
...
+# Windows specific options
+lib_args = []
+if host_machine.system() == 'windows'
+  if is_shared_lib
+    lib_args = ['-DFLUIDSYNTH_DLL_EXPORTS']
+  endif
+  # Specific workaround on MSYS2 to obtain static dependencies of intl, such as iconv
+  # This is required as of meson 0.59
+  if get_option('try-static-deps') and cc.get_id() in ['gcc', 'clang'] 
+    ext_deps += [dependency('intl' , static : true)]
```
shermp, line 108: Is there any problem of doing this check without bumping the minimum required Meson version?

The only platform that requires this (MSYS2) has Meson 0.59 available.

eli-schwartz: The platforms which don't require it, will still fail to "find" it, and error out.

If you use `required: false` then on versions of meson less than 0.60 it will just fail to find it at all. You may want to detect the specific case of MSYS2 though, and in that case set it to `required: true`.

shermp:
@eli-schwartz do you need to tag the new glib release in the wrapdb for CI to work properly?

eli-schwartz:
There is a github action to tag the new glib release, but that just uploads it to https://wrapdb.mesonbuild.com and running the wrapdb test project will use unreleased versions of a wrap which are available in `subprojects/`.

shermp:
I asked because I've had trouble in the past with `sanity_check` failing due to missing tags in the repo. (I had forgotten to get tags when I synced with upstream)

eli-schwartz:
That should only be a problem when releases.json is out of date, hence the rebase.

shermp:
Ok, rebased and cleaned history again.

@kcgen has indicated he will check there hasn't been any regressions on Mac's.

kcgen:
Confirming the branch builds and links without issue on macOS.
``` text
Apple clang version 13.0.0 (clang-1300.0.29.3)
Target: arm64-apple-darwin20.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

subprojects/packagefiles/fluidsynth/meson.build
shermp: The fact that meson doesn't allow getting environment variables is a pain. That is by far the easiest way to detect MSYS2.

I may have to detect if `sh` is available, and if so, check with a shell one-liner.

shermp: Added a proper MSYS2 check. This branch should only be run when host is Windows and MSYS2 is detected.

eli-schwartz: you can check with `python -c 'import os; print(os.getenv("MSYSTEM", ""))'` and python is more likely installed than sh on Windows... and guaranteed if you use meson, less guaranteed if you use https://sr.ht/~lattis/muon but whatever, it still works as long as python is installed.

subprojects/packagefiles/fluidsynth/meson.build
```diff
@@ -9,6 +9,33 @@ cc = meson.get_compiler('c')
...
+  message('Checking if Windows host is MSYS2 based')
+  shell_cmd = find_program('sh', required : false)
+  if shell_cmd.found() 
+    r = run_command('sh', '-c', '[ ! -z "$MSYSTEM" ] && [ "$MSYSTEM" != "MSYS" ] && exit 0; exit 1')
+    if r.returncode() == 0
+      message('MSYS2 based environment detected')
+      host_is_msys2 = true
+    endif
+  endif
+endif
```
shermp, line 38: I debated on whether to use `sh` or `python` for this.

Ended up using `sh`, because it's guaranteed to exist on MSYS2, so therefore if it can't be found, the environment is definitely not MSYS2.

eli-schwartz: All the messages here seem not very useful to the person seeing it on the console.

Consider:

```meson
shell_cmd = find_program('sh', required : false, disabler: true)
host_is_msys2 = run_command('sh', '-c', '[ ! -z "$MSYSTEM" ] && [ "$MSYSTEM" != "MSYS" ] && exit 0; exit 1', check: false).returncode() == 0
```

If the program is not found, it turns into a disabler object and any code that tries to evaluate it becomes a no-op (including the later branch `ext_deps += [dependency('intl' , static : true)]`)

shermp: That simplifies things a lot. Thanks for the suggestion.

New implementation still hides the check behind a 'windows' check, just in case some *nix based system had defined `MSYSTEM` for some reason.

eli-schwartz:
Can you squash the 3 commits which progressively add/modify the MSYS2 probing and workaround?

shermp:
> Can you squash the 3 commits which progressively add/modify the MSYS2 probing and workaround?

Done

shermp:
Thank you very much for your assistance with this @eli-schwartz

PR #218, opened 2021-11-01
SDL2: Upgrade to 2.0.16

kcgen:


kcgen:
The as-is SDL2 wrap hard-codes support for a very limited set of SDL2's features on Linux and Windows. These features aren't dynamically checked at setup time.

It also doesn't support:
 - Wayland
 - CPU SIMD acceleration (SSE, 3DNow, Altivec, ARM Neon, ..)
 - Accelerated hardware video ouptut (OpenGL, GLES, Vulkan, Metal, KMS DRM, ..)  
 - Audio interfaces (ALSA, Pulse, Pipewire, Fusion, ..)

Work to-do:

The list of `set_option` and `dep_options` found in [SDL2's cmake list](https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt) need to be converted to `Meson_options` entries, with default settings matching cmake's, and dynamic setup-time checks need to be added for each feature in `meson.build`.

Without audio or hardware acceleration, the wrap is effectively not viable for actual by game or emulator projects.

kcgen:
Closing this PR, as this work is beyond what I can take on.
Hopefully someone (or a company using Meson and SDL2) can sponsor such an effort!

jpakkane:
Not everything needs to be fixed in one go. Even minor updates are helpful.

PR #220, opened 2021-11-02
Fix project() 'version' argument in sdl2_image

germandiagogomez:
sdl2_image wrap project patch does not include version number. Fixed in this PR, was creating trouble for a dependency in my project.


jpakkane:
You need to update the version json file as well.

germandiagogomez:
@jpakkane is that the right way to do it?

eli-schwartz:
Yes, but it should be squashed into one commit.

germandiagogomez:
Done

jpakkane:
The ubuntu test failure is due to missing deps? The Windows one is unclear, but probably not caused by this change?

germandiagogomez:
Is there any way in which I can help? Indeed the Ubuntu one seems like a missing dependency in the machine running the CI. 

No idea what happens on Windows side.

eli-schwartz:
The sdl2 wrap needs a better CI config, I guess. But see also #218

PR #221, opened 2021-11-05
imgui: update to 1.85

stephanlachnit:
Same as https://github.com/mesonbuild/imgui/pull/13.


stephanlachnit:
Ok nice, thanks to the CI the dx backends are test-compiled. The webgpu support via emscripten can't be tested due to missing emscripten in ubuntu though.

subprojects/packagefiles/imgui/meson.build
```diff
@@ -30,21 +30,30 @@ if dx12_dep.found()
     sources += 'backends/imgui_impl_dx12.cpp'
     dependencies += dx12_dep
 endif
-metal_dep = dependency('metal', required: get_option('metal'))
-if get_option('metal').enabled()
+metal_dep = cpp.find_library('metal', required: get_option('metal'))
```
eli-schwartz, line 33: IIRC metal is an apple frameworks library, so if it is added as https://mesonbuild.com/Dependencies.html#appleframeworks it will only ever be found on macOS. Adding it as an ordinary library is probably not the same thing.

stephanlachnit: Done

ci_config.json
```diff
@@ -32,6 +32,18 @@
       "glib:tests=false"
     ]
   },
+  "imgui": {
+    "_comment": ["- Disable sdl_renderer in CI as version 2.0.17+ is required."],
```
eli-schwartz, line 36: It's a bit unfortunate that this apparently.needs the development version of sdl2?

Anyway this should probably be a version: kwarg on the dependency lookup.

eli-schwartz: Also yes, adding the version kwarg to the first sdl2 lookup will cause the subproject to not be a valid satisfier
 But the second sdl2 lookup will work fine as it doesn't have a version restriction, so sdl2 will be enabled but not sdl_renderer.

stephanlachnit: > It's a bit unfortunate that this apparently needs the development version of sdl2?

Yes a bit unlucky, but better to implement now  (https://github.com/ocornut/imgui/blob/55d35d8387c15bf0cfd71861df67af8cfbda7456/backends/imgui_impl_sdlrenderer.cpp#L1-L2).

> Anyway this should probably be a version: kwarg on the dependency lookup.

Good point, done!

PR #223, opened 2021-11-06
Update google-benchmark wrap to release 1.6.0

matta:


eli-schwartz:
The sanity checks fail because it doesn't have a provide section. Based on google-benchmark's README.md, the dependency name should be `benchmark` (they install https://github.com/google/benchmark/blob/main/cmake/benchmark.pc.in and benchmarkConfig.cmake)

matta:
Hi Eli, thanks for looking at this. I know almost nothing about meson or cmake so I am flying mostly blind here and appreciate the help. A few things:

I didn't know how to run the wrapdb CI checks before sending the PR. Should https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#passing-automatic-validation mention tools/sanity-checks.py?

I don't know how to manually test the new wrap against my project. Is it possible?

Upstream names the two libraries 'benchmark' and 'benchmark_main', so I've changed the wrap's meson.build to match. I don't know why meson's wrap chose different names, but I found it confusing.

I added the provide section, but now sanity-check fails because the prior versions did not (I think).

I'd be happy to first work on a 1.5.2-2 fix to address the wrapping modernization, then push the 1.6.0-1 that does the version upgrade separately. Or, I can squash this down to one commit. I'll defer to your judgement on that.

matta:
Ah, I see:

subprojects/benchmark-1.6.0/meson.build:34:0: ERROR: Target name 'benchmark' is reserved for Meson's internal use. Please rename.

eli-schwartz:
benchmark_main doesn't provide a usable dependency, the *-config.cmake file and pkg-config files are only provided for "benchmark". It's fine to just list that one thing in the provide section.

It's fine to have discrete commits in this PR for each change you're making, but only bump the version/release in releases.json once in the final commit.

matta:
Ok. I removed benchmark_main and dropped the library renames. Ready for another attempt.

eli-schwartz:
Oh right, almost forgot...

https://github.com/mesonbuild/wrapdb#how-to-contribute-new-wraps you need to add the provided dependencies in the most recent version to the releases.json too -- needed for `meson wrap search`.

matta:
Done. I also squashed it down to one commit.

eli-schwartz:
Failing to link on Windows. :(

matta:
The missing function comes from the Shlwapi library on Windows. I attempted to add that conditional dep

subprojects/packagefiles/google-benchmark/meson.build
```diff
@@ -28,13 +29,19 @@ benchmark_sources = [
 benchmark_main_sources = ['src/benchmark_main.cc']
 inc = include_directories('include')
 
-thread_dep = dependency('threads')
+
+lib_benchmark_deps = [ dependency('threads') ]
+
+if target_machine.system() == 'windows'
```
eli-schwartz, line 35: The meson documentation describes host_machine as the environment that the built library/program runs on, and target_machine as the environment that the built program (running on host_machine) generates code for.

>  *target machine* is the machine on which the compiled binary's output will run, *only meaningful* if the program produces machine-specific output.

eli-schwartz: Yes, cross-compilation is super confusing. https://mesonbuild.com/Cross-compilation.html

matta: Yeah, so I took this to mean that the target machine is what is relevant.  I.e. a hypothetical cross compile on a Linux machine to a windows binary would need the dep.  Similarly, a cross compile on a windows machine to a linux binary would not.  Am I wrong?

matta: I see that most everybody on github uses host_machine == 'windows' but I think everybody is basically wrong.  :-)  It just doesn't matter in practice because if either host or target is 'windows' then they both are (in practice).

eli-schwartz: build_machine is typically used for helper programs which are built and then run as part of the build process to do other things.

host_machine is typically used for defining the machine you're cross-compiling for.

target_machine is basically used only when the project you are building is a cross-compiler, and you are e.g. cross-compiling from your build_machine laptop running linux, trying to cross-compile a host_machine Windows compiler, and then when you have the built compiler, you're going to distribute it to Windows users, who are going to use that compiler to cross-compile the target_machine macOS programs.

matta: Oh, I get it.  target machine is yet a third thing and host machine is what I thought target was!

eli-schwartz: The terms used are very fun!

I like to ironically think of them like this:

build_machine is the machine you use as a buildbot host

host_machine is the machine you're building for.

target_machine is the machine that your target is targeting.

Also, this is basically the case with autotools too, but personally I don't trust any code I've written to be cross-compile safe unless I've double checked the manual right then to be sure.

matta:
I'm at a loss about the remaining VisualStudio build errors.

Earlier you said this package doesn't provide a usable benchmark-main (I don't understand why FWIW), and the current error seems related to it. Shall we remove the benchmark-main from meson.build? I pushed a commit to do that.

eli-schwartz:
Well, the benchmark-main library itself is probably usable. The problem is that there is no dependency interface upstream for it, so you cannot find it with `dependency(...)`. I suppose people could use `cc.find_library('benchmark_main')` to heuristically scrape for individual libraries.

eli-schwartz:
https://mesonbuild.com/FAQ.html#my-project-works-fine-on-linux-and-mingw-but-fails-to-link-with-msvc-due-to-a-missing-lib-file-fatal-error-lnk1181-why

This indicates the actual Windows version of benchmark is broken, whether benchmark-main is removed or not.

eli-schwartz:
Ah, apparently the upstream CMake build defines https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html to autogenerate a .def file that globs all symbols and exports the entire world.

matta:
Ah, so this is in part https://github.com/mesonbuild/meson/issues/2132.

Yes, looks like the google-benchmark wrap has never worked for Windows, since the DLL built by meson never exported any symbols. Upstream relies entirely on CMake's WINDOWS_EXPORT_ALL_SYMBOLS feature to deal with this. I know that Google typically statically links release binaries, so this is probably how the upstream Bazel build works (if it works on Windows at all).

I see that the abseil-cpp wrap, which faces a similar problem of building a C++ from the same company, uses static_library and avoids this problem (intentionally or not). I've gone with that approach for google-benchmark as well. E.g.

https://github.com/mesonbuild/wrapdb/blob/bbe3c37a8e1202980525c6251aaf7ec9efd49710/subprojects/packagefiles/abseil-cpp/meson.build#L197

eli-schwartz:
Thanks for persevering! The changes look great now.

The final issue is that this wrap now works on Windows, which means the suppression list marking this as expected-broken for the mass build test (that tries to build every wrap when we are preparing for a new release candidate of meson) should be updated to no longer list this wrap as broken.

So that's a yay. ;)

matta:
> The final issue is that this wrap now works on Windows, which means the suppression list marking this as expected-broken for the mass build test (that tries to build every wrap when we are preparing for a new release candidate of meson) should be updated to no longer list this wrap as broken.

I think I made the right tweak to ci_config.json to address that.  Thanks for all the help!

eli-schwartz:
Thanks!

Issue #222, opened 2021-11-06
PCRE subproject needs a new download URL


ebassi:
According to the [PCRE website](http://pcre.org/):

> Note that the former ftp.pcre.org FTP site is no longer available.

This has broken the PCRE wrap. Sadly, it appears that the only location for PCRE tarballs is a SourceForge mirror.

PR #224, opened 2021-11-07
Update the location of the PCRE tarball

ebassi:
The ftp.pcre.org server is gone, according to pcre.org. The next best
thing is the SourceForge mirror.

Fixes: #222


eli-schwartz:
This will need a revision bump in releases.json in order to update the wrap file provided by `meson wrap install pcre`.

jpakkane:
Might need a rebase as well.

ebassi:
I've rebased and updated `releases.json`.

xclaesse:
Was going to say just update to pcre2, but https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1451.

PR #225, opened 2021-11-11
WIP: Add GStreamer project

xclaesse:
Notes:
- I had to name it `gstreamer-full.wrap` because we cannot have 2 subprojects with the same name, and gstreamer monorepo has itself a subproject named `gstreamer`.
- Currently uses git because 1.19.3 release does not have tarball for the monorepo, only its subprojects. @tp-m I think we should make gstreamer-full.tar.gz as part of the official release for 1.20, what do you think?
- The `[provide]` section contains only libraries we always build, I took the list of pc files generated when configuring with `-Dauto_features=disabled`.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -365,6 +365,29 @@
       "1.0.2-1"
     ]
   },
+  "gstreamer-full": {
+    "dependency_names": [
+      "gstreamer-1.0",
+      "gstreamer-base-1.0",
+      "gstreamer-pbutils-1.0",
+      "gstreamer-rtsp-1.0",
+      "gstreamer-allocators-1.0",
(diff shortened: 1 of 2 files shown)


xclaesse:
Hmm, gtk4 checks for `gstreamer-player-1.0` first, I think that one is important enough that we should include it here even if it's not guaranteed that it's going to be built, if user disable the option.

eli-schwartz:
Why should we add a monorepo aggregation here? Why not add the individual projects instead?

jpakkane:
I refer to gst devs on what is the best way to package their stuff for downstream usage.

PR #229, opened 2021-11-16
Updated json-c to 0.15

igaw:


igaw:
FWIW, I am trying to get the CI build working for nvme-cli/libnvme. The provided json-c version in ubuntu-latest is too old.

https://github.com/linux-nvme/nvme-cli/runs/4227608084?check_suite_focus=true

xclaesse:
CI fails with MSVC:
```
json-c| Checking for size of "ssize_t" : -1
```

igaw:
Okay, hopefully I can figure out this bit as well. Thanks for the pointer.

igaw:
Unfortunately, I can't really test the MSVC part locally. So this is just a shoot in the dark. Maybe I am lucky :)

igaw:
The SSIZE_T thingy seems to be resolved with last version.  But now the linker is unhappy with

```
json_object.c.obj : error LNK2019: unresolved external symbol __sync_add_and_fetch referenced in function json_object_get
json_object.c.obj : error LNK2019: unresolved external symbol __sync_sub_and_fetch referenced in function json_object_put
```

Hmm, looking at the json-c code, there is a compile switch which disables the builtins. I guess I need some compiler checks for this as well.

xclaesse:
Probably that check: https://github.com/json-c/json-c/blob/master/CMakeLists.txt#L265

igaw:
As previously stated, the upstream project has fallback code in case the atomics are not supported. And the meson.build file sports this here:

https://github.com/mesonbuild/wrapdb/blob/0c908f8b6c3f1c58afbf8cf7a4f24ff9912e6c07/subprojects/packagefiles/json-c/config/meson.build#L101
```
if cc.compiles('int main() { int i, x = 0; i = __sync_add_and_fetch(&x,1); return x; }')                                            
  conf.set10('HAVE_ATOMIC_BUILTINS', true)                                                                                          
endif    
```

So in theory,  the fallback should be used for MSVC. There must be something wrong with this but I can't spot it right now.

igaw:
Let's see if 'cc.links()' does the trick.

eli-schwartz:
The following might help figure out whether that is actually being detected...

```diff
diff --git a/subprojects/packagefiles/json-c/config/meson.build b/subprojects/packagefiles/json-c/config/meson.build
index 1bdedf4..298a69f 100644
--- a/subprojects/packagefiles/json-c/config/meson.build
+++ b/subprojects/packagefiles/json-c/config/meson.build
@@ -98,11 +98,11 @@ foreach type : check_types
   conf.set('SIZEOF_' + type.underscorify().to_upper(), size)
 endforeach
 
-if cc.compiles('int main() { int i, x = 0; i = __sync_add_and_fetch(&x,1); return x; }')
+if cc.compiles('int main() { int i, x = 0; i = __sync_add_and_fetch(&x,1); return x; }', name: 'atomic builtins')
   conf.set10('HAVE_ATOMIC_BUILTINS', true)
 endif
 
-if cc.compiles('__thread int x = 0; int main() { return 0; }')
+if cc.compiles('__thread int x = 0; int main() { return 0; }', name: '__thread')
   conf.set10('HAVE___THREAD', true)
 endif
```

Note that the CMake function is more correctly comparable to meson's `cc.links()`

igaw:
Thanks for the feedback. 'cc.links' did the trick :) I'll add the 'name' part. Seems useful.
Right, now there is another failure...

eli-schwartz:
Unfortunately, we cannot use cc.has_function('__sync_add_and_fetch') because meson's implementation will then only look for `__has_builtin(__builtin___sync_add_and_fetch)`.

This is a special case of:
> The following built-in functions are intended to be compatible with those described in the Intel Itanium Processor-specific Application Binary Interface, section 7.4. ***As such, they depart from normal GCC practice by not using the ‘__builtin_’ prefix*** and also by being overloaded so that they work on multiple types.

eli-schwartz:
>  Right, now there is another failure...

json-c was detected on the list of "known failures" and blacklisted from the mega-test of all wraps (which we run for regression testing before new meson releases). It's very simple to remove it, thus marking this wrap as fixed. :D We want people who fix broken wraps to mark them as fixed, obviously.

xclaesse:
> Right, now there is another failure...

Just remove json-c from the list in ci_config.json and it should be all fine now, thanks!

igaw:
Yep, removed 'json-c' from the ci_config.json file. *fingers crossed*

eli-schwartz:
And in commit 7bae011d19cca3252793ae5ab79c0f59c2d23055 I improved the error message for that.

Artificially adding json-c to skip_linux, I get the following sanity checks failure:

```
[14/14] Linking target subprojects/json-c-0.15/libjson-c.so
No tests defined.
.
FAIL: test_releases (__main__.TestReleases) [If this works now, please remove it from skip_linux!] (step='check_new_release', name='json-c')
  File "/home/eschwartz/git/mesonbuild/wrapdb/./tools/sanity_checks.py", line 172, in test_releases
    self.assertNotIn(name, self.skip)
AssertionError: 'json-c' unexpectedly found in ['arduinocore-avr', 'cpr', 'glew', 'hinnant-date', 'imgui-sfml', 'jsoncpp', 'libtomcrypt', 'libxmlpp', 'openal-soft', 'openh264', 'pcre2', 'protobuf-c', 'pugixml', 'quazip', 'sdl2', 'sdl2_image', 'sdl2_mixer', 'sdl2_net', 'sdl2_ttf', 'zstd', 'json-c']

Ran 3 tests in 6.074s

FAILED (failures=1, skipped=1)
```

PR #230, opened 2021-11-17
json-c: Update to current pkgconfig.generate() API

igaw:
meson warns with:

  Library json-c was passed to the "libraries" keyword argument of a
  previous call to generate() method instead of first positional
  argument.

Update the call to the current function signature requirements. This
increases the minimal dependency on meson to 0.47.


eli-schwartz:
Normally I try to avoid bumping the minimum versions of meson when it's not strictly needed, aided by the fact that deprecation warnings are typically guarded by minimum-version checks. e.g. Ubuntu 18.04 doesn't have 0.46.

What's interesting here is that this warning actually gets triggered by the design of the json-c project, but the warning itself comes from a different project entirely (the one consuming json-c)... that issue is a bit difficult to spot in the sanity checks!

So this LGTM, but...

Your commit message references 0.47 while the change itself (and the pkgconfig module docs) is for 0.46.

igaw:
Oh, let me update the commit message with the right version number. I am pretty good in getting numbers wrong...

igaw:
FTR, this is what I currently see:

> json-c| WARNING: Project targeting '>=0.37' but tried to use feature introduced in '0.46.0': pkgconfig.generate optional positional argument.
> json-c| Build targets in project: 1
> json-c| WARNING: Project specifies a minimum meson_version '>=0.37' but uses features which were added in newer versions:
> json-c| * 0.46.0: {'pkgconfig.generate optional positional argument'}
>

eli-schwartz:
> I am pretty good in getting numbers wrong...


Aren't we all? :D

PR #233, opened 2021-11-23
Add RtAudio

ntonnaett:
5.2.0 is the first version with upstream meson support and was just released some days ago.

https://github.com/thestk/rtaudio

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -54,6 +54,13 @@
       "rdkafka:WITH_ZSTD=disabled"
     ]
   },
+  "rtaudio": {
+    "debian_packages": [
+      "libpulse-dev",
+      "libasound2-dev",
+      "libjack-dev"
+    ]
+  },
   "sfml": {
     "debian_packages": [
       "libgl-dev",
--- a/releases.json
+++ b/releases.json
@@ -981,6 +981,14 @@
       "3.0.1-1"
     ]
   },
+  "rtaudio": {
+    "dependency_names": [
+      "rtaudio"
+    ],
+    "versions": [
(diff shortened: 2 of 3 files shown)


xclaesse:
Seems their release tarballs does not contain meson.build files. I guess they use autotools to make release and forgot EXTRA_DIST?

ntonnaett:
I first checked if it builds and afterwards switched to wrap-file. So I need an overlay for 5.2.0 and try to add EXTRA_DIST upstream. Thanks for the hint!

This should also work on Windows and macOS.

ntonnaett:
Nope. It's in another subdirectory...

eli-schwartz:
Maybe upstream could release a .1 tarball that contains the full EXTRA_DIST?

eli-schwartz:
@xclaesse

>  A set of C++ classes that provide a common API for realtime audio input/output across Linux (native ALSA, JACK, PulseAudio and OSS), Macintosh OS X (CoreAudio and JACK), and Windows (DirectSound, ASIO, and WASAPI) operating systems.

ntonnaett:
I thought the directory line is for where the files get unpacked.

eli-schwartz:
This should really be one commit I think.

ntonnaett:
done

The error about exporting of a derived std::runtime_error class is bothering me for quite some time. Upstream only tests mingw.

eli-schwartz:
Is that actually an  error? It links correctly and the api test binary passes.

ntonnaett:
No, it's a warning. And probably no problem because the implementation is in the header file and actually doesn't need to be exported.

PR #234, opened 2021-11-24
fluidsynth: fix error when `sh` is not available

benoit-pierre:
The setup step would abort with `ERROR: Program or command 'sh' not found or not executable`.


eli-schwartz:
I guess the CI runner happened to have it installed, lol...

Thanks for catching that.

benoit-pierre:
It's the `run_command('sh', …)` invocation that fails if there's no `sh` executable available.

benoit-pierre:
There does not seem to be a point in using `disabler: true` in this case.

benoit-pierre:
Also, it might make more sense for the run command to actually use `shell_cmd`, no?

eli-schwartz:
Aha...

The run_command should be running shell_cmd instead of `'sh'` so that it reuses the found program object rather than creating an implicit second one. This will also prevent meson from logging twice that it tries to find that program.

benoit-pierre:
Ah, yes, it should be:
```python
  shell_cmd = find_program('sh', required : false, disabler: true)
  host_is_msys2 = run_command(shell_cmd, '-c', '[ ! -z "$MSYSTEM" ] && [ "$MSYSTEM" != "MSYS" ] && exit 0; exit 1', check: false).returncode() == 0
```

benoit-pierre:
Will amend.

benoit-pierre:
Done.

benoit-pierre:
It's [fixed upstream](https://gitlab.gnome.org/GNOME/glib/-/commit/e5dc299701df7189685d322cb709649ce3c9c81d), but not released yet, would that be an accepted exception to the `it is generally not accepted to override upstream files` rule?

eli-schwartz:
That's fixed on the glib maintenance branch which is scheduled for release at the beginning of December.

Actually, that release will also include a thing to make it use the same wrap name as the wrapdb itself -- which means that meson can prioritize using the wrapdb copy which we can independently fix the url for.

benoit-pierre:
OK.

eli-schwartz:
Actually I can just merge this with failing CI, as the change is verifiably correct via eyeball review.

benoit-pierre:
Well, that part does work as expected yes (I'm hitting other issues).

PR #236, opened 2021-11-30
libebml + libmatroska

neheb:


neheb:
So I fixed the sanity_checks error by replacing the patch with two simple sed commands.

I have no idea what's going on with VisualStudio though. Maybe it needs the two packages to be submitted separately?

neheb:
so the build creates subprojects/libebml-release-1.4.2/libebml.dll but libmatroska wants subprojects/libebml-release-1.4.2/libebml.lib ???

neheb:
is skip_windows broken?

eli-schwartz:
It's not documented in README.md, the json comment says:
> lists of wraps that are known to fail on the CI and should get repaired whenever possible

and the CI fails with:

```
FAIL: test_releases (__main__.TestReleases) [If this works now, please remove it from skip_windows!](step='check_new_release', name='libebml')

[...]

    self.assertNotIn(name, self.skip)
```

That first part, i.e. "it is not documented in README.md", means "this is not actually allowed to be used, so do not use it".

The other two parts serve to explain *why*. The purpose of `skip_windows` is to flag certain wraps as automatically failing CI, therefore it's not even worth bothering to check that they work. Specifically, they are excluded from ***regression*** tests. After all, why check for regressions in a wrap that you know fails either way? ;)

...

tl;dr This is *not* a setting for "I know this wrap doesn't work on Windows, but I want it to pass CI anyway, so that it can be merged".

Wraps which do not work on Windows should probably be fixed, because otherwise how do people use them?

neheb:
Note that it's only broken with MSVC and not with Clang on Windows. Tested all MSYS2 combinations. I've reported it upstream but I doubt I'll get a response soon: https://github.com/Matroska-Org/libebml/issues/91

neheb:
```
libebml| WARNING: You should add the boolean check kwarg to the run_command call.
```

is this a joke?

eli-schwartz:
Why would it be a joke?

Now, it may be unexpected but that is because this CI uses meson prereleases and 0.61.0rc1 is now released as of today.

neheb:
Alright so upstream says that shared builds on Windows are unsupported. I currently have the library being build as static_library. I assume this is fine? libebml is only used by libmatroska anyway.

subprojects/packagefiles/libebml/meson.build
```diff
@@ -0,0 +1,88 @@
+project('libebml', 'cpp',
+  version : '1.4.2',
+  default_options : ['warning_level=3', 'cpp_std=c++14'])
+
+#hacks to fix compilation with GCC11. They go away with the next version
+run_command('sed', 's/cassert/limits/g', '-i', 'src/EbmlString.cpp', check : true)
+run_command('sed', 's/cassert/limits/g', '-i', 'src/EbmlUnicodeString.cpp', check : true)
```
eli-schwartz, line 7: `sed` may be quite unavailable on Windows.

On BSD systems, and on macOS which uses the BSD sed, this will fail -- `src/EbmlString.cpp` is parsed as the value of `-i <extension>`. You need a different syntax for using GNU sed or BSD sed, and POSIX doesn't standardize any way to edit a file in place.

This also modifies the source directory which seems... very gross. Handling it as a custom_target which creates a new, modified file in builddir/ might make more sense...

And for context, this is a "backport" of https://github.com/Matroska-Org/libebml/pull/81
So... is there a status update on when upstream is going to release this?

subprojects/packagefiles/libebml/meson.build
```diff
@@ -0,0 +1,88 @@
...
+
+#hacks to fix MSVC compilation
+run_command('sed', 's/uint64 UnixEpochDelay;/uint64 UnixEpochDelay = 978307200;/g', '-i', 'ebml/EbmlDate.h', check : true)
+run_command('sed', 's/const uint64 EbmlDate::UnixEpochDelay = 978307200;//g', '-i', 'src/EbmlDate.cpp', check : true)
```
eli-schwartz, line 11: This cross references to https://github.com/Matroska-Org/libebml/issues/91

This seems like a relevant point made in that issue:
> when compiled with BUILD_SHARED_LIBS set these static variables are not exported, even when using EBML_DLL_API in their definition.

Wouldn't this be a better solution? I'm highly skeptical of sed-patching the upstream sources with unmerged changes.

neheb: https://github.com/Matroska-Org/libebml/pull/94 fixes it.

With static_library, this is not needed I think.

subprojects/packagefiles/libebml/meson.build
neheb: There is not.

My initial attempt was to use a .patch file and using patch but I guess even that is unavailable on Windows...

neheb:
removed MSVC sed line as it's only relevant for shared windows libraries, which is unsupported upstream. commented out GCC11+ fixes as they are not relevant to the CI.

PR #238, opened 2021-12-07
Add wayland-protocols

dcbaker:
These use meson upstream, and are mostly just headers.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -8,7 +8,7 @@
     "protobuf-c", "pugixml", "quazip", "sdl2", "sdl2_image", "sdl2_mixer",
     "sdl2_net", "sdl2_ttf", "zstd"
   ],
-  "skip_windows":[
+  "skip_windows": [
     "arduinocore-avr", "box2d", "cjson", "cpr", "docopt",
     "emilk-loguru", "enet", "facil", "flatbuffers", "freetype2",
     "fuse", "gdbm", "google-snappy", "google-woff2",
--- a/ci_config.json
+++ b/ci_config.json
@@ -68,5 +68,12 @@
       "libudev-dev",
       "libxrandr-dev"
     ]
+  },
+  "wayland-protocols": {
+    "linux_only": true,
+    "debian_packages": [
+      "libwayland-bin",
+      "libwayland-dev"
+    ]
   }
 }
--- a/releases.json
+++ b/releases.json
@@ -1227,6 +1227,14 @@
       "1.2.158-1"
     ]
   },
+  "wayland-protocols": {
+    "dependency_names": [
+      "wayland-protocols"
+    ],
+    "versions": [
+        "1.24-1"
+    ]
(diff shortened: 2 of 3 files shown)


ci_config.json
```diff
@@ -18,7 +18,8 @@
     "libxml2", "libxmlpp", "lua", "lzo2", "mocklibc", "mpdecimal", "netstring-c",
     "nng", "openh264", "openssl", "pcre2", "protobuf", "protobuf-c", "pugixml",
     "quazip", "rdkafka", "reflex", "sdl2", "sdl2_image", "sdl2_mixer",
-    "sdl2_net", "sdl2_ttf", "termbox", "unit-system", "xxhash", "zstd"
+    "sdl2_net", "sdl2_ttf", "termbox", "unit-system", "wayland-protocols",
```
eli-schwartz, line 21: This is intended to mark packages as "auto-fail CI, and in exchange mark as known-broken for the manually triggered build_all.yml".

For software that fails on Windows, but is intended to since it is linux-specific, see how liburing works later in this file.

releases.json
```diff
@@ -1227,6 +1227,12 @@
       "1.2.158-1"
     ]
   },
+  "wayland-protocols": {
+    "dependency_names": "wayland-protocols",
```
eli-schwartz, line 1231: This value needs to be a list, not a string.

eli-schwartz (changes requested):

ci_config.json
```diff
@@ -68,5 +68,8 @@
       "libudev-dev",
       "libxrandr-dev"
     ]
+  },
+  "wayland-protocols": {
+    "linux_oinly": true
```
eli-schwartz, line 73:
```suggestion
    "linux_only": true
```

ci_config.json
```diff
@@ -8,7 +8,7 @@
     "protobuf-c", "pugixml", "quazip", "sdl2", "sdl2_image", "sdl2_mixer",
     "sdl2_net", "sdl2_ttf", "zstd"
   ],
-  "skip_windows":[
+  "skip_windows": [
```
eli-schwartz, line 11: Still stray diff.

releases.json
```diff
@@ -1227,6 +1227,14 @@
       "1.2.158-1"
     ]
   },
+  "wayland-protocols": {
+    "dependency_names": [
+      "wayland-protocols"
+    ],
+    "versions": [
+        "1.24"
```
eli-schwartz, line 1235: Needs to have a wrap revision too, not just an upstream release version.

```suggestion
        "1.24-1"
```

eli-schwartz (changes requested):

dcbaker:
well, other than still running on windows this seems happy

eli-schwartz:
Let's see if that fixes the CI. The linux_only feature seems to have been broken...

PR #240, opened 2021-12-09
Add wrap files for slirp-4.6.1

kcgen:
This PR depends on the Glib 2.70.2, which is submitted under: https://github.com/mesonbuild/wrapdb/pull/241

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -69,6 +69,13 @@
       "libxrandr-dev"
     ]
   },
+  "slirp": {
+    "_comment": ["- Limited to linux_only pending the addition of other POSIX-ish environments to the test matrix."],
+    "linux_only": true,
+    "build_options": [
+      "glib:tests=false"
+    ]
+  },
   "wayland-protocols": {
     "linux_only": true,
     "debian_packages": [
--- a/releases.json
+++ b/releases.json
@@ -1067,6 +1067,14 @@
       "2.5.1-1"
     ]
   },
+  "slirp": {
+    "dependency_names": [
+      "slirp"
+    ],
+    "versions": [
+      "4.6.1-1"
+    ]
+  },
   "spdlog": {
     "dependency_names": [
       "spdlog"
--- a/subprojects/slirp.wrap
+++ b/subprojects/slirp.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = libslirp-v4.6.1
+source_url = https://gitlab.freedesktop.org/slirp/libslirp/-/archive/v4.6.1/libslirp-v4.6.1.tar.gz
+source_filename = libslirp-v4.6.1.tar.gz
+source_hash = 69ad4df0123742a29cc783b35de34771ed74d085482470df6313b6abeb799b11
+
+# the libslirp project uses a pkg-config name of slirp instead of libslirp
+[provide]
+slirp = slirp_dep


kcgen:
This PR depends on the Glib 2.70.2 (submitted under a separate PR).

kcgen:
I've added included `skip_windows = [ ..., "slirp", ... ]` in **`ci_config.json`**:




However, despite this, the Visual Studio CI job is still being run, and *still* failing (even with Glib 2.70.2):



The big lesson learned for me is: if you want a project to be cross-platform, stay well away from Glib!

kcgen:
Going on four hours wrestling with Glib and its brokenness in CI:

I've included slirp in the `skip_windows` list and even set `linux_only = true` for slirp, but the CI chain here continues to attempt Windows builds:



How can I de-activate this? Slirp + Glib + Windows is off the books right now, and I atleast want a half-usable wrap for the Linux side (even if Glib unit tests are still broken). 

Thank you for any tips or help.

eli-schwartz:
It's a CI error to specify skip_windows. Wraps must not be skipped. Adding them to the skip list is an admission of failure, and is used to track known-broken wraps to exclude them from the regression tests (test_all.yml).

Software that is linux-only by design rather than due to a bug, should use the linux-only key, and no other.

kcgen:
Slirp depends on a POSIX environment (and headers), and it does build under an msys2 environment. 

It does not build under visual studio.  

Other projects that ship libslirp functionaily use msys2 with modern gcc and clang (or previously, mingw gcc 4.x). 

> It's a CI error to specify skip_windows. Wraps must not be skipped. 

Thanks. Slirp is now removed from skip_windows. 

> Adding them to the skip list is an admission of failure,

Under Windows, we expect visual studio to be unable to compile Slirp (and this is confirmed here as well).  We do expect msys2 compilers to build Slirp. 

> Software that is linux-only by design rather than due to a bug, should use the linux-only key, and no other.

Slirp is in a grey zone, being posix-by-design, meaning it will definitely not build on Windows with Microsoft's compilers.  But it will build when given a mock-posix environment such as provided by cygin / mingw / msys2. 


How can we handle this gray zone?

eli-schwartz:
I would call this linux_only, pending an expansion of the test matrix to define other unixy test runtimes. You could leave a comment in the config file to that effect.

kcgen:
Thanks for the direction @eli-schwartz 👍 
Comment updated in the CI files and I improved the commit message as well. 
Looking forward to using Slirp in Staging!

eli-schwartz:
"502 bad gateway". Rerunning CI...

kcgen:
@eli-schwartz , 

Need some advice on the slirp wrap:

1. dep name: 
 - In pkg-config, it's called `slirp` (so meson looks for `slirp_dep`)
 - When build as a wrap, it provides `libslirp_dep`.

Argh.

2. include prefix directory
 - When installed in pkg-config, C++ code: `#include <slirp/libslirp.h>`
 - When built as a wrap,there is no parent `slirp`. C++ code: `#include <libslirp.h>`

Arghhhh.

This is the chunk from libslirp's meson.build:

``` python
if install_devel
  install_headers(['src/libslirp.h'], subdir : 'slirp')

  pkg = import('pkgconfig')

  pkg.generate(
    version : version,
    libraries : lib,
    requires : [
      'glib-2.0',
    ],
    name : 'slirp',
    description : 'User-space network stack',
    filebase : 'slirp',
    subdirs : 'slirp',
  )
else
  if get_option('default_library') == 'both'
    lib = lib.get_static_lib()
  endif
  libslirp_dep = declare_dependency(
    include_directories: include_directories('.', 'src'),
    link_with: lib)
endif
```

This makes it extremely painful (if not impossible) to use without making big messes of logical checks and redefinitions both on the meson side and the source-code code.  Certainly, we don't want to foist this on users (myself included).

eli-schwartz:
> Inconsistent dep names :
> 
>  * When installed in `pkg-config`, it's called `slirp` (so meson looks for `slirp_dep`)
> 
>  * When built as a wrap, it provides `libslirp_dep`.

It's not inconsistent at all. But clearly I should have reviewed it from a computer, not from my phone, and told you "of course this has to be `slirp = libslirp_dep` because the left hand side is the pkg-config name and the right-hand side is the meson variable name it is equal to, but there is no such meson variable".

(Even better would be if we could lint this in CI, but I am not sure how to programmatically get a variable name from outside the meson DSL.)

Meson looks for a random free-form string, and you incorrectly told it that the random free-form string would be the byte "s", followed by the byte "l", followed by the byte "i", followed by et cetera.



> Inconsistent `#include` prefix directories:
> 
>  * When installed in pkg-config, C++ code needs `#include <slirp/libslirp.h>`
> 
>  * When built as a wrap,there is no parent `slirp`, C++ code needs `#include <libslirp.h>`

This seems like a wrong interpretation of what is going on there.

```
$ pkg-config --cflags slirp
-I/usr/include/slirp -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include 
$ ls /usr/include/slirp
libslirp.h  libslirp-version.h
```

You're very much supposed to not use the "slirp/" part. If you do, and libslirp is installed in pkg-config mode into a non-default prefix, then you will be unable to find the headers, because pkg-config says it should be called `libslirp.h`, you are using `slirp/libslirp.h`, and it only sometimes accidentally worked if `/usr/include` was on the default search path ***and*** the software was installed there as well.

I guess to be fair it *usually* accidentally worked.

kcgen:
@eli-schwartz - thank you for coming to the rescue with a quick fixup, and for generously providing such a detailed explanation so I (and possibly others) can do a better job at creating wraps, and know what's going on under the hood.  :+1:

eli-schwartz:
I really wish I had any idea how to find out what variables are available in a meson project so I could lint that in CI, though. :D

PR #241, opened 2021-12-09
glib: update to 2.70.2 maintenance release

kcgen:


kcgen:
I tried re-activating unit tests to see if 2.70.2 will pass.  
But there are failures under Linux and Windows.. so I've dropped that commit:

### glib-2.70.2 unit test results on Linux



### glib-2.70.2 unit test results on Windows

eli-schwartz:
The Linux failures are due to missing dependencies, I think.

As the CI config says, upstream does not expect the tests to pass on Windows. People should stop trying to re-enable them, it's a total waste of time.

PR #242, opened 2021-12-10
Update Fluidsynth wrap to 2.2.4

shermp:
Updates the fluidsynth wrap to 2.2.4. ~~Also sneaks in another change to get 'try-static-deps' working with `iconv`. Previous config compiled fine standalone, but not when compiling with dosbox-staging, so was missed.~~


eli-schwartz:
Dang it, I was hoping that it was a short enough time period that no one would have pulled it. :/

shermp:
> Dang it, I was hoping that it was a short enough time period that no one would have pulled it. :/

No worries. Just a stroke of poor luck

Hmm, let's see if I can rebase my fork master and try again.

subprojects/packagefiles/fluidsynth/meson.build
```diff
@@ -112,7 +112,9 @@ if host_machine.system() == 'windows'
   # Specific workaround on MSYS2 to obtain static dependencies of intl, such as iconv
   # This is required as of meson 0.59
   if get_option('try-static-deps') and host_is_msys2
-    ext_deps += [dependency('intl' , static : true)]
+    # 'iconv' can be found via pkg-config on MSYS2, so no need for the 'iconv' specific
+    # dependency available in Meson 0.60
+    ext_deps += [dependency('iconv' , static : true)]
```
eli-schwartz, line 117: So, refreshing my memory on what is going on here...

https://github.com/mesonbuild/wrapdb/pull/217#pullrequestreview-789982285

Before 0.60, this will not be found, generally -- as I noted there. But it is found in the specific case of msys2, which this branch is under... and in meson 0.60, this should be found directly in meson.

But per the comments there, what you actually want is intl, not iconv. What's the specific problem solved by switching to iconv now?

shermp: Because I *still* get `iconv` linking errors when compiling `dosbox-staging` with `intl`, as-in `intl` doesn't appear to provide `iconv` as a required dependency. (This is for static linking only). The whole thing is a bit of a PITA to be honest...

shermp:
> Ugh, apologies for the remote merge commit @eli-schwartz , your force-push to master messed things up.

> Dang it, I was hoping that it was a short enough time period that no one would have pulled it. :/

Looks like I managed to rebase my fork successfully, so a clean history again

subprojects/packagefiles/fluidsynth/meson.build
eli-schwartz: The original 0.58 addition of the intl dependency does not check for static, but in 0.59 it learned to handle that and internally does `cc.find_library('iconv')`.

Which version of meson were you using?

shermp: Huh, swore I was having problems with 0.59. Just tried again with fresh fluidsynth wrap patch and clean build dir with meson 0.60.1, and it compiles fine.

I'll drop that commit. Thanks for prodding me to have another, closer look.

PR #244, opened 2021-12-12
Fix GTest build on Mac

Maratyszcza:
Specify C++11 target to avoid "no member named 'make_tuple' in namespace 'std'" errors when building GTest on Mac.


eli-schwartz:
Thanks, this looks like it makes sense based on https://github.com/google/googletest/blob/1b18723e874b256c1e39378c6774a90701d70f7a/googletest/cmake/internal_utils.cmake#L196

However, you should bump the revision in releases.json so that it gets released as a new version.

Maratyszcza:
@eli-schwartz PTAL

PR #245, opened 2021-12-12
fluidsynth: Add setup control for OpenMP feature

kcgen:


eli-schwartz:
This seems wrong, why not use a feature option and pass it into the openmp dependency lookup as `required: `?

kcgen:
Thanks for the quick suggestion @eli-schwartz.

eli-schwartz:
Seems to make sense, thanks!

PR #248, opened 2021-12-14
htslib: update to 1.14

SoapGentoo:


subprojects/packagefiles/htslib/test/meson.build
```diff
@@ -0,0 +1,119 @@
...
+
+test('test_hfile',
+  executable(
+    'test_hfile', [
+      htslib_config_h,
+      files(['hfile.c'])],
+    link_with : htslib_test_lib,
+    include_directories: '..',
+    install : false),
+  workdir : meson.project_source_root())
```
eli-schwartz, line 78: FWIW you could get away with a lower version of meson by using join_paths(meson.current_source_dir(), '..') or by assigning current_source_dir from the parent meson.build and reusing that variable. 0.48 is a lot more widespread, especially on stable distros (Debian buster, Ubuntu 20.04 primarily) and it seems fairly easy to do...

PR #252, opened 2021-12-31
vulkan-headers: update to 1.2.203

neheb:
Fixed actual usage by adding a provide line.


eli-schwartz:
Vulkan-Headers upstream doesn't provide a pkg-config file or a cmake file or an anything. How does this provide section fit into expected usage? People cannot do `dependency('vulkan-headers')` because it cannot be satisfied except via fallback... what should they do instead?

neheb:
If dependency cannot be used, how is this meant to be used?

eli-schwartz:
I don't know!

What I'd like to know is how the upstream Khronos Group expects people to consume this. Let's say it is installed as a system dependency, how do projects detect *that*?

neheb:
removed provides line.

PR #253, opened 2021-12-31
opencl-headers: add

neheb:
Added OpenCLHeaders as the name as upstream uses that for its cmake
configuration files.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -868,6 +868,14 @@
       "1.21.0-1"
     ]
   },
+  "opencl-headers": {
+    "dependency_names": [
+      "openclheaders"
+    ],
+    "versions": [
+      "2021.06.30-1"
+    ]
+  },
   "openh264": {
     "versions": [
       "1.7.0-1"
--- a/subprojects/packagefiles/opencl-headers/meson.build
+++ b/subprojects/packagefiles/opencl-headers/meson.build
@@ -1,0 +1,38 @@
+project(
+  'OpenCLHeaders',
+  'c',
+  license : 'Apache-2.0',
+  version : '2021.06.30',
+)
+
+install_headers(
+  'CL/cl_d3d10.h',
+  'CL/cl_d3d11.h',
+  'CL/cl_dx9_media_sharing.h',
+  'CL/cl_dx9_media_sharing_intel.h',
+  'CL/cl_egl.h',
+  'CL/cl_ext.h',
+  'CL/cl_ext_intel.h',
+  'CL/cl_gl_ext.h',
+  'CL/cl_gl.h',
+  'CL/cl.h',
+  'CL/cl_half.h',
+  'CL/cl_icd.h',
+  'CL/cl_layer.h',
+  'CL/cl_platform.h',
+  'CL/cl_va_api_media_sharing_intel.h',
+  'CL/cl_version.h',
+  'CL/opencl.h',
+  subdir : 'CL',
+)
+
+pkgc = import('pkgconfig')
+pkgc.generate(name: 'OpenCLHeaders',
+  version: '2.2',
+  description: 'C language headers for the OpenCL API'
+)
+
+opencl_headers_dep = declare_dependency(
+  include_directories : include_directories('.'),
+  version : '2.2',
+)
--- a/subprojects/opencl-headers.wrap
+++ b/subprojects/opencl-headers.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = OpenCL-Headers-2021.06.30
+source_url = https://github.com/KhronosGroup/OpenCL-Headers/archive/refs/tags/v2021.06.30.tar.gz
+source_filename = OpenCL-Headers-2021.06.30.tar.gz
+source_hash = 6640d590c30d90f89351f5e3043ae6363feeb19ac5e64bc35f8cfa1a6cd5498e
+patch_directory = opencl-headers
+
+[provide]
+OpenCLHeaders = opencl_headers_dep


releases.json
```diff
@@ -868,6 +868,14 @@
       "1.21.0-1"
     ]
   },
+  "opencl-headers": {
+    "dependency_names": [
+      "OpenCLHeaders"
+    ],
+    "versions": [
+      "2021-06.30-1"
```
eli-schwartz, line 876:
```suggestion
      "2021.06.30-1"
```

subprojects/packagefiles/opencl-headers/meson.build
```diff
@@ -0,0 +1,16 @@
+project(
+  'OpenCLHeaders',
+  'c',
+  license : 'Apache-2.0',
+  version : '3.0',
```
eli-schwartz, line 5: Where does 3.0 come from? Upstream uses `2.2`.

eli-schwartz (changes requested):
Seems to be missing the wrap file. :p

subprojects/packagefiles/opencl-headers/meson.build
neheb: Correct. sort of.

README.md says it should be 3.0. My guess is that someone forgot to update CMakeLists.txt.

eli-schwartz: Perhaps. The problem then is that OpenCLHeadersConfigVersion.cmake` sets the version as `set(PACKAGE_VERSION "2.2")` so these will disagree.

neheb:
```
FAIL: test_releases (__main__.TestReleases) (step='have_same_provides', name='opencl-headers')
  File "/home/mangix/devstuff/wrapdb/./tools/sanity_checks.py", line 151, in test_releases
    self.assertEqual(sorted(deps), sorted(info.get('dependency_names', [])))
AssertionError: Lists differ: ['openclheaders'] != ['OpenCLHeaders']

First differing element 0:
'openclheaders'
'OpenCLHeaders'

- ['openclheaders']
?   ^   ^^^

+ ['OpenCLHeaders']
?   ^   ^^^


FAIL: test_releases (__main__.TestReleases) (step='check_meson_version', name='opencl-headers')
  File "/home/mangix/devstuff/wrapdb/./tools/sanity_checks.py", line 95, in check_meson_version
    self.assertEqual(subproject['version'], version)
AssertionError: '3.0' != '2021.06.30'
- 3.0
+ 2021.06.30

```

hmmm no idea how to fix these.

eli-schwartz:
I think that wrap provides are case insensitive so just lower-casing it and searching for `dependency('OpenCLHeaders')` should work?

The version discrepancy is entirely confusing, because upstream tags a release according to date and then declares a project version according to the *specification* version, which seems  a bit wrong...

In general it *is* possible to have `pkgconfig.generate()` and `declare_dependency()` declare a version different from the project version. This is typically done in cases where a library API version differs from a project version, but in this case it is less "differs" and more "wasn't thought through" I guess.

neheb:
dependency('OpenCLHeaders') works with the cmake files. dependency('openclheaders') does not with the cmake files. No idea about pkgconfig.

subprojects/packagefiles/opencl-headers/meson.build
neheb: Replaced with 2.2.

neheb:
Lower-cased dependency_names.

eli-schwartz:
pkg-config and cmake lookups are case sensitive.

They query the case-sensitive name via the cmake or pkg-config tools, which themselves look up files with a specific name and filesystems, well, the better class of filesystem, are case sensitive.

Meson itself doesn't need to care though, so a provide section can be looked up either way. So can dependencies provided as custom dependency factories e.g. "system" lookups. In that case, the python code that implements this will declare a (case-sensitive) name mapping for pkg-config or cmake.

subprojects/packagefiles/opencl-headers/meson.build
```diff
@@ -0,0 +1,16 @@
...
+)
+
+pkgc = import('pkgconfig')
+pkgc.generate(name: 'OpenCLHeaders',
+  version: '2.2',
+  description: 'C language headers for the OpenCL API'
+)
+
+opencl_headers_dep = declare_dependency(
+  include_directories : include_directories('CL'),
```
eli-schwartz, line 15: This should declare the version as well, since it defaults to the project version which is not what we want to use here.

That way, `dependency('OpenCLHeaders', version: '==2.2')` will work correctly.

eli-schwartz (changes requested):

subprojects/packagefiles/opencl-headers/meson.build
```diff
@@ -0,0 +1,16 @@
...
+  'OpenCLHeaders',
+  'c',
+  license : 'Apache-2.0',
+  version : '2021.06.30',
+)
+
+pkgc = import('pkgconfig')
+pkgc.generate(name: 'OpenCLHeaders',
+  version: '2.2',
+  description: 'C language headers for the OpenCL API'
```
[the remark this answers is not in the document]
neheb: It does now.

subprojects/packagefiles/opencl-headers/meson.build
neheb: Done.

subprojects/packagefiles/opencl-headers/meson.build
```diff
@@ -0,0 +1,38 @@
...
+)
+
+opencl_headers_dep = declare_dependency(
+  include_directories : include_directories('CL'),
```
eli-schwartz, line 36: Hmm, I think that the include directory should be `'.'` -- the same way we install `CL/*.h` to `includedir/CL/*.h` but don't specify:

>  `subdirs` which subdirs of `include` should be added to the header search path, for example if you install headers into `${PREFIX}/include/foobar-1`, the correct value for this argument would be `foobar-1`

Indeed, the cmake config specifies /usr/include as the correct -I flag too, and the project documentation says to use `#include <CL/opencl.h>`.

eli-schwartz (changes requested):

subprojects/packagefiles/opencl-headers/meson.build
neheb: Good catch.

eli-schwartz:
Since commit 3cd3b2db62d15a530242830218509d3a3549970c there is no need to include a copy of LICENSE.build as the COPYING file from the root of the repo will be used when creating the uploaded release. And it is the same license. :)

neheb:
Great

eli-schwartz:
Thanks!

PR #254, opened 2022-01-02
add libyaml

alex-tee:


subprojects/packagefiles/libyaml/meson.build
```diff
@@ -0,0 +1,54 @@
...
+  'YAML_VERSION_MINOR' : yaml_minor_version,
+  'YAML_VERSION_PATCH' : yaml_patch_version,
+})
+cdata.set_quoted('YAML_VERSION_STRING', meson.project_version())
+
+config_h = configure_file(
+  output: 'config.h',
+  configuration: cdata)
+
+yaml_lib = library('yaml',
```
eli-schwartz, line 27: This is wrong. :(

Actually, this is confusing as heck.

`src/Makefile.am` passes the `-release $(YAML_LT_RELEASE)` flag to libtool, producing a library named `yaml-0` and linked to via... `-lyaml`

subprojects/packagefiles/libyaml/meson.build
```diff
@@ -0,0 +1,54 @@
...
+  'src/loader.c',
+  'src/parser.c',
+  'src/reader.c',
+  'src/scanner.c',
+  'src/writer.c',
+  include_directories: [
+    include_directories('.'),
+    ],
+  version: meson.project_version(),
+  soversion: yaml_major_version + '.' + yaml_minor_version,
```
eli-schwartz, line 40: Upstream library versioning uses libtool compatible versioning `-version-info 2:9:0` which translates to `version: '2.0.9'` and the soversion is directly inferred from version as `soversion: '2'`. it does not include multiple components.

subprojects/packagefiles/libyaml/meson.build
```diff
@@ -0,0 +1,54 @@
...
+  implicit_include_directories: false,
```
eli-schwartz, line 41: As a matter of curiosity, why is this necessary, and why do you then add `include_directories('.')`?

eli-schwartz (dismissed):
This installs:
```
        ├── libyaml.so -> libyaml.so.0.2
        ├── libyaml.so.0.2 -> libyaml.so.0.2.5
        ├── libyaml.so.0.2.5
```

Upstream installs:
```
        ├── libyaml.so -> libyaml-0.so.2.0.9
        ├── libyaml-0.so.2 -> libyaml-0.so.2.0.9
        ├── libyaml-0.so.2.0.9
```

I have... no idea how to handle that "yaml -> yaml-0" part. It won't break pkg-config but it will break people manually using `-lyaml`?

subprojects/packagefiles/libyaml/meson.build
alex-tee: I copy-pasted this part from another package then I saw that the `.` directory is needed. force-pushed a fix

subprojects/packagefiles/libyaml/meson.build
eli-schwartz: This cannot be replicated in meson. The soname (ABI experience) is named `yaml-0`, while the pkg-config file exposes `yaml`.

The perfect is the enemy of the good -- so let's merge this using `yaml-0`. This is compatible in the soname table, and even generates non-equal but still *compatible* pkg-config files, and only breaks people using the equivalent of `cc.find_library('yaml')`. At this point we can say "you probably should build with autotools if you are doing system packaging.

Can you leave a comment to the effect of "this cannot avoid breaking the use of `-lyaml` due to strange libtool magic, but everything else works" so that no one forgets why this is happening? If upstream is ever interested in meson, they can debate the issue at that time.

alex-tee: done

eli-schwartz:
I'd also like to see the testsuite at https://github.com/yaml/libyaml/tree/master/tests get run. It's just some simple binaries linked to libyaml and run without options.

alex-tee:
> I'd also like to see the testsuite at https://github.com/yaml/libyaml/tree/master/tests get run. It's just some simple binaries linked to libyaml and run without options.

I added some of the tests and they execute but they expect input to do anything meaningful. at least we know you can link and run binaries

eli-schwartz:
FWIW upstream only defines two of them (test-version test-reader) as check_PROGRAMS, and the rest are just defined as noinst_PROGRAMS which are probably meant to be run interactively.

Even though both cmake and autotools *build* the other programs, they don't run them as tests. I'm indifferent as to whether they are built at all.

alex-tee:
> FWIW upstream only defines two of them (test-version test-reader) as check_PROGRAMS

I made it build those 2 only

PR #255, opened 2022-01-03
libnpupnp: add

neheb:

### DIFF
--- a/subprojects/packagefiles/libnpupnp/meson.build
+++ b/subprojects/packagefiles/libnpupnp/meson.build
@@ -1,0 +1,213 @@
+project('libnpupnp', 'cpp',
+  version : '4.2.1',
+  default_options : [ 'cpp_std=c++11', 'warning_level=1' ]
+)
+
+cpp = meson.get_compiler('cpp')
+
+deps = []
+if host_machine.system() == 'windows'
+    deps += cpp.find_library('ws2_32')
+    deps += cpp.find_library('iphlpapi')
+elif host_machine.system() == 'sunos'
+    deps += cpp.find_library('nsl')
+    deps += cpp.find_library('socket')
+endif
+
+deps += dependency('threads')
+deps += dependency('libmicrohttpd')
+deps += dependency('libcurl')
+expat_dep = dependency('expat', required : get_option('expat'))
+deps += expat_dep
+
+if (host_machine.system() == 'windows')
+  add_project_arguments('-DDLL_EXPORT', language : 'cpp')
+endif
+
+add_project_arguments('-DSMALLUT_NO_REGEX', language : 'cpp')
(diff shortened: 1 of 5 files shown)


neheb:
did not expect that...

neheb:
No idea how to proceed.

subprojects/packagefiles/libnpupnp/meson.build
```diff
@@ -0,0 +1,215 @@
...
+if get_option('tools')
+  npupnp_sources += [
+    'src/api/upnptools.cpp',
+  ]
+endif
+
+libnpupnp = library('npupnp',
+  npupnp_sources,
+  include_directories : libnpupnp_incdir,
+  dependencies : [ winsock_dep, iphlpapi_dep, nsl_dep, socket_dep, threads_dep, libmicrohttpd_dep, libcurl_dep, expat_dep ],
```
eli-schwartz, line 86: This also has a library version upstream which should be retained, `-version-info 6:1:2` in libtool terms is  `version: '4.2.1'` in meson terms (and produces `libnpupnp.so.4.2.1`).

subprojects/packagefiles/libnpupnp/meson.build
```diff
@@ -0,0 +1,215 @@
+project('npupnp', 'cpp',
```
eli-schwartz, line 1: Hmm, shouldn't this be libnpupnp? That is what is in AC_INIT().

eli-schwartz (commented):
The libmicrohttpd thing on Windows could be solved either by adding libmicrohttpd to the WrapDB too, or disabling an optional component if not found... but libmicrohttpd doesn't seem to be optional.

jpakkane:
Since the deps are not available as wraps, you need to disable building on Windows in the json files at the root.

jpakkane:
It's a bit unexpected, but you need to set `linux_only` to `true` rather than adding the name to `skip_windows`.

neheb:
Set,

In other news, I tried to make a meson.build file for libmicrohttp but it seems a lot more work than this...

eli-schwartz:
I kinda... really dislike the idea of using linux_only, which is *supposed* to indicate that the source code is written for POSIX and thus isn't expected to build with MSVC, for "don't build on Windows because missing dependencies".

At the very least, this needs an explanatory comment for why on earth the software doesn't build on Windows even though it's supposed to.

(I would much prefer a libmicrohttpd port obviously.)

neheb:
Added comment.

jpakkane:
⚔️ conflicts.

neheb:
Joy.

neheb:
Fixed conflicts.

jpakkane:
Thanks.

PR #256, opened 2022-01-03
lua: put interpreter behind a compile option

Jan200101:
Only build the lua interpreter (lua, luac) if needed.

If a project only depends on Lua as a library for scripting then there is no reason to also build the interpreter


subprojects/packagefiles/lua/meson_options.txt
```diff
@@ -6,3 +6,7 @@ option(
   'line_editing', type: 'boolean', value: false,
   description: 'Link with GNU readline (or libedit) to add history to the Lua interpreter.'
 )
+option(
+  'interpreter', type: 'boolean', value: true,
+  description: 'Build the Lua interpreter.'
+)
```
eli-schwartz, line 12: Looks like this has a bad line ending.

Jan200101: fixed, no idea how this got past me

eli-schwartz:
Seems reasonable, and behavior-preserving by default... :)

eli-schwartz:
In order to do a release of the changes, you need to bump the revision in releases.json though.

Jan200101:
right, apologies for rushing things

eli-schwartz:
Argh, Windows...

```
[36/36] Linking target subprojects/lua-5.4.3/lua.exe
FAILED: subprojects/lua-5.4.3/lua.exe subprojects/lua-5.4.3/lua.pdb 
"link"  /MACHINE:x64 /OUT:subprojects/lua-5.4.3/lua.exe subprojects/lua-5.4.3/lua.exe.p/src_lua.c.obj "/nologo" "/release" "/nologo" "/DEBUG" "/PDB:subprojects\lua-5.4.3\lua.pdb" "subprojects\lua-5.4.3\lua.lib" "/SUBSYSTEM:CONSOLE" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
LINK : fatal error LNK1181: cannot open input file 'subprojects\lua-5.4.3\lua.lib'
```

Jan200101:
love it when it does that.
I'll see what I can do when I get a hold of a Windows machine

eli-schwartz:
The good news is that it doesn't seem to actually be your fault. The last update to this wrap was before we added Windows CI, and the build_all regression tests marked lua as known-broken now that I look at it (ci_config.json).

Jan200101:
Tested against MinGW64 and came across no issue so this seems to be caused by Visual Studio

Jan200101:
Okay I cannot figure out whats going wrong.
This CI issue existed even before this PR so its not something new.

Essentially, the `executable("lua")` is expecting the previously defined `library("lua")` to be static for whatever reason
I am not well versed with meson but I have essentially tried to strip the build.meson as far down as I could without getting anything to work.

Is there anything I am missing?

Jan200101:
I have taken the aproach of creating a internal static library to link the interpreter.
That solves the issue and works

eli-schwartz:
I suppose that seems fine. In theory one could link to the static library only on MSVC, and link to the default library otherwise.

eli-schwartz:
⚔️ conflicts

Jan200101:
After some digging this seems to be expected
https://github.com/mesonbuild/meson/blob/master/docs/markdown/FAQ.md#my-project-works-fine-on-linux-and-mingw-but-fails-to-link-with-msvc-due-to-a-missing-lib-file-fatal-error-lnk1181-why

It seems that lua does not support MSVC at all so compiling them statically if msvc is detected is the best way.

eli-schwartz:
Cool, thanks for figuring this out!

PR #258, opened 2022-01-04
sdl2: update to 2.0.18

Jan200101:
Tested with:
- gcc11 Fedora 35 x86_64
- gcc11 MinGW64 x86_64


Jan200101:
Line endings got messed up somewhere along the way

eli-schwartz:
Thanks.

A precious attempt at this was in #218, unfortunately getting it to run in CI is... complicated, by the fact that the original implementation of this wrap never did thorough feature checks. The CI could get further if you installed some Linux dependencies via ci_config.json

Jan200101:
I see a way of fixing the Linux CI, but not the Windows CI.
How do you suggest proceeding?

jpakkane:
The following patch on top of this makes it compile for me:

```
diff --git a/meson.build b/meson.build
index ccef14e..d2e7dac 100644
--- a/meson.build
+++ b/meson.build
@@ -143,7 +143,7 @@ endforeach
 cdata.set('HAVE_LIBC', 1)

 if cc.compiles('''#include<math.h>
-double dummy = M_PI
+double dummy = M_PI;
 ''', name : 'M_PI')
   cdata.set('HAVE_M_PI', 1)
 endif
diff --git a/src/core/windows/meson.build b/src/core/windows/meson.build
index 9c4a688..a959c9f 100644
--- a/src/core/windows/meson.build
+++ b/src/core/windows/meson.build
@@ -1 +1,4 @@
-all_sources += files('SDL_windows.c', 'SDL_xinput.c')
+all_sources += files('SDL_windows.c',
+    'SDL_xinput.c',
+    'SDL_hid.c',
+    )
diff --git a/src/hidapi/meson.build b/src/hidapi/meson.build
index 499c074..2017606 100644
--- a/src/hidapi/meson.build
+++ b/src/hidapi/meson.build
@@ -2,4 +2,8 @@ all_sources += files(
   'SDL_hidapi.c',
 )

-core_inc += include_directories('hidapi')
\ No newline at end of file
+core_inc += include_directories('hidapi')
+
+  core_deps += [cc.find_library('setupapi')]
+endif
diff --git a/src/joystick/hidapi/meson.build b/src/joystick/hidapi/meson.build
index 3a14de3..58092b4 100644
--- a/src/joystick/hidapi/meson.build
+++ b/src/joystick/hidapi/meson.build
@@ -1,10 +1,14 @@
 all_sources += files(
   'SDL_hidapi_gamecube.c',
+  'SDL_hidapi_luna.c',
   'SDL_hidapi_ps4.c',
+  'SDL_hidapi_ps5.c',
   'SDL_hidapi_rumble.c',
   'SDL_hidapi_steam.c',
+  'SDL_hidapi_stadia.c',
   'SDL_hidapi_switch.c',
   'SDL_hidapi_xbox360.c',
+  'SDL_hidapi_xbox360w.c',
   'SDL_hidapi_xboxone.c',
   'SDL_hidapijoystick.c',
 )
\ No newline at end of file
diff --git a/src/joystick/meson.build b/src/joystick/meson.build
index 38a1d22..a56cf46 100644
--- a/src/joystick/meson.build
+++ b/src/joystick/meson.build
@@ -5,6 +5,7 @@ all_sources += files(

 subdir('hidapi')
 subdir('steam')
+subdir('virtual')

 if host_machine.system() == 'linux'
   subdir('linux')
diff --git a/src/joystick/windows/meson.build b/src/joystick/windows/meson.build
index f3fdeb8..581a618 100644
--- a/src/joystick/windows/meson.build
+++ b/src/joystick/windows/meson.build
@@ -1,5 +1,9 @@
 all_sources += files(
   'SDL_dinputjoystick.c',
+  'SDL_rawinputjoystick.c',
+  'SDL_windows_gaming_input.c',
   'SDL_windowsjoystick.c',
   'SDL_xinputjoystick.c'
 )
+
+core_deps += [cc.find_library('cfgmgr32')]
diff --git a/src/meson.build b/src/meson.build
index 06f2a03..c23c313 100644
--- a/src/meson.build
+++ b/src/meson.build
@@ -27,7 +27,7 @@ subdir('power')
 subdir('render')
 subdir('sensor')
 subdir('stdlib')
-subdir('test')
+#subdir('test')
 subdir('thread')
 subdir('timer')
 subdir('video')
diff --git a/src/render/direct3d11/meson.build b/src/render/direct3d11/meson.build
index f6cf157..ca15625 100644
--- a/src/render/direct3d11/meson.build
+++ b/src/render/direct3d11/meson.build
@@ -1,3 +1,4 @@
 all_sources += files(
   'SDL_render_d3d11.c',
+  'SDL_shaders_d3d11.c',
 )
diff --git a/src/sensor/meson.build b/src/sensor/meson.build
index 8d419f7..db2cd08 100644
--- a/src/sensor/meson.build
+++ b/src/sensor/meson.build
@@ -2,4 +2,8 @@ all_sources += files(
   'SDL_sensor.c',
 )

+  subdir('windows')
+endif
+
 subdir('dummy')
diff --git a/src/test/meson.build b/src/test/meson.build
index 471a980..4649f68 100644
--- a/src/test/meson.build
+++ b/src/test/meson.build
@@ -1,3 +1,4 @@
+# FIXME, these should be executables instead.
 all_sources += files(
   'SDL_test_assert.c',
   'SDL_test_common.c',
diff --git a/src/thread/windows/meson.build b/src/thread/windows/meson.build
index f978eb3..a7c854b 100644
--- a/src/thread/windows/meson.build
+++ b/src/thread/windows/meson.build
@@ -1,7 +1,8 @@
 all_sources += files(
+  'SDL_syscond_cv.c',
+  'SDL_sysmutex.c',
+  'SDL_syssem.c',
   'SDL_systhread.c',
   'SDL_systls.c',
-  'SDL_syssem.c',
-  'SDL_sysmutex.c',
   '../generic/SDL_syscond.c',
 )
```

Jan200101:
Okay that seems to get it further, but I am still facing issues when compiling with MinGW.
After looking further it seems that it uses a preconfigured SDL_config.h which defines things that are not available on MinGW.

Jan200101:
Amazingly Windows compiles now, but most of the features that were previously present had to be cut out.

I know stuff is messy right now and there needs to be a lot of changing to get it back to the previous state, but then it will work on MSVC and MinGW

jpakkane:
> Amazingly Windows compiles now, but most of the features that were previously present had to be cut out.

Why is this? The version I did did not require removing anything.

Jan200101:
> > Amazingly Windows compiles now, but most of the features that were previously present had to be cut out.
> 
> Why is this? The version I did did not require removing anything.

The version you provided may have compiled against MSVC but not against MinGW because it lacks headers from the WinSDK.

The issue is that SDL2 was configured to use the official SDL2_config_windows.h when we really needed something to give us more control on what to enable.

I will have to manually go through the list of definitions and add them to the check_functions and check_headers arrays

jpakkane:
Since the original did not support MinGW either (I think, didn't test it though), if doing all that work takes a lot of time, an alternative is to submit the current version that works with VS as-is and then add the MinGW bits later in a separate pr.

Jan200101:
The previous one did support MinGW which is why I aim to continue support.

Its not going to take a very long time its just a matter of matching up headers with flags

eli-schwartz:
As a completely random aside: what are the advantages of having one meson.build per subdirectory, versus listing most sources in one file using filenames with directory components. I am more familiar with the latter, and it should be just as functional... Is it actually easier to write the current way? 👀

jpakkane:
In the end it is a matter of taste, really.

That being said, this approach is better if you need to be selective in things you pick up from the subdirs. Having the check on whether to use something be in the same dir as the file itself (as opposed to some random dir uptree) is easier.

Jan200101:
> As a completely random aside: what are the advantages of having one meson.build per subdirectory, versus listing most sources in one file using filenames with directory components. I am more familiar with the latter, and it should be just as functional... Is it actually easier to write the current way? eyes

It... depends

Honestly I am considering rewriting it as a single meson.build to achieve parity with the official CMakelists.txt because they simply wildcard a bunch of files
https://github.com/libsdl-org/SDL/blob/c477768e6f926c38d4cdc236cab8376faf9f6789/CMakeLists.txt#L462-L483

Jan200101:
Ubuntu CI seemingly randomly failed. I'll force repush in a  bit to try again

Jan200101:
CI passes and tested locally using MinGW

jpakkane:
Great, thanks.

PR #259, opened 2022-01-04
pcre2: increase the stack size for test builds

Jan200101:
the reason tests on Windows were failing was because the program was running out of stack space while expanding a regex operation.

### DIFF
--- a/subprojects/packagefiles/pcre2/meson.build
+++ b/subprojects/packagefiles/pcre2/meson.build
@@ -68,7 +68,7 @@ check_headers = [
 config_h_defs = []
 foreach h : check_headers
   if c_compiler.has_header(h)
-    config_h_defs += ['-DHAVE_' + h.underscorify().to_upper(]
+    config_h_defs += ['-DHAVE_' + h.underscorify().to_upper()]
   endif
 endforeach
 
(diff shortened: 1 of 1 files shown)


Jan200101:
upstream issue https://github.com/PhilipHazel/pcre2/issues/65

jpakkane:
You need to add a new release to the main json file.

Jan200101:
I figured this was not needed since I assumed the pcre2 branch was mainly for trying to get it updated

subprojects/packagefiles/pcre2/meson.build
```diff
@@ -147,10 +147,16 @@ pkg.generate(libraries : pcre2_32_lib,
 #### tests
 
 if not meson.is_cross_build() # wine wrappers are not bat-friendly, this would need better testing
+  link_args = []
+  if host_machine.system() == 'windows'
+    link_args += ['/STACK:2500000']
```
eli-schwartz, line 152: This will technically work, even though it sets `link_args = [ ['/STACK:2500000'] ]`, because it will be automatically flattened to a one-dimensional array in the process of being consumed by `executable()`. However, it's unnecessary. :D

eli-schwartz:
Adding a new release is not necessary -- because the primary release on that branch is not yet released, so bumping it to -2 means we'd have a -1 that was never released.

And, thanks for looking into this!

eli-schwartz:
I'll rebase the pcre2 branch later today (and drop the debug commit) then merge this PR into it and release the combined update.

subprojects/packagefiles/pcre2/meson.build
Jan200101: was not aware of this, I don't really use meson in the past this seemed like the right behavior.

PR #265, opened 2022-01-07
Silence warnings on msvc

barcharcraz:
msvc emits more warnings than gcc at equivalent meson warning_levels, silence the ones coming from zlib.

(maybe we should just set the warning level to zero in the wrap instead ...)


jpakkane:

barcharcraz:
Alright I updated that file, and it looks like the rest of the instructions for contributions don't apply to simple updates.

eli-schwartz:
Really, the changes should be squashed into one commit...

Other than that it seems fine.

jpakkane:
I'll merge this with Github's squash thingy.

PR #270, opened 2022-01-09
sdl2: fix building on MacOS

Jan200101:
It was painful to figure this out and get it back working

Non Working build (before this PR)
https://github.com/lite-xl/lite-xl/runs/4753284675?check_suite_focus=true

Working build (after this PR
https://github.com/Jan200101/lite-xl/runs/4753459468?check_suite_focus=true


subprojects/packagefiles/sdl2/meson.build
```diff
@@ -184,6 +184,7 @@ check_functions = [
 # check token HAVE_SEM_TIMEDWAIT
   ['HAVE_INOTIFY_INIT', 'inotify_init', '#include<sys/inotify.h>'],
   ['HAVE_INOTIFY_INIT1', 'inotify_init1', '#include<sys/inotify.h>'],
+  ['HAVE_DLOPEN', 'dlopen', '#include<dlcfn.h>'],
```
Jan200101, line 187: All the dlopen logic came from https://github.com/libsdl-org/SDL/blob/7967c97618fba7326028005e72ead740fa6e6589/cmake/sdlchecks.cmake#L32

subprojects/packagefiles/sdl2/src/joystick/meson.build
```diff
@@ -13,6 +13,7 @@ elif host_machine.system() == 'windows'
   subdir('windows')
 elif host_machine.system() == 'darwin'
   subdir('darwin')
+  subdir('iphoneos')
```
Jan200101, line 16: despite being called `iphoneos` we still need it because it defines `SDL_IOS_JoystickDriver` which is used across MacOS and iOS

jpakkane:
Nice, thanks.

PR #271, opened 2022-01-09
Add libsxlt.

jpakkane:


eli-schwartz:
https://gitlab.gnome.org/GNOME/libxslt/-/merge_requests/14

eli-schwartz:
(sorry, misclick)

Instead of trying to fix defines piecemeal, what is wrong with the upstream MR I linked? Can we try to use that instead? Especially given, it actually implements options and stuff :)

jpakkane:
I'm doing this because:

- that MR is over a year old with no recent action
- I need this now
- This works for me in an LO build

People are welcome to submit better versions (including the one stuck upstream) once this is in.

eli-schwartz:
I disagree that we should merge this PR, without first at least testing whether or not the upstream MR serves as a better, more feature-complete base.

If it is low effort to still merge a better solution, we should not merge a less feature complete one without even *trying* the other one. It cannot possibly take long to check.

eli-schwartz:
4d15bf1b8186e50f2344c0fe78eb527e628e7770 seems to basically work, except that it exposes a bug in the libxml2 port...

jpakkane:
If you add my metadata fixes to that, then sure, we can merge that instead.

jpakkane:
Actually I already did that.

eli-schwartz:
Which metadata fixes do you mean?

(My branch does have a fix to libxml2 as well.)

PR #272, opened 2022-01-11
nonstd-expected-lite: add overrides

dcbaker:
So that explicit fallbacks are not required

### DIFF
--- a/subprojects/packagefiles/nonstd-expected-lite/meson.build
+++ b/subprojects/packagefiles/nonstd-expected-lite/meson.build
@@ -3,3 +3,7 @@ project('nonstd-expected-lite', 'cpp', version: '0.3.0', license: 'BSL-1.0')
 includes = include_directories('include')
 
 nonstd_expected_lite_dep = declare_dependency(include_directories: includes)
+
+if meson.version().version_compare('>= 0.54.0')
+  meson.override_dependency('expected-lite', nonstd_expected_lite_dep)
+endif
--- a/releases.json
+++ b/releases.json
@@ -861,7 +861,11 @@
     ]
   },
   "nonstd-expected-lite": {
+    "dependency_names": [
+      "expected-lite"
+    ],
     "versions": [
+      "0.3.0-2",
       "0.3.0-1"
     ]
   },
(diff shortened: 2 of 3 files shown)


eli-schwartz:
Upstream provides cmake config files. They misuse and abuse variables, but it looks like the expected use is `dependency('expected-lite', method: 'cmake')` and as such the provide section in this PR is wrong.

subprojects/packagefiles/nonstd-expected-lite/meson.build
```diff
@@ -3,3 +3,5 @@ project('nonstd-expected-lite', 'cpp', version: '0.3.0', license: 'BSL-1.0')
 includes = include_directories('include')
 
 nonstd_expected_lite_dep = declare_dependency(include_directories: includes)
+
+meson.override_dependency('nonstd-expected-lite', nonstd_expected_lite_dep)
```
eli-schwartz, line 7: Overrides are not critical functionality for something with a primary use of "use a provide section in a wrap file" and cause this wrap to suddenly require meson 0.54, which is not reflected in the project+) meson_version.

Meson doesn't warn about this for projects with no specific minimum, even though in cases like this it is effectively a semantic "minimum is at least as far back as 0.3x and probably even further back but how do we actually test it".

eli-schwartz (changes requested):

subprojects/packagefiles/nonstd-expected-lite/meson.build
dcbaker: I've changed the minimum to 0.54.0, it's pretty old at this point so it doesn't seem harmful. But I can change that if you'd prefer.

eli-schwartz: I prefer to be conservative just in case. And 0.54 may be pretty old, but it's still too for the latest Ubuntu LTS, which only has 0.53.

In this case, the feature basically doesn't have any utility. It exists for the sake of projects that provide multiple dependencies, when you want to force meson to be consistent about not picking up system deps in another subproject after evaluating this meson.build, because linking to multiple versions of a shared library is painful disaster. But none of that is relevant to a project which provides *one* exported dependency and it's a header-only dependency at that.

I have no problem using override_dependency anyway... as long as some other feature which has practical use already requires meson 0.54, because if the feature is needed that trumps "doesn't work on Ubuntu 20.04 LTS".

dcbaker: okay, I've wrapped the override_dependency call inside a version check

eli-schwartz: Thanks for humoring me. :)

PR #276, opened 2022-01-13
Add support for ludocode-mpack v1.1.

schmidtw:


releases.json
```diff
@@ -759,6 +759,7 @@
   "ludocode-mpack": {
     "versions": [
       "1.0-1"
+      "1.1-1"
```
jpakkane, line 762: The newest one should be at the top, and this is missing a comma between entries.

subprojects/ludocode-mpack.wrap
```diff
@@ -1,7 +1,7 @@
 [wrap-file]
-directory = mpack-1.0
-source_url = https://github.com/ludocode/mpack/archive/v1.0.tar.gz
-source_filename = ludocode-mpack-1.0.tar.gz
-source_hash = f80fc0539554dc58eaa22f6260e91709366475b01b2d004372e98a59d08388ac
+directory = mpack-1.1
+source_url = https://github.com/ludocode/mpack/archive/v1.1.tar.gz
+source_filename = ludocode-mpack-1.1.tar.gz
+source_hash = a62a3954dbbd77cf7f3048e8d7e89718d45a851d55003c37c51277c71ca968e6
 patch_directory = ludocode-mpack
```
jpakkane, line 6: You'll need to add a `[provide]` entry to this wrap. It was not a requirement when the original was created, but the the rules have changed since then.

releases.json
schmidtw: Thanks for pointing this out - I forgot it was a json file & not yaml so it "looked" right.  The error message in the tools/sanity_checks.py when I ran it locally looked like it was a python issue vs. an actual problem with my code.  I'm working on fixing this and the provide.

eli-schwartz: If you'd like to add a commit that updates `tools/sanity_checks.py` to wrap that json.load in a try/except block and have it raise a slightly more informative error -- say, for example, "file 'releases.json' is malformed", then people would not make that mistake in the future, and future you might thank present you the next time you submit a wrap update. :D

PR #277, opened 2022-01-13
imgui: Update to 1.86

stephanlachnit:
Update to 1.86

Removed support for Marmalade (upstream removed it).


stephanlachnit:
Note: the emscripten option does not work. While I found some stuff on the internet on how to use emscripten with Meson, but didn't get it to work. If you have any idea, please let me know.

PR #278, opened 2022-01-13
Cleanup PCRE Windows native build

neodesys:
Add flags for Microsoft compiler to not spam some non-vital level-4
warnings during compilation.


subprojects/packagefiles/pcre/meson.build
```diff
@@ -2,6 +2,13 @@ project('pcre', 'c')
 
 cc = meson.get_compiler('c')
 
+dftables_c_args = ['-DHAVE_CONFIG_H']
+if cc.get_id() == 'msvc'
+  dftables_c_args += ['-D_CRT_SECURE_NO_WARNINGS', '/wd4244']
+  add_project_arguments('-D_CRT_SECURE_NO_WARNINGS',
+    '/wd4244', '/wd4701', '/wd4703', '/wd4267', language : 'c')
```
eli-schwartz, line 9: Why are you adding these both to dftables_c_args and to the project-wide arguments?

neodesys: When setting only the flags to the project with add_project_arguments(), they are not taken into account when building the dftables executable.

Warnings are correctly hidden for building the rest of the targets but at the moment of building the dftables executable msvc is called without the corresponding flags.

Setting explicitly the c_args on the target solves the issue. I suppose that the default behavior is changed by having the target with 'native: true'. Unless this may be a bug with meson, I don't know. To be honest my knowledge of meson is very light :smile:

neodesys: I've rewritten it in a more elegant way following the same style I've seen in zlib wrapper.
It is also more complete as I've removed also the warnings for the other compilers (MinGW/gcc) and fixed a dependency issue when building with -Dshared_lib=true.

eli-schwartz: Due to the whole native/non-native thing, I think this might be a bigger issue than currently expected.

You're checking for supported arguments using `meson.get_compiler('c')` (defaulting to `native: false`) and giving those arguments to the `native: true` target. I think maybe you want to check for the available options twice.

neodesys: You're right 👍 I didn't pay enough attention to this detail. This is now fixed.

jpakkane:
You need to add a new release to the json file.

jpakkane:
That's strange. Try rebasing against current master, hopefully it will fix this.

neodesys:
👍 rebased, hope it will work 😄

PR #280, opened 2022-01-14
taglib: add

neheb:

### DIFF
--- a/subprojects/packagefiles/taglib/meson.build
+++ b/subprojects/packagefiles/taglib/meson.build
@@ -1,0 +1,311 @@
+project('taglib', 'cpp',
+  version : '1.12',
+  meson_version : '>=0.46.0',
+  default_options : ['warning_level=0', 'cpp_std=c++98']
+)
+
+cpp = meson.get_compiler('cpp')
+
+add_project_arguments('-DHAVE_CONFIG_H', language : 'cpp')
+if get_option('default_library') == 'static'
+  add_project_arguments('-DMAKE_TAGLIB_LIB', language : 'cpp')
+else
+  add_project_arguments('-DTAGLIB_STATIC', language : 'cpp')
+endif
+
+cdata = configuration_data()
+if cpp.compiles('int main() { int i, x = 0; i = __sync_add_and_fetch(&x,1); return x; }', name : 'atomic builtins')
+  cdata.set('HAVE_GCC_ATOMIC', 1)
+elif cpp.compiles('#include <libkern/OSAtomic.h> int main() { int i, x = 0; i = OSAtomicIncrement32Barrier(&x); return x; }', name : 'macOS atomic')
+  cdata.set('HAVE_MAC_ATOMIC', 1)
+elif cpp.compiles('#include <windows.h> int main() { LONG i, x = 0; i = InterlockedIncrement(&x); return x; }', name : 'Windows atomic')
+  cdata.set('HAVE_WIN_ATOMIC', 1)
+endif
+
+if cpp.compiles('int main() { __builtin_bswap64(0); return 0; }', name : 'gcc builtin byteswap')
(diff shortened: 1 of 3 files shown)


neheb:
there's one huge problem with this PR as is. I cannot use it as a subproject:

```
../src/metadata/taglib_handler.h:37:10: fatal error: tbytevector.h: No such file or directory
   37 | #include <tbytevector.h>
      |          ^~~~~~~~~~~~~~~
```

I can't figure out why. the pkgconfig file has the proper path.

neheb:
Ah got it. Fixed now.

subprojects/packagefiles/taglib/meson.build
```diff
@@ -0,0 +1,316 @@
...
+  'taglib/riff/aiff',
+  'taglib/riff/wav',
+  'taglib/s3m',
+  'taglib/toolkit',
+  'taglib/trueaudio',
+  'taglib/wavpack',
+  'taglib/xm',
+]
+
+taglib = library('tag',
```
eli-schwartz, line 200: Upstream installs this with a version of `1.18.0` (soname defaulting to `1`).

subprojects/packagefiles/taglib/meson.build
```diff
@@ -0,0 +1,316 @@
...
+  'taglib/wavpack/wavpackproperties.cpp',
+  'taglib/xm/xmfile.cpp',
+  'taglib/xm/xmproperties.cpp',
+  include_directories : taglib_includes,
+  install : true
+)
+
+pkg = import('pkgconfig')
+pkg.generate(taglib,
+  subdirs : 'taglib'
```
eli-schwartz, line 309: This produces `tag.pc`, which doesn't match `taglib`. You need to set the `filebase: 'taglib'` since the default name matching the library name is wrong.

eli-schwartz (commented):
For some reason upstream also installs:
- /usr/include/taglib/tlist.tcc
- /usr/include/taglib/tmap.tcc

Also, there is a bindings/c/ directory which may be useful, it provides:
- /usr/include/taglib/tag_c.h
- /usr/lib/libtag_c.so.0.0.0
- /usr/lib/pkgconfig/taglib_c.pc

neheb:
Updated.

subprojects/packagefiles/taglib/meson.build
```diff
@@ -0,0 +1,316 @@
...
+  'taglib/ogg/speex/speexproperties.h',
+  'taglib/ogg/vorbis/vorbisfile.h',
+  'taglib/ogg/vorbis/vorbisproperties.h',
+  'taglib/ogg/oggfile.h',
+  'taglib/ogg/oggpage.h',
+  'taglib/ogg/oggpageheader.h',
+  'taglib/ogg/xiphcomment.h',
+  'taglib/riff/aiff/aifffile.h',
+  'taglib/riff/aiff/aiffproperties.h',
+  'taglib/riff/riffutils.h',
```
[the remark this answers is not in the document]
eli-schwartz: You removed rifffile instead of riffutils. :)

rifffile is/should be installed.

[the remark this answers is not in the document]
neheb: whoops.

neheb:
I think this is about right.

Those byteswap checks could be simplified. I would think only the builtin byteswap and MSC ones are used.

subprojects/packagefiles/taglib/meson.build
```diff
@@ -0,0 +1,310 @@
+project('taglib', 'cpp',
+  version : '1.12',
```
eli-schwartz, line 2: You are using one feature that requires meson 0.46.0

Because I am a nitpicker (sorry) I actually think it would be neat if we can use `meson_version:` here to declare that.

neheb: I think this should do it.

PR #283, opened 2022-01-17
Add graphite2.

jpakkane:


subprojects/packagefiles/graphite2/meson.build
```diff
@@ -0,0 +1,49 @@
...
+  'src/gr_face.cpp',
+  'src/gr_features.cpp',
+  'src/gr_font.cpp',
+  'src/gr_logging.cpp',
+  'src/gr_segment.cpp',
+  'src/gr_slot.cpp',
+  'src/json.cpp',
+  include_directories: ['include', 'src'],
+  cpp_args: extra_args,
+  gnu_symbol_visibility: 'hidden',
```
eli-schwartz, line 45: This really should have set the library version to 3.2.1

Also I would have liked to see the tests/ directory get built and tested. :/

PR #285, opened 2022-01-17
Add Harfbuzz.

jpakkane:


ci_config.json
```diff
@@ -128,5 +128,13 @@
       "libwayland-bin",
       "libwayland-dev"
     ]
+  },
+  "harfbuzz": {
+    "linux_only": true,
+    "debian_packages": [
+      "ragel",
```
eli-schwartz, line 135: I was wondering why we needed this!

Then I realized:

```
warning('You have to install ragel if you are going to develop HarfBuzz itself')
```
Ewwww. That's not what a warning is for.

ci_config.json
```diff
@@ -128,5 +128,13 @@
       "libwayland-bin",
       "libwayland-dev"
     ]
+  },
+  "harfbuzz": {
+    "linux_only": true,
```
eli-schwartz, line 133: json, what a beautiful language. commas are errors after one item, but required between two items.

eli-schwartz (changes requested):
Windows is supported upstream, so this should not be linux_only.

The silly `warning()` when ragel isn't available to "Develop harfbuzz" can be solved on Windows by passing build_options of `"harfbuzz:ragel_subproject=true"`

jpakkane:
For a truly proper solution, Ragel upstream should add an `override_find_program` and then it could be put in WrapDB itself with proper dependencies and all that. It would simplify Harfbuzz upstream build files, too.

But on the other hand wraps are not really meant for development, only for consuming. Needing to build a tool to generate sources which are already there is useless extra work in this particular case.

eli-schwartz:
Yes, but the WrapDB CI at least can override this pointless option just so we can get Windows testing to complete.

I submitted a ticket to add an option that suppresses the pointless warning, with a suggestion that it shouldn't be the default behavior at all.

It's completely unnecessary, because releases have the generated files anyway...

eli-schwartz:
> For a truly proper solution, Ragel upstream

Harfbuzz wrote their own meson.build patch for Ragel anyway, so this would be trivial to solve there.

subprojects/harfbuzz.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = harfbuzz-3.2.0
+source_url = https://github.com/harfbuzz/harfbuzz/releases/download/3.2.0/harfbuzz-3.2.0.tar.xz
```
xclaesse, line 3: Could be updated to 4.3.0.

subprojects/harfbuzz.wrap
```diff
@@ -0,0 +1,9 @@
...
+source_filename = harfbuzz-3.2.0.tar.xz
+source_hash = 0ada50a1c199bb6f70843ab893c55867743a443b84d087d54df08ad883ebc2cd
+
+[provide]
+harfbuzz = libharfbuzz_dep
+harfbuzz-icu = libharfbuzz_icu_dep
```
xclaesse, line 9: There are a few more:
harfbuzz = libharfbuzz_dep
harfbuzz-subset = libharfbuzz_subset_dep
harfbuzz-icu = libharfbuzz_icu_dep
harfbuzz-gobject = libharfbuzz_gobject_dep

ci_config.json
```diff
@@ -105,6 +105,21 @@
...
   },
+  "harfbuzz": {
+    "_comment": ["- Currently builds with no warnings other than an incorrect warning() function",
+                 "  tracked at https://github.com/harfbuzz/harfbuzz/issues/3378",
+                 "  Absolutely impossible to fix on Windows."],
+    "fatal_warnings": false,
+    "build_options": [
+      "glib:tests=false",
+      "cairo:tests=disabled",
+      "pixman:tests=disabled"
```
eli-schwartz, line 116: Since commit 6f7eac16c18d8fbf088ded8015d3e6da314cdd22 it should not be necessary to manually suppress tests for other subprojects anymore.

ci_config.json
xclaesse: That should be fixed in next harfbuzz release: https://github.com/harfbuzz/harfbuzz/pull/3617.

xclaesse:
Harfbuzz tests are failing when run as subproject, made a fix upstream: https://github.com/harfbuzz/harfbuzz/pull/3630.

tristan957:
Seems like this release has the fix @xclaesse wrote: https://github.com/harfbuzz/harfbuzz/releases/tag/4.4.1

neheb:
```
2022-07-22T02:28:01.2746320Z subprojects/glib-2.72.2/glib/meson.build:406:8: ERROR: Tried to override dependency 'glib-2.0' which has already been resolved or overridden at /Users/runner/work/wrapdb/wrapdb/subprojects/harfbuzz-4.4.1/meson.build:91:
```

eli-schwartz:
The macOS CI finds glib itself as a system dep, but cannot find the mkenums program and therefore falls back to building it via the subproject, resulting in two versions of glib that conflict each other. Weird.

eli-schwartz:
https://github.com/Homebrew/homebrew-core/blob/bb1d0720ea1c3e805b5d8e2375131cb77715ee4e/Formula/glib.rb#L112-L117

```ruby
def caveats
  <<~EOS
    Python executables and libraries are no longer included with this formula, but they are available separately:
      brew install glib-utils
  EOS
end
```

PR #286, opened 2022-01-18
Clarify json error

schmidtw:
@eli-schwartz I put this small change together in the spirit of trying to make the error I ran into more clear.  I'm not sure if I made it better or simply different.  I am interested in your thoughts.


tools/sanity_checks.py
```diff
@@ -53,10 +53,17 @@ def setUpClass(cls):
...
+            fn = 'releases.json'
+            with open(fn, 'r') as f:
+                cls.releases = json.load(f)
+            fn = 'ci_config.json'
+            with open(fn, 'r') as f:
+                cls.ci_config = json.load(f)
+        except json.decoder.JSONDecodeError as err:
+            what = "Error in file '%s': %s" % (fn, err)
+        finally:
+            raise RuntimeError(what)
```
dcbaker, line 66: this is going to result in `RuntimeError` *always* being raised, which doesn't seem like what you want. I think you meant
```suggestion
        except json.decoder.JSONDecodeError as err:
            raise RuntimeError(f'{Error in file "{fn)": {err}') from err
```

Which will throw a new exception from the old one, and otherwise pass. A `finally` block is guaranteed to happen regardless of whether there is an exception, and are useful for cases like:
```python
state = make_state()
try:
    state.do_thing()
finally:
    state.cleanup()
```

schmidtw: Thanks for this catch.  I'm normally not developing in python so I'm quite rusty.

schmidtw:
The output for this change looks like this if there is a failure due to JSON decode error:

```
ERROR: setUpClass (__main__.TestReleases)
  File "/home/schmidtw/sync/work/xmidt/all/wrapdb/./tools/sanity_checks.py", line 63, in setUpClass
    cls.ci_config = json.load(f)
  File "/usr/lib64/python3.10/json/__init__.py", line 293, in load
    return loads(fp.read(),
  File "/usr/lib64/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.10/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 2 column 1 (char 2)

The above exception was the direct cause of the following exception:

  File "/home/schmidtw/sync/work/xmidt/all/wrapdb/./tools/sanity_checks.py", line 66, in setUpClass
    raise RuntimeError(f'Error in file "{fn}": {err}') from err
RuntimeError: Error in file "ci_config.json": Expecting property name enclosed in double quotes: line 2 column 1 (char 2)
```

This code adds the last bit of description.  Given the unit test structure available in the function where this error occurs, I'm not sure there is a better way to output errors.  I'm also not sure how valuable the 1 extra line is.  Thoughts?

eli-schwartz:
Repeating the internal error description of the json decoder doesn't seem particularly useful, what about instead just raising `RuntimeError(f'file {fn} is malformed')`?

eli-schwartz:
```
ERROR: setUpClass (__main__.TestReleases)
  File "/home/eschwartz/git/mesonbuild/wrapdb/./tools/sanity_checks.py", line 59, in setUpClass
    cls.releases = json.load(f)
  File "/usr/lib/python3.10/json/__init__.py", line 293, in load
    return loads(fp.read(),
  File "/usr/lib/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 245 column 7 (char 3586)

During handling of the above exception, another exception occurred:

  File "/home/eschwartz/git/mesonbuild/wrapdb/./tools/sanity_checks.py", line 64, in setUpClass
    raise RuntimeError(f'file {fn} is malformed')
RuntimeError: file releases.json is malformed

```

As far as UI goes, this looks fine to me.

The PR needs to be rebased on git master, though, to get rid of all the merge commits and suchlike. Please make sure there is one discrete commit implementing the change.

PR #287, opened 2022-01-19
lmdb: add 0.9.29

git-bruh:


git-bruh:
ci should be fixed now, the code is in `libraries/liblmdb` so i thought adding it to `directory` would `cd` into that before executing meson

eli-schwartz:
A better idea would be to use a proper release tarball rather than a tarball of something else which then vendors a copy of this.

git-bruh:
No, the tarball is released in this manner only. It doesn't have anything except lmdb but they prefer to keep it inside subdirs for some reason

eli-schwartz:
> for some reason

There is not and cannot be a reason. It's a bug in their project management.

Unfortunately it's also a bug in your meson.build file, because your declare_dependency is now wrong.

subprojects/packagefiles/lmdb/meson.build
```diff
@@ -0,0 +1,29 @@
...
+cc = meson.get_compiler('c')
+
+src = [
+  'libraries/liblmdb/mdb.c',
+  'libraries/liblmdb/midl.c',
+]
+
+c_args = [
+  '-W',
+  '-Wall',
```
eli-schwartz, line 12: What is this for?

Note that -Wall specifically should be set by project default_options specifying a meson `warning_level`.

subprojects/packagefiles/lmdb/meson.build
```diff
@@ -0,0 +1,29 @@
...
+  '-Wno-unused-parameter',
+  '-Wbad-function-cast',
+  '-Wuninitialized'
+]
+
+pthread_dep = cc.find_library('pthread', required: true)
+incdir = include_directories('.')
```
eli-schwartz, line 19: This no longer adds libraries/liblmdb/.

subprojects/packagefiles/lmdb/meson.build
```diff
@@ -0,0 +1,29 @@
...
+
+lmdb = library(
+  'lmdb',
+  src,
+  c_args: c_args,
+  dependencies: pthread_dep,
+  include_directories: incdir,
```
eli-schwartz, line 26: You don't ever need to add `'.'` as an include directories, since meson does that by default.

You also do not need to add any include directory at all, for the compiler to find a header in the same directory as the source file itself is located in.

subprojects/packagefiles/lmdb/meson.build
```diff
@@ -0,0 +1,29 @@
...
+
+c_args = [
+  '-W',
+  '-Wall',
+  '-Wno-unused-parameter',
+  '-Wbad-function-cast',
+  '-Wuninitialized'
+]
+
+pthread_dep = cc.find_library('pthread', required: true)
```
eli-schwartz, line 18: See https://mesonbuild.com/FAQ.html#what-is-the-correct-way-to-use-threads-such-as-pthreads

eli-schwartz (changes requested):

eli-schwartz:
This should also create the mtest executable and launch it under meson's test framework. Maybe the mtestX programs too (not sure why the upstream Makefile doesn't run those).
A full fledged port would presumably mean building the main executables too, and installing the man pages.

subprojects/packagefiles/lmdb/tests/meson.build
```diff
@@ -0,0 +1,16 @@
...
+  'mtest5'
+]
+
+rm = find_program('rm')
+mkdir = find_program('mkdir')
+
+foreach t : tests
+  exe = executable(t, '../libraries/liblmdb/@0@.c'.format(t),
+    dependencies: [lmdb_dep])
+  test(t, exe)
```
git-bruh, line 15: The tests need a `testdb` directory created before they are run, so the flow goes like `mkdir testdb; ./test; rm -rf testdb` not able to find how to achieve this in meson

eli-schwartz: The test can be something like (untested) `'python', '-c', import sys, os, subprocess; os.makedirs("testdb", exist_ok=True); subprocess.check_call([sys.argv[1], "testdb")', exe`

subprojects/packagefiles/lmdb/meson.build
```diff
@@ -0,0 +1,42 @@
...
+  c_args: c_args,
+  dependencies: threads_dep,
+)
+
+lmdb_dep = declare_dependency(include_directories: incdir, link_with: lmdb)
+install_man(docs)
+
+if get_option('programs')
+  subdir('programs')
+endif
```
eli-schwartz, line 38: If you are going to make an option to install the programs, then the man pages for those programs should only be installed when the programs are.

subprojects/packagefiles/lmdb/meson.build
```diff
@@ -0,0 +1,42 @@
...
+
+if get_option('tests')
+  subdir('tests')
+endif
```
eli-schwartz, line 42: If you are going to make an option to build and run the tests (note that running tests from another project can exclude / include tests by requesting a suite) then the WrapDB CI must set the option to true (in ci_config.json).

eli-schwartz (commented):
The constant use of source paths that always go "../libraries/liblmdb/" feels awkward and makes me wonder why you don't just descend into that directory and do all work there.

subprojects/packagefiles/lmdb/tests/meson.build
git-bruh: Did it with `sh` since it's easier to do stuff. Also `rm -rf` is run each time since lmdb tests expect a clean database.

git-bruh:
> descend into that directory

Not sure how to do this, also it would again require constant changing back and forth between dirs  (If the parent meson.build changes the lookup path of subdirs aswell)

All other feedback should have been addressed. Also added installation of library and headers which i forgot.

git-bruh:
All the subdir() calls are useless for this project, will get rid of them.

git-bruh:
Added `program_names` to the `lmdb.wrap` file and fixed the directory.

jpakkane:
The CI failure should go away if you rebase.

git-bruh:
Thanks. I also forgot to add the programs to releases.json which causes another failure, will fix it.

git-bruh:
Should be fixed now, ran the sanity checks locally.

jpakkane:
cl : Command line error D8021 : invalid numeric argument '/Wno-unused-parameter'

subprojects/packagefiles/lmdb/libraries/liblmdb/meson.build
```diff
@@ -0,0 +1,68 @@
+cc = meson.get_compiler('c')
+
+src = ['mdb.c', 'midl.c']
+
+c_args = ['-Wno-unused-parameter', '-Wbad-function-cast', '-Wuninitialized']
```
eli-schwartz, line 5: Maybe instead of adding this manually to every target, you should just use `add_project_arguments`.

While you are at it, you can fix the MSVC error by using [`cc.get_supported_arguments()`](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerget_supported_arguments) to only pass those warning flags if the compiler supports them.

eli-schwartz (changes requested):

subprojects/packagefiles/lmdb/libraries/liblmdb/meson.build
git-bruh: Fixed

eli-schwartz:
Okay, that's interesting. One of the programs is failing to compile this code:

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include "lmdb.h"

#ifdef	_WIN32
#define	Z	"I"
#else
#define	Z	"z"
#endif
```

... because unistd.h is a POSIX header and Windows doesn't include it, but apparently the code does try to support WIN32?

git-bruh:
The main library does support windows, should I just disable the programs if building for windows?

jpakkane:
Seems reasonable.Just add a comment above saying why they are disabled on Windows.

git-bruh:
Not sure what's wrong now.. Seems to be creating a .dll but the link needs a .lib

jpakkane:
The symbol failure is [probabably due to this issue](https://mesonbuild.com/FAQ.html#my-project-works-fine-on-linux-and-mingw-but-fails-to-link-with-msvc-due-to-a-missing-lib-file-fatal-error-lnk1181-why). If the project has the necessary macro definitions for symbol exporting then you need to set the correct `-D<something>` to enable them. If it does not have them, then it can't be built on Windows as a shared library and you need to add `--default-library=static` to the CI configuration file for this project.

eli-schwartz:
It might make more sense to use `build_target('lmdb', ..., target_type: host_machine.system() == 'windows ? 'static_library' : 'library')`

git-bruh:
Hopefully everything works now.. :smiling_face_with_tear:

eli-schwartz:
Yay!

PR #288, opened 2022-01-20
Add Orcus.

jpakkane:


subprojects/packagefiles/orcus/meson.build
```diff
@@ -0,0 +1,19 @@
...
+  default_options: 'cpp_std=c++17')
+
+varr = meson.project_version().split('.')
+version_major = varr[0]
+version_minor = varr[1]
+version_micro = varr[2]
+
+boost_dep = dependency('boost')
+boost_fs_dep = dependency('boost', modules: ['filesystem'])
+thread_depå = dependency('threads')
```
eli-schwartz: That is a fun typo!

subprojects/packagefiles/orcus/src/liborcus/meson.build
```diff
@@ -0,0 +1,107 @@
...
+cdata.set('ORCUS_MINOR_VERSION', version_minor)
+cdata.set('ORCUS_MICRO_VERSION', version_micro)
+
+
+configure_file(input: 'constants.inl.in',
+  output: 'constants.inl',
+  configuration: cdata)
+
+orcus_lib = library('orcus-0.17',
+  # core
```
eli-schwartz, line 12: These should probably all have `version: '0.0.0'` corresponding to upstream. Silly that they have a useless soname on top of having a version in the library name, but what can you do...

jpakkane: FWICT this library has exaclty one user, and that is Libreoffice. They use system libs on Linux so this would only really be used on Windows and Mac so this does not really matter.

PR #293, opened 2022-01-25
unity: add 2.5.2

git-bruh:


eli-schwartz:
It would also be nice if upstream installed a pkg-config file in addition to their current cmake-config file. :D

PR #296, opened 2022-01-27
Fix xslt dependency include directories.

jpakkane:
It seems that existing code uses `#include<libxslt/foo.h>` instead of just `#include<foo>`.


jpakkane:
It turns out I had an old broken version and the wrapdb one was alredy fixed. Closing.

PR #297, opened 2022-01-27
openssl: openssl.wrap does not use openssl_cli

sc108-lee:
install: true always trigger openssl(which is cli bin) installation
since "build_by_default: false" for that,
it cause error for "meson install" from main project that uses openssl.wrap for subproject

Signed-off-by: Steven Seungcheol Lee <sc108.lee@samsung.com>


jpakkane:

sc108-lee:
@jpakkane Thanks for checking, It was my first commit for this repo
so didn't notice that.
updated :)

eli-schwartz:
Obsoleted by #322

PR #304, opened 2022-02-03
Report the filename when json decoding fails

schmidtw:
Add a simple try/catch block that allows reporting the filename of the
JSON file that failed to be decoded.  This will hopefully help folks
find the root cause of the failure faster.


dcbaker (approved):

eli-schwartz:
Looks good, but I do wonder why it took 3 PRs...

PR #305, opened 2022-02-04
Adds hinnant-date 3.0.1 release

jeandet:


jpakkane:
You need to add a provide section. They were not needed when the original was added to wrapdb, but they are required now.

jeandet:
Not sure why it fails :/

jpakkane:
You need to edit the ci configuration. For Linux you need to add the `libcurl-dev` package and for Windows you need to set it to linux-only with a comment saying it needs a dependency not in wrapdb. Grep for `_comment` in the json file to find examples.

releases.json
```diff
@@ -432,6 +432,7 @@
   },
   "hinnant-date": {
```
jpakkane, line 433: You need to add `dependency_names` here. See one of the other entries.

eli-schwartz:
Curl is only needed when setting the tzdb to automatically download, so I would think a better idea than disabling Windows CI is to in fact configure it to use a manual tzdb in ci_config.json so it doesn't link to curl.

This *yet again* demonstrates why I do not like treating linux_only as "well I couldn't figure out how to get the CI working on Windows after taking a quick look at it". It is *meant* to be for software that is written to build exclusively against POSIX APIs.

jeandet:
@eli-schwartz, I tend to agree, but at some point we have to chose what is the purpose of this CI. Is this supposed to be an exhaustive test for wraps? Should we validate build options? If so, we are actually far from this. Most wraps are distributed with at best one build config tested (+looks like there is no OSX). 
Here if I understand correctly, build_options are for both linux and windows, that implies we wouldn't test this wrap with its default options on both linux and windows. I'm not sure this is something we want.

So I can set use_manual_tzdb and add [windowsZones.xml](https://raw.githubusercontent.com/unicode-org/cldr/master/common/supplemental/windowsZones.xml) for CI if that's OK.

Then, should we think about extending this wrap later to do this by default on Windows and ship the XML file with the wrap?

jpakkane:
The guidelines I've been using thus far have been:

- Having something working reasonably in WrapDB is better than not having anything at all
- A reasonable amount of effort should be spent on making the wraps work on all major platforms
- Iterative development is better than huge single chunks (they spread out the work to more people too, which is nice)

I.e. "It's better to have something usable for some people than nothing for everybody."

There is, of course, a the debate on how much is "a reasonable amount of effort", but that is a different discussion.

eli-schwartz:
> I tend to agree, but at some point we have to chose what is the purpose of this CI. Is this supposed to be an exhaustive test for wraps? Should we validate build options? If so, we are actually far from this. Most wraps are distributed with at best one build config tested (+looks like there is no OSX).

I think generally which options are enabled and what combinations isn't especially interesting. The main goal should be twofold:
- validate that linking actually works on Windows, which is actually known to be unreliable (thanks MSVC) and in general that build issues of the type "only works on the OS I originally tested" don't happen; these tend to be *very* apparent when they happen
- verify that a baseline configuration (or, better, one that includes all optional features) works, to get broad coverage

We don't necessarily need to be perfect. And trying to be perfect anyway will probably lead to a tremendous amount of 100% useless work (I don't believe there is value in adding a matrix of all possible options). All we can do is try to avoid common mistakes, and improve over time.

> I.e. "It's better to have something usable for some people than nothing for everybody."

I mostly agree with this philosophy, but would probably place somewhat higher priority on checking that it works on all major platforms, based on the rationale that while something usable for some people because it doesn't implement everything that other people need, is understandable... something that *seems* usable for some people, but stops working when you try using it on other OSes, is kind of a frustrating bug to hit.

So I would understand "as the wrap submitter/updater I don't believe it is feasible to test this in CI for Windows", but I would *personally* prefer it to only be a weapon of last resort, and I'm willing to devote effort into helping overcome that obstacle if the "Eli Schwartz yardstick of I think this could be gotten to work" qualifies.

I think this one here does. We can check that hinnant-date works in the manual tzdb case, which fulfills the use case of many people if not everyone, and we can check that it works on Windows (which is a popular use case).

eli-schwartz:
... Actually come to think of it, the current meson_options.txt is... inconvenient. We could be using a feature option for curl, we could default to saner options rather than making the logical choice on Linux be disabled, we could do configure-time checks for what upstream bizarrely delegates to a runtime check in discover_tz_dir, we could avoid having multiple options all do the same thing and having their *absence* enable the curl...

Honestly, this seems to be mostly copied semantics from the upstream CMakeLists.txt, which is *horrible* because the upstream CMakeLists.txt is horrible. :thinking: perhaps that should be fixed.

neheb:
https://github.com/mesonbuild/wrapdb/pull/397

eli-schwartz:
As far as I can tell, that PR is a duplicate of this one and doesn't change anything other than dropping it from broken_windows, which was never the problem with this PR.

jeandet:
@eli-schwartz @neheb, feel free to work on this, I won't have time to work on this anytime soon. Building stuff on windows is never fun and takes time.
Another issue that should be discussed is that, after all the TZ lib is is just an extension that isn't built by default with CMake:
https://github.com/HowardHinnant/date/blob/master/CMakeLists.txt#L37
Maybe TZ lib should be disabled by default first then investigating how to build it and get TZ data on window could be done later.

neheb:
Should this be closed?

PR #306, opened 2022-02-05
Add range-v3 provide section in the wrap file.

B4mbus:
Added a provide section for the range-v3 wrap file and dependency name in the releases file.


eli-schwartz:
Thanks, but in order to trigger a new release you will need to bump the version in `releases.json` as well.

(The part after the `-` is the wrapdb build number added on to the upstream stable tag.)

eli-schwartz:
Actually now might be a good time to bump it to the upstream 0.11.0 stable tag. You can do that in a second commit.

B4mbus:
@eli-schwartz it is done now. Please tell me know if I made any mistake or need to change something again, thanks.

eli-schwartz:
Perfect, thanks!

PR #309, opened 2022-02-16
HiGHS: add wrap

lukas-weber:
This is a wrap for version 1.2.0 of the HiGHS linear programming solver library. I did a manual port of the build system which builds the library and unit tests, but not the application, fortran or julia bindings.

I ran into one problem. Upstream uses CMake’s `generate_export_header()` to generate some macros in `libhighs_export.h`. As far as I can tell the contents of this are completely system independent and my work around was to include the output from CMake as a simple file in the patch. This causes errors in the wrapdb CI, so I wonder what I should do instead.


eli-schwartz:
`generate_export_header` is fun... my recollection is that the purpose of this is to support specifically MSVC. The problem with MSVC is that there are ***three*** different compiler attribute states you need depending on how the library is compiled and whether you are the consumer:
- when compiling a shared library, you have to mark all the function headers as one declspec
- when *linking* to a shared library, you have to edit the headers to use a different declspec
- when compiling or linking to a static library, you have to do something different?

Some people solve this problem by using compiler defines when compiling, and allowing the headers to operate normally by default via if/else.

Apparently def files are different and let you skip at least the first, but it's not clear whether you can skip the second.

The cmake generated header has this boilerplate, because it is "super hard" I suppose. But the boilerplate in the file you submitted in this PR is only for GCC, it uses visibility attributes. GCC is a sane compiler, you just use `__attribute__((visibility("default")))` regardless of whether you are building static or shared or linking to someone else's anything. It just works. It's one thing, and you just need to know whether the function is supposed to be public or private.

Sadly, this file won't work when building on Windows using MSVC, because apparently cmake optimizes the generated file to NOT be "completely system independent". It should check for the define `_MSC_VER` to see if it should use declspec, and do some complicated nonsense switching, and otherwise if `__GNUC__ >= 4` just unconditionally set the sane thing.

Really since this is code is so static once generated, it would behoove upstream to check that content into their public headers rather than have cmake do file IO for it and have the process be weirdly tied to the cmake build system.

subprojects/packagefiles/highs/check/meson.build
```diff
@@ -0,0 +1,77 @@
+
+catch_dep = declare_dependency(
+  include_directories : include_directories('../extern')
+)
```
eli-schwartz, line 4: We should ignore the vendored code and make use of dependency('catch2'). This is already in the WrapDB. If you are interested in submitting meson support upstream, then they could vendor catch2.wrap as well.

subprojects/packagefiles/highs/meson.build
```diff
@@ -0,0 +1,102 @@
...
+subdir('src')
+highs_deps = [dependency('threads')]
+if ipx_on
+  highs_deps += ipx_dep
+endif
+
+libhighs = library(
+  'highs', highs_sources,
+  dependencies : highs_deps,
+  include_directories : highs_include_directories)
```
eli-schwartz, line 94: I assume this is probably installable... ?

eli-schwartz (changes requested):
Have only taken a quick look at this, but only spotted two minor issues.

...

Note that releases.json should be updated to include information about the new wrap, and you should have https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section

subprojects/packagefiles/highs/check/meson.build
```diff
@@ -0,0 +1,77 @@
...
+
+test_sources = files([
```
tristan957, line 6: The array here is redundant.

tristan957: Same comment for include_directories() further down

subprojects/packagefiles/highs/meson.build
lukas-weber: What is the policy on installing things if they are a subproject? I remember there was once the problem that such libraries will be installed along with their consuming projects and then collide with system-wide installations of the library.

Is this no longer a problem or do I need to include some check?

eli-schwartz: A check would look something like

```
not meson.is_subproject() or get_option('default_library') == 'static'
```

i.e. only install if highs is being built standalone as its own project, OR if it's configured to build and link as a shared library which must be available at runtime.

Meson did recently add support for a manual `meson install` option that only installs runtime files -- i.e. shared libraries but not static libraries. This CANNOT be handled per-subproject but if the main project doesn't install any devel files anyway it doesn't matter. It's an imperfect solution but may be good enough for you. In that case, just install and don't bother with checks.

Which one do you prefer? My usual approach is to defer the decision to the people who are affected, since neither solution seems wrong to me.

xclaesse: This is controversial topic. I very strongly disagree with @eli-schwartz, IMHO there should be no special case for subproject, they should install everything just like any project. What user actually want is to not install devel files, such as static libraries, and that can be done using meson install --tags.

What @eli-schwartz suggest is completely inconsistent with every other meson projects.

xclaesse: > Which one do you prefer? My usual approach is to defer the decision to the people who are affected, since neither solution seems wrong to me.

The decision is not for the maintainer to make, it is fundamentally a user decision what they install.

lukas-weber: I do not have a strong opinion on this, so I will skip the check.

eli-schwartz: I have explained ad nauseam that users ***cannot*** decide, the existing options are insufficient and you're coming from the perspective of someone that primarily uses subprojects in the opposite way from what the wrapdb intends. So it's no surprise that you disagree with me.

I'm grateful that the situation isn't as bad as it used to be. But it's by no means perfect.

Your disagreement is duly noted. Meanwhile I will continue to prioritize the needs of actual users rather than validation of one's own proposed solution that is insufficient for user needs.

eli-schwartz: @lukas-weber great, that makes things wonderfully simple.

xclaesse: > Meanwhile I will continue to prioritize the needs of actual users rather than validation of one's own proposed solution that is insufficient for user needs.

I have absolutely have no idea at all what is missing for users, please provide suggestions in a separate issue.

subprojects/packagefiles/highs/check/meson.build
lukas-weber: I have a problem here now because the version they vendor shadows the dependency. I need to include the `extern` include dir also for other headers needed by the source and this bleeds into the test build.
…especially inconvenient because their old version of catch2 does not build on my system.

eli-schwartz: Those other bits of code in extern/ are also external projects -- you could try adding doctest and pdqsort to the wrapdb, but I have no idea what filereaderlp is, it's not googleable, and none of the files contain accreditation for the source location they are pulled from.

lukas-weber: I think filereaderlp is `https://github.com/feldmeier/FilereaderLP`.
But actually ipm, basislu and a bunch of other parts could also be interpreted as extern projects, but they are melded into the codebase with various kinds of glue. Maybe the vendored filereaderlp also has or will in the future have similar hacks in it.

As a third party build system, I would rather not dig deeply into these details. It would be easier to drop support for building the test suite. The current implementation is not complete anyway. It does not build the binary or the julia bindings and the tests associated with those.

eli-schwartz: If the vendored filereaderlp becomes a fork, then it should be moved out of extern/ surely. I don't see this as a problem.

Thanks for tracking down the provenance of that code. It includes a pkg-config file so it seems designed to be a stable public interface ;) definitely makes sense to use it as a subproject.

eli-schwartz:
Update: current versions of HiGHS don't use generated export headers (or symbol visibility at all).

HaoZeke:
Also [upstream](https://github.com/ERGO-Code/HiGHS) has a `meson.build` which is usable, so this could be simplified now.

PR #310, opened 2022-02-18
libffmpegthumbnailer: add

neheb:

### DIFF
--- a/subprojects/packagefiles/libffmpegthumbnailer/meson.build
+++ b/subprojects/packagefiles/libffmpegthumbnailer/meson.build
@@ -1,0 +1,72 @@
+project('ffmpegthumbnailer', 'cpp',
+  version : '2.2.2',
+  default_options : ['warning_level=0', 'cpp_std=c++11'],
+  meson_version: '>=0.47.0'
+)
+
+libavcodec_dep = dependency('libavcodec')
+libavfilter_dep = dependency('libavfilter')
+libavformat_dep = dependency('libavformat')
+libavutil_dep = dependency('libavutil')
+libjpeg_dep = dependency('libjpeg', required : get_option('jpeg'))
+libpng_dep = dependency('libpng', required : get_option('png'))
+
+ffmpegthumbnailer_deps = [ libavcodec_dep, libavfilter_dep, libavformat_dep, libavutil_dep, libjpeg_dep, libpng_dep ]
+
+cdata = configuration_data()
+cpp = meson.get_compiler('cpp')
+
+cdata.set('HAVE_PNG', libjpeg_dep.found())
+cdata.set('HAVE_JPEG', libpng_dep.found())
+
+cfile = configure_file(
+  output : 'config.h',
+  configuration : cdata,
+)
+
+libffmpegthumbnailer_incdir = include_directories([ '.', 'libffmpegthumbnailer' ])
+
+compfiles = [
+  'libffmpegthumbnailer/filmstripfilter.cpp',
+  'libffmpegthumbnailer/moviedecoder.cpp',
+  'libffmpegthumbnailer/rgbwriter.cpp',
+  'libffmpegthumbnailer/stringoperations.cpp',
+  'libffmpegthumbnailer/videothumbnailer.cpp',
+  'libffmpegthumbnailer/videothumbnailerc.cpp',
+]
+
+if libjpeg_dep.found()
+  compfiles += 'libffmpegthumbnailer/jpegwriter.cpp'
+  ffmpegthumbnailer_deps += libjpeg_dep
+endif
+
+if libpng_dep.found()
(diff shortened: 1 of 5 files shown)


subprojects/packagefiles/libffmpegthumbnailer/meson.build
```diff
@@ -0,0 +1,71 @@
...
+
+libffmpegthumbnailer = library('ffmpegthumbnailer',
+  compfiles,
+  dependencies : ffmpegthumbnailer_deps,
+  install : true,
+  version : '4.15.1',
+)
+
+pkg = import('pkgconfig')
+pkg.generate(libffmpegthumbnailer)
```
eli-schwartz, line 55: This could also use `description: 'Generates thumbnails for movie files'`

subprojects/packagefiles/libffmpegthumbnailer/meson_options.txt
```diff
@@ -0,0 +1,7 @@
+option('jpeg', type : 'feature',
+  description : 'JPEG support',
+)
```
eli-schwartz, line 3: IMHO this probably would look a bit nicer as one line.

subprojects/packagefiles/libffmpegthumbnailer/meson.build
```diff
@@ -0,0 +1,71 @@
+project('ffmpegthumbnailer', 'cpp',
+  version : '2.2.2',
```
eli-schwartz, line 2: Currently uses `meson_version: '>=0.47.0'`, could we add this explicitly?

eli-schwartz (changes requested):
The `test/` directory should be easy to wire up, same with main.cpp. You'd need to add a couple more bits to config.h for that.

subprojects/packagefiles/libffmpegthumbnailer/meson_options.txt
neheb: Done.

subprojects/packagefiles/libffmpegthumbnailer/meson.build
neheb: Done.

eli-schwartz:
This will somehow need ffmpeg installed, could be done in ci_config.json for the Linux tests.

There's a gstreamer port of ffmpeg, but I'm not sure what the status of it is, @ tp-m would know I suppose. it would be great if it one day ended up in the wrapdb!

neheb:
doesn't microsoft have vcpkg?

eli-schwartz:
I have no clue how that works, if you want to set it up for our Visual Studio test workflow then color me interested!

neheb:
Neither do I. I've wasted many hours trying to set it up on Windows.

xclaesse:
> There's a gstreamer port of ffmpeg, but I'm not sure what the status of it is, @tp-m would know I suppose. it would be great if it one day ended up in the wrapdb!

Yes, it's there: https://gitlab.freedesktop.org/gstreamer/meson-ports. Works great and has been regularly used by GStreamer for a long time already. I would like to see that in wrapdb as well, but I don't think it makes sense to copy the build system into wrapdb. I would instead recommend if @tp-m or @nirbheek could make a release tarball of that "fork" and use that in wrapdb. I think they follow the same rule as wrapdb and only keep build system related patches, no modification of upstream code. But I could be wrong.

nirbheek:
Actually looking again, we do have some patches but they're for build fixes. No one has the patience to deal with FFmpeg upstream to submit them.

neheb:
ffffffffffffffffffffffffffffffff

Failing because it needs to be updated for ffmpeg 5.0.

neheb:
```
 ==> ffmpeg@4
ffmpeg@4 is keg-only, which means it was not symlinked into /usr/local,
because this is an alternate version of another formula.

If you need to have ffmpeg@4 first in your PATH, run:
  echo 'export PATH="/usr/local/opt/ffmpeg@4/bin:$PATH"' >> /Users/runner/.bash_profile

For compilers to find ffmpeg@4 you may need to set:
  export LDFLAGS="-L/usr/local/opt/ffmpeg@4/lib"
  export CPPFLAGS="-I/usr/local/opt/ffmpeg@4/include"

For pkg-config to find ffmpeg@4 you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/ffmpeg@4/lib/pkgconfig"
```

hmmm. I'm unsure how to proceed.

neheb:
Hrm so upstream fixed compilation with ffmpeg 5. Unfortunately, a new release needs to be cut. For reference: https://github.com/dirkvdb/ffmpegthumbnailer/pull/198

neheb:
Fixed finally.

neheb:
ping @jpakkane

PR #311, opened 2022-02-19
Add wrap for ff-nvcodec-headers version 11.1.5.1-0

nirbheek:
https://git.videolan.org/?p=ffmpeg/nv-codec-headers.git

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -231,6 +231,14 @@
       "2.0.2-0"
     ]
   },
+  "ff-nvcodec-headers": {
+    "dependency_names": [
+      "ffnvcodec"
+    ],
+    "versions": [
+      "11.1.5.1-0"
+    ]
+  },
   "flac": {
     "dependency_names": [
       "flac"
--- a/subprojects/ff-nvcodec-headers.wrap
+++ b/subprojects/ff-nvcodec-headers.wrap
@@ -1,0 +1,9 @@
+[wrap-git]
+directory = ff-nvcodec-headers
+# Web UI: https://git.videolan.org/?p=ffmpeg/nv-codec-headers.git
+url = https://git.videolan.org/git/ffmpeg/nv-codec-headers.git
+revision = n11.1.5.1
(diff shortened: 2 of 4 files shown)


eli-schwartz:
This PR should NOT have been merged, as it explicitly violates the WrapDB policy -- indeed, in order to get it to pass the CI, you had to code a special exemption into the sanity checks, which is a sure sign that it was doing something wrong.

It was merged without review and without the opportunity for anyone to raise objections.

nirbheek:
Okay, so I did something you disagreed with. I don't see a review of the changes in this PR from you, and instead what I see is a revert that was _also_ merged without discussion with me or anyone else. Nothing was broken, there was no urgency to revert, so why did you do that?

I hope you realize that what you did was incredibly rude. You basically took a unilateral decision to revert the work of someone else without discussion with them, with anyone else, nor for any reason other than "I didn't like it".

jpakkane:
The reversion was a bit on the rough side, but on the other hand this PR has a policy code change and it was merged within one hour of creation without any external review. That is also not cool.

To prevent this sort of thing from happening in the future, let's add a policy that any pull requests that change the code of wrapdb itself may not be merged without an approval from someone that is not the submitter itself.

nirbheek:
> this PR has a policy code change and it was merged within one hour of creation without any external review. That is also not cool.

Yes, that was my bad, and I freely admit it. I thought the sanity checks were extra-strict to make review safer, and that it was expected to punch holes in it when required by a wrap. Didn't realize that it was explicit policy to only allow wrap-file (ref. discussion on #mesonbuild).

My issue has been only with Eli's response to my mistake. I would have expected him to leave me a message asking me about it, or that the revert was done after discussion on the channel. I would've also been ok with it if he had at the very least left comments on my PR presenting an alternative after reverting it.

An immediate revert is a very aggressive move in my view, and an action of last-resort in the case of breakage. On top of that, not even leaving a review is just plain insulting.

PR #313, opened 2022-02-20
Revert "Revert "Add wrap for ff-nvcodec-headers version 11.1.5.1-0""

nirbheek:
This reverts commit e26eaed80e4de47587aaa24d7963666fca263612.

@eli-schwartz I expect a review from you on this to tell me exactly what you would do differently.

### DIFF
--- a/subprojects/ff-nvcodec-headers.wrap
+++ b/subprojects/ff-nvcodec-headers.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = nv-codec-headers-11.1.5.1
+source_url = https://github.com/FFmpeg/nv-codec-headers/releases/download/n11.1.5.1/nv-codec-headers-11.1.5.1.tar.gz
+source_filename = nv-codec-headers-11.1.5.1.tar.gz
+source_hash = a28cdde3ac0e9e02c2dde7a1b4de5333b4ac6148a8332ca712da243a3361a0d9
+patch_directory = ff-nvcodec-headers
+
+[provide]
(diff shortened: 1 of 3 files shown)


subprojects/ff-nvcodec-headers.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-git]
```
jpakkane, line 1: Git wraps are not permitted in wrapdb, only tarballs.

jpakkane:
Strange, the tests should not be passing because `sanity_checks.py` has `self.assertEqual(config.sections()[0], 'wrap-file')`. Maybe it is not being run for new wraps.

subprojects/ff-nvcodec-headers.wrap
nirbheek: ff-nvcodec-headers does not release tarballs, and I do not expect them to ever do that. That's why I added it and pointed it to a tag.

nirbheek:
> Strange, the tests should not be passing because `sanity_checks.py` has `self.assertEqual(config.sections()[0], 'wrap-file')`. Maybe it is not being run for new wraps.

I added a new variable in the sanity checks file called ` PER_PROJECT_PERMITTED_WRAP_GIT` for wraps that are permitted to use git repos since I do not expect this project to ever have a release tarball.

nirbheek:
Actually, I was wrong. There is a tarball release on the github mirror. I will update this to use that.

PR #314, opened 2022-02-27
fix xxhash configuration for Windows and Clang

ninjaoflight:
Fix xxhash dynamic library configuration for native clang in windows
details:
the xxhash package only adds the build arg `-DXXH_EXPORT` on dynamic library configuration when using `msvc`, however on `native clang` this is also required to generate the `.lib`, this patch adds support for this, however it could be also added for `xxhash 0.8.1` patch when it is ready


eli-schwartz:
What happens if we just always pass that define? I would assume it would be a no-op for Unix.

Given the entire problem here was incorrectly gating on handpicked compilers... Adding another handpicked compiler seems wrong IMHO.

Alternatively we could define it based on the `host_machine.system()` being Windows.

ninjaoflight:
Sounds good to me, I changed it, so it now checks for Windows instead of clang and MSVC.

eli-schwartz:
Superseded by #315

Issue #247, opened 2021-12-13
Update xxhash to 0.8.1


tristan957:
Tracking issue for myself unless someone else gets to it before me. Feel free to assign it to me.

ninjaoflight:
this was fixed on #315

PR #316, opened 2022-02-28
Sqlite3 3.38.0

ninjaoflight:
- Update SQLite from 3.34.1 to 3.38.1
- Define `-DSQLITE_API=__declspec(dllexport)` in `Windows` when compiling as Dynamic Library instead of only when using `MSVC` (similar to #247 windows fix)


eli-schwartz:
Thanks!

PR #318, opened 2022-03-01
htslib: update to 1.15

dpar39:


eli-schwartz:
Sigh, upstream proposed to add a Windows/MSVC .def file in https://github.com/samtools/htslib/commit/3f7e13e0688e50fca61e6f043e6655a165d596a2 but it is generated and not available in the release tarball. Oh well, guess this is a routine version bump after all... :D

PR #322, opened 2022-03-06
OPENSSL: making openssl_cli to be always compiled

### LINKED ISSUE #320: openssl wrap install fails
see https://github.com/linux-nvme/libnvme/issues/268
```
RuntimeError: File 'subprojects/openssl-1.1.1l/openssl' could not be found
```
from https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/openssl/meson.build
```
openssl_cli = executable(
  'openssl',
  build_by_default: false,
  dependencies: dependencies + [openssl_dep],
  sources: openssl_cli_sources,
  include_directories: include_directories,
  c_args: c_args,
  install: true,
)
```
what should happen if `build_by_default: false` and `install: true` is set. I haven't found anything related...

@nazar-pc @xclaesse @jpakkane any advice ?

glimchb:
Fixes #320

build_by_default=false is incompatible with install=true

Since openssl_cli takes no time to build, let's always build it

Signed-off-by: Boris Glimcher <Boris.Glimcher@emc.com>


nazar-pc:
I think we can also have an option to chose whether executable needs to be built or not. It should cover all cases then. The default could be not building, since most common case would be to use this as a library.

eli-schwartz:
What's the downside of building it?

nazar-pc:
Wasting CPU resources, CI time, disk space when only libraries are necessary.

eli-schwartz:
It's 2.3 mb of disk space out of 28 and compiles 1/14th the number of files needed to build the rest of OpenSSL.

The wall time to build the rest of openssl is 41 seconds. The wall time to additionally build the openssl_cli is an additional 5.

I'm not opposed to adding an option anyway, but if so it should be enabled in `ci_config.json` so that the WrapDB CI checks that it works.

glimchb:
I agree with @eli-schwartz the compilation time and size are neglectable in this case. 
@nazar-pc can we merge this as is and add option in a separate PR/release ? 
I'm not an expert and can not handle that path of addition an option and CI by myself

nazar-pc:
There is one option in `subprojects/packagefiles/openssl/meson_options.txt` already, you can add another one by analogy.

I would strongly prefer to have an option, otherwise folks will write shell scripts to remove unnecessary build artifacts from installations, etc.

glimchb:
@nazar-pc added an option following the `asm` example, please review
What about CI as @eli-schwartz mentioned?

nazar-pc:
I think you just need to add an entry with build options for `openssl` in `ci_config.json` file to enable this new feature in CI. Also add `"linux_only": true` as it doesn't work with Windows CI right now anyway.

glimchb:
thanks @nazar-pc . done, please review

nazar-pc (approved):
Makes sense to me, someone needs to trigger CI though

glimchb:
@eli-schwartz can you please trigger CI and review the changes ? Thanks

eli-schwartz:
https://mesonbuild.com/Build-options.html#build-option-types

You want a boolean type option.

glimchb:
@eli-schwartz fixed

eli-schwartz (approved):

glimchb:
@eli-schwartz thanks for guidance and merge. How this gets published to wrap website ?

nazar-pc:
Automatically, give it a bit of time to build

eli-schwartz:
It was uploaded a couple minutes before your question. :)

PR #323, opened 2022-03-12
SDL2: Update to 2.0.20, adopt meson.build from Taisei

Jan200101:
This new, more complete meson.build has been adopted from <https://github.com/taisei-project/SDL> provided by the Taisei Project.


Jan200101:
The CI is not being nice today

subprojects/packagefiles/sdl2/find-dylib-name.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+
+import subprocess
+import os
+import re
+import sys
+import pathlib
+
+
+def verbose(*args):
```
jpakkane, line 10: This file should have a comment explaining what it does. It is not obvious from the code.

Jan200101: I added a comment at the top of the file to explain what it does and when it is used

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -10,375 +14,1034 @@ version_micro = varr[2]
...
+    else
+        cdata.set('SDL_@0@_DISABLED'.format(ss_name.to_upper()), 1)
+        message('Subsystem "@0@" is DISABLED'.format(ss_name))
+    endif
+endforeach
+
+# sigh...
+# https://github.com/mesonbuild/meson/issues/1889
+# Let's try to cure the superproject std
+unfuck_std = '''
```
jpakkane, line 409: This variable name should be something different.

Jan200101: removed the obscene language and updated the comment to be more descriptive on what its doing

xclaesse:
Are you merging review fixes back to the original repository? Let's not have a fork of a fork...

For this kind of big ports I'm wondering if we aren't better to just take tarball from taisei GitHub instead of keeping a big overlay in wrapdb. Feels like the same situation as for ffmpeg fork gstreamer team is maintaining.

eli-schwartz:
Both ffmpeg and SDL would be wrapdb policy violations, since we don't permit using forks as sources.

Jan200101:
> Are you merging review fixes back to the original repository? Let's not have a fork of a fork...
> 
> For this kind of big ports I'm wondering if we aren't better to just take tarball from taisei GitHub instead of keeping a big overlay in wrapdb. Feels like the same situation as for ffmpeg fork gstreamer team is maintaining.

I had talked to Akari before submitting this, and apparently the intention was to submit the meson stuff to wrapdb but current world events have rendered them unable to do so.

xclaesse:
> Both ffmpeg and SDL would be wrapdb policy violations, since we don't permit using forks as sources.

Policies can change and have expectations. For ffmpeg I'm certain we should add the fork in wrapdb, but in the case of SDL I don't know what else that fork contains other than meson definition.

Jan200101:
Personally I have no issue with using the Taisei fork
but I can see many issues crop up in regards to the authenticity of the original code.

If the decision is made that using a fork is better, I suggest talking to the Taisei Developers first to see if they are willing to take such a workload.

eli-schwartz:
> For ffmpeg I'm certain we should add the fork in wrapdb

And I am equally certain we should not, so we are at an impasse I guess.

Meanwhile, the gst port has no actual need of the wrapdb so I don't quite get why it matters...

> but I can see many issues crop up in regards to the authenticity of the original code.

Exactly.

Akaricchi:
I'm a Taisei developer and maintainer of the SDL fork.

> but in the case of SDL I don't know what else that fork contains other than meson definition.

Currently just the meson definition, but relying on our fork directly for wrapdb is not wise. We have carried local patches in the past, and the meson definitions are written with Taisei-specific use-cases in mind; e.g. installing isn't handled at all.

> Are you merging review fixes back to the original repository? Let's not have a fork of a fork...

I'm happy to merge the improvements back, but that probably won't happen soon. I live in Ukraine and had to flee since the front line got really close to my house, so I'm pretty much incapacitated for the near future. I might be able to do a review though, no promises.

jpakkane:
> I live in Ukraine and had to flee since the front line got really close to my house, so I'm pretty much incapacitated for the near future. I might be able to do a review though, no promises.

If I was in your situation, reviewing pull requests would not be particularly high on my priority list. Respect! I hope you stay safe.

eli-schwartz:
Yeah, please do keep yourself safe and put PR reviews at the bottom of your to-do list.

jpakkane:
I don't know if anyone really has the requisite expertise to review this properly (if someone is, do speak up). As this comes from a known working source then the right thing to do is to merge this and fix issues if and when they turn up. Opinions?

eli-schwartz:
There is one simple and obvious question to look at, which is that this uses `build_by_default: not meson.is_subproject()`, so the WrapDB CI is not actually testing that it compiles properly. Also it is `install: false` and while I understand Taisei's motivation for this, I think `install: get_option('default_library') != 'static' or not meson.is_subproject()` would be more correct?

jpakkane:
Build by default should always be true for the main library (even better, the kwarg should be removed altogether). If you have SDL in your project, you'd typically want to build it.

eli-schwartz:
Sure, those were my thoughts exactly.

Akaricchi:
> Also it is install: false and while I understand Taisei's motivation for this, I think install: get_option('default_library') != 'static' or not meson.is_subproject() would be more correct?

The actually correct thing to do would be to remove this line and [fix meson to make the superproject decide what gets installed by default](https://github.com/mesonbuild/meson/pull/10009#issuecomment-1042498653). Since wrapdb has no policy about this (as far as I know) this line should probably just be removed either way (for the wrapdb version, but not for the Taisei fork).

> Build by default should always be true for the main library (even better, the kwarg should be removed altogether). If you have SDL in your project, you'd typically want to build it.

I think meson should ignore build by default in subprojects and treat it as always false. It maybe doesn't matter in SDL's case, but generally there should be no reason to build a target that's ultimately not used by the superproject. But I digress.

Jan200101:
> Build by default should always be true for the main library (even better, the kwarg should be removed altogether). If you have SDL in your project, you'd typically want to build it.

If you are using it as a subproject you are probably going to depend on it, which means it gets build.
Is there any reason it should be build, even when nothing consumes it?

> The actually correct thing to do would be to remove this line and [fix meson to make the superproject decide what gets installed by default](https://github.com/mesonbuild/meson/pull/10009#issuecomment-1042498653).

I agree.
a subproject can never know what exactly the main project wants to do and it is best not to make assumptions.

eli-schwartz:
> Is there any reason it should be build, even when nothing consumes it?

Because this premature optimization is, as you yourself point out, not going to work -- you cannot actually suppress it from being built, because it will ***always be linked to when used as a dependency fallback***.

The ***only*** thing `build_by_default: false` can do here is break the wrapdb CI (which builds as a gnome-style metaproject instead) and make me refuse to merge it.

> a subproject can never know what exactly the main project wants to do and it is best not to make assumptions.

If you say "do install", you're assuming "yes". If you say "don't install", you're assuming "no".

Let's leave the philosophy out of this PR. Meson currently doesn't have an existing capacity to "not make assumptions". Refusal to make a decision is itself a decision.

I have pointed out a robust heuristic, please use it. Thanks. :)

eli-schwartz:
```
sdl2| Run-time dependency gl found: NO (tried pkgconfig and system)
sdl2| Run-time dependency glesv2 found: NO (tried pkgconfig and cmake)
sdl2| Message: Using embedded GLESv2 headers
sdl2| Run-time dependency egl found: NO (tried pkgconfig and cmake)
sdl2| Run-time dependency egl found: NO (tried pkgconfig and cmake)
sdl2| Message: Using embedded EGL headers
```

Why is it failing at build time if configure time dependencies marked as `required: false` aren't found? Maybe some component is supposed to be switched off in that case but isn't?

eli-schwartz:
Windows:
```
[64/313] Compiling C object subprojects/SDL2-2.0.20/sdl2.dll.p/src_haptic_windows_SDL_windowshaptic.c.obj
FAILED: subprojects/SDL2-2.0.20/sdl2.dll.p/src_haptic_windows_SDL_windowshaptic.c.obj 
"cl" "-Isubprojects\SDL2-2.0.20\sdl2.dll.p" "-Isubprojects\SDL2-2.0.20" "-I..\subprojects\SDL2-2.0.20" "-Isubprojects\SDL2-2.0.20\include" "-I..\subprojects\SDL2-2.0.20\include" "-I..\subprojects\SDL2-2.0.20\src\video\khronos" "/MDd" "/nologo" "/showIncludes" "/utf-8" "/W2" "/Od" "/Zi" "-DUSING_GENERATED_CONFIG_H" "/Fdsubprojects\SDL2-2.0.20\sdl2.dll.p\src_haptic_windows_SDL_windowshaptic.c.pdb" /Fosubprojects/SDL2-2.0.20/sdl2.dll.p/src_haptic_windows_SDL_windowshaptic.c.obj "/c" ../subprojects/SDL2-2.0.20/src/haptic/windows/SDL_windowshaptic.c
D:\a\wrapdb\wrapdb\subprojects\SDL2-2.0.20\src\haptic\windows\SDL_windowshaptic_c.h(61): error C2061: syntax error: identifier 'XINPUT_VIBRATION'
D:\a\wrapdb\wrapdb\subprojects\SDL2-2.0.20\src\haptic\windows\SDL_windowshaptic_c.h(63): error C2059: syntax error: '}'
../subprojects/SDL2-2.0.20/src/haptic/windows/SDL_windowshaptic.c(314): error C2027: use of undefined type 'haptic_hweffect'
D:\a\wrapdb\wrapdb\subprojects\SDL2-2.0.20\src\haptic\windows\SDL_windowshaptic_c.h(54): note: see declaration of 'haptic_hweffect'
../subprojects/SDL2-2.0.20/src/haptic/windows/SDL_windowshaptic.c(319): error C2027: use of undefined type 'haptic_hweffect'
D:\a\wrapdb\wrapdb\subprojects\SDL2-2.0.20\src\haptic\windows\SDL_windowshaptic_c.h(54): note: see declaration of 'haptic_hweffect'
```

"syntax error" oh no...

Jan200101:
The thing I love the most about CI is the fact that testing it locally is a pain.

I will figure this stuff out and test it on a local debian machine.

jpakkane:
I tested this on Windows and the error is because `HAVE_XINPUT_H` is unset. This workaround fixes it for me:

```meson
    cdata.set('HAVE_XINPUT_H', 1)
endif
```

Jan200101:
> I tested this on Windows and the error is because `HAVE_XINPUT_H` is unset. This workaround fixes it for me:
> 
> ```meson
> if host_machine.system() == 'windows'
>     cdata.set('HAVE_XINPUT_H', 1)
> endif
> ```

Issue was `SDL_HAPTIC_XINPUT` being set despite xinput not being present.
I also went ahead and fixed some MinGW incompatibilities that would only occur on a very unique setup

eli-schwartz:
> CI is green and issues raised seem to be fixed. If issues arise they can be dealt with in followup PRs. Thanks,

Not all issues were fixed, I raised a review comment:

> Also it is `install: false` and while I understand Taisei's motivation for this, I think `install: get_option('default_library') != 'static' or not meson.is_subproject()` would be more correct?

It was switched to remove the install kwarg altogether, which is implict `false`, which is what I wasn't happy about.

PR #331, opened 2022-03-17
add MacOS CI

Jan200101:
This will need some fleshing out before it can get merged, which is why I am keeping as a draft for now.

Since there are now more than 2 CI options, `linux_only` no longer represents a binary choice.
Instead of adding a `<os>_only` for every single one, I suggest an array under `build_on` containing any combination of `linux`, `windows` and `darwin`.

Because up until now no one tested on MacOS, some work needs to be done to populate the `broken_darwin` array.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -1,10 +1,10 @@
 {
   "_comment": ["This is intentionally undocumented in README.md, do not use it.",
                "Using it is an automatic CI error.",
-               "skip_linux and skip_windows are lists of wraps that are known",
+               "broken_linux, broken_windows and broken_macos are lists of wraps that are known",
                "to be broken on the CI and ***MUST*** get repaired before updates",
-               "will be accepted. If upstream does not support Windows, set that",
-               "wrap to linux_only instead.",
+               "will be accepted. If upstream does not support an Operating System,",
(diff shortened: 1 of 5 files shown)


Jan200101:
CI is passing, I'm force pushing again to fix some newlines that are annoying me

ci_config.json
```diff
@@ -1,10 +1,10 @@
 {
   "_comment": ["This is intentionally undocumented in README.md, do not use it.",
                "Using it is an automatic CI error.",
-               "skip_linux and skip_windows are lists of wraps that are known",
+               "broken_linux, broken_windows and broken_macos are lists of wraps that are known",
                "to be broken on the CI and ***MUST*** get repaired before updates",
-               "will be accepted. If upstream does not support Windows, set that",
-               "wrap to linux_only instead.",
+               "will be accepted. If upstream does not support an Operating System, set that",
+               "explicitly disable that Operating System in build_on.",
```
eli-schwartz, line 7: "set that explicitly disable that Operating System"

The "set that" part is left over from the previous wording.

eli-schwartz:
Thank you for your hard work here!

PR #333, opened 2022-03-17
SDL fixes

Jan200101:
- define `DLL_EXPORT` when doing non static Windows builds to properly export symbols
- move sdl2main from `link_whole` to `link_with` so it can be used with libraries (fixes building SDL2image on Windows)
- disable the offscreen video backend by default, like the offial SDL2 CMakeLists.txt. This should fix MacOS builds while not straying too far from the offical logic


Jan200101:
Force pushing to trigger MacOS CI

PR #335, opened 2022-03-20
libupnp: add

neheb:

### DIFF
--- a/subprojects/packagefiles/libupnp/meson.build
+++ b/subprojects/packagefiles/libupnp/meson.build
@@ -1,0 +1,296 @@
+project('pupnp', 'c',
+  version : '1.14.12',
+  default_options : [ 'c_std=gnu99', 'warning_level=1' ]
+)
+
+libupnp_incdir = include_directories([ '.', 'upnp/inc', 'upnp/src/inc', 'upnp/src/threadutil' ])
+libixml_incdir = include_directories([ 'ixml/inc', 'ixml/src/inc' ])
+
+libixml = static_library('ixml',
+  'ixml/src/attr.c',
+  'ixml/src/document.c',
+  'ixml/src/element.c',
+  'ixml/src/ixml.c',
+  'ixml/src/ixmldebug.c',
+  'ixml/src/ixmlmembuf.c',
+  'ixml/src/ixmlparser.c',
+  'ixml/src/namedNodeMap.c',
+  'ixml/src/node.c',
+  'ixml/src/nodeList.c',
+  include_directories : [ libixml_incdir, libupnp_incdir ],
+)
+
+upnp_sources = [
+  'upnp/src/api/UpnpActionComplete.c',
+  'upnp/src/api/UpnpActionRequest.c',
+  'upnp/src/api/UpnpDiscovery.c',
+  'upnp/src/api/UpnpEvent.c',
+  'upnp/src/api/UpnpEventSubscribe.c',
+  'upnp/src/api/UpnpExtraHeaders.c',
+  'upnp/src/api/UpnpFileInfo.c',
+  'upnp/src/api/UpnpStateVarComplete.c',
+  'upnp/src/api/UpnpStateVarRequest.c',
+  'upnp/src/api/UpnpSubscriptionRequest.c',
+  'upnp/src/api/UpnpString.c',
+  'upnp/src/api/upnpapi.c',
+  'upnp/src/api/upnpdebug.c',
+  'upnp/src/genlib/client_table/GenlibClientSubscription.c',
+  'upnp/src/genlib/client_table/client_table.c',
+  'upnp/src/genlib/miniserver/miniserver.c',
+  'upnp/src/genlib/net/sock.c',
+  'upnp/src/genlib/net/http/httpparser.c',
+  'upnp/src/genlib/net/http/httpreadwrite.c',
+  'upnp/src/genlib/net/http/parsetools.c',
+  'upnp/src/genlib/net/http/statcodes.c',
+  'upnp/src/genlib/net/http/webserver.c',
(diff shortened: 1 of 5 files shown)


subprojects/packagefiles/libupnp/meson.build
```diff
@@ -0,0 +1,292 @@
...
+  'upnp/inc/UpnpStateVarRequest.h',
+  'upnp/inc/UpnpStdInt.h',
+  'upnp/inc/UpnpString.h',
+  'upnp/inc/UpnpSubscriptionRequest.h',
+  'upnp/inc/UpnpUniStd.h',
+  ufile,
+]
+
+if get_option('tools')
+  'upnp/inc/upnptools.h'
```
eli-schwartz, line 237: What's this?

subprojects/packagefiles/libupnp/meson.build
```diff
@@ -0,0 +1,292 @@
...
+endif
+
+install_headers(headers)
```
eli-schwartz, line 240: It seems that upstream installs all these to a 'upnp' subdirectory of the includedir?

subprojects/packagefiles/libupnp/meson.build
```diff
@@ -0,0 +1,292 @@
+project('pupnp', 'c',
+  version : '1.14.12',
+  default_options : [ 'c_std=gnu99', 'warning_level=1' ]
+)
+
+libupnp_incdir = include_directories([ '.', 'upnp/inc', 'upnp/src/inc', 'upnp/src/threadutil' ])
+libixml_incdir = include_directories([ 'ixml/inc', 'ixml/src/inc' ])
+
+libixml = static_library('ixml',
```
eli-schwartz, line 9: Why is this specifically a static library, especially considering it is installed?

subprojects/packagefiles/libupnp/meson_options.txt
```diff
@@ -0,0 +1,74 @@
...
+)
+
+option('optssdp', type : 'boolean',
+  value : true,
+  description : 'optional SSDP headers support',
+)
+
+option('open_ssl', type : 'boolean',
+  value : false,
+  description : 'open-ssl support',
```
eli-schwartz, line 48: Huh, I am not used to seeing that spelled with a dash/underscore in the name. :p

eli-schwartz (changes requested):
Just took a very quick look at this, will try to take a closer look tomorrow.

subprojects/packagefiles/libupnp/meson_options.txt
neheb: Upstream does this. No idea why.

subprojects/packagefiles/libupnp/meson.build
neheb: Because nothing besides libupnp uses it.

subprojects/packagefiles/libupnp/meson.build
neheb: Mistake. Should be added to headers.

subprojects/packagefiles/libupnp/meson.build
neheb: I'll need to check this.

subprojects/packagefiles/libupnp/meson.build
eli-schwartz: Then in that case, why is it installed? And why does it also install public headers?

neheb: Beats me.

neheb:
Wonder how upstream handles compilation for MSVC...

neheb:
Anyone have any ideas? The issue seems to be that windows.h is implicitly included somewhere making the whole thing work with CMake.

neheb:
Umm, never mind.

```
2022-03-21T03:57:19.1186474Z D:\a\wrapdb\wrapdb\subprojects\libupnp-1.14.12\upnp\inc\ithread.h(51): fatal error C1083: Cannot open include file: 'pthread.h': No such file or directory
```

No idea how I would even fix that.

eli-schwartz:
Well, idk what that horror is.

The official pthreads4w project is here: https://sourceforge.net/p/pthreads4w/code/ci/master/tree/Makefile#l229

That's an NMake makefile, and honestly it looks quite simple to port to meson.

xclaesse:
>  Please note:- whilst PThreads4W can be built and run by it, MinGW64 includes it's own default POSIX Threads library called "winpthreads". The two are not compatible and in order to build and run PThreads4W (formerly PThreads-WIn32) MinGW64 must be installed without win32pthreads. If you want or need to build and run with PThreads4W then you need to choose win32 threading instead of POSIX when you install MinGW64 to not install the conflicting winpthreads include and library files.

When I though the pthread situation couldn't possibly get worse... everyday we learn new ways being disappointed by the Windows platform... I'm currious if libupnp could be interested by porting to winpthread maybe?

eli-schwartz:
Since they currently support MSVC and we have MSVC CI that doesn't seem like a critical blocker.

And it's not terribly surprising that competing versions of the same library will clobber each other's headers.

neheb:
OTOH there's already libnpupnp, which replaces this mess with std::thread. Which puts into question the value of libupnp honestly...

xclaesse:
> Which puts into question the value of libupnp honestly

I am totally unaware of that situation, it seems indeed that libnpupnp replaced libupnp for a while already and we have a wrap for it. Why are you adding this wrap then?

neheb:
In somewhat interesting news, libupnp 1.16 will be C++11. Meaning this pthread stuff can probably be replaced by std::thread.

neheb:
Fixed.

PR #336, opened 2022-03-21
libnpupnp: test

neheb:


neheb:
cpp_std=c++11 is specified but is not being applied to macOS.

xclaesse:
That's because `cpp_std` is a global option that cannot be defined in a subproject. We need that PR to fix this: https://github.com/mesonbuild/meson/pull/9394.

As workaround for wrapdb CI, you can add this in ci_config.json:
```
    "build_options": [
      "cpp_std=gnu++11"
    ]
```

neheb:
Fixed.

PR #338, opened 2022-03-21
inih: Fix INIReader to upper case

stephanlachnit:
Reported here: https://github.com/Tachi107/cloudflare-ddns/issues/10#issuecomment-1073641944

Upper case is also consistent with the pkg-config file.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -508,9 +508,10 @@
   "inih": {
     "dependency_names": [
       "inih",
-      "inireader"
+      "INIReader"
     ],
     "versions": [
+      "r53-2",
       "r53-1",
       "r52-1",
       "r51-1"
--- a/subprojects/inih.wrap
+++ b/subprojects/inih.wrap
@@ -7,4 +7,4 @@ source_hash = 01b0366fdfdf6363efc070c2f856f1afa33e7a6546548bada5456ad94a516241
 
 [provide]
 inih = inih_dep
-inireader = inireader_dep
+INIReader = INIReader_dep


stephanlachnit:
I think this should be merged even though the CI fails, as the lower case version never worked anyway and people should use the upper case version to be consistent with the pkg-config file.

eli-schwartz:
I'm pretty sure the CI is telling you that you're wrong and the lower case version actually does work. :)

Did you try it first?

For the record, the CI error is saying that the keys in the wrap file are *still* in lowercase, and they don't match the uppercase dependency names in the JSON file.

With the context that changing the case in the wrap file ***didn't actually take hold*** you can determine in short order that python's stdlib configparser deletes case sensitivity if it is found in keys (but not values).

This is fine, because Meson explicitly handles this: it treats dependency names as case insensitive (but case preserving) both for wrap provide lookups and for dependencies with custom lookup functionality.

However, the case must be correct in order to successfully run the `pkg-config` or `cmake` tools to probe for a dependency, which is often problematic because cmake has weird capitalization compared to the cmake names, so you can find one but not both unless you use 0.60-compatible "multiple dependency names", or have a custom lookup in Meson itself that accepts any case and encodes the exact string to pass to pkg-config/cmake.

stephanlachnit:
> Did you try it first?

Actually no, I sanity-checked the meson file upstream, and just figured it wouldn't work. However, I tried it now. MWE in an empty diretory:

```sh
echo "project('test')" > meson.build
echo "dependency('INIReader')" >> meson.build
mkdir subprojects
meson wrap search INIReader  # doesn't find inih
meson wrap install inih
meson setup builddir --wrap-mode=nofallback  # works if INIReader.pc is available
rm -rf builddir
meson setup builddir --wrap-mode=forcefallback  # fails, see below
```

Log:
```
The Meson build system
Version: 0.61.2
Source dir: /tmp/coding/test
Build dir: /tmp/coding/test/builddir
Build type: native build
Project name: test
Project version: undefined
Host machine cpu: x86_64
Looking for a fallback subproject for the dependency INIReader because:
Use of fallback dependencies is forced.
Downloading inih source from https://github.com/benhoyt/inih/archive/r53.tar.gz
Downloading file of unknown size.
Downloading inih patch from https://wrapdb.mesonbuild.com/v2/inih_r53-1/get_patch
Download size: 769
Downloading: ..........

Executing subproject inih 

inih| Project name: inih
inih| Project version: undefined
inih| C compiler for the host machine: ccache cc (gcc 11.2.0 "cc (Debian 11.2.0-18) 11.2.0")
inih| C linker for the host machine: cc ld.bfd 2.38
inih| C++ compiler for the host machine: ccache c++ (gcc 11.2.0 "c++ (Debian 11.2.0-18) 11.2.0")
inih| C++ linker for the host machine: c++ ld.bfd 2.38
inih| Build targets in project: 2
inih| Subproject inih finished.

WARNING: Variable 'inireader_dep' in the subproject 'subprojects/inih-r53' is not found
Dependency INIReader from subproject subprojects/inih-r53 found: NO

meson.build:2:0: ERROR: Dependency 'INIReader' is required but not found.

A full log can be found at /tmp/coding/test/builddir/meson-logs/meson-log.txt
```

> For the record, the CI error is saying that the keys in the wrap file are _still_ in lowercase, and they don't match the uppercase dependency names in the JSON file.

If I get it right, it says that the dependency names changed?

> With the context that changing the case in the wrap file _**didn't actually take hold**_ you can determine in short order that python's stdlib configparser deletes case sensitivity if it is found in keys (but not values).

Right, I see that the case actually doesn't matter for dependency names. However this is not true for variable names.

I guess I will rework the PR to only change the variable name, but not the dependency name. Can you reopen it?

eli-schwartz:
That is a good point.

> Can you reopen it?

You force pushed your branch, so GitHub doesn't offer that option. Maybe it would re-enable that option if you reverted back to the commit I closed it on.

stephanlachnit:
> You force pushed your branch, so GitHub doesn't offer that option. Maybe it would re-enable that option if you reverted back to the commit I closed it on.

Don't have it locally anymore, I'll open a new PR then.

eli-schwartz:
You probably do have it in `git reflog` :) but it's not a big deal, and I don't even know whether it would work, so it's not really worth bothering *that much* about it.

eli-schwartz:
Also note that the Windows job failed for two reasons, one of which was a compile error.

stephanlachnit:
> You probably do have it in `git reflog` :)

Indeed I do! Thanks, didn't know that feature :)

PR #281, opened 2022-01-14
Upgrade OpenSSL to 3.0.2

nazar-pc:
Changes:
* Upgrade to 3.0.2
* riscv64 architecture support
* cleanups

PR #339, opened 2022-03-21
Fix OpenSSL CI release due to missing dependency

nazar-pc:
Fix for package missing in https://github.com/mesonbuild/wrapdb/pull/281 that led to error during release publish.

Would be nice for it to install packages from `debian_packages` automatically though.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,8 +10,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Install nasm for OpenSSL generator
-        run: sudo apt-get install --no-install-recommends nasm
       - name: Upload release assets
         run: |
           git fetch --unshallow --tags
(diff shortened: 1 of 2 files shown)


eli-schwartz:
> Would be nice for it to install packages from `debian_packages` automatically though.

Only for the one package that's also configured to have a release generator though.

Anyway, please implement that change so we can catch the mistake next time. ;)

nazar-pc:
Python is not my strongest side, I'd really appreciate help here.

eli-schwartz:
In `tools/create_release.py` there is a line

```python
if generator.exists():
```

Inside that block it runs the generator.

Add another couple of lines which use python's `json` library to load ci_config.json and get the list of packages, which you can then pass to another subprocess command, this time running apt to install the packages.

There is prior art in doing all this in the sanity_checks.py script, and json module documentation at https://docs.python.org/3/library/json.html

nazar-pc:
OK, I'll try to take care of it later today

nazar-pc:
This update seems to work properly in my fork

eli-schwartz:
Thanks!

PR #341, opened 2022-03-21
fmt: update to 8.1.1

neheb:
Fix compilation with MSVC and macOS

the CI config.


neheb:
Rebased.

PR #342, opened 2022-03-21
libmatroska: fix compilation with MSVC and macOS

neheb:

the CI config.


eli-schwartz:
Thanks, fix confirmed via CI.

It is now safe to remove the release bump from releases.json (which, like the other PR, is not needed if no effective meson.build changes are made).

neheb:
Done.

PR #343, opened 2022-03-21
protobuf: fix compilation with MSVC and macOS

neheb:

the CI config.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -17,8 +17,8 @@
     "google-woff2", "hinnant-date", "imgui-sfml", "jbigkit", "libexif",
     "libgpiod", "liblbfgs", "liblzma", "libwebsockets", "libxmlpp",
     "lzo2", "mocklibc", "mpdecimal", "netstring-c", "nng", "openh264",
-    "protobuf", "protobuf-c", "quazip", "rdkafka", "reflex", "termbox",
-    "unit-system", "zstd"
+    "protobuf-c", "quazip", "rdkafka", "reflex", "termbox","unit-system",
+    "zstd"
   ],
   "broken_darwin": [
     "abseil-cpp", "arduinocore-avr", "box2d", "check", "cli11", "cpr",
@@ -26,7 +26,7 @@
     "google-benchmark", "google-woff2", "graphite2", "hinnant-date",
     "imgui", "imgui-sfml", "libgpiod", "libwebsockets", "libxmlpp",
     "lmdb", "lz4", "mocklibc", "mpdecimal", "openal-soft",
-    "openh264", "protobuf", "protobuf-c", "quazip", "rdkafka", "re2",
+    "openh264", "protobuf-c", "quazip", "rdkafka", "re2",
     "rtaudio", "sdl2_image", "sfml", "tinyply", "tinyxml2", "unit-system",
     "zstd"
   ],
@@ -179,6 +179,11 @@
     ],
     "build_options": [
       "cpp_std=c++17"
+    ]
+  },
+  "protobuf" : {
+    "build_options": [
+      "cpp_std=c++14"
     ]
   },
   "rdkafka": {
--- a/subprojects/packagefiles/protobuf/meson.build
+++ b/subprojects/packagefiles/protobuf/meson.build
@@ -1,6 +1,14 @@
-project('protobuf', 'cpp', version: '3.12.2', license: 'protobuf')
+project('protobuf', 'cpp', version: '3.20.0', license: 'protobuf')
 
 cc = meson.get_compiler('cpp')
+
+ttype = cc.get_id() == 'msvc' ? 'static_library' : 'library'
+
+if get_option('default_library') != 'static'
+  add_project_arguments('-DPROTOBUF_USE_DLLS', language : 'cpp')
+  add_project_arguments('-DLIBPROTOBUF_EXPORTS', language : 'cpp')
+  add_project_arguments('-DLIBPROTOC_EXPORTS', language : 'cpp')
+endif
 
 flags = cc.get_supported_arguments([
   '-DHAVE_PTHREAD',
@@ -17,20 +25,25 @@ deps = dependency('threads')
 libprotobuf_lite_src = [
   'src/google/protobuf/any_lite.cc',
   'src/google/protobuf/arena.cc',
+  'src/google/protobuf/arenastring.cc',
+  'src/google/protobuf/arenaz_sampler.cc',
   'src/google/protobuf/extension_set.cc',
   'src/google/protobuf/generated_enum_util.cc',
-  'src/google/protobuf/generated_message_table_driven_lite.cc',
+  'src/google/protobuf/generated_message_tctable_lite.cc',
   'src/google/protobuf/generated_message_util.cc',
   'src/google/protobuf/implicit_weak_message.cc',
+  'src/google/protobuf/inlined_string_field.cc',
   'src/google/protobuf/io/coded_stream.cc',
   'src/google/protobuf/io/io_win32.cc',
   'src/google/protobuf/io/strtod.cc',
   'src/google/protobuf/io/zero_copy_stream.cc',
   'src/google/protobuf/io/zero_copy_stream_impl.cc',
   'src/google/protobuf/io/zero_copy_stream_impl_lite.cc',
+  'src/google/protobuf/map.cc',
   'src/google/protobuf/message_lite.cc',
   'src/google/protobuf/parse_context.cc',
   'src/google/protobuf/repeated_field.cc',
+  'src/google/protobuf/repeated_ptr_field.cc',
   'src/google/protobuf/stubs/bytestream.cc',
   'src/google/protobuf/stubs/common.cc',
   'src/google/protobuf/stubs/int128.cc',
@@ -43,11 +56,12 @@ libprotobuf_lite_src = [
   'src/google/protobuf/stubs/time.cc',
   'src/google/protobuf/wire_format_lite.cc',
 ]
-libprotobuf_lite = library(
+libprotobuf_lite = build_target(
   'protobuf-lite',
   libprotobuf_lite_src,
   cpp_args: flags,
   dependencies: deps,
+  target_type: ttype,
   include_directories: include_directories('src'),
 )
 protobuf_lite_dep = declare_dependency(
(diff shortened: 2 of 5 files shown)


neheb:
Decided to update the version as well.

neheb:
```
Found ninja-1.10.2 at "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.EXE"
WARNING: Could not create compilation database.
```

What?

eli-schwartz:
That's, uh, bizarre. I don't even...

eli-schwartz:
So a bit of debugging with that fix above ^^ and it turns out the compdb failure is because it errors out with:

```
ninja: error: build.ninja:907: multiple rules generate subprojects/protobuf-3.18.2/protoc.pdb [-w dupbuild=err]
```

and without --fatal-meson-warnings this will proceed with one warning to successfully configure... then immediately error out when trying to build.

eli-schwartz:
I guess this happens because we create:
- a lib named "protoc"
- an executable named "protoc"

so protoc.dll and protoc.exe clash on the pdb. Lovely! Thank you Microsoft. <3

dcbaker:
We lie about c++11 in msvc and just give you c++14 with a warning

eli-schwartz:
@dcbaker, so we don't need to upgrade the std in default_options, or... ?

@neheb,

I would still prefer to have this actually work on Windows, which should be possible by moving one or the other into its own directory.

neheb:
Maybe.

eli-schwartz (changes requested):
Can we get this to be compatible with MSVC?

jpakkane:
⚔️ conflicts.

neheb:
Rebased.

neheb:
@eli-schwartz looks like it builds now.

ci_config.json
```diff
@@ -176,6 +175,15 @@
       "cpp_std=c++17"
     ]
   },
+  "protobuf" : {
+    "_comment": ["- PDB can't handle protobuf.dll and protobuf.exe"],
+    "build_on": {
+      "windows": false
```
eli-schwartz, line 181: You forgot to remove this, so it still isn't being tested on Windows. I assume it *should* work now, yes...

subprojects/protobuf.wrap
```diff
@@ -1,6 +1,11 @@
...
+directory = protobuf-3.18.2
+source_url = https://github.com/protocolbuffers/protobuf/releases/download/v3.18.2/protobuf-all-3.18.2.tar.gz
+source_filename = protobuf-all-3.18.2.tar.gz
+source_hash = 5675aff486e15353cb20fe0abe1fddf9d5fabc3ac2b9e84406a333cb14f7984f
 patch_directory = protobuf
+
+[provide]
+protobuf = protobuf_dep
+protobuf_lite = protobuf_lite_dep
+protoc = protoc_dep
```
eli-schwartz, line 11: protoc_dep isn't a public dep AFAICT, it's only used internally by the protoc binary. e.g. it doesn't install a pkg-config file.

On the other hand, I think you do want to add in a `meson.version.version_compare()` block, an `override_find_program` for protoc_native, and then a `program_names = protoc`.

eli-schwartz (changes requested):

subprojects/protobuf.wrap
eli-schwartz: Also I just double checked and protobuf-lite had a dash, not an underscore.

neheb: I don't see this meson.build installing pkgconfig files.

neheb:
Any idea on the failure?

neheb:
So to build properly, MSVC needs to use static_library. library and both_libraries do not work.

eli-schwartz:
Actually, spelunking through the upstream code I believe you just need to define:
- PROTOBUF_USE_DLLS if building shared
- LIBPROTOC_EXPORTS and LIBPROTOC_EXPORTS when building this library, but not as part of its declare_dependency or pkgconfig flags

This also means that when building with GCC/clang it sets up `__attribute__((visibility("default")))`, which in turn means I believe that you can use `gnu_symbol_visibility: 'hidden'`.

eli-schwartz:
If it *was* indeed impossible for Windows, I would then ask that you use `library()` on non-Windows, and `static_library()` with MSVC -- you can do that with e.g. `build_target(cc.get_id() == 'windows' ? 'static_library' : 'library')`

neheb:
```
 LINK : fatal error LNK1181: cannot open input file **'subprojects\protobuf-3.20.0\protobuf-lite.lib'**
```

subprojects/packagefiles/protobuf/meson.build
```diff
@@ -1,7 +1,15 @@
-project('protobuf', 'cpp', version: '3.12.2', license: 'protobuf')
+project('protobuf', 'cpp', version: '3.20.0', license: 'protobuf')
 
 cc = meson.get_compiler('cpp')
 
+if get_option('default_library') != 'static'
+  add_project_arguments('-DPROTOBUF_USE_DLLS', language : 'cpp')
+  if cc.get_id() == 'msvc'
+    add_project_arguments('LIBPROTOBUF_EXPORTS', language : 'cpp')
+    add_project_arguments('LIBPROTOC_EXPORTS', language : 'cpp')
```
eli-schwartz, line 9: These are missing `-D`, and also you want to define them everywhere since they are used by GCC too.

neheb:
Yeah no. This reminds me of libebml. I don't see shared libraries being buildable on MSVC.

eli-schwartz:
This is baffling, because it seems to work here: https://github.com/protocolbuffers/protobuf/blob/b5a35bcc7e4cc4ff00eb105449b300fca3a8b0d2/cmake/libprotobuf-lite.cmake#L113-L117

eli-schwartz:
Oh, the error is drowned out in a sea of warnings. The real error is this:
```
src_google_protobuf_any.cc.obj : error LNK2019: unresolved external symbol "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl google::protobuf::internal::GetTypeUrl(class google::protobuf::stringpiece_internal::StringPiece,class google::protobuf::stringpiece_internal::StringPiece)" (?GetTypeUrl@internal@protobuf@google@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@VStringPiece@stringpiece_internal@23@0@Z) referenced in function "public: bool __cdecl google::protobuf::internal::AnyMetadata::PackFrom(class google::protobuf::Arena *,class google::protobuf::Message const &,class google::protobuf::stringpiece_internal::StringPiece)" (?PackFrom@AnyMetadata@internal@protobuf@google@@QEAA_NPEAVArena@34@AEBVMessage@34@VStringPiece@stringpiece_internal@34@@Z)
src_google_protobuf_wrappers.pb.cc.obj : error LNK2001: unresolved external symbol "class google::protobuf::internal::ExplicitlyConstructed<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >,8> google::protobuf::internal::fixed_address_empty_string" (?fixed_address_empty_string@internal@protobuf@google@@3V?$ExplicitlyConstructed@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@$07@123@A)
src_google_protobuf_util_message_differencer.cc.obj : error LNK2001: unresolved external symbol "class google::protobuf::internal::ExplicitlyConstructed<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >,8> google::protobuf::internal::fixed_address_empty_string" (?fixed_address_empty_string@internal@protobuf@google@@3V?$ExplicitlyConstructed@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@$07@123@A)
src_google_protobuf_util_time_util.cc.obj : error LNK2001: unresolved external symbol "class google::protobuf::internal::ExplicitlyConstructed<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >,8> google::protobuf::internal::fixed_address_empty_string" (?fixed_address_empty_string@internal@protobuf@google@@3V?$ExplicitlyConstructed@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@$07@123@A)
src_google_protobuf_util_type_resolver_util.cc.obj : error LNK2001: unresolved external symbol "class google::protobuf::internal::ExplicitlyConstructed<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >,8> google::protobuf::internal::fixed_address_empty_string" (?fixed_address_empty_string@internal@protobuf@google@@3V?$ExplicitlyConstructed@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@$07@123@A)
```

A bunch of useless warnings are suppressed via https://github.com/protocolbuffers/protobuf/blob/b5a35bcc7e4cc4ff00eb105449b300fca3a8b0d2/CMakeLists.txt#L250-L262

neheb:
yeah so with libebml I avoided these issues by making it a static_library everywhere. a completely fine solution as libebml is only used by libmatroska and nothing else. not the case with protobuf.

Upstream calls these defines deprecated. So probably shouldn't bother with them and just build static with MSVC. I doubt upstream even tests shared libraries (dlls).

eli-schwartz:
> Upstream calls these defines deprecated.

What? How sad.

I'll take your word for it (please do leave a comment explaining why msvc uses a `static` build target though).

neheb:
For OpenWrt, I have to pass this to get shared libraries https://github.com/openwrt/packages/blob/master/libs/protobuf/Makefile#L79 . Which means the default is static libraries.

Upstream's comment: https://github.com/protocolbuffers/protobuf/blob/3bdcc1266b65e10a61d398ba15bfbec612cbdb5d/src/google/protobuf/stubs/port.h#L91

eli-schwartz:
I was looking at this which does not say anything about being deprecated: https://github.com/protocolbuffers/protobuf/blob/3bdcc1266b65e10a61d398ba15bfbec612cbdb5d/src/google/protobuf/port_def.inc#L412

The other comment *looks* like it is referring to third-party users using the protobuf internal export defines... in their third-party projects, which *sounds* insane.

I think the BUILD_SHARED_LIBS is just a side effect of the way they test for it by cloning the option to protobuf_BUILD_SHARED_LIBS_DEFAULT and then again to protobuf_BUILD_SHARED_LIBS.

neheb:
sure but without it I get no shared libraries.

PR #347, opened 2022-03-22
SDL2: enable installation, fix linking with sdl2main

Jan200101:
- add the install kwarg as [eli wanted](https://github.com/mesonbuild/wrapdb/pull/323#issuecomment-1075552763)
- hopefully finally fix sdl2main being a pain
  - on msvc we need to link_with sdl2main otherwise we lose symbols and it cannot resolve `_SDL_main`
  - on any other toolchain we need link_whole otherwise we don't link with `--whole-archive` which results in some things like WinMain being thrown out


subprojects/packagefiles/sdl2/meson.build
```diff
@@ -1030,13 +1030,16 @@ if platform_is_winrt or platform_is_haiku
     all_sources += [cxx_sources]
 endif
 
+do_install = get_option('default_library') != 'static' or not meson.is_subproject()
```
tristan957, line 1033: Shouldn't the not meson.is_subproject() be removed in favor of meson install --skip-subprojects?

eli-schwartz: --skip-subprojects is badly behaved if `default_library=shared`, since then you must install libsdl2.so but it doesn't get installed.

This PR correctly installs the library if either of the following conditions are true:
- sdl2 is being built standalone, so you care about installing it as its own unique thing
- the default_library is such that you need it installed at runtime

Jan200101: this was taken from eli's comment [here](https://github.com/mesonbuild/wrapdb/pull/323#issuecomment-1066197960) so it is best you discuss that with him

xclaesse: I think it's wrong to make such special case here. Why is SDL special compared to all other projects? The real fix is install --tags runtime. You could want to install static libraries to be consumed by another project via pkg+config.

eli-schwartz: SDL isn't special, let's update *all* wraps to do like this one.

Thanks for the suggestion.

xclaesse: That makes absolutely no sense. If you really think that's the right thing to do, should be done in meson by default. But it's wrong.

eli-schwartz: I may be wrong, but at least I'm consistently wrong, because I am indeed on record as saying it should be done in Meson by default.

It's an old discussion at this point, frankly. The relevant keywords are "gst-build" and "metaproject".

tristan957: For the record, I agree with @xclaesse. This makes no sense to me. I understand you are the maintainer, but shouldn't we be pushing Meson best practices in the WrapDB and not what our personal ideas are?

PR #348, opened 2022-03-22
liblzma: update to 5.2.5

neheb:


eli-schwartz:
There are new defines, BTW: HAVE_ENCODERS / HAVE_DECODERS

https://github.com/xz-mirror/xz/commit/2296778f3c9a1e3a8699973b09dd3610b8baa402

eli-schwartz:
Per https://github.com/xz-mirror/xz/blob/master/windows/INSTALL-MSVC.txt#L13 the command line utilities are not supported on MSVC and should not be added as build targets.

neheb:
Fixed.

subprojects/packagefiles/liblzma/src/xz/meson.build
```diff
@@ -1,31 +1,33 @@
...
-  'signals.c',
-  'suffix.c',
-  'util.c',
-  '../common/tuklib_open_stdxxx.c',
-  '../common/tuklib_progname.c',
-  '../common/tuklib_exit.c',
-  '../common/tuklib_mbstr_width.c',
-  '../common/tuklib_mbstr_fw.c',
-]
+if cc.get_id() != 'msvc'
```
eli-schwartz, line 1: This might be simpler to do by either checking in src/meson.build before subdir'ing in, or just checking here and doing subdir_done().

IMHO it is slightly clearer to read than figuring out what is covered by the indented code, and also, it avoids having to change lots of lines just to move the indentation. :D

neheb: Done.

PR #349, opened 2022-03-22
gdbm: update to 1.23

neheb:


eli-schwartz:
GNU gdbm apparently doesn't support non-POSIX platforms like MSVC (should work okay on cygwin/msys2 apparently).

It would be nice if upstream added a pkg-config file as that is currently meaningless here -- users cannot use `dependency('gdbm')` anyway. I'd just ignore that CI error.

neheb:
No longer builds on CI.

PR #351, opened 2022-03-24
Fetch libxml2 subproject using http instead of ftp 

JohanSternerup:
In restricted networks FTP is sometimes blocked and then a HTTP
fallback is needed.


xclaesse:
@JohanSternerup ah you need to add a revision in releases.json.

eli-schwartz:
As libxml2 is now maintained by  Nick Wellnhofer of gnome rather than Daniel Veillard, and hosted on downloads.gnome.org instead of xmlsoft.org (which is a redirect to libvirt's domain with broken SSL), the tarballs are officially rehosted on the gnome infrastructure and should probably use that first instead of ftp.

xclaesse:
Btw, upstream meson support is/was being worked on: https://gitlab.gnome.org/GNOME/libxml2/-/issues/328.

PR #352, opened 2022-03-25
Fixup source_filename for libxml2 to use xz instead of gz

JohanSternerup:
When changing source_url in previous
commit (6cf1e10d213130f90191a3cffc1884b563392ad2) the url changed from
gz to xz but source_filename wasn't changed.


xclaesse:
oops, good catch! Looks like we could maybe do a bit more sanity checking in the CI.

PR #353, opened 2022-03-25
libopenjp2: Fix static build on Windows

xclaesse:
Applications that build static libopenjp2 needs OPJ_STATIC define when
including its header.


xclaesse:
@fasterthanlime I saw you did similar in other projects (e.g. https://github.com/frida/proxy-libintl/pull/2). Seems we have to do that trick pretty much in all wraps...

eli-schwartz:
That's probably also relevant to pkg-config.

xclaesse:
@eli-schwartz Hm good point. I've never seen a pkg config file doing that, but we could start that trend I guess... Not that this is only relevant to static build for Windows. At least glib define that in their public headers so it's not needed for that lib. Not sure how common that is.

eli-schwartz:
> I've never seen a pkg config file doing that, but we could start that trend I guess...

Any pkg-config file which intends to support Windows would need to do that. There are definitely pkg-config files that in fact do so.

I know I've seen it, I just forget where... they probably aren't in the linux generated versions of the .pc files so I cannot trivially find them. :p

xclaesse:
@eli-schwartz done, and did the same for libxml2: https://github.com/mesonbuild/wrapdb/pull/354

xclaesse:
Since the same fix got approved in other wraps, let's just merge it now.

PR #357, opened 2022-03-26
fluidsynth: update to 2.2.6

kcgen:


neheb (approved):

neheb:
Not relevant, but can you also fix up the broken_windows section?

eli-schwartz:
> Not relevant, but can you also fix up the broken_windows section?

Next time please do not do this! It is line noise. Only remove the one that you are marking as no longer broken. If you want to "fix up" something that is not "broken", please submit a separate PR to reflow the list, and revert the reflowing you've already done.

This makes it easier to read the changes in this PR. Instead of looking at and trying to review a 32-line diff, you only need to look at a *one* line diff.

kcgen:
Great point @eli-schwartz ; a good case where one-item-per line works well (alphabetic sort).  I think code reviews and diffs are the perfect scenario to encourage their use.

neheb:
@eli-schwartz I suggest running clang-format on all the json files. Gets rid of these problems. Probably would be good to add a CI check for that too.

PR #360, opened 2022-03-27
clang-format json files

neheb:


eli-schwartz:
I'm not really a fan of putting every package on its own line.

jpakkane:
The alternative is having poorly readable diffs. Neither is perfect, but this is probably better as it can be automated.

eli-schwartz:
Diffs are only poorly readable if you *allow* people to reflow the entire file on trivial changes instead of having them remove one word from the middle of the line. :) It worked fine that way until a single PR yesterday.

eli-schwartz:
Also making every per-package comment 3 lines instead of 1 feels bad. Actually they probably don't even need to be an array, it's just cargo-culting from the much longer comment on the `broken_*` keys. We don't actually read the `_comment` keys ever, so we don't care what type they contain or if they are sometimes a string and sometimes an array of strings.

xclaesse:
I'm personally I'm favor of this PR, I would even suggest to check that in ci.

neheb:
https://github.com/DoozyX/clang-format-lint-action

eli-schwartz:
I'm pretty firmly against checking it in CI if it is going to fail the CI any time someone does the *sensible* thing and puts a single-item list on a single line.

neheb:
I mean even git has git clang-format for this.

eli-schwartz:
```
s/git has/clang has/
```

(of course clang-format will advertise its own tool, not sure what that has to do with anything.)

jpakkane:
I think you can do a precommit hook that prevents you from committing in malformed files. Which would be cool.

eli-schwartz:
It's less cool if you have GCC installed but not clang(-format), or have to be told to install a githook because those aren't automatically enabled for security reasons.

neheb:
Rebased. You guys want a simplified version so this can be merged?

neheb:
Did just that.

neheb:
@eli-schwartz how about now?

neheb:
Rebased.

jpakkane:
More automation is better than manual labor.

PR #362, opened 2022-03-30
Add FTXUI.

jpakkane:


subprojects/packagefiles/ftxui/meson.build
```diff
@@ -0,0 +1,96 @@
...
+    default_options: ['cpp_std=c++17'],
+)
+
+if host_machine.system() == 'windows'
+  add_project_arguments('/utf-8', '-DUNICODE', language: 'cpp')
+endif
+
+# Upstream CMake supports only static linking,
+# so we do the same.
+
```
eli-schwartz, line 13: This seems wrong. https://github.com/ArthurSonzogni/FTXUI/commit/57a5512a22aa1a9af570b932fbf8751b20039bdb

eli-schwartz (changes requested):

subprojects/packagefiles/ftxui/meson.build
eli-schwartz: Maybe it's just broken on MSVC? We do have `build_target(..., target_type: host_machine.system() == 'windows' ? 'static_library' : 'library')` for that...

xclaesse: They don't have any `dllexport` so I guess that's why they can't build shared on Windows. Sounds like a job for https://github.com/mesonbuild/meson/pull/10199.

eli-schwartz: That won't help in the slightest, it just creates a define that projects then have to care enough to use -- which they incredibly often do not. :(

eli-schwartz: Anyway, I think it would be best to use the build_target() with target_type trick here so that at least linux and macOS users can defer to the builtin options.

jpakkane: On the other hand I'm a big fan of simplicity. This has the same behaviour on all platforms and based on what we know of usage, most WrapDB users seem to prefer static linking.

eli-schwartz: Hmm... the main reason I'd want to do regular library() for linux is to make it closer to upstream and thus more useful if attempting to submit as a PR to upstream, anyway.

That may not be super important here after all.

xclaesse: > it just creates a define that projects then have to care enough to use

Ah, they indeed don't have a define at all yet, I was expecting they would have one for at least gnu visibility, or maybe using the cmake generator. Indeed that would need to change header files. Upstream could be interested to have it though.

> On the other hand I'm a big fan of simplicity. This has the same behaviour on all platforms and based on what we know of usage, most WrapDB users seem to prefer static linking.

I agree, if upstream does not deal with symbol visibility at all better not try doing shared library.

eli-schwartz: > I agree, if upstream does not deal with symbol visibility at all better not try doing shared library.

TBH on Linux symbol visibility is *just* a performance optimization and prevents accidental use of non-public API. And the latter is just as much of a problem when linking statically.

The effort needed to bend over backward is probably a bigger concern.

xclaesse: > the main reason I'd want to do regular library() for linux is to make it closer to upstream and thus more useful if attempting to submit as a PR to upstream, anyway.

That's a valid reason tbh. Maybe just use library() and add `if host_system == windows and default_library!=static error('not supported')` ? And link to an upstream bug report for adding the symbol visibility define.

subprojects/ftxui.wrap
```diff
@@ -0,0 +1,11 @@
...
+directory = FTXUI-2.0.0
+source_url = https://github.com/ArthurSonzogni/FTXUI/archive/refs/tags/v2.0.0.tar.gz
+source_filename = FTXUI-2.0.0.tar.gz
+source_hash = d891695ef22176f0c09f8261a37af9ad5b262dd670a81e6b83661a23abc2c54f
+patch_directory = ftxui
+
+[provide]
+ftxui-screen = screen_dep
+ftxui-dom = dom_dep
+ftxui-component = component_dep
```
eli-schwartz, line 11: Upstream installs *one* dependency, `fxtui`, and would require finding via meson via `dependency('fxtui', method: 'cmake', modules: [fxtui::screen'])`.

How utterly impossible to map to pkg-config at all.

I guess this is the best we could do.

jpakkane:
It's not perfect but it's working, usable and can be improved later if need be.

PR #364, opened 2022-04-02
Freetype: Update to 2.12.0 release

satmandu:
Update freetype to `2.12.0` release.


xclaesse:
Our policy for wrapdb is to not use wrap-git and only accept wraps based on released tarball.

satmandu:
Is this more acceptable?

satmandu:
(And should a PR update releases.json too?)

subprojects/freetype2.wrap
```diff
@@ -1,8 +1,9 @@
 [wrap-file]
-directory = freetype-2.11.1
-source_url = https://download.savannah.gnu.org/releases/freetype/freetype-2.11.1.tar.gz
-source_filename = freetype-2.11.1.tar.gz
-source_hash = f8db94d307e9c54961b39a1cc799a67d46681480696ed72ecf78d4473770f09b
+directory = freetype
```
jpakkane, line 2: This should be the directory that the tar file has, so probably `freetype-2.12.0`.

subprojects/freetype2.wrap
```diff
@@ -1,8 +1,9 @@
...
+source_url = https://download.savannah.gnu.org/releases/freetype/freetype-2.12.0.tar.xz
+source_filename = freetype-2.12.0.tar.xz
+source_hash = ef5c336aacc1a079ff9262d6308d6c2a066dd4d2a905301c4adda9b354399033
 
 [provide]
 freetype2 = freetype_dep
+freetype = freetype_dep
```
jpakkane, line 9: FWICT this is not correct. The pkg-config name of Freetype is `freetype2` so this should match that. Just remove this line.

jpakkane (changes requested):

jpakkane:
> (And should a PR update releases.json too?)

Yes.

satmandu:
> > (And should a PR update releases.json too?)
> 
> Yes.

I see releases are suffixed with a "-1" or higher.  So I assume this should be '2.12.0-1' ?

xclaesse:
Exact

satmandu:
done

eli-schwartz:
After the suggested review changes, we are back to the original freetype2.wrap file, but with the version number bumped and the hash updated.

Please squash all changes into one commit. Commit message should probably be something like `Freetype: update to 2.1.12 release`

satmandu:
Adjusted. (Also fixed the title, as we are talking about version 2.12.0...)

eli-schwartz:
I see upstream actually provides both gz and xz, makes sense to switch to the latter indeed.

eli-schwartz:
The macOS CI is erroring out by telling you that this update fixes CI there, so you should remove it from the known-broken list.

satmandu:
Where is that?

eli-schwartz:
The CI message calls out the string `broken_darwin`, you can find this in the file `ci_config.json`.

satmandu:
Done.

eli-schwartz:
Thanks!

PR #365, opened 2022-04-02
Add cairo subproject as used in harfbuzz

satmandu:
As per https://github.com/harfbuzz/harfbuzz/blob/main/subprojects/cairo.wrap
and also
https://github.com/harfbuzz/harfbuzz/pull/3501/files

### DIFF
--- a/subprojects/cairo.wrap
+++ b/subprojects/cairo.wrap
@@ -1,0 +1,7 @@
+directory=cairo
+url=https://gitlab.freedesktop.org/cairo/cairo.git
+depth=1
+revision=1.17.6
+
+[provide]
+dependency_names = cairo, cairo-ft


eli-schwartz:
Like the other PR -- this must be a wrap-file, the WrapDB does not accept git based wraps.

Thirdparty projects are welcome to do whatever they want.

subprojects/cairo.wrap
```diff
@@ -0,0 +1,7 @@
+directory=cairo
```
eli-schwartz, line 1: This doesn't have a section header at all, hmm.

eli-schwartz:
... and of course CI doesn't even catch the issue, because releases.json doesn't contain the new wrap.

satmandu:
Since https://github.com/mesonbuild/wrapdb/pull/366 has been opened for `cairo`, and seems more comprehensive, I will defer to that PR.

PR #190, opened 2021-10-08
abseil-cpp: fix `hash` library that included test cpp file with `main()`

nazar-pc:

PR #368, opened 2022-04-03
Fix abseil-cpp with duplicated `main` again

nazar-pc:
I already fixed it once in https://github.com/mesonbuild/wrapdb/pull/190/commits/3e1f3c6159f28b4510001072b4e4173fc75ce489, but it was re-added again in #262 for some reason and caused the same exact problem.

That file is not necessary and contains `main()` function, breaking any chance to use `absl_hash` as a library.

cc @inglor


jpakkane:
You need to add

```
  "abseil-cpp": {
    "build_options": [
      "cpp_std=c++17"
    ]
  },
```

to `ci_config.json` because apparently macOS defaults to c++98 or something.

nazar-pc:
But `'cpp_std=c++11',` is specified in `meson.build` already.
And we use previous version of the wrap `20210324.2-2` with CI passing on all platforms, including macOS 10.15 and macOS 11.
I'm confused...

eli-schwartz:
It probably depends on the cpp_std of the main project, since https://github.com/mesonbuild/meson/pull/9394 in meson git master if it is set in the subproject default_options then that will be respected.

nazar-pc:
Should work now

PR #369, opened 2022-04-08
libobsd: update to 1.1.0

guijan:
I ported it to MSVC so I removed it from ci_config.json


neheb:
@guijan can you also submit this to OpenWrt?

guijan:
> @guijan can you also submit this to OpenWrt?

Yes, I'll give it a try.

PR #373, opened 2022-04-13
spdlog: update to 1.10.0

neheb:


neheb:
strange that this succeeds. Should be failing on macOS.

neheb:
hahahahaha fixed.

Turns out it's broken upstream.

neheb:
Fixed.

PR #374, opened 2022-04-13
new wrap: dragonbox

HassanSajjad-302:
Added https://github.com/jk-jeon/dragonbox/

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -169,6 +169,14 @@
       "0.6.3-1"
     ]
   },
+  "dragonbox": {
+    "dependency_names": [
+      "dragonbox"
+    ],
+    "versions": [
+      "1.1.2-1"
+    ]
+  },
   "eigen": {
     "dependency_names": [
       "eigen3"
--- a/subprojects/packagefiles/dragonbox/meson.build
+++ b/subprojects/packagefiles/dragonbox/meson.build
@@ -1,0 +1,13 @@
+project('dragonbox',
+    'cpp',
+    version : '1.1.2',
+    license : 'Boost',
+    default_options : ['cpp_std=c++17']
+)
+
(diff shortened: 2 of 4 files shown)


ilmari-lauhakangas:
Ping @jpakkane

releases.json
```diff
@@ -1660,5 +1660,13 @@
       "1.3.3-2",
       "1.3.3-1"
     ]
+  },
+  "dragonbox": {
+    "dependency_names": [
+      "dragonbox"
+    ],
+    "versions": [
+      "1.1.2"
```
jpakkane, line 1669: This needs to be `1.1.2-1`. The number after the dash signifies the "wrapdb version". There may be several for any given upstream release.

subprojects/packagefiles/dragonbox/meson.build
```diff
@@ -0,0 +1,13 @@
...
+    'cpp',
+    version : '1.1.2',
+    license : 'Boost',
+    default_options : ['cpp_std=c++17']
+)
+
+dragonbox_lib = library('dragonbox', 'source/dragonbox_to_chars.cpp', include_directories : include_directories('include/'))
+
+dragonbox_dep = declare_dependency(
+    include_directories : include_directories('include/'),
```
jpakkane, line 11: This should not have the trailing slash.

jpakkane (changes requested):

HassanSajjad-302:
Changes Requested Applied.

jpakkane:
You need to set the C++ version in `ci_config.json`. Due to reasons it defaults to empty and on the compiler we use for running CI it means either C++11 or something even older.

In addition the list of items in the JSON must be sorted, so you need to add the entry to the correct place, not at the end of the array.

HassanSajjad-302:
Updated.

PR #380, opened 2022-04-16
Use C++17 by default.

jpakkane:
Most C++ dependencies seem to require a newer stdlib version than is the default so let's use 17 by default to cut down on boilerplate.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -32,19 +32,15 @@
   ],
 
   "abseil-cpp": {
-    "_comment": ["Remove this once https://github.com/mesonbuild/meson/pull/9394 is released"],
     "build_options": [
-      "cpp_std=c++14"
     ]
   },
   "box2d": {
     "build_options": [
-      "cpp_std=c++14"
     ]
   },
   "dragonbox": {
     "build_options": [
-      "cpp_std=c++17"
     ]
   },
   "epoxy": {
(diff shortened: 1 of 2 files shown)


ci_config.json
```diff
@@ -34,17 +34,14 @@
   "abseil-cpp": {
     "_comment": ["Remove this once https://github.com/mesonbuild/meson/pull/9394 is released"],
```
neheb, line 35: Comment here says it all.

Might as well get rid of these package entries completely.

xclaesse:
IMHO we should instead set cpp_std in each wrap default_option, that way they will work when used by different projects. At least when 0.63 gets released.

jpakkane:
> IMHO we should instead set cpp_std in each wrap default_option

They already do. If we wanted to be thorough, then we should build each project twice: once as a subproject and once as the toplevel project. But in the mean time this change will simplify settings for everybody.

meson.build
```diff
@@ -1,5 +1,6 @@
 project('wrapdb',
   meson_version: '>=0.55.0'
+  default_options: ['cpp_std=c++17'],
```
krnowak, line 3: This broke meson.build, because of missing comma at the end of `meson_version` line.

xclaesse:
> But in the mean time this change will simplify settings for everybody.

I don't think we should set it globally because it hides the problem if we forget to set it on subproject. We should at least revert this when meson 0.63 gets released.

PR #381, opened 2022-04-16
quazip: update to 1.3

neheb:

### DIFF
--- a/subprojects/packagefiles/quazip/meson.build
+++ b/subprojects/packagefiles/quazip/meson.build
@@ -1,12 +1,17 @@
 project(
   'QuaZip',
   ['c', 'cpp'],
-  version: '0.7.3',
+  version: '1.3',
   license: 'LGPL-2.1 with static linking exception'
 )
 
-qt5 = import('qt5')
-qt5_dep = dependency('qt5', modules: 'Core')
+qt_dep = dependency('qt6', modules: [ 'Core', 'Core5Compat' ], required : false)
+if qt_dep.found()
+  qt = import('qt6')
+else
+  qt_dep = dependency('qt5', modules: 'Core')
+  qt = import('qt5')
+endif
 zlib_dep = dependency('zlib', fallback: ['zlib', 'zlib_dep'])
 
 subdir('quazip')
--- a/subprojects/packagefiles/quazip/quazip/meson.build
+++ b/subprojects/packagefiles/quazip/quazip/meson.build
@@ -16,6 +16,7 @@ quazip_src = [
   'qioapi.cpp',
   'JlCompress.cpp',
   'quaadler32.cpp',
+  'quachecksum32.cpp',
   'quacrc32.cpp',
   'quagzipfile.cpp',
   'quaziodevice.cpp',
@@ -28,15 +29,14 @@ quazip_src = [
   'zip.c'
 ]
 
-quazip_pp = qt5.preprocess(moc_headers: quazip_moc_headers)
+quazip_pp = qt.preprocess(moc_headers: quazip_moc_headers)
 
 quazip_lib = library(
   'quazip',
   quazip_src,
   quazip_pp,
-  dependencies: [qt5_dep, zlib_dep],
+  dependencies: [qt_dep, zlib_dep],
   cpp_args: quazip_args,
-  override_options: ['cpp_std=c++11']
 )
 
 quazip_incdir = include_directories('.')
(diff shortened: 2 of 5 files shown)


neheb:
```
 subprojects/quazip-1.3/quazip/meson.build:32:0: ERROR: MOC sources specified and couldn't find uic-qt6, please check your qt6 installation
```

No idea what's going on.

jpakkane:
Could maybe be [this](https://github.com/mesonbuild/meson/pull/9988)?

Unrelated to that, since you are active here would you like to get dev rights on this repo? All it basically means that you can review and merge other people's merge requests. (Not your own, though, they still need to be reviewed by someone else. :smile: )

neheb:
Sure

subprojects/packagefiles/quazip/meson.build
```diff
@@ -1,12 +1,18 @@
...
 
-qt5 = import('qt5')
-qt5_dep = dependency('qt5', modules: 'Core')
+qt_dep = dependency('qt5', modules: 'Core', required : false)
+if qt_dep.found()
+  qt = import('qt5')
+else
+  qt = import('qt6')
+  qt_dep = dependency('qt6', modules: 'Core')
+endif
```
eli-schwartz, line 15: Interestingly, upstream has a build variable (not option, remember, "what are options") to select the version of Qt to use, and if unspecified it will go look for all 3 in the order: qt6, qt5, qt4

subprojects/packagefiles/quazip/quazip/meson.build
```diff
@@ -28,15 +29,14 @@ quazip_src = [
   'zip.c'
 ]
 
-quazip_pp = qt5.preprocess(moc_headers: quazip_moc_headers)
+quazip_pp = qt.compile_moc(headers: quazip_moc_headers)
```
eli-schwartz, line 32: I think switching from a non-deprecated method to a method new in 0.59 isn't really best for the WrapDB and we should offer broader support than that.

eli-schwartz (changes requested):

neheb:
Fixed.

eli-schwartz:
Since Meson's support for Qt6 doesn't correctly handle 6.1 yet, it seems reasonable to install qt5 in the macOS CI just like we're forced to do in the Ubuntu CI.

(I'll refrain from commenting about the actual state of reluctance in the Qt ecosystem as a whole.)

neheb:
The issue becomes similar to https://github.com/mesonbuild/wrapdb/pull/310

homebrew installs it to a different directory.

eli-schwartz:
Good grief...

It certainly doesn't log any sort of warning to that effect. And it's impossible to tell what the contents of qt--6.2.3_1.big_sur.bottle.tar.gz are, normally I would just download the urls listed in the logs but Github Packages just says "authentication failed", hides the ..../packages page from the repository main page, and then refuses to provide download urls for absolutely anything ever. So I cannot even wget it and check the tarball file listing. Brilliant!

...

Shouldn't qmake be reporting the "different directory" though?

neheb:
```
 Undefined symbols for architecture x86_64:
  "QTextCodec::codecForName(QByteArray const&)", referenced from:
      QTextCodec::codecForName(char const*) in quazip.cpp.o
  "QTextCodec::codecForLocale()", referenced from:
      QuaZipPrivate::QuaZipPrivate(QuaZip*) in quazip.cpp.o
      QuaZipPrivate::getDefaultFileNameCodec() in quazip.cpp.o
      QuaZipPrivate::QuaZipPrivate(QuaZip*, QString const&) in quazip.cpp.o
      QuaZipPrivate::QuaZipPrivate(QuaZip*, QIODevice*) in quazip.cpp.o
  "QTextCodec::fromUnicode(QString const&) const", referenced from:
      QuaZip::close() in quazip.cpp.o
      QuaZipFile::open(QFlags<QIODeviceBase::OpenModeFlag>, QuaZipNewInfo const&, char const*, unsigned int, int, int, bool, int, int, int) in quazipfile.cpp.o
  "QTextCodec::toUnicode(QByteArray const&) const", referenced from:
      QuaZip::getComment() const in quazip.cpp.o
      QuaZip::getCurrentFileName() const in quazip.cpp.o
      QuaZip::getCurrentFileInfo(QuaZipFileInfo64*) const in quazip.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Makes no sense...

neheb:
Missing module. Builds now.

eli-schwartz:
Nice, thanks for debugging that.

PR #383, opened 2022-04-17
lzo2: fix compilation on windows

neheb:
Same fix as the vcpkg package.


neheb:
```
6/6 lzo2 / lzotest directory check FAIL             0.49s   exit status 127
```

something tells me this only happens with CI.

eli-schwartz:
> something tells me this only happens with CI.

Something tells me this should not disable testing in CI! Just make that one test not be defined on Windows. :/

eli-schwartz:
Fixed via 4feafbacad11a20ea79dc158cba195c362b61788

Issue #214, opened 2021-10-23
flac wrap broken on non-x86


neheb:
https://github.com/MusicPlayerDaemon/MPD/commit/45f3dd8b7acbb1cd34b3a02a4ccb3087447f1783

eli-schwartz:
https://github.com/mesonbuild/wrapdb/blob/3f766fa3d62352530fc57aeeb4c5cdf879ea87bc/subprojects/packagefiles/flac/meson.build#L12-L14

and used by https://github.com/mesonbuild/wrapdb/blob/3f766fa3d62352530fc57aeeb4c5cdf879ea87bc/subprojects/packagefiles/flac/meson.build#L29-L33

In theory this should be working fine. In practice, the mesondefine gets translated into "#define FOO 0" and the project itself uses `#ifdef FOO`, rather than `#if FOO`.

Aside: I do wonder why we don't use `config.cmake.h.in` here.

neheb:
I tested this on an ARMv7 system. Compiles fine.

PR #385, opened 2022-04-18
flac: try to fix on non x86

neheb:
The code uses ifdef instead of if for these. Adjust to not define them
if not found.


neheb:
Fixes: https://github.com/mesonbuild/wrapdb/issues/214

PR #386, opened 2022-04-18
onqtam-doctest: Update to 2.4.8

krnowak:
Update doctest to latest stable version, 2.4.8. In the meantime the project has moved from github.com/onqtam/doctest to its own org - github.com/doctest/doctest. So not sure if the wrap file should be renamed (doctest-doctest.wrap?). I don't know if there is any policy around renaming wraps.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -172,6 +172,14 @@
       "0.6.3-1"
     ]
   },
+  "doctest": {
+    "dependency_names": [
+      "doctest"
+    ],
+    "versions": [
+      "2.4.8-1"
+    ]
+  },
   "dragonbox": {
     "dependency_names": [
       "dragonbox"
--- a/subprojects/doctest.wrap
+++ b/subprojects/doctest.wrap
@@ -1,0 +1,8 @@
+[wrap-file]
+directory = doctest-2.4.8
+source_url = https://github.com/doctest/doctest/archive/refs/tags/v2.4.8.tar.gz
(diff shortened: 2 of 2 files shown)


jpakkane:
> I don't know if there is any policy around renaming wraps.

There isn't one, really. Just add the new `doctest.wrap` on its own and leave the old one be. It needs to remain so that old projects keep on building. When people are ready to update, they'll delete the old wrap file and add a new one.

krnowak:
> > I don't know if there is any policy around renaming wraps.
> 
> There isn't one, really. Just add the new `doctest.wrap` on its own and leave the old one be. It needs to remain so that old projects keep on building. When people are ready to update, they'll delete the old wrap file and add a new one.

Done.

This gave me a thought that probably adding some way of marking wraps as deprecated could help here, like:
```ini
[wrap-file]
deprecated_for = doctest # or doctest.wrap
deprecated_message = "Switch to doctest"
…
```

eli-schwartz:
It's not entirely clear to me why it had the former name to begin with, honestly.

krnowak:
> It's not entirely clear to me why it had the former name to begin with, honestly.

No idea, maybe to avoid confusion with python doctest? Not that this would make sense.

jpakkane:
I did some web searching and it seems that there is no other package called `doctest` for compiled code. There is one for Haskell and another for Python but that seems to be it. So we can probably call this just `doctest` without name clashes.

PR #387, opened 2022-04-18
Add a missing comma

krnowak:
It broke in #380.


jpakkane:
Thanks, but since I broke it I'll also get it fixed in #388.

krnowak:
> Thanks, but since I broke it I'll also get it fixed in #388.

Sorry, just spotted it while working on doctest update, so did the drive-by fix.

jpakkane:
No prob. This one got in fact merged before I could do mine so it's all good.

PR #392, opened 2022-04-19
protobuf-c: update to 1.4.1

neheb:

### DIFF
--- a/subprojects/packagefiles/protobuf-c/meson.build
+++ b/subprojects/packagefiles/protobuf-c/meson.build
@@ -1,45 +1,13 @@
-project('protobuf-c', 'c', 'cpp', version : '1.3.0', license : 'BSD-2')
+project(
+  'protobuf-c',
+  'c',
+  version: '1.4.1',
+  license: 'BSD-2',
+  meson_version: '>=0.50.0',
+)
 
-c = meson.get_compiler('c')
-cc = meson.get_compiler('cpp')
+incdir = include_directories('.')
 
-libprotobuf_c_src = [
-  'protobuf-c/protobuf-c.c',
-]
-
-protobuf_deps = dependency('protobuf',
-                           fallback: ['protobuf', 'protobuf_dep'],
-                           version: '>= 3.0.0')
-libprotobuf_c = library('protobuf-c',
-                        libprotobuf_c_src,
-                        cpp_args: [],
-                        dependencies: protobuf_deps,
-                        include_directories: include_directories('protobuf-c'))
-
-protoc_c_src = [
-  'protoc-c/c_bytes_field.cc',
-  'protoc-c/c_enum.cc',
-  'protoc-c/c_enum_field.cc',
-  'protoc-c/c_extension.cc',
-  'protoc-c/c_field.cc',
-  'protoc-c/c_file.cc',
-  'protoc-c/c_generator.cc',
-  'protoc-c/c_helpers.cc',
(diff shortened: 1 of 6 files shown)


subprojects/packagefiles/protobuf-c/meson.build
```diff
@@ -1,20 +1,23 @@
...
+  'protobuf',
+  fallback: ['protobuf', 'protobuf_dep'],
+  version: '>= 3.0.0'
+)
+
+libprotobuf_c = library(
+  'protobuf-c',
+  libprotobuf_c_src,
+  dependencies: protobuf_deps,
+  include_directories: include_directories('.')
```
eli-schwartz, line 19: This changed from "protobuf-c" apparently? Note that "." is already added by default...

subprojects/packagefiles/protobuf-c/meson.build
```diff
@@ -33,13 +36,19 @@ protoc_c_src = [
...
+)
 
-protoc_c = executable('protoc-c',
-                      protoc_c_src,
-                      cpp_args: '-DPACKAGE_STRING="protobuf-c"',
-                      include_directories: include_directories('protoc-c',
-                                                               'protobuf-c'),
-                      dependencies: protoc_deps)
+run_command('protoc', '--cpp_out', meson.current_build_dir(), '--proto_path', meson.current_source_dir(), meson.current_source_dir() + '/protobuf-c/protobuf-c.proto', check : true)
+run_command(protoc_c, meson.current_build_dir() + '/protobuf-c/protobuf-c.pb.cc', meson.current_build_dir() + '/protobuf-c/protobuf-c.pb.h')
```
eli-schwartz, line 54: What is this doing??? You cannot use a build target in run_command anyway, but on general principle why would this be a run_command rather than a custom_target?

However, it is also not entirely clear to me what it is trying to accomplish. Maybe it is supposed to test running the built software? Wouldn't that be better as a `test()`?

eli-schwartz (changes requested):

subprojects/packagefiles/protobuf-c/meson.build
neheb: eh no. It's building files needed for compilation.

eli-schwartz: It's not clear to me where/when this is needed for compilation, and it still needs to be a custom_target (or generator) in that case.

neheb: sounds about right.

neheb:
@eli-schwartz 

```
@BUILD_ROOT@/subprojects/protobuf-c-1.4.0/
```

feels like a bad hack. anyway to get the subproject root or something?

neheb:
```
subprojects/protobuf-c-1.4.0/meson.build:11:0: ERROR: C++ shared or static library 'protoc' not found
```

completely lost as to why.

neheb:
```
subprojects/protobuf-c-1.4.1/meson.build:5:0: ERROR: C++ shared or static library 'protoc' not found
```

I get the same error on Debian even though:

```
~> find /usr | grep protoc.s
/usr/lib/arm-linux-gnueabihf/libprotoc.so.23.0.4
/usr/lib/arm-linux-gnueabihf/libprotoc.so.23
```

I can't figure out why meson is unable to find it.

neheb:
@jpakkane this is ready

PR #393, opened 2022-04-19
argparse: update to 2.3

ericonr:
This is a draft for now, pending on a release that includes https://github.com/p-ranav/argparse/pull/170 , so we don't introduce any new warnings for users.


releases.json
```diff
@@ -42,7 +42,8 @@
       "argparse"
     ],
     "versions": [
-      "2.2-1"
+      "2.2-1",
+      "2.3-1"
```
eli-schwartz, line 46: The order of these needs to be reversed.

eli-schwartz (changes requested):

releases.json
ericonr: Thanks!
And hi :)

eli-schwartz: hi :)

PR #397, opened 2022-04-20
hinnant-date: update to 3.0.1

neheb:

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -7,13 +7,13 @@
                "explicitly disable that Operating System in build_on.",
                "Please remove projects from those lists once they are updated."],
   "broken_linux": [
-    "arduinocore-avr", "cpr", "glew", "hinnant-date", "imgui-sfml",
+    "arduinocore-avr", "cpr", "glew", "imgui-sfml",
     "libxmlpp", "openal-soft", "openh264",
     "protobuf-c", "quazip", "zstd"
   ],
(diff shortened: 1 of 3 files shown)


jpakkane:
LGTM, but currently ⚔️ conflicts.

jpakkane:
Still does...

neheb:
The snappy commit broke it. Disadvantage of avoiding clang-format on ci_config.json.

neheb:
Fixed again.

jpakkane:
Great, thanks.

eli-schwartz:
I'm disappointed that this PR was merged despite it being *literally* a duplicate of #305

If it was acceptable to merge an update that intentionally disables CI because the wrap is conceptually flawed, then we should have merged #305 -- if not, then my review comments for #305 still apply to this PR. Either way it should not be merged.

:/

eli-schwartz:
So I just want to clarify -- "then we should have merged #305".

The main problem I have with this PR is it seems fundamentally dishonest to reject someone's work "because foo", then have someone else resubmit the same PR under their own name, not fix "foo", and be golden.

In short, I'm disappointed that you left that PR to rot instead of responding *there* that "this is a 1:1 port of upstream". Regardless of anything to do with specific wraps, I'd greatly appreciate it if we could do better than that in the future.

PR #399, opened 2022-04-21
Adds tracy

tmayoff:


eli-schwartz:
Seems like upstream directly sets c++11 per target, but this technically doesn't work on MSVC. You might need to PR a fix for that directly to tracy (check compiler.get_id() == 'msvc')

tmayoff:
I fixed this upstream (https://github.com/wolfpld/tracy/pull/370) but until they release a new version I've added a patch here

neheb (approved):

eli-schwartz:
@jpakkane I'm slightly confused why we're including a patched version of the upstream meson.build? This is generally considered... suboptimal... in the WrapDB?

neheb:
@eli-schwartz isn't this the situation with zstd?

jpakkane:
It is a bit unoptimal, yes. But since:

- the fix is upstream
- the change is only for the build system, not for functionality
- having a slightly modified version in WrapDB means people can use it (as opposed to having nothing)

I figured merging would be better than waiting until the next release (which might take a while).

eli-schwartz:
@neheb zstd is actually a very weird case, because upstream provides meson.build files but only in a subdirectory of the project (which meson doesn't allow). At the end of the day, they don't want it in the project root because they don't want it to be "official", so in order to use it you need to acquire zstd, then copy the topmost meson.build up a few directories so meson can detect it.

They won't budge, and meson won't work without the changes. At least zstd upstream is amenable to structuring things so that most of the logic isn't in that topmost meson.build, so we only need to include a stub loader.

@jpakkane I guess... I'm curious how long it's likely to be until a release.

tmayoff:
Before adding the patch I did look through the releases to try and find some 'schedule' or pattern, however the last release 0.8.1 was 4 days ago, before that was 0.8 which was almost a year after the release before. It seems to be anywhere from a month to  a lot longer than that. I can contact the maintainer to see if there is a plan

eli-schwartz:
I guess it sounds like it's not a high priority to wait for a rapid release, then. :) Thanks for confirming.

PR #401, opened 2022-04-22
abseil-cpp: define NOMINMAX with all msvc-style compiler frontends.

lukester1975:
As per subject, really. Currently fails with clang-cl. Obviously various ways to fix, but here's one...

Thanks


jpakkane:
You need to add a new release in the release json file.

neheb:
Sounds like most cc.get_id() calls here would need to be changed...

lukester1975:
OK, now with updated release.json.

As for other msvc test, it's not actually necessary to change it (`abseil-cpp| Checking if "GCC atomic builtins" : links: YES`; at least with clang-cl 13) but can do if that's preferred...

neheb:
```
FAIL: test_releases_json (__main__.TestReleases)
  File "./tools/sanity_checks.py", line 77, in test_releases_json
    self.assertIn(version, self.releases[name]['versions'], f"for {name}")
AssertionError: '5.0.1-1' not found in ['3.0.1-2', '3.0.1-1'] : for libxmlpp
```

What a strange error...

jpakkane:
I _think_ that should be fixed with a rebase.

PR #403, opened 2022-04-22
Add new wrap for TartanLlama/expected

punytroll:
- header only library
- without tests

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -250,6 +250,14 @@
       "2.2.5-1"
     ]
   },
+  "expected": {
+    "dependency_names": [
+      "expected"
+    ],
+	"versions": [
+	  "1.0.0-1"
+	]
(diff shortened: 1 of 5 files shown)


jpakkane:
The [corresponding optional library] is called `tl-optional` in WrapDB, so this should be called `tl-expected` for consistency.

jpakkane:
I can squash this now, but in the future please keep MRs to a single commit.. Thanks.

punytroll:
Thanks for the squash merge.

I saw, that other pull requests always force-pushed their changes resulting in one commit only but I am not familiar with the workflow. Could you please give me a pointer? Tanks!

jpakkane:
Basically you do your changes and then:

    git commit -a --amend

and then

    git push --force

PR #404, opened 2022-04-24
libccp4c: Update source URL

taw10:
The old FTP server broke and went away, but was replaced with an HTTP server at the same location.


jpakkane:

taw10:
Whoops, sorry!

PR #406, opened 2022-04-25
Add json-glib wrap

xclaesse:


ci_config.json
```diff
@@ -86,6 +86,12 @@
       "htslib:tests=true"
     ]
   },
+  "json-glib": {
+    "_comment": "gettext is not optional: https://gitlab.gnome.org/GNOME/json-glib/-/merge_requests/50",
+    "debian_packages": [
+      "gettext"
```
eli-schwartz, line 92: Tbh it basically is optional. It's historically been optional with a warning, for people that build projects in specialized workflows before gettext is available. Your PR would make it optional via a project option instead.

What's the eventual end goal of the wrapdb here? To shed this dependency? It seems like a mild one but eh, to each his own.

xclaesse: The warning makes the CI fail because of --fatal-mesin-warnings. But indeed maybe it wasn't worth doing a PR in json-glib for that.

Fun fact, it works for visual studio CI because perl provide xgettext...

xclaesse: > What's the eventual end goal of the wrapdb here? To shed this dependency? It seems like a mild one but eh, to each his own.

I wrote the wrap for gegl that I've been trying to build recently, as you saw. My end goal is to have native meson projects in wrapdb, so we can have automatic fallback without having to copy and maintain wrap files everywhere: https://github.com/mesonbuild/meson/pull/8941.

PR #411, opened 2022-04-27
tree-sitter: runtime library

albfan:
https://github.com/tree-sitter/tree-sitter

a library to generate AST trees

This only provides the runtime

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1563,6 +1563,14 @@
       "2.5.0-1"
     ]
   },
+  "tree-sitter": {
+    "dependency_names": [
+      "tree-sitter"
+    ],
+    "versions": [
+      "0.15.7"
+    ]
+  },
   "tracy": {
     "dependency_names": [
       "tracy"
--- a/subprojects/packagefiles/tree-sitter/meson.build
+++ b/subprojects/packagefiles/tree-sitter/meson.build
@@ -1,0 +1,29 @@
+project('tree-sitter', 'c',
+  version: '0.15.7',
+  license: 'MIT',
+  meson_version: '>=0.46.0',
+)
+
+tree_sitter = library('tree-sitter', 'lib/src/lib.c',
+  include_directories: [
+    include_directories('lib/src'),
+    include_directories('lib/include'),
+    include_directories('lib/utf8proc')
+  ],
+  soname: '0.0',
+  install:true,
+)
+
+install_headers('lib/include')
+
+tree_sitter_dep = declare_dependency(
+  link_with: tree_sitter,
+  include_directories: include_directories('lib/include'),
+)
+
+
+pkg = import('pkgconfig')
+pkg.generate(tree_sitter,
+             version : '0.15.7',
+             name : 'libtree-sitter',
+             description : 'An incremental parsing system for programming tools.')
--- a/subprojects/tree-sitter.wrap
+++ b/subprojects/tree-sitter.wrap
@@ -1,0 +1,10 @@
+[wrap-file]
+directory = tree-sitter-0.15.7
+source_url = https://github.com/tree-sitter/tree-sitter/archive/refs/tags/0.15.7.tar.gz
+source_filename = tree-sitter-0.15.7.tar.gz
+source_hash = 459e69606d396ffc92492b8cbbeb2ced2153f005b87be370f6b42fd412b527fe
+patch_directory = tree-sitter
+
+[provide]
+libtree-sitter = libtree-sitter_dep
+


albfan:
cc @tristan957

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,17 @@
+project('tree-sitter', 'c')
+
+tree_sitter_lib = static_library('tree-sitter-runtime', [
+  'lib/src/lib.c',
+  ]
```
eli-schwartz, line 5: Upstream's Makefile creates and installs this as 'tree-sitter', and generates both static and shared libraries, so you should not hardcode it as static but rather allow the user to select it via `-Ddefault_library=`.

lib/src/lib.c is an amalgamated file that is updated less frequently than lib/src/*.c, I haven't checked when/how often it gets updated. :/ Not sure whether that matters. The Makefile only uses the amalgamation with `make AMALGAMATED=1`.

This should have an install target, including installing lib/include/tree_sitter/ to includedir, and generate a pkg-config file.

subprojects/tree-sitter.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = tree-sitter-0.15.7
+source_url = https://github.com/tree-sitter/tree-sitter/archive/refs/tags/0.15.7.tar.gz
+source_filename = 0.15.7.tar.gz
+source_hash = 98e91c7c6bf162bf90e4e70fdbc41a8188b9fa8de5ad840c401198014406ce9e
+patch_directory = tree-sitter
+
+[provide]
+libzstd = libzstd_dep
```
eli-schwartz, line 9: Copy/paste error?

eli-schwartz (changes requested):

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,17 @@
...
+,
```
tristan957, line 6: clean up your white space. Lots of misalignment

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,17 @@
+project('tree-sitter', 'c')
+
+tree_sitter_lib = static_library('tree-sitter-runtime', [
```
tristan957, line 3: use `library`

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,17 @@
+project('tree-sitter', 'c')
```
tristan957, line 1: Add a version and the license please :)

subprojects/packagefiles/tree-sitter/meson.build
tristan957, line 3: I would almost say remove the `_lib` suffix for this variable. Maybe others disagree

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,17 @@
...
+  include_directories: [
+    include_directories('lib/src'),
+    include_directories('lib/include'),
+    include_directories('lib/utf8proc')
+    ],
+  )
+
+tree_sitter_dep = declare_dependency(
+  link_with: tree_sitter_lib,
+  include_directories: include_directories(join_paths('lib', 'include')),
```
tristan957, line 16:
```suggestion
  include_directories: include_directories('lib/include'),
```

subprojects/tree-sitter.wrap
tristan957, line 9: whoops :)

tristan957 (changes requested):

tristan957:
Off to a good start. Are there any build options for tree-sitter that the Meson port should be aware of? WrapDB usually also requires the tests to also be built.

Does tree-sitter install a pkg-config file?

subprojects/packagefiles/tree-sitter/meson.build
albfan: ok, makes sense, thanks. Working on it

subprojects/tree-sitter.wrap
albfan: yep

subprojects/packagefiles/tree-sitter/meson.build
albfan: done

subprojects/packagefiles/tree-sitter/meson.build
albfan: ok, solved

albfan: sure

subprojects/tree-sitter.wrap
albfan: solved

subprojects/packagefiles/tree-sitter/meson.build
albfan: solved

eli-schwartz:
@tristan957 speaking as the author of the upstream Makefile:

- Yes, one option (it chooses to use the amalgamated source)
- there are none
- yes

albfan:
I'm working on fix the amalgamated.

About testing I have examples to parse json or similar, https://gitlab.gnome.org/albfan/vala-tree-sitter/-/blob/master/src/main.vala#L36 but I'm unsure what to test on this runtime itself, it should be the plumbings to use parsers on it like #412

albfan:
I provided pkg-config in latest commit

tristan957:
Cool. So should the amalgamated source be a custom_target() of some sort?

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,25 @@
...
+)
+
+tree_sitter_dep = declare_dependency(
+  link_with: tree_sitter,
+  include_directories: include_directories('lib/include'),
+)
+
+
+pkg = import('pkgconfig')
+pkg.generate(libraries : tree-sitter,
```
tristan957, line 20: you can't have a - there

subprojects/packagefiles/tree-sitter/meson.build
albfan: done

subprojects/packagefiles/tree-sitter/meson.build
albfan: opps

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,28 @@
+project('tree-sitter', 'c'
```
eli-schwartz, line 1: syntax error

albfan:
Honestly looks like reinventing the wheel. As library uses a custom Makefile, didn't look into it much.

Point to sources should clarify sources and what to build.

latest version is 0.20.6, but several parsers fail there, so I stick to 0.15.7

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,28 @@
...
+  version: '0.15.7',
+  license: 'MIT',
+)
+
+tree_sitter = library('tree-sitter', [
+  'lib/src/lib.c',
+  ],
```
eli-schwartz, line 8: Source files don't need to be wrapped in arrays, the parser expands them into varargs, see https://mesonbuild.com/Syntax.html#argument-flattening

Internally, it just sees `library('tree-sitter', 'lib/src/lib.c', include_directories: ..., ...)`

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,28 @@
...
+)
+
+tree_sitter_dep = declare_dependency(
+  link_with: tree_sitter,
+  include_directories: include_directories('lib/include'),
+)
+
+
+pkg = import('pkgconfig')
+pkg.generate(libraries : tree_sitter,
```
eli-schwartz, line 23:
```suggestion
pkg.generate(tree_sitter,
```

See the pkgconfig module documentation:
> Since 0.46 a StaticLibrary or SharedLibrary object can optionally be passed as first positional argument. If one is provided a default value will be provided for all required fields of the pc file:

eli-schwartz: You will need to then add `meson_version: '>=0.46.0',` to the project() declaration.

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,28 @@
...
+             subdirs : 'lib/include',
+             version : '0.15.7',
+             name : 'libtree-sitter',
+             filebase : 'simple',
```
eli-schwartz, line 27: What is this filebase doing here?

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,28 @@
...
+
+tree_sitter_dep = declare_dependency(
+  link_with: tree_sitter,
+  include_directories: include_directories('lib/include'),
+)
+
+
+pkg = import('pkgconfig')
+pkg.generate(libraries : tree_sitter,
+             subdirs : 'lib/include',
```
eli-schwartz, line 24: Subdirs are supposed to reflect the installed location, which is /usr/include -- there is no /usr/include/lib/include subdir, and anyway upstream uses `Cflags: -I${includedir}` so there is no subdir.

subprojects/tree-sitter.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = tree-sitter-0.15.7
+source_url = https://github.com/tree-sitter/tree-sitter/archive/refs/tags/0.15.7.tar.gz
+source_filename = 0.15.7.tar.gz
```
eli-schwartz, line 4: This local save-to-filename should include "tree-sitter" at the beginning, rather than just containing the version.

subprojects/tree-sitter.wrap
```diff
@@ -0,0 +1,10 @@
...
+source_hash = 98e91c7c6bf162bf90e4e70fdbc41a8188b9fa8de5ad840c401198014406ce9e
```
eli-schwartz, line 5: This is another zstd.wrap copy paste error.

eli-schwartz (changes requested):

subprojects/packagefiles/tree-sitter/meson.build
eli-schwartz: The install target means also `install: true` on the library.

albfan: apart from specify all detailed sources I think this is done

eli-schwartz: Upstream also has a `soname: '0.0'` equivalent in the Makefile.

releases.json
```diff
@@ -1563,6 +1563,14 @@
       "2.5.0-1"
     ]
   },
+  "tree-sitter": {
+    "dependency_names": [
+      "tree-sitter"
+    ],
+    "versions": [
+      "0.15.7"
```
eli-schwartz, line 1571: All versions in this file must end in a hyphen followed by an integer indicating the WrapDB revision, starting at 1.

eli-schwartz (changes requested):

eli-schwartz:
Now I remember... lib/src/lib.c isn't a "proper" amalgamation anyway, it's really just a unity build lol.

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,29 @@
...
+
+tree_sitter_dep = declare_dependency(
+  link_with: tree_sitter,
+  include_directories: include_directories('lib/include'),
+)
+
+
+pkg = import('pkgconfig')
+pkg.generate(tree_sitter,
+             version : '0.15.7',
```
tristan957, line 27: Make spacing consistent like other lines

subprojects/packagefiles/tree-sitter/meson.build
```diff
@@ -0,0 +1,29 @@
...
+  include_directories: [
+    include_directories('lib/src'),
+    include_directories('lib/include'),
+    include_directories('lib/utf8proc')
+  ],
+  soname: '0.0',
+  install:true,
+)
+
+install_headers('lib/include')
```
tristan957, line 17: I haven't looked, but does this need to be under a subdir?

Also does install_headers() actually work like this?

tristan957:
May want to add meson.override_dependency(). Not sure what the WrapDB policy is

subprojects/packagefiles/tree-sitter/meson.build
eli-schwartz: It doesn't, and once the CI error in releases.json is fixed the CI will actually run meson and provide error messages to the effect that this isn't a file on disk.

albfan:
Will work on this locally and reopen when is at least working for me.

PR #413, opened 2022-04-27
libuv: Require C99 for C++ style comments

xclaesse:
The c_std=c89 option on subproject was previously ignored by Meson, but
with master (used by build-all CI) this started to fail.

### DIFF
--- a/subprojects/packagefiles/libuv/meson.build
+++ b/subprojects/packagefiles/libuv/meson.build
@@ -1,7 +1,7 @@
 project('libuv', 'c',
   version : '1.43.0',
   license : 'libuv',
-  default_options : [ 'c_std=c89' ],
+  default_options : [ 'c_std=c99' ],
 )
 
 # get compiler type


xclaesse:
A bunch of wraps regressed on Windows too because of per subproject std: pugixml, sdl2_mixer, sdl2_net, sfml, taglib.

neheb:
what happened with taglib?

xclaesse:
@neheb
```
2022-04-24T01:03:06.9138603Z subprojects\taglib-1.12\meson.build:1:0: ERROR: Value "c++98" (of type "string") for combo option "C++ language standard to use" is not one of the choices. Possible choices are (as string): "none", "c++11", "vc++11", "c++14", "c++latest", "vc++latest", "vc++14", "c++17", "vc++17", "c++20", "vc++20".
```

You can see logs in CI: https://github.com/mesonbuild/wrapdb/actions/workflows/build_all.yml

eli-schwartz:
This would be one of those cases where what we actually want is an allowable range of std versions rather than a single hardcoded one, I guess...

neheb:
Fascinating. I wonder if taglib compiles as C++14 or w/e.

eli-schwartz:
For example pugixml, does not specify a std at all. It worked in February, but likely got broken by #380

The failing logs show that pugixml is miscompiling with command lines that include `"/std:c++17"`.

neheb:
Seems this is fixed in the next version: https://github.com/zeux/pugixml/commit/1312a46d91623d7a970f8248b0655e7ac28a8463

xclaesse:
> edit: that is, version 1.12 is present currently. 1.12.1 fixes it.

mind doing a PR to update it please?

eli-schwartz:
That pugixml commit claims to be present in 1.10 too.

neheb:
Huh yeah. Wonder why it fails...

neheb:
There's a new version. I'll do the update.

PR #414, opened 2022-04-27
glib: Rename gio-win32-2.0 to gio-windows-2.0

xclaesse:
This is to match the pkg-config dependency name. Unfortunately the
override_dependency() was wrong in glib too, will be fixed in the next
stable release. See
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2616.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -429,7 +429,7 @@
       "gmodule-2.0",
       "glib-2.0",
       "gio-2.0",
-      "gio-win32-2.0",
+      "gio-windows-2.0",
       "gio-unix-2.0"
     ],
     "program_names": [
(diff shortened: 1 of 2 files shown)


xclaesse:
This cannot be merged until we get a glib release with the fix.

xclaesse:
Glib 2.72.2 got released with the fix.

xclaesse:
@eli-schwartz any objection on this?

eli-schwartz:
Makes perfect sense. I'm pretty sure I filled out the dependency_names with a quick grep for override_dependency, and these definitely should be kept in sync with what is actually provided!
Nice catch on the pkg-config/override_dependency mismatch.

PR #415, opened 2022-04-27
SDL2: update to 2.0.22

Jan200101:


Jan200101:
I believe the CI failure is caused by a bug in SDL2.
I'm unsure if it truly is and what the right way of fixing it would be, so I reported it upstream https://github.com/libsdl-org/SDL/issues/5580

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -231,7 +231,7 @@ wayland_libdecor_dep = dependency('libdecor-0', required : opt_video_wayland_lib
 
 if wayland_scanner_dep.found()
     wayland_scanner = find_program(
-        wayland_scanner_dep.get_pkgconfig_variable('wayland_scanner'),
+        'wayland-scanner',
```
Jan200101, line 234: Naively looking for `wayland-scanner` is how both [SDL (CMake)](https://github.com/libsdl-org/SDL/blob/05bd225a80556412cff052d74a72c5bf8333e403/cmake/sdlchecks.cmake#L616) and and [GTK (Meson)](https://github.com/GNOME/gtk/blob/ff4b5c89965da95eb70441d1e2b1818154cc59c8/gdk/wayland/meson.build#L41) do it

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -840,6 +840,11 @@ elif platform_is_windows
         sys_joystick += 'rawinput'
     endif
 
+    # workaround for https://github.com/libsdl-org/SDL/issues/5580
+    if cdata.get('HAVE_XINPUT_H', 0) == 0
+        cdata.set('HAVE_WINDOWS_GAMING_INPUT_H', 0)
+    endif
+
```
Jan200101, line 847: This is the best we can do to deal with an SDL bug.

subprojects/packagefiles/sdl2/meson.build
eli-schwartz: If you're already looking up wayland-scanner as a dependency then it should have the right location for you. If you want to look it up the naive way, why look up that dep at all?

Jan200101: This is mainly intended as future proofing since dependency.get_pkgconfig_variable has been deprecated with 0.56.0 according to the meson output

```
sdl2| NOTICE: Future-deprecated features used:
sdl2| * 0.56.0: {'dependency.get_pkgconfig_variable'}
sdl2| * 0.58.0: {'meson.get_cross_property'}
```

Since this is also the behavior that is used upstream as well as in other meson projects I believe this is the right way to go forward

Jan200101:
I tried to work around the previously mentioned bug and explicitly disabled WGI when XInput was  unavailable but that only showed me another error.

wrapdb will probably have to skip out on 2.0.22

subprojects/packagefiles/sdl2/meson.build
eli-schwartz: Deprecated doesn't mean you cannot use it. It doesn't even mean you should not use it.

It just means that in the future, you should switch to `dependency.get_variable(pkgconfig: 'wayland_scanner')`, but cannot yet (because that requires a new minimum version of Meson). They behave the same, except the new one is more flexible.

subprojects/packagefiles/sdl2/meson.build
Jan200101: I tried moving this into th header, but that just exposed another SDL bug

subprojects/packagefiles/sdl2/meson.build
Jan200101: I am aware
the question is when will it get removed and how far back does SDL need to support meson.

But looking a bit further, it seems SDL uses BOTH solutions with autoconf and CMake
https://github.com/libsdl-org/SDL/blob/05bd225a80556412cff052d74a72c5bf8333e403/configure.ac#L1561

At which point the question becomes, what is the right way of doing it?

I'll look into it some more and if I cannot find the suggested upstream way of getting the wayland-scanner I'll revert this

eli-schwartz: > the question is when will it get removed and how far back does SDL need to support meson.

It requires effectively no maintenance and we are loath to remove support for functionality people are using. Plus, 0.56 is quite recent. I wouldn't worry about it.

At the very least we're not going to remove support for anything that was deprecated later than the very oldest versions of Meson people are still using. For example, the oldest supported Ubuntu version of meson for at least another year is 0.45, and there may be older systems out there, *and* we don't want to break projects that are no longer actively developed.

jpakkane: There are no rules on how old versions of Meson wraps need to support. There are some that require the latest release even. I suspect that people using wraps also have fairly recent Meson versions, especially since we changed the web API at some point and new wraps only go to the new system. I'm not sure but AFAICR the Meson version in Ubuntu LTS is too old to support the new web API so they could not use these even if they wanted to.

eli-schwartz: Right, but that was not my point. My point is that you don't need to worry about upgrading existing WrapDB uses of the older API if your only reason to do so is fear that Meson is going to remove the API, because we are basically not going to remove the API in the foreseeable future.

(Personally, I prefer to make them support the oldest versions of Meson that it is easy and practical to support, but use newer versions of Meson at any time if they seem to provide desired functionality that doesn't exist in older versions of Meson. For example, I encourage the use of feature options which don't work with Ubuntu 18.04 LTS, but I don't encourage source_root -> project_source_root because they are the same and it's not worth it.)

Jan200101: I've looked at various other projects to make up my mind on what the more widely used solution is and I have come to the conclusion that naively looking for the binary is less common outside of a few circles, thus I think its worth reverting this

subprojects/packagefiles/sdl2/meson.build
Jan200101: The bugs have been fixed with https://github.com/libsdl-org/SDL/pull/5592, now its only a matter of waiting for a release

eli-schwartz: Great!

Any timeframe on the release? Given the current release is barely a week old, will we be waiting for a while?

takase1121:
There is a problem with statically building and linking SDL to a program in Windows, whereby the program tries to load libiconv dll while everything should be statically linked. [This line](https://github.com/mesonbuild/wrapdb/blob/174ad7cab61a34f84263f865eae5c4fa9d6a6158/subprojects/packagefiles/sdl2/meson.build#L339) might be the reason, instead of simply adding it as optional system library, it should be [something like this](https://github.com/libsdl-org/SDL/blob/05bd225a80556412cff052d74a72c5bf8333e403/CMakeLists.txt#L929), checking if iconv exists and setting the macro.

Jan200101:
> There is a problem with statically building and linking SDL to a program in Windows, whereby the program tries to load libiconv dll while everything should be statically linked. [This line](https://github.com/mesonbuild/wrapdb/blob/174ad7cab61a34f84263f865eae5c4fa9d6a6158/subprojects/packagefiles/sdl2/meson.build#L339) might be the reason,

I do not see how this would cause it to be dynamically linked

> instead of simply adding it as optional system library, it should be [something like this](https://github.com/libsdl-org/SDL/blob/05bd225a80556412cff052d74a72c5bf8333e403/CMakeLists.txt#L929), checking if iconv exists and setting the macro.

We do a similar check

subprojects/packagefiles/sdl2/meson.build
Jan200101: there is no publicly available timeframe, but I'd say about 2-4 months at most.

neheb:
new version of SDL2 is out.

Guldoman:
> new version of SDL2 is out.

The 2.23.1 version is a prerelease version, so it might be better to wait for 2.24.0

neheb:
2.24 was released.

PR #425, opened 2022-05-06
Add `wrapdb_version` field when generating wraps

xclaesse:


xclaesse:
This will be used by Meson to know if a wrap with no patch comes from WrapDB and which version does it have. See https://github.com/mesonbuild/meson/pull/10357.

PR #426, opened 2022-05-10
mpdecimal: Update to release 2.5.1

dnicolodi:
Windows build definition not implemented yet.

This is in preparation to update to the new upstream release 2.5.1: the work is done but my understanding is that the CI needs to be enabled for other updates to be considered.


eli-schwartz:
> the CI needs to be enabled for other updates to be considered.

You could do both in one PR, surely?

What was the specific problem with Windows, by the way?

subprojects/packagefiles/mpdecimal/meson.build
```diff
@@ -194,9 +194,9 @@ mpd_header_config_fullcov = '''
 machine = get_option('machine')
 if machine == 'auto'
   if cc.sizeof('size_t') == 8
-    if conf.get('HAVE_GCC_ASM_FOR_X64') == 1
+    if conf.get('HAVE_GCC_ASM_FOR_X64', 0) == 1
```
eli-schwartz, line 197: None of these `conf.get()`s are actually used by the software. They match the configure.ac from upstream, which defines them as `AC_DEFINE` but never use them... they are only checked later in configure.ac to set the machine as we do here.

eli-schwartz: I think 0ca96efc582e55711fdb23edf0412b0a411a4932 is a better general fix for this.

dnicolodi:
Thank you for the review and the work on improving the build.

The only reason for not submitting the improved 2.5.0 build and the 2.5.1 build based on it in the same PR is that the Ci is not happy if I do so. The only reason not to include the Windows build in the PR is that I don't have an host where to test the build easily available and I am not familiar enough with MSVC compilers to write the build recipe blindly and use the CI to validate it: it would be a too loose iterative process. However, I see that you already done this work too.

Unfortunately, release 2.5.1 changes the build more extensively that expected. I'll wait for you to merge your patches and then I'll rebase the 2.5.1 work on top.

By the way, does Meson support AIX? mpdecimal 2.5.1 introduces support for AIX, but it seems that Meson does not support building on AIX yet.

eli-schwartz:
AIX support: https://mesonbuild.com/Release-notes-for-0-56-0.html#preliminary-aix-support

It's not exactly tested in CI, either for Meson or the WrapDB, and we're entirely dependent on third-party contributions for that, but adding untested support code for it and hoping it works is fine if you want to do it.

dnicolodi:
I inferred that AIX was not supported from the table here https://mesonbuild.com/Reference-tables.html#operating-system-names

It seems that most of mpdecimal build support for AIX is dealing with the xlC compiler, thus I think there isn't much to be done in Meson for that, until support for this compiler appears.

eli-schwartz:
That makes sense, yeah.

eli-schwartz:
> I'll wait for you to merge your patches and then I'll rebase the 2.5.1 work on top.

I pushed my fixes, please rebase now. :)

dnicolodi:
> I pushed my fixes, please rebase now. :)

Shouldn't a 2.5.0-3 release have been created with your changes?

eli-schwartz:
We're in the middle of upgrading to 2.5.1, though.

Is there a compelling reason to offer the .0 version with these changes? Who is the target use case that isn't just going to use the .1 version?

IMHO this is similar to the reason we don't typically bother updating version X.Y to X.Y+1 if X.Y+2 is available and we can go straight to that. We have done it on request though, if people said that they believe they do/will need it.

dnicolodi:
It would be just for completeness and for having the CI run on the last changes.

dnicolodi:
Is there a policy about downloading material as part of the build? I would prefer not to do it, but the mpdecimal tests depend on test data that is not distributed as part of the mpdecimal tarball and the original test runner downloads them as needed.

dnicolodi:
I pushed the upgrade to release 2.5.1 and I added on top two commits that allow to build the examples and the tests. Running the tests requires downloading the test corpus. I added a python script that does that. I am not sure this is acceptable.

jpakkane:
We generally don't want to download things behind the users' backs. An alternative approach would be that there is no option but instead the code uses the filesystem module to check if the needed data files are in some expected path. If the user has manually downloaded the data then it will be used automatically and if it's not there then the corresponding parts are not built or tun.

dnicolodi:
How would that work for running the tests as part of the wrapdb CI?

dnicolodi:
Also, would the test for the required data be at configure time or at the time the tests are run? I was thinking to use a `custom_target` to execute the script that downloads the test data, but this requires the check at run time.

dnicolodi:
I removed support for the tests that require additional data files.

subprojects/packagefiles/mpdecimal/meson.build
```diff
@@ -202,66 +200,120 @@ if machine == 'auto'
   endif
 endif
 
+
+add_project_arguments(cc.get_supported_arguments('-Wno-unknown-pragmas'), language: ['c', 'cpp'])
+add_project_arguments('-DNDEBUG', language: ['c', 'cpp'])
```
eli-schwartz, line 205: Meson has a `default_options: 'b_ndebug=...']` option.

subprojects/packagefiles/mpdecimal/meson.build
```diff
@@ -202,66 +200,120 @@ if machine == 'auto'
...
-  add_project_arguments('-DUNIVERSAL', language : 'c')
-  conf.set('MPD_HEADER_CONFIG', mpd_header_config_universal)
+  add_project_arguments('-DUNIVERSAL', language: 'c')
+  conf.set('MPD_HEADER_CONFIG', mpd_header_config_UNIVERSAL)
+  if host_machine.system() == 'windows'
+    error('machine "universal" not supported on Windows')
+  endif
+
+else
+  error('unknown machine type "{0}"', machine)
```
eli-schwartz, line 297: This doesn't work in either python or Meson. You need to format it.

subprojects/packagefiles/mpdecimal/meson.build
```diff
@@ -312,13 +365,15 @@ libmpdecpp_sources = files(
...
   'mpdec++',
   libmpdecpp_sources, mpdecimal_h,
-  dependencies : [m_dep, mpdec_dep],
-  version : library_version,
-  darwin_versions : [compat_version, library_version])
+  dependencies: [m_dep, mpdec_dep],
+  version: library_version,
+  soversion: library_soversion,
+  darwin_versions: [compat_version, library_version],
+  override_options: ['optimization=3'])
```
eli-schwartz, line 372: Why is this optimization needed, instead of using whatever the user specified or setting it in default_options?

subprojects/packagefiles/mpdecimal/tests/meson.build
```diff
@@ -0,0 +1,24 @@
...
+
+testdata = custom_target(
+  'testdata',
+  command: [
+    find_program('gettests.py'),
+    join_paths(meson.current_source_dir(), 'testdata_dist'),
+    join_paths(meson.current_build_dir(), 'testdata')],
+  # The output file is not used but custom_target() requires to specify one.
+  output: 'testdata.stamp',
+  build_always_stale: true)
```
eli-schwartz, line 11: I don't like this. Always running a python program in a no-op build is wasteful. Why not just create the stamp file?

subprojects/packagefiles/mpdecimal/tests/gettests.py
```diff
@@ -0,0 +1,23 @@
...
+
+src, dst, *download = sys.argv[1:]
+
+if not os.path.isfile(os.path.join(dst, 'baseconv.decTest')):
+    shutil.copytree(src, dst, copy_function=shutil.copyfile, dirs_exist_ok=True)
+
+if download and not os.path.isfile(os.path.join(dst, 'add.decTest')):
+    with urllib.request.urlopen('http://speleotrove.com/decimal/dectest.zip') as f:
+        data = io.BytesIO(f.read())
+        assert hashlib.sha256(data.getbuffer()).hexdigest() == SHA256
```
eli-schwartz, line 21: Python assert is not a functionality for checking whether or not to do something. It's a functionality for running debug code that double-checks whether the program is performing as expected. It can be optimized out to nothing.

Just use if/raise. :)

subprojects/packagefiles/mpdecimal/meson.build
dnicolodi: The autotools build sets explicit and different optimization levels for the C and C++ libraries. I don't know if it is needed, but I thought the exercise is to replicate as close as possible what the native build rules do.

subprojects/packagefiles/mpdecimal/meson.build
dnicolodi: The check is kind of silly as the values for the machine option are constrained. I just removed the `error()`.

subprojects/packagefiles/mpdecimal/tests/meson.build
dnicolodi: Indeed. I was too focuses on matching the native build exactly.

subprojects/packagefiles/mpdecimal/tests/gettests.py
eli-schwartz: It would be technically more correct as a `raise ValueError` than as a `raise AssertionError` although, indeed, either one promotes this from "logically impossible" to "error checking".

So that's acceptable.

subprojects/packagefiles/mpdecimal/meson.build
```diff
@@ -202,66 +201,117 @@ if machine == 'auto'
...
-build_dll = []
-use_dll = []
+
+build_mpdec_compiler_args = []
 if host_machine.system() == 'windows'
   add_project_arguments('/wd4200', '/wd4204', '/wd4221', '/D_CRT_SECURE_NO_WARNINGS', '/nologo', language: 'c')
   if get_option('default_library') != 'static'
-    build_dll = ['/DBUILD_DLL']
-    use_dll = ['/DUSE_DLL']
+    build_mpdec_compiler_args = ['-DBUILD_LIBMPDEC']
```
eli-schwartz, line 303: Looking at the changes upstream, it seems that there is still a `_DLL` macro guard which determines if `__declspec(dllimport)` is part of the use flags.

I think we may still want that?

subprojects/packagefiles/mpdecimal/tests++/meson.build
```diff
@@ -0,0 +1,27 @@
...
+  conf.set('HAVE_PTHREAD_H', 1)
+endif
+
+configure_file(
+  output: 'config.h',
+  configuration: conf)
+
+threads_dep = dependency('threads')
+
+runtest = executable('runtest', files('runtest.cc', 'test.cc'), dependencies: [mpdecpp_dep, threads_dep])
```
eli-schwartz, line 13: `files()` does not need to be used here, because it's intended to be used in cases where you save the files to a variable and use them in another subdir. They then remember the directory they originally came from.

Here, they are inlined in the definition of executable() so there's no advantage of using it (and it makes the line length longer for no real advantage).

subprojects/packagefiles/mpdecimal/tests++/meson.build
```diff
@@ -0,0 +1,27 @@
...
+# These tests require to download test data.
+# test(
+#   'libmpdec++/official', runtest,
+#   args: files('official.topTest'),
+#   depends: testdata,
+#   workdir: join_paths(meson.current_build_dir(), '..', 'tests'))
+
+test(
+  'libmpdec++/additional', runtest,
+  args: files('additional.topTest'),
```
eli-schwartz, line 24: (Although this does need to be files() because it is added as string arguments to a test, and hence needs to remember the directory it came from.)

subprojects/packagefiles/mpdecimal/meson.build
eli-schwartz: I dunno. There may be an upstream reason that this specific optimization is needed, there may not be. Mainly I just wanted to make sure that this was being done for a reason -- the one you gave is fine.

subprojects/packagefiles/mpdecimal/tests/meson.build
eli-schwartz: Maybe I'm missing something, but it doesn't look like the stamp file actually gets created by the script even after your change?

subprojects/packagefiles/mpdecimal/meson.build
dnicolodi: I didn't find where in `Makefile.vc` that is set, thus I didn't include it:
```
~/src/wrapdb/subprojects/mpdecimal-2.5.1$ find . -name Makefile.vc | xargs grep _DLL
~/src/wrapdb/subprojects/mpdecimal-2.5.1$
```

subprojects/packagefiles/mpdecimal/tests++/meson.build
dnicolodi: It does not hurt either, does it? It is just 5 characters longer and I find it more explicit.

subprojects/packagefiles/mpdecimal/tests++/meson.build
dnicolodi: This needs to be `files()` otherwise it does not work, because if the `workdir` parameter, I think.

subprojects/packagefiles/mpdecimal/meson.build
eli-schwartz: The Makefile.vc doesn't provide any export interface such as pkg-config, nor does it really have usage documentation... I think that this is just left for users to blindly guess at. :(

subprojects/packagefiles/mpdecimal/tests++/meson.build
eli-schwartz: Yes, I did say that :) to clarify that my above comment doesn't apply here.

subprojects/packagefiles/mpdecimal/meson.build
dnicolodi: The examples and tests work fine without it, thus I think we don't need it.

subprojects/packagefiles/mpdecimal/tests++/meson.build
eli-schwartz: I find it less readable, also, this line is already 105 characters and reducing it to 98 characters helps a *lot*.

dnicolodi: Whatever. It is nice that the review for wraps has been tightened so much after all the crap that entered the repo...

eli-schwartz: It's a work in progress. :p

A few times I've just... gone through old wraps trying to systematically fix particular issues that bothered me. But there's a lot of old stuff. :/

dnicolodi: Unless you see any other issue with the PR, can you please merge this?

eli-schwartz:
The current build succeeds on Windows, but with the following warnings:
```
[34/48] Compiling C++ object subprojects/mpdecimal-2.5.1/mpdec++-3.dll.p/libmpdec++_decimal.cc.obj
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(165): warning C4273: 'decimal::context_template': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(223): note: see previous definition of 'context_template'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(176): warning C4273: 'decimal::getcontext': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(225): note: see previous definition of 'getcontext'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(186): warning C4273: 'decimal::MaxContext': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(425): note: see previous definition of 'MaxContext'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(193): warning C4273: 'decimal::IEEEContext': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(426): note: see previous definition of 'IEEEContext'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(206): warning C4273: 'decimal::Context::raiseit': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(234): note: see previous definition of 'raiseit'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(229): warning C4273: 'decimal::Context::repr': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(421): note: see previous definition of 'repr'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(253): warning C4273: 'decimal::operator <<': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(422): note: see previous definition of '<<'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(265): warning C4273: 'decimal::Decimal::exact': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(1125): note: see previous definition of 'exact'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(280): warning C4273: 'decimal::Decimal::exact': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(1126): note: see previous definition of 'exact'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(286): warning C4273: 'decimal::Decimal::ln10': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(1127): note: see previous definition of 'ln10'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(299): warning C4273: 'decimal::Decimal::radix': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(1128): note: see previous definition of 'radix'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(305): warning C4273: 'decimal::Decimal::repr': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(1180): note: see previous definition of 'repr'
../subprojects/mpdecimal-2.5.1/libmpdec++/decimal.cc(313): warning C4273: 'decimal::operator <<': inconsistent dll linkage
D:\a\wrapdb\wrapdb\subprojects\mpdecimal-2.5.1\libmpdec++\decimal.hh(1232): note: see previous definition of '<<'
```

I don't know how important they are, and I don't know whether this is somehow related to the _DLL define, because I'm not the most knowledgeable about MSVC.

Maybe @skrah knows.

dnicolodi:
> I don't know how important they are, and I don't know whether this is somehow related to the _DLL define, because I'm not the most knowledgeable about MSVC.

It cannot be related to the `_DLL` define as the native `Makefile.vc` does not have mention of this define. I don't have an easy way to test the native build on Windows. It may be related to the fact that `libmpdec++` is built only as a static DLL on Windows in the native build.

dnicolodi:
The warnings are gone. I forgot to  add `/DBUILD_MPDDECXX` to the compiler flags when building the c++ library. It would be nice if the CI would run `meson compile` with the `-v` flag to more easily compare the executed commands. I added it in a branch in my personal fork, but it is not very practical.

dnicolodi:
See #433

skrah:
@dnicolodi: Thanks for this!

Regarding the AIX work: The majority of the AIX support code deals with the preferred way in which AIX builds shared libraries:

The convention is  to create versioned shared library objects, e.g., shr4.o, shr4_64.o and created a single archive `libmpdec.a` that also contains the static objects. So all of shared/static libraries are in a single archive. Moreover, one can put previous versions of shared libraries in the same archive. New builds link against the most recent version, old binaries continue to function, thereby preventing DLL hell. The AIX linker also automatically sets the equivalent of an RPATH in the binaries, so the correct library is always found. It is actually pretty nice.

eli-schwartz:
Neat, thanks for clarifying.

This is something meson has to handle itself as it's in charge of producing .so files and archives.

We actually do have a code contribution from an AIX person a while back that deals with archiving shared libraries. It's not something I'm familiar with, but per the contribution, on AIX, we at least stick a shared library into an archive. I'm not sure that on default_library=both builds, the static *.o files are also in the same archive -- maybe they are, maybe they aren't. I don't have a machine to test it on, anyway...

dnicolodi:
@skrah I'll try to find some time to update this for the 4.0 release.

skrah:
@dnicolodi Nice. The 4.0.0 build changes look scary, but I think they can all be ignored for meson, i.e,  they are already implemented in meson or could potentially be implemented in meson itself (as @eli-schwartz mentioned for AIX).

The core logic is the same, in fact the meson build works with just bumping the `version` and `SOVERSION`.

`config.h` can be ignored except for `HAVE_PTHREAD_H` for the tests.

PR #429, opened 2022-05-13
GLFW v3.3.7-1

Gio:
This wrap currently only works for *nix-type systems. There's enough of a base for myself or another enterprising soul to continue the project for anything not currently covered.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -75,6 +75,30 @@
       "libglu1-mesa-dev"
     ]
   },
+  "glfw": {
+    "_comment": "- Requires POSIX platform.",
+    "debian_packages": [
+      "libxrandr-dev",
+      "libxinerama-dev",
+      "libxcursor-dev",
+      "libxi-dev",
+      "libxkbcommon-dev"
+    ],
+    "brew_packages": [
+      "libxrandr",
+      "libxinerama",
+      "libxcursor",
+      "libxi",
+      "libxkbcommon"
+    ],
+    "build_options": [
+      "glfw:tests=true",
+      "glfw:examples=true"
+    ],
+    "build_on": {
+      "windows": false
+    }
+  },
   "glib": {
     "_comment": "- Disable unit tests because upstream does not run them on Windows and some are flaky on Linux.",
     "build_options": [
--- a/subprojects/packagefiles/glfw/examples/meson.build
+++ b/subprojects/packagefiles/glfw/examples/meson.build
@@ -1,0 +1,37 @@
+###  GLFW examples
+#@ Constants
+ex_include = include_directories('../include', '../src', '../deps')
+glad_gl = '../deps/glad_gl.c'
+getopt = '../deps/getopt.c'
+tinycthread = '../deps/tinycthread.c'
+
+#@ Flags
+c_flags = []
+if is_posix
+  c_flags += '-D_XOPEN_SOURCE=700'
+endif
+if host_machine.system() == 'darwin'
+  c_flags += '-D_DARWIN_C_SOURCE'
+endif
+
+#@ Example executables
+examples = [['offscreen', glad_gl],
+            ['boing', glad_gl],
+            ['gears', glad_gl],
+            ['heightmap', glad_gl],
+            ['particles', [tinycthread, getopt, glad_gl]],
+            ['sharing', glad_gl],
+            ['simple', glad_gl],
+            ['splitview', glad_gl],
+            ['wave', glad_gl]]
+foreach e : examples
+  name = e[0]
+  src = [e[0] + '.c', e[1]]
+  executable(name, src,
+             include_directories : ex_include,
+             c_args : c_flags,
+             dependencies : [glfw_dep, deps],
+             build_by_default : true,
+             install : get_option('install'),
+             install_dir : opt_libdir / 'GLFW' / 'examples')
+endforeach
--- a/subprojects/packagefiles/glfw/meson.build
+++ b/subprojects/packagefiles/glfw/meson.build
@@ -1,0 +1,228 @@
+#####  GLFW  #####
+#@ Project file
+project('glfw', 'c',
+         version : '3.3.7',
+         license : 'zlib',
+         default_options : ['default_library=static',
+                            'b_ndebug=if-release',
+                            'c_std=c99',
+                            'warning_level=0'],
+         meson_version : '>=0.49.0')
+#@ Project variables
+sys_os = host_machine.system()
+sys_cc = meson.get_compiler('c')
+opt_libdir = get_option('libdir')
+opt_display = get_option('display-api')
+is_posix = (sys_os == 'linux' or sys_os == 'sunos' or sys_os == 'dragonfly' or sys_os.endswith('bsd') or sys_os == 'darwin')
+
+
+### Dependencies ###
+deps = []
+#@ Backend API
+#  OSMesa
+if opt_display == 'osmesa'
+  deps += dependency('osmesa')
+#  X11
+elif opt_display == 'x11'
+  foreach x : ['x11', 'xrandr', 'xinerama', 'xcursor', 'xi', 'xkbcommon']
+    deps += dependency(x, required : true)
+  endforeach
+#  Wayland
+elif opt_display == 'wayland'
+  warning('Wayland support is experimental & incomplete.')
+  warning('GLFW < 3.4 cannot detect Wayland if both x11 and Wayland exist on-system!')
+  foreach w : ['wayland-client', 'wayland-cursor', 'wayland-server', 'wayland-egl', 'xkbcommon']
+    deps += dependency(w, required : true)
+  endforeach
+  wl_protocols = dependency('wayland-protocols')
+  wl_dir = wl_protocols.get_variable('pkgdatadir')
+  wl_scan_dep = dependency('wayland-scanner',
+                            required : false,
+                            native : true)
+  if wl_scan_dep.found()
+    wl_scanner = find_program(wl_scan_dep.get_variable(pkgconfig : 'wayland_scanner'), native : true)
+  else
+    wl_scanner = find_program('wayland-scanner', native : true)
+  endif
+  s_wayland = []
+  protocols = [['idle-inhibit', 'unstable'],
+               ['pointer-constraints', 'unstable'],
+               ['relative-pointer', 'unstable'],
+               ['viewporter', 'stable'],
+               ['xdg-decoration', 'unstable'],
+               ['xdg-shell', 'stable']]
+  foreach p : protocols
+    x_path = join_paths(wl_dir, p[1], p[0], p[0])
+    if p[1] == 'stable'
+      x_name = p[0]
+      xml = files(x_path + '.xml')
+    else
+      x_name = p[0] + '-unstable-v1'
+      xml = files(x_path + '-unstable-v1.xml')
+    endif
+    foreach f : ['client-header', 'server-header', 'private-code']
+      if (f == 'client-header' and p[0] == 'xdg-decoration')
+        outfile = 'wayland-' + p[0] + '-client-protocol.h'
+      elif f == 'client-header'
+        outfile = 'wayland-' + x_name + '-client-protocol.h'
+      elif f == 'server-header'
+        outfile = 'wayland-' + p[0] + '-server-protocol.h'
+      elif (f == 'private-code' and p[0] == 'xdg-decoration')
+        outfile = 'wayland-' + p[0] + '-protocol.c'
+      else
+        outfile = 'wayland-' + x_name + '-protocol.c'
+      endif
+      s_wayland += custom_target('@0@ @1@'.format(p[0], f),
+                                  command : ['wayland-scanner', f, '@INPUT@', '@OUTPUT@'],
+                                  input : xml,
+                                  output : outfile)
+    endforeach
+  endforeach
+#  Win32
+elif opt_display == 'win32'
+  #——TODO:
+  # ☐ Implement Win32 build
+  error('Win32 build is not currently implemented.')
+#  Cocoa
+elif opt_display == 'cocoa'
+  #——TODO:
+  # ☐ Implement Cocoa build
+  error('Cocoa build is not currently implemented.')
+endif
+#@ System Dependencies
+foreach l : ['m', 'rt', 'dl']
+  deps += sys_cc.find_library(l, required : false)
+endforeach
+deps += dependency('threads')
+
+
+### Configuration ###
+cfg_data = configuration_data()
+foreach d : ['osmesa', 'x11', 'wayland', 'win32', 'cocoa']
+  cfg_data.set('_GLFW_' + d.to_upper(), opt_display == d)
+endforeach
+cfg_file = configure_file(configuration : cfg_data,
+                          output : 'glfw_config.h')
+
+
+### Targets ###
+#@ Primary target:
+##   GLFW library
+#  Common files
+s_common = [cfg_file,
+           'src/context.c',
+           'src/init.c',
+           'src/input.c',
+           'src/monitor.c',
+           'src/vulkan.c',
+           'src/window.c']
+#  Backend files
+if opt_display == 'osmesa'
+  s_display = ['src/null_init.c',
+               'src/null_monitor.c',
+               'src/null_window.c',
+               'src/null_joystick.c',
+               'src/posix_time.c',
+               'src/posix_thread.c',
+               'src/osmesa_context.c']
+elif opt_display == 'x11'
+  s_display = ['src/x11_init.c',
+               'src/x11_monitor.c',
+               'src/x11_window.c',
+               'src/xkb_unicode.c',
+               'src/posix_time.c',
+               'src/posix_thread.c',
+               'src/osmesa_context.c',
+               'src/egl_context.c',
+               'src/glx_context.c']
+  if sys_os == 'linux'
+    s_display += 'src/linux_joystick.c'
+  else
+    s_display += 'src/null_joystick.c'
+  endif
+elif opt_display == 'wayland'
+  s_display = [s_wayland,
+              'src/wl_init.c',
+              'src/wl_monitor.c',
+              'src/wl_window.c',
+              'src/xkb_unicode.c',
+              'src/posix_time.c',
+              'src/posix_thread.c',
+              'src/osmesa_context.c',
+              'src/egl_context.c']
+  if sys_os == 'linux'
+    s_display += 'src/linux_joystick.c'
+  else
+    s_display += 'src/null_joystick.c'
+  endif
+elif opt_display == 'win32'
+#——TODO:
+# ☐ Implement Win32 build routine
+  s_display = ['src/win32_init.c',
+               'src/win32_joystick.c',
+               'src/win32_monitor.c',
+               'src/win32_time.c',
+               'src/win32_thread.c',
+               'src/win32_window.c',
+               'src/wgl_context.c']
+elif opt_display == 'cocoa'
+#——TODO:
+# ☐ Implement Cocoa build routine
+  s_display = ['src/cocoa_init.m',
+               'src/cocoa_joystick.m',
+               'src/cocoa_monitor.m',
+               'src/cocoa_window.m',
+               'src/cocoa_time.c',
+               'src/posix_thread.c',
+               'src/nsgl_context.m']
+endif
+srcfiles = [s_common, s_display]
+incdirs = include_directories('include', 'src')
+#@ Flags
+c_flags = []
+c_flags += '-D_GLFW_USE_CONFIG_H'
+if host_machine.system() == 'darwin'
+  c_flags += '-D_DARWIN_C_SOURCE'
+endif
+#  Build
+glfw_lib = library('glfw3',
+                    srcfiles,
+                    c_args : c_flags,
+                    include_directories : incdirs,
+                    dependencies : deps,
+                    version : meson.project_version(),
+                    build_by_default : true,
+                    pic : true,
+                    install : get_option('install'))
+glfw_dep = declare_dependency(include_directories : incdirs, link_with : glfw_lib)
+if meson.version().version_compare('>=0.54.0')
+  meson.override_dependency('glfw3', glfw_dep)
+endif
+#  Headers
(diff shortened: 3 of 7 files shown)


subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
+#####  GLFW  #####
+#@ Project file
+project('glfw', 'c'
+       , version : '3.3.7'
+       , license : 'zlib'
+       , default_options : ['default_library=static'
+                           ,'b_ndebug=if-release'
+                           ,'c_std=c99'
+                           ,'warning_level=0'
+                           ,'werror=false']
```
eli-schwartz, line 10: Werror defaults to false.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+#
+opt_prefix = get_option('prefix')            #  Returns the install path prefix
+opt_bindir = get_option('bindir')            #  Returns the binary subdirectory
+opt_libdir = get_option('libdir')            #  Returns the library subdirectory
+opt_incdir = get_option('includedir')        #  Returns the include subdirectory
+opt_datadir = get_option('datadir')          #  Returns the data subdirectory
+#
+opt_docs = get_option('build-docs')          #  Build documentation? T/F
+opt_tests = get_option('build-tests')        #  Build test programs? T/F
+opt_examples = get_option('build-examples')  #  Build example programs? T/F
```
eli-schwartz, line 27: These variables are used... once? They should be inlined where used.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+opt_install = get_option('install')          #  Install GLFW? T/F
+opt_display = get_option('display-api')      #  Returns the selected backend API
+#@ Extra variables
+if (sys_os == 'dragonfly' or sys_os.endswith('bsd'))  #
+  bsd = ['bsd']                                           #  Unifies the BSDs in one reference
+endif                                                     #
```
eli-schwartz, line 33: At the time this is used, it may be undefined because it's only created in a conditional. Also why is it an array?

You may want to define an is_bsd variable instead.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+  endforeach
+#  Wayland
+elif opt_display == 'wayland'
+  warning('Wayland support is experimental & incomplete.')
+  warning('GLFW < 3.4 cannot detect Wayland if both x11 and Wayland exist on-system!')
+  foreach w : ['wayland-client', 'wayland-cursor', 'wayland-server', 'wayland-egl', 'xkbcommon']
+    deps += dependency(w, required : true)
+  endforeach
+  wl_protocols = dependency('wayland-protocols')
+  wl_dir = wl_protocols.get_variable(pkgconfig : 'pkgdatadir')
```
eli-schwartz, line 64: This shouldn't be pkg-config specific. It can be used as a subproject with the same variable name.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+  wl_scan_dep = dependency('wayland-scanner'
+                          , required : false
+                          , native : true)
```
eli-schwartz, line 67: The entire build uses commas at the beginning of a line instead of the end of the previous line. Please change it.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+elif opt_display == 'cocoa'
+  #——TODO:
+  # ☐ Implement Cocoa build
+  error('Cocoa build is not currently implemented.')
+endif
+#@ System Dependencies
+if (sys_os == 'linux' or sys_os == 'sunos')
+  foreach l : ['m', 'rt', 'dl']
+    deps += sys_cc.find_library(l, required : true)
+  endforeach
```
eli-schwartz, line 122: These shouldn't be per OS. Some of these are guaranteed by POSIX, just find them with `required: false` md if they aren't found, they are built into libc.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+foreach d : ['osmesa', 'x11', 'wayland', 'win32', 'cocoa']
+  cfg_data.set('_GLFW_' + d.to_upper(), opt_display == d)
+endforeach
+cfg_file = configure_file(configuration : cfg_data
+                         ,output : 'glfw_config.h')
+
+
+### Targets ###
+#@ Install key
+I = opt_install
```
eli-schwartz, line 143: Why is this defined statically once, and then reassigned to another variable? Just directly assign it. But really this is very inscrutable.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+  s_display = ['src/cocoa_init.m'
+              ,'src/cocoa_joystick.m'
+              ,'src/cocoa_monitor.m'
+              ,'src/cocoa_window.m'
+              ,'src/cocoa_time.c'
+              ,'src/posix_thread.c'
+              ,'src/nsgl_context.m']
+endif
+srcfiles = [s_common, s_display]
+incdirs = [include_directories('include'), include_directories('src')]
```
eli-schwartz, line 215: You can pass multiple dirs to this function.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+#  Build
+glfw_lib = build_target('glfw3'
```
eli-schwartz, line 217: Why use build_target instead of directly using the type? The purpose of this function is when you need to dynamically choose which target_type to use.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,259 @@
...
+#@ Secondary target:
+##   GLFW examples
+subdir('examples')
+
+
+### Post
+#@ Subproject
+if prj_sub == true
+  meson.override_dependency('glfw3', glfw_dep)  #  Overrides any installed GLFW library
+endif
```
eli-schwartz, line 250: Do this unconditionally, it causes no harm when it's not a subproject.

subprojects/packagefiles/glfw/tests/meson.build
```diff
@@ -0,0 +1,249 @@
+###  GLFW test programs
+#@ Build check
+if opt_tests == false
+  subdir_done()
+endif
+
+#@ Constants
+relpath = '../'
+incdirs = [include_directories(relpath + 'include')
+          ,include_directories(relpath + 'src')
```
eli-schwartz, line 10: This relpath stuff is:
- a constant
- indirect, thus confusing
- leads to longer lines

I don't understand the goal.

subprojects/packagefiles/glfw/tests/meson.build
```diff
@@ -0,0 +1,249 @@
...
+          ,include_directories(relpath + 'deps')]
+glad_gl = [relpath + 'deps/glad_gl.c']
+glad_vulkan = [relpath + 'deps/glad_vulkan.c']
+getopt = [relpath + 'deps/getopt.c']
```
eli-schwartz, line 14: Why are there spaces in the pathnames?

eli-schwartz (changes requested):

subprojects/packagefiles/glfw/meson.build
Gio: Oh! It's an array because at one point I had it as `['bsd', true]` depending on how I wanted to reference it, then realized that it was unnecessary.

subprojects/packagefiles/glfw/meson.build
Gio: Why, indeed? As part of your first comment about inlining variables used once, I went through and removed occurrences like this.

subprojects/packagefiles/glfw/tests/meson.build
Gio: I am not certain what you mean. `'../' + 'deps/<filename>' = '../deps/<filename>'` ..correct? They are in unnecessary arrays, however, so I'll fix those up.

subprojects/packagefiles/glfw/tests/meson.build
Gio: When initially writing and testing, I discovered that Meson was working from the current subdir (I had thought the working directory would remain at the level of the original meson.build). At the time, it was easier to add `relpath` to everything.

subprojects/packagefiles/glfw/tests/meson.build
eli-schwartz: When I reviewed this on my phone, I could swear I saw stuff like ` getopt = [relpath + ' deps / getopt.c ']`. On the other hand, I could have been insane -- or alternatively, the website was giving my phone a lot of trouble and it may have mis-rendered. idk anymore.

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,226 @@
...
+#  Cocoa
+elif opt_display == 'cocoa'
+  #——TODO:
+  # ☐ Implement Cocoa build
+  error('Cocoa build is not currently implemented.')
+endif
+#@ System Dependencies
+if (sys_os == 'linux' or sys_os == 'sunos' or is_bsd == true)
+  foreach l : ['m', 'rt', 'dl']
+    deps += sys_cc.find_library(l, required : false)
```
eli-schwartz, line 98: You can do this on any system, not just linux/sunos/bsd. :)

dl specifically is not POSIX and may not actually exist on all systems, so not-found could either mean it's built into libc or that it is missing from the OS.

There's a very recent Meson 0.62.0 custom dependency for it added in https://github.com/mesonbuild/meson/pull/9927 though.

subprojects/packagefiles/glfw/tests/meson.build
```diff
@@ -0,0 +1,226 @@
...
+getopt = '../deps/getopt.c'
+tinycthread = '../deps/tinycthread.c'
+
+#@ Flags
+if sys_os == 'linux'
+  c_flags = ['-D_XOPEN_SOURCE=700']
+endif
+
+#@ Test executables
+tst_clipboard = executable('clipboard',
```
eli-schwartz, line 20: Similar to what I mentioned by the examples... it might be easier to define all these in a foreach loop as they are mostly the same.

eli-schwartz (changes requested):
There's a merge commit in here somehow... it needs to be rebased out before we can merge this.

ci_config.json
```diff
@@ -71,6 +71,20 @@
       "windows": false
     }
   },
+  "glfw": {
```
eli-schwartz, line 74: The CI should enable any build options that are false by default.

subprojects/packagefiles/glfw/tests/meson.build
```diff
@@ -0,0 +1,133 @@
...
+         ['tearing', glad_gl, tst_tearing],
+         ['threads', [tinycthread, glad_gl], tst_threads],
+         ['timeout', glad_gl, tst_timeout],
+         ['title', glad_gl, tst_title],
+         ['triangle-vulkan', glad_vulkan, tst_trianglevulkan],
+         ['windows', [getopt, glad_gl], tst_windows]]
+foreach t : tests
+  name = t[0]
+  src = [t[0] + '.c'] + t[1]
+  t[2] = executable(name, src,
```
eli-schwartz, line 43: The lexer doesn't like this: "Assignment target must be an id."

You cannot indirectly assign *to* an array element like this, but that being said, the array contains nonexistent variables.

I would instead suggest assigning it to `exe =`, then defining the test immediately inside the loop as well.

subprojects/packagefiles/glfw/examples/meson.build
```diff
@@ -0,0 +1,37 @@
+###  GLFW examples
+#@ Build check
+if opt_examples == false
```
eli-schwartz, line 3: This is not defined anywhere.

Also, the option is already a boolean, so no need to compare to `== false`
```suggestion
if not get_option('build-examples')
```

eli-schwartz (changes requested):

ci_config.json
eli-schwartz: I don't think this got implemented, by the way?

Gio: Oh! Let me fix that.

subprojects/packagefiles/glfw/tests/meson.build
```diff
@@ -0,0 +1,51 @@
...
+tinycthread = '../deps/tinycthread.c'
+
+#@ Flags
+if is_posix == true
+  c_flags = ['-D_XOPEN_SOURCE=700']
+endif
+
+#@ Test executables
+tests = [['clipboard', [getopt, glad_gl], ['console', 'glfw']],
+         ['events', [getopt, glad_gl], ['console', 'glfw']],
```
eli-schwartz, line 21: Every single test seems to be tagged with the "glfw" suite, which is redundant because Meson includes a builtin scope for the subproject itself, hence `--suite glfw` will already select that. :)

subprojects/packagefiles/glfw/examples/meson.build
```diff
@@ -0,0 +1,37 @@
...
+endif
+
+#@ Constants
+ex_include = include_directories('../include', '../src', '../deps')
+glad_gl = '../deps/glad_gl.c'
+getopt = '../deps/getopt.c'
+tinycthread = '../deps/tinycthread.c'
+
+#@ Flags
+if is_posix == true
```
eli-schwartz, line 14: This too can drop the `== true`

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,221 @@
+#####  GLFW  #####
+#@ Project file
+project('glfw', 'c',
+         version : '3.3.7',
+         license : 'zlib',
+         default_options : ['default_library=static',
+                            'b_ndebug=if-release',
+                            'c_std=c99',
+                            'warning_level=0'],
+         meson_version : '>=0.61.0')
```
eli-schwartz, line 10: The most recent meson version needed for this, I think, is 0.54?

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,221 @@
...
+                    srcfiles,
+                    c_args : ['-D_GLFW_USE_CONFIG_H'],
+                    include_directories : incdirs,
+                    dependencies : deps,
+                    version : meson.project_version(),
+                    build_by_default : true,
+                    pic : true,
+                    install : I)
+glfw_dep = declare_dependency(include_directories : incdirs, link_with : glfw_lib)
+meson.override_dependency('glfw3', glfw_dep)
```
eli-schwartz, line 194: If this was wrapped in

```suggestion
if meson.version().version_compare('>=0.54.0')
    meson.override_dependency('glfw3', glfw_dep)
endif
```

then the minimum version of Meson could drop to 0.49.0 instead.

Gio:
Since the "tests" were not automated, I changed them to behave as the "examples" do.

subprojects/packagefiles/glfw/examples/meson.build
```diff
@@ -0,0 +1,37 @@
...
+#@ Constants
+ex_include = include_directories('../include', '../src', '../deps')
+glad_gl = '../deps/glad_gl.c'
+getopt = '../deps/getopt.c'
+tinycthread = '../deps/tinycthread.c'
+
+#@ Flags
+if is_posix
+  c_flags = ['-D_XOPEN_SOURCE=700']
+endif
```
eli-schwartz, line 16: Is this actually being used by anything?

subprojects/packagefiles/glfw/tests/meson.build
```diff
@@ -0,0 +1,49 @@
...
+tst_include = include_directories('../include', '../src', '../deps')
+glad_gl = '../deps/glad_gl.c'
+glad_vulkan = '../deps/glad_vulkan.c'
+getopt = '../deps/getopt.c'
+tinycthread = '../deps/tinycthread.c'
+
+#@ Flags
+if is_posix
+  c_flags = ['-D_XOPEN_SOURCE=700']
+endif
```
eli-schwartz, line 17: Or this?

eli-schwartz (commented):
Other than some possibly unused code, looks good to me.

subprojects/packagefiles/glfw/examples/meson.build
Gio: Yes! Several of the examples and tests will fail to build without it.

eli-schwartz: No, I mean, when is this variable ever used by passing it as a kwarg to a function?

subprojects/packagefiles/glfw/tests/meson.build
Gio: Same as with the examples, some tests will not build correctly. While GLFW itself is c99, the ancillary programs seem to take a few more liberties.

subprojects/packagefiles/glfw/examples/meson.build
Gio: Oops! Fixed.

You pointed out that tinycthread should set "-D_XOPEN_SOURCE=500", but that doesn't seem to work. =\

subprojects/packagefiles/glfw/meson_options.txt
```diff
@@ -0,0 +1,26 @@
...
+        value : false,
+        description : 'Build GLFW test programs?')
+option('build-examples',
+        type : 'boolean',
+        value : false,
+        description : 'Build GLFW example programs?')
+option('install',
+        type : 'boolean',
+        value : false,
+        description : 'Install GLFW?')
```
dnicolodi, line 19: I am not sure I understand the need for this option, can you please elaborate?

subprojects/packagefiles/glfw/meson.build
```diff
@@ -0,0 +1,223 @@
...
+foreach d : ['osmesa', 'x11', 'wayland', 'win32', 'cocoa']
+  cfg_data.set('_GLFW_' + d.to_upper(), opt_display == d)
+endforeach
+cfg_file = configure_file(configuration : cfg_data,
+                          output : 'glfw_config.h')
+
+
+### Targets ###
+#@ Install key
+I = get_option('install')
```
dnicolodi, line 110: Maybe this variable can get a more descriptive name.

subprojects/packagefiles/glfw/docs/meson.build
```diff
@@ -0,0 +1,7 @@
+###  GLFW documentation
+#@ Build check
+if not get_option('build-docs')
+  subdir_done()
+endif
+
+warning('Documentation build is not implemented yet!')
```
dnicolodi, line 7: What's the point of having a build option for enabling a feature that is not implemented? Anyhow, it would make more sense to move the conditional to the main `meson.build` and make `subdir(...)` the body of the conditional.

subprojects/packagefiles/glfw/meson_options.txt
```diff
@@ -0,0 +1,26 @@
+#####  GLFW build options
+
+#@ Build options
+option('build-docs',
+        type : 'boolean',
+        value : false,
+        description : 'Build GLFW documentation?')
```
dnicolodi, line 7: See above on why I think this option is not necessary.

subprojects/packagefiles/glfw/meson_options.txt
```diff
@@ -0,0 +1,26 @@
...
+option('build-tests',
+        type : 'boolean',
+        value : false,
+        description : 'Build GLFW test programs?')
+option('build-examples',
+        type : 'boolean',
+        value : false,
+        description : 'Build GLFW example programs?')
```
dnicolodi, line 15: Looking at other wraps it seems that the most common spelling for this kind of options is `tests` and `examples` without the `build-` prefix, in line with the Meson style recommendation to not name options with the `enable-` prefix. It may be nice to standardize on the same name for options with the same intent.

subprojects/packagefiles/glfw/meson_options.txt
Gio: If a user wants to install the library, its headers, and the various examples, they should be able to choose to do so, **even if** it is a subproject.

subprojects/packagefiles/glfw/meson_options.txt
Gio: This is true. I'll get those changed.

subprojects/packagefiles/glfw/docs/meson.build
Gio: > Anyhow, it would make more sense to move the conditional to the main `meson.build` and make `subdir(...)` the body of the conditional.

Is it a necessary change? I suppose it saves time, and I have no good argument other than "I preferred to keep my 'subdir(...)' logic inside their own build files."

subprojects/packagefiles/glfw/meson_options.txt
Gio: > What's the point of having a build option for enabling a feature that is not implemented?

It provides a framework for those that follow, the same as the `win32` and `cocoa` display API options. Should those be removed? Is it _necessary_? No, it is not. Is it _useful_? I think so.

Gio:
Without a way to properly test MacOS builds, I believe that I should disable the MacOS CI. It's had too many failures and I don't know what changes are needed to fix them.

neheb:
```
 ../subprojects/glfw-3.3.7/tests/glfwinfo.c:422:21: error: implicitly declaring library function 'strcasecmp' with type 'int (const char *, const char *)' [-Werror,-Wimplicit-function-declaration]
                if (strcasecmp(optarg, API_NAME_OPENGL) == 0)
```

:). Reminds me of the flac wrap. Try passing

```
  add_project_arguments('-D_DARWIN_C_SOURCE', language: 'c')
endif
```

Something similar to that.

Gio:
> :). Reminds me of the flac wrap. Try passing

Alright! Giving it a shot.

eli-schwartz (approved):

neheb (approved):

PR #430, opened 2022-05-14
clang-format again

neheb:


jpakkane:
This is good but it will always be broken unless this is gated in CI with a test. Could you add one to this?

neheb:
IIRC @eli-schwartz had an issue with that. I think the CI job could be limited to one file though. I'll see.

jpakkane:
Instead of clang-format, could we just use Python's JSON dumper with `sort_keys: True` and `indent: 4`? It would not need any external dependencies.

neheb:
No idea what that is. I just use git clang-format. I'll look into this.

eli-schwartz:
I also think policing this is a waste of time and there's nothing wrong with the current state of these files in git master.

jpakkane:
There are really only two different ways to do this:

- no enforcement at all
- having a script that does the formatting automatically and then enforce that in CI

The latter is better here simply because having the list of projects in an alphabetical order is a _big_ usability improvement.

So basically if there is a script, say `formatjsons.py` that does the formatting and also does `formatjsons.py --is-formatted` that returns non-zero if the files are not formatted then all manual work goes away. Which is what you want.

eli-schwartz:
This PR isn't doing anything to the alphabetical order of the list of projects anyway?

neheb:
Rebased.

PR #431, opened 2022-05-15
Add imgui_stdlib.cpp to the sources.

apoorv569:


jpakkane:
You need to add a new entry to the releases json file.

apoorv569:
> You need to add a new entry to the releases json file.

I have added the new entry.

jpakkane:
Please keep the commits squashed to a single one.

neheb:
entry should be removed from broken_darwin in ci_config.json

eli-schwartz:
It actually does appear to be broken though.

```

**Comment by neheb** [comment]
sounds like objc needs to be added to the project call.

**Comment by eli-schwartz** [comment]
Just conditionally where .mm sources are added for darwin only.

**Comment by apoorv569** [comment]
> Please keep the commits squashed to a single one.

Fixed.

**Comment by jpakkane** [comment]
After `project()` do

```meson
    add_languages('objcpp')
endif
```

jpakkane:
I don't know enough about objcpp to meaningfully debug that. It is possible this is an upstream issue cause by Apple making things stricter. I'm fine with marking this as not passing on Darwin like it was. (Leave the code changes in, though, they will be useful to the next person.)

eli-schwartz:
From searching on the upstream github repository this seems to be https://github.com/ocornut/imgui/pull/5123

So implementing the change here is sufficient to make sure that on the next upstream release it magically works. Hopefully.

PR #432, opened 2022-05-15
libwebsockets: update to 4.0.21

neheb:
Ran meson.build file through muon's fmt_unstable to get rid of
problematic tab usage in meson.build.

### DIFF
--- a/subprojects/packagefiles/libwebsockets/meson.build
+++ b/subprojects/packagefiles/libwebsockets/meson.build
@@ -1,13 +1,13 @@
 project('libwebsockets', 'c', version: '4.0.13', license: 'MIT')
 
 lws_library_version = meson.project_version()
-lws_library_version_major  = '4'
-lws_library_version_minor  = '0'
-lws_library_version_patch  = '13'
+lws_library_version_major = '4'
+lws_library_version_minor = '0'
+lws_library_version_patch = '13'
 lws_library_version_number = 16
 
-lws_incdirs = [ ]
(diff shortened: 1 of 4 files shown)


subprojects/packagefiles/libwebsockets/meson.build
```diff
@@ -1,23 +1,22 @@
...
+lws_dependencies = []
 lws_c_args = []
 
-
 # ToDo: detect platform properly
 esp_platform = false
 ios = false
 apple = false
 mingw = false
 win32 = build_machine.system() == 'win32'
```
dnicolodi, line 18: `.system()` doe not return `win32` but `windows`

eli-schwartz: Fantastic... so many reasons this wrap was broken on Windows and generally stuff that never got tested back in the old-style wrapdb "test it by running it on the maintainer's personal laptop" approach to CI. :D

dnicolodi: CI runs also on personal forks thus it would be maybe better to open a PR when at least the most obvious errors are fixed. It would allow reviewer to focus on PRs that are a bit more polished...

neheb: Indeed.

neheb:
draft until I can figure out a good way to fix MSVC.

neheb:
Good enough.

PR #433, opened 2022-05-15
ci: Run meson compile with the -v option

dnicolodi:
Having the executed commands reported in the CI logs allows to more easily inspect the actual compiler flags used. This is especially useful when a local build environment is not available.


neheb (approved):

neheb:
In a different place, it was requested that I do the same. Makes sense to have a verbose CI.

PR #438, opened 2022-05-17
imgui: try to fix for macOS

neheb:


subprojects/packagefiles/imgui/meson.build
```diff
@@ -34,7 +44,7 @@ if dx12_dep.found()
     sources += 'backends/imgui_impl_dx12.cpp'
     dependencies += dx12_dep
 endif
-metal_dep = dependency('appleframeworks', modules : 'metal', required: get_option('metal'))
+metal_dep = dependency('appleframeworks', modules : [ 'Foundation', 'AppKit', 'GameController', 'Metal' ], required: get_option('metal'))
```
jpakkane, line 47: Are those other ones only needed when using Metal? They would seem to be necessary even without it, for example if using OpenGL.

neheb: I get linking errors without them.

Foundation for example is completely necessary. AppKit and GameController have APIs that are used here.

jpakkane: Yes, but again those should probably be used always on macOS. The way things are currently set up they are only looked up when metal is enabled. Or, in other words, should those be split into a generic `mac_dep` which has everything except Metal and `metal_dep` which only has metal.

neheb: Erm no.

Only the metal files use those Apple libraries and APIs. Everything else is standard C++.

neheb: Also note that these APIs are ObjC(++) APIs. There's an add_language('objcpp') call in case of metal. It's used nowhere else.

PR #440, opened 2022-05-17
emilk-loguru: try to fix for windows

neheb:

### DIFF
--- a/subprojects/packagefiles/emilk-loguru/meson_options.txt
+++ b/subprojects/packagefiles/emilk-loguru/meson_options.txt
@@ -104,8 +104,7 @@ option(
 
 option(
   'loguru_stacktraces',
-  type: 'boolean',
-  value: true,
+  type: 'feature',
   description: 'Print stack traces on abort',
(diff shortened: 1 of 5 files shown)


subprojects/packagefiles/emilk-loguru/meson_options.txt
```diff
@@ -104,8 +104,7 @@ option(
 
 option(
   'loguru_stacktraces',
-  type: 'boolean',
-  value: true,
+  type: 'feature',
```
eli-schwartz, line 107: We could add `meson_version: '>=0.47.0'` to the project() declaration too, I guess.

neheb: feature is already used elsewhere. but sure.

eli-schwartz: Ah, in that case it *definitely* can't hurt. :D

neheb:
Fixed.

PR #441, opened 2022-05-18
docopt: remove from broken CI

neheb:


subprojects/packagefiles/docopt/meson.build
```diff
@@ -3,7 +3,20 @@ project('docopt.cpp', 'cpp',
         license: 'Boost',
         default_options: 'cpp_std=c++11')
 
-doclib = library('docopt', 'docopt.cpp')
+cpp = meson.get_compiler('cpp')
+if cpp.get_argument_syntax() == 'msvc'
+  if get_option('buildtype') != 'static'
```
eli-schwartz, line 8: This could be on one line as `x and y`.

The workaround below doesn't need to repeat the conditional either, it could be folded into this same if/endif.

neheb: Fixed.

PR #442, opened 2022-05-19
Add wrap files for speexdsp-1.2.0

kcgen:
A relatively feature-complete wrap that includes:
 - basic support for SIMD (SSE2 and NEON)
 - options to build the library with fixed-point or floating-point numerical representation
 - options to build sample programs
 - user-focussed defaults: floating point and minimal warnings

``` text
  Library Summary
    Optimization level: 3
    SIMD instructions : SSE2
    Numerical type    : floating-point
    FFT library       : OggVorbis FFT (built-in)
```


subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,85 @@
...
+  default_options: ['warning_level=2',
+                    'c_std=gnu99',
+                    'buildtype=release'],
+  license : 'bsd-like')
+
+lib_current = 1
+lib_revision = 0
+lib_age = 2
+lib_soversion = 1
+lib_version = '@0@.@1@.@2@'.format(lib_soversion, lib_age, lib_revision)
```
dnicolodi, line 12: Isn't this the same as the project version? Can't you do `lib_version = meson.project_version()`?

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,85 @@
+ project('speexdsp', 'c',
+  version : '1.2.0',
+  default_options: ['warning_level=2',
+                    'c_std=gnu99',
+                    'buildtype=release'],
+  license : 'bsd-like')
+
+lib_current = 1
+lib_revision = 0
+lib_age = 2
```
dnicolodi, line 10: I am not sure of the versioning scheme used by this project, but it seem that this could be
```
v = meson.project_version().split('.')
lib_current = v[0]
lib_age = v[1]
lib_revision = v[2]
```
avoiding to encode the project version in two places. But `lib_current`, `lib_age`, and `lib_revision` seem to be used only to define `lib_version`, thus see below.

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,85 @@
...
+endif
+
+foreach arg : c_args
+  if cc.has_argument(arg)
+    add_project_arguments(arg, language: 'c')
+  endif
+endforeach
+
+# Check for C99 variable-size arrays, or alloca() as fallback
+msg_use_alloca = false
```
dnicolodi, line 44: What is this variable for? It seems it is set but never used again.

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,85 @@
...
+if cc.compiles('''static int x;
+                  char some_func (void) {
+                    char a[++x];
+                    a[sizeof a - 1] = 0;
+                    int N;
+                    return a[0];
+                  }''', name : 'C99 variable-size arrays')
+  conf_data.set('VAR_ARRAYS', 1)
+  msg_use_alloca = 'NO (using C99 variable-size arrays instead)'
```
dnicolodi, line 53: Meson is not happy if you try to change the type of a variable. `msg_use_alloca` is initialized to a bool and you re assigning a string to it here.

subprojects/packagefiles/speexdsp/meson.build
eli-schwartz: The versioning scheme used by this project is autotools, so prepare for great confusion and annoyance.

The lib_current / lib_revision / lib_age is wrong as defined here. Upstream calculates this as `6, 1, 5`, and passes it to libtool as `-version-info 6:1:5`.

libtool then does math (specifically, `(C - A).(A).(R)`), and sets the equivalent of Meson's `version: '1.5.1'`, which you'll notice is not the same as either the libtool numbers (because math!!!) or the project version.

Given the project version and the library version are totally unconnected, you MUST encode it in two places, the only question is whether to do it via math or as a string constant.

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,85 @@
...
+
+elif host_cpu_family == 'aarch64'
+  c_args += cc.get_supported_arguments(['-mfpu=neon-fp-armv8'])
+  conf_data.set10('USE_NEON', true)
+endif
+
+foreach arg : c_args
+  if cc.has_argument(arg)
+    add_project_arguments(arg, language: 'c')
+  endif
```
eli-schwartz, line 40: You already did get_supported_arguments, you don't need to use has_argument again.

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,85 @@
...
+                      int foo=10;
+                      int * array = alloca(foo);
+                    }''', name : 'alloca (std)')
+  conf_data.set('USE_ALLOCA', true)
+  msg_use_alloca = true
+endif
+
+# Symbol visibility definition
+if cc.get_id() == 'msvc'
+  conf_data.set('EXPORT', '')
```
eli-schwartz, line 74: You can use the dllexport declspec here, surely?

subprojects/packagefiles/speexdsp/libspeexdsp/meson.build
```diff
@@ -0,0 +1,53 @@
...
+  'mdf.c',
+  'preprocess.c',
+  'resample.c',
+  'scal.c',
+  'smallft.c',
+]
+
+cc_is_not_msvc = cc.get_id() != 'msvc'
+
+m_dep = cc.find_library('m', required : cc_is_not_msvc)
```
eli-schwartz, line 17: https://mesonbuild.com/howtox.html#add-math-library-lm-portably

It's mandated by POSIX, it must exist but may be built into libc rather than existing as a distinct file. So this will either be found, or be unneeded, and you can simply unconditionally consider it `required: false`.

subprojects/packagefiles/speexdsp/libspeexdsp/meson.build
```diff
@@ -0,0 +1,53 @@
...
+
+speexdsp = library('speexdsp',
+  sources,
+  dependencies: m_dep,
+  include_directories : incdir,
+  vs_module_defs: '../win32/libspeexdsp.def',
+  version: lib_version,
+  soversion: cc_is_not_msvc ? lib_soversion : '',
```
eli-schwartz, line 25: Why is this msvc check needed?

subprojects/packagefiles/speexdsp/libspeexdsp/meson.build
```diff
@@ -0,0 +1,53 @@
...
+)
+
+pkgconfig = import('pkgconfig')
+pkgconfig.generate(speexdsp,
+  description: 'SpeexDSP is a patent-free, Open Source/Free Software DSP library',
+)
+
+# Unit Tests [TODO]
+#  - Figure out how to run all of them. Currently 'denoise',  'echo',
+#    and 'resample' don't run or sit waiting for stdin.
```
eli-schwartz, line 40: This is guarded by the upstream autotools as `if BUILD_EXAMPLES` so those may not be intended to work as unittests.

kcgen:
Thanks for the review and feedback, @dnicolodi and @eli-schwartz. Will get this PR tidied up tomorrow!

subprojects/packagefiles/speexdsp/meson.build
kcgen: I started this wrap based on the `ogg` files, so that's where this `lib_` stuff came from -- I'm removing all of it here because it's confusing and not adding any value that I can see. Thanks for the explanation @eli-schwartz !

subprojects/packagefiles/speexdsp/meson.build
kcgen: Yes; adding.

subprojects/packagefiles/speexdsp/meson.build
kcgen: Fixed

subprojects/packagefiles/speexdsp/libspeexdsp/meson.build
kcgen: It came from the `ogg` wrap:

``` text
subprojects/packagefiles/ogg/src/meson.build:  soversion: cc.get_id() == 'msvc' ? '' : lib_soversion,
```

(but now that I've stripped the auto-tools `lib_`* baggage, this is all going away)

subprojects/packagefiles/speexdsp/libspeexdsp/meson.build
kcgen: Thanks for catching that, @eli-schwartz!
The filename start with `test<feature>.c`, which threw me off. 
I've removed them.

subprojects/packagefiles/speexdsp/libspeexdsp/meson.build
kcgen: Perfect; now the wrap is compiler-agnostic. Thanks @eli-schwartz.

subprojects/packagefiles/speexdsp/meson.build
kcgen: Thanks @dnicolodi; this game from the existing `ogg` wrap.
I'm blowing all of this away - all C99 compilers must support vararrays (and the wrap requires  `c_std=c99` compliance), so these checks aren't needed.

subprojects/packagefiles/speexdsp/meson.build
kcgen: Thanks @dnicolodi; this game from the existing `ogg` wrap.
It's now gone.

subprojects/packagefiles/speexdsp/meson.build
kcgen: This auto-tools baggage has been tossed.

subprojects/packagefiles/speexdsp/meson.build
kcgen: Tried, along with a couple variations but the builds fails with

```text
error C2375: 'speex_preprocess_state_init': redefinition; different linkage
```

(For every function)

Checked how vcpkg is handling the `EXPORT` define, and they also make it an empty string for msvc:

https://github.com/microsoft/vcpkg/blob/314018fef43c7e3a52a3e46599313b70ace3977a/ports/speexdsp/CMakeLists.txt#L53

Going with that.

subprojects/packagefiles/speexdsp/meson.build
eli-schwartz: The ogg wrap does the math in question, and rationalizes using the confusion because... it allows trivially copying the values of C, A, and R from the configure.ac on updates. This may or may not be more or less confusing than checking the final filename of an autotools build, and then copying that as a statically coded string.

You can't remove *all* of it, you do need to pass the `version: ` kwarg somehow, whether you statically code it as `'1.5.1'` or copy in the C, A, R values from configure.ac.

For reference, the ogg files aren't *that* confusing. Here's what they do:

```meson
# See configure.ac and CMakeLists.txt
lib_current = 8
lib_revision = 5
lib_age = 8
lib_soversion = lib_current - lib_age
lib_version = '@0@.@1@.@2@'.format(lib_soversion, lib_age, lib_revision)
```

and the first 3 variables can be copy/pasted from configure.ac to suit.

subprojects/packagefiles/speexdsp/libspeexdsp/meson.build
eli-schwartz: IMO you should still build them, they are a useful verification that the programs *build* and especially link correctly against the library.

They just aren't useful as a unittest.

subprojects/packagefiles/speexdsp/meson.build
eli-schwartz: Ah, I get it now.

On GCC / clang, they use visibility attributes, and on MSVC they just use the vs_module_defs file which I see you already included. Makes sense. I think upstream should not require `define`ing this though, sad...

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,60 @@
...
+
+# Configure EXPORT of symbols
+if cc.get_id() == 'msvc' and get_option('default_library') != 'static'
+  cdata.set('EXPORT', '')
+elif cc.has_argument('-fvisibility=hidden')
+  add_project_arguments('-fvisibility=hidden', language: 'c')
+  cdata.set('EXPORT', 'extern __attribute__ ((visibility ("default")))')
+else
+  cdata.set('EXPORT', 'extern')
+endif
```
eli-schwartz, line 47: Instead of has_argument, please use [`cc.has_function_attribute()`](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerhas_function_attribute)

Instead of using add_project_arguments, you can use `gnu_symbol_visibility: 'hidden'` on the actual library

subprojects/packagefiles/speexdsp/meson.build
kcgen: thanks for going into more detail, @eli-schwartz .

Going with the C, A, and R copy-and-paste approach [from the speexdsp-1.2.0's configure.ac](https://github.com/xiph/speexdsp/blob/68311d46785be76d2a186c75578d51108bff6dfb/configure.ac#L3-L14):

```
lib_current = 6   
lib_revision = 1
lib_age = 5
lib_soversion = lib_current - lib_age
lib_version = '@0@.@1@.@2@'.format(lib_soversion, lib_age, lib_revision)
```

The math escapes me:

 - `lib_soversion` = 6 - 5 => **1**
 - `lib_age` = **5**
 - `lib_revision` = **1**

Putting this together into `lib_version =` gives me `1.5.1`, which I suppose matches the major version  of SpeexDSP's 1.2.0 -- but I've been even more confused by other autotool'isms than this :sweat_smile:

subprojects/packagefiles/speexdsp/meson.build
kcgen: Thanks for the tips - both are now in place.

eli-schwartz:
Thinking again about this `#define EXPORT` shenanigans... really, this should be solved by upstream. It really, really should. The proper way to do this is with the following preprocessor snippet:

```c
#if __GNUC__ >= 4
    #define EXPORT extern __attribute__ ((visibility ("default")))
#endif
```

As for speexdsp, it checks this in configure.ac and then... sticks this into the generated config.h, matching what we do here. But probably a better way would be to simply include this *exact* snippet, but with an `#else` branch, into config.h.meson. We can then skip doing a compile check in meson.

Ideally, I'd want to submit an improvement upstream to handle this define in the existing headers so that we don't need to handle it at all.

subprojects/packagefiles/speexdsp/meson.build
eli-schwartz: From the method description:
> [This table](https://mesonbuild.com/Reference-tables.html#gcc-__attribute__) lists all of the supported attributes.

You just use the attribute name (`visibility`), not the name of some command-line flag. :)

kcgen: (eventually realized that.. and now switched back to github :slightly_smiling_face: and see your tip!)
Working good now.

kcgen:
> Ideally, I'd want to submit an improvement upstream

Fully agree, although I've had limited luck upstream:
 - Refused to address Clang static analysis issues: https://github.com/xiph/opus/issues/146
 - Refused to acknowledge include path parent-child issues (and inconsistencies against their own other Xiph projects) https://github.com/xiph/opusfile/issues/10

eli-schwartz:
> * Refused to acknowledge include path parent-child issues (and inconsistencies against their own other Xiph projects)

I agree with their reasoning, though!
> An application can specify `#include <opus/opus.h>`, while only specifying the parent include path, such as `/usr/include` or` /usr/local/include`, knowing that the opus directory is inside.

An application must not do this, because:
- if the application is *manually* specifying the include path they can just as easily use the correct, expected one
- if the application is passing no paths, but relying on /usr/include and /usr/local/include being in the default include path, this will not work if opus is installed to a custom prefix

Additionally, the suggestion you offered in that ticket and which they rejected, will outright break in three cases:
- building opus as a meson subproject, `<opus/opus.h>` will not exist (`include/opus.h` does exist)
- building opusfile as a meson subproject, opusfile.h will not be able to include `"opus_multistream.h"` from the same directory
- building and installing opus and opusfile to two *different* prefixes will not be able to include `"opus_multistream.h"` from the same directory

Note that this is not a matter of being inconsistent with their parent project. The parent project uses `"opus_types.h"` specifically due to the fact that it is semantically correct for two header files distributed in the same source tree and installed to the same location in the same `make install`, to use the double quotes ("first check the same directory as this file") include style rather than the angle brackets include style. The reason is because they are shipped together, not because they are going to end up in the same installed location.

I don't think that's really a good reason to say that upstream isn't open to accepting improvements, especially if given a patch. But I can look into that myself.

kcgen:
> The parent project uses "opus_types.h" specifically due to the fact that it is semantically correct for two header files distributed in the same source tree and installed to the same location in the same

Thanks for the detailed response, @eli-schwartz . 

Indeed, it's the flip scenario (when projects are separate) that made me now realize that these headers are infact from separate projects (which was a point I didn't appreciate when I filed that bug):

/usr/include/opus/
```
-rw-r--r-- 1 root root  35K Mar 24 09:28 opus_defines.h
-rw-r--r-- 1 root root 114K Nov 24 01:33 opusfile.h
-rw-r--r-- 1 root root  48K Mar 24 09:28 opus.h
-rw-r--r-- 1 root root  33K Mar 24 09:28 opus_multistream.h
-rw-r--r-- 1 root root  28K Mar 24 09:28 opus_projection.h
-rw-r--r-- 1 root root 5.1K Mar 24 09:28 opus_types.h
```

 - **opusfile.h**: `# include <opus_multistream.h>`
 - **opus_projection.h**: `#include "opus_multistream.h"`

So even though the `opusfile.h` include seems inconsistent when taken a face-value (post-installation), it's actually fine given it originates from a separate project, and doesn't actually have `opus_multistream.h` as a "peer" header - so needs to find it externally, like you've explained.

:+1: 

> I don't think that's really a good reason to say that upstream isn't open to accepting improvements, especially if given a patch. But I can look into that myself.

I agree!

kcgen:
Ps- decided to delete my comment regarding upstream, as it's not relevant and my position was biased from my incorrect assessment; sorry for that.

subprojects/packagefiles/speexdsp/libspeexdsp/meson.build
kcgen: Decided to make the wrap more feature-complete with `meson_options`, and added `with-example-programs` to the options.  The executables are also placed inside an `./examples/` directory, to be extra-clear as to what they are.

kcgen:
Added options for numeric representation and building sample programs.

``` text
speexdsp 1.2.0

  Implementation Summary
    SIMD instructions     : SSE2
    Numerical type        : floating-point
    FFT library           : OggVorbis FFT (built-in)

  Build Summary
    Build type            : debug
    Link time optimization: False
    Library type          : both
    Install prefix        : /usr/local
    Example programs      : False
```

or with `-Duse-fixed-point=true -Dwith-example-programs=true`:

``` text
speexdsp 1.2.0

  Implementation Summary
    SIMD instructions     : SSE2
    Numerical type        : fixed-point
    FFT library           : Kiss FFT (built-in)

  Build Summary
    Build type            : debug
    Link time optimization: False
    Library type          : both
    Install prefix        : /usr/local
    Example programs      : True
```

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,113 @@
+project(
+  'speexdsp',
+  'c',
+  version : '1.2.0',
+  license : 'bsd-like',
```
eli-schwartz, line 5: This seems to be, specifically, the `BSD-3-Clause` license, FWIW.

subprojects/packagefiles/speexdsp/meson_options.txt
```diff
@@ -0,0 +1,2 @@
+option('use-fixed-point', type : 'boolean', value : false, description : 'Compile fixed-point library')
+option('with-example-programs', type : 'boolean', value : false, description : 'Build example programs')
```
eli-schwartz, line 2: Maybe these option names could be simplified to e.g. `fixed-point` and `examples`? The "with" feels a bit redundant in particular.

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,113 @@
...
+endif
+
+summary('SIMD instructions', simd_type,                           section : 'Implementation Summary')
+summary('Numerical type',    numerical_type,                      section : 'Implementation Summary')
+summary('FFT library',       fft_library,                         section : 'Implementation Summary')
+summary('Build type',        get_option('buildtype'),             section : 'Build Summary')
+summary('Link time optimization', get_option('b_lto'),            section : 'Build Summary')
+summary('Library type',      get_option('default_library'),       section : 'Build Summary')
+summary('Install prefix',    get_option('prefix'),                section : 'Build Summary')
+summary('Example programs',  get_option('with-example-programs'), section : 'Build Summary')
```
eli-schwartz, line 113: Are these particularly interesting in the summary?

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,113 @@
...
+
+# Populate config.h from the above data
+configure_file(input : 'config.h.meson',
+               output : 'config.h',
+               configuration : cdata)
+
+# Add accumulated flags to the build
+foreach arg : c_args
+  add_project_arguments(arg, language: 'c')
+endforeach
```
dnicolodi, line 90: I don't think the foreach loop here is necessary, you can pass an array of options to `add_project_arguments`. Also, you could avoid to accumulate options in an array, just call `add_project_arguments` directly when discovering the required options.

subprojects/packagefiles/speexdsp/meson_options.txt
dnicolodi, line 2: Other wraps use an option called simply `examples` for this. It would be nice to keep consistency, unless there are arguments for deviating from it.

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,113 @@
...
+endif
+
+# Set numerical representation config data
+numerical_type = ''
+fft_library = ''
+if get_option('use-fixed-point')
+  # the fixed-point implementation doesn't use SSE intrinsics (and arch.h
+  # checks for this), however we still let the compiler pack floats into
+  # SIMD calls instructions whenever its cost model permits. 
+  cdata.set('USE_SSE', false)
```
dnicolodi, line 55: This seems to be changing the `USE_SSE` coniguration data variable but the `summary()` call will still report the use of SSE2, unless I am missing something. Is this intended?

dnicolodi (commented):
The indentation is also inconsistent: sometimes you use two spaces and sometimes a variable number of chars.

subprojects/packagefiles/speexdsp/meson.build
kcgen: It's intended; I've tried to describe this better in the comment:

``` text
  # The fixed-point implementation isn't written using SSE intrinsics
  # and "arch.h" checks that USE_SSE is disabled when using fixed-point,
  # so we explicitly disable it here to pass that check:
  cdata.set('USE_SSE', false)
  # However, we still let the compiler pack integer operations into SIMD
  # calls whenever possible and we inform the user that SIMD is active in
  # the Summary block.
```

subprojects/packagefiles/speexdsp/meson_options.txt
kcgen: Switched to `examples`, thanks.

subprojects/packagefiles/speexdsp/meson.build
kcgen: Thanks; now passing in the `c_args` array, and also performing a single `get_supported_arguments` call here too.  So that lightens the load up-top.

I could have added `add_project_arguments` and `get_supported_arguments` at each point in the code that involves a flag, however that's a bit more cumbersome on the eyes (and repetitive), so I prefer accumulating a simple list of values across the entire `.build` followed by a single point that checks and adds the arguments. 

(It fits with other wraps, and gives future contributors a bit less load to deal with if they're adding a couple more features).

eli-schwartz: I'm perfectly okay with accumulating with an array, yeah.

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,136 @@
...
+        get_option('b_lto'),
+        section : 'Build Summary')
+
+summary('Library type',
+        get_option('default_library'),
+        section : 'Build Summary')
+
+summary('Install prefix',
+        get_option('prefix'),
+        section : 'Build Summary')
```
eli-schwartz, line 26: From the docs for summary:

> Instead of calling summary as summary(key, value), it is also possible to directly pass a dictionary to the [summary()](https://mesonbuild.com/Reference-manual_functions.html#summary) function, as seen in the example below.

You could do e.g.

```meson
summary({
    'Build type': get_option('buildtype'),
    'Link time optimization': get_option('b_lto'),
    'Library type': get_option('default_library'),
    'Install prefix': get_option('prefix'),
}, section: 'Build Summary')
```

kcgen:
> The indentation is also inconsistent: sometimes you use two spaces and sometimes a variable number of chars.

It now consistent:
 - non-returning function calls are aligned off the last `(`.
    ``` python
    function(arg1,
             arg2,
             arg3)
    ```
 - The one exception is configuration data, are paired are side-by-side
    ``` python
    cdata.set("FEATURE", true)
    ```
 - Literal lists are indented by two
    ``` python
    list = [
      "item1",
      "item2",
      "item3"]
    ```
 - Closing `]` and `)`s are always adjacent to the last value, to visually keep the indent level tidy (I agree this is a bit more tedious if someone's frequently changing the items in the list, but I prefer the small readability improvement over this cost).

subprojects/packagefiles/speexdsp/meson.build
kcgen: Fixed; thanks.

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,113 @@
...
+  'speexdsp',
+  'c',
+  version : '1.2.0',
+  license : 'bsd-like',
+  default_options : [
+    'c_std=c99',
+    'buildtype=release',
+    'default_library=both',
+    'b_lto=true',
+    'warning_level=0'])
```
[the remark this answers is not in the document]
kcgen: SpeexDSP is performance-critical. With DOSBox Staging, even when optimized it impacts our global FPS by ~10%.

I want to provide the most informed build options out-of-the-box against the most probable use case, with the theory that it should be easy to get a "good result" and hard to get a bad result.

This means:
 - Providing both static and shared libraries (maximizes flexibility out-of-the-box)
 - Providing an optimized build (Meson gives us LTO too and speexDSP works fine with it)
 - Eliminating visual clutter (reducing warnings: they're still there.. and if someone wants to dig in and help a project, they can turn them on and fix them).

subprojects/packagefiles/speexdsp/meson_options.txt
kcgen: Changed, thanks!

subprojects/packagefiles/speexdsp/meson.build
kcgen: Yes, we often see copy-and-paste console logs, and seeing these (specifically the first three) help a lot. 
For example, if a library is shared and the user's reporting unresolved .so problems, then they probably moved their binary (so this is a big help). 

We often have users reporting stuttering or frame hitching - and we need to drill in to realize they have some dependencies not optimized; seeing that up-front here is really helpful.

For informed users, having that little status block showing them that their build is optimized (or not), or that they're building shared libs instead of static is all the help they need to dive into the Meson docs and learn those options.

It's important that the summary is provided at the end of the setup phase instead of scolled off their console (some users are on non-GUI framebuffers, and scroll back doesn't even work).

subprojects/packagefiles/speexdsp/meson.build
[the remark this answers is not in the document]
eli-schwartz: Why is it flexible to provide both static and shared? When linking to it as a subproject, it will only ever use the shared version anyway. And in practice, if you build vendored wraps as a subproject, you want to link statically so that the built program is more relocatable.

As far as lto goes, I totally agree lto is great and you should use it, but... as a per-subproject setting it's entirely useless, because Meson ignores it.

Instead, you should be setting lto in the superproject (e.g. dosbox) and it will apply to the entire build, including subprojects.

As far as warnings go, I'm indifferent to them :) but note that you can always turn them off with `dependency('foobar', default_options: ['warning_level=0'])` even if you're using a wrap someone else contributed which has it set higher.

subprojects/packagefiles/speexdsp/meson.build
eli-schwartz: I am particularly wondering about the first 3. Build type isn't really interesting compared to the debug/optimization options, I generally assume that shared/static isn't very interesting except to people thinking about distribution, and lto, particularly, is a useless option because the subproject values are ignored and only the superproject value matters, so... there's nothing to print here.

subprojects/packagefiles/speexdsp/meson.build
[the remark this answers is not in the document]
kcgen: > Why is it flexible to provide both static and shared? 

If there's anyone using the meson files as-is (not as vendored in wraps) or if there are repo packagers using the meson build files, then this would generate both static and shared libraries, which is often lacking (and missed) by packagers. Being able to satisfy both shared linking and static lib requests ==  more flexible. But perhaps this usecase isn't common.

> When linking to it as a subproject, it will only ever use the shared version anyway. And in practice, if you build vendored wraps as a subproject, you want to link statically so that the built program is more relocatable.

Indeed; for our release builds (and when giving users build options), we already turn on all the static-lib options (they're quite a mess of them).

Maybe it makes sense for me to default to a static library -- what do you suggest?

> As far as lto goes, I totally agree lto is great and you should use it, but... as a per-subproject setting it's entirely useless, because Meson ignores it. 

Yes, we use LTO explicitly; so this setting is useless and will never be active in any "default" build scenario?  In that case.. I'll knock it out - but will hold off for you to confirm.

> Instead, you should be setting lto in the superproject (e.g. dosbox) and it will apply to the entire build, including subprojects.

We do set it explicitly on the command line, but not by default at the project level - because those builds are expensive and can't be ccache'd.  But LTO is cheap for speexDSP, so I'll explicitly set it for that in our defaults.

> As far as warnings go, I'm indifferent to them :) but note that you can always turn them off with dependency('foobar', default_options: ['warning_level=0']) even if you're using a wrap someone else contributed which has it set higher.

:+1: , good to know - and I will add that for our other wraps (which also flood the console).

subprojects/packagefiles/speexdsp/meson.build
kcgen: Knocked out the entire Build Summary. Default config is now:

``` python
        default_options : [
          'c_std=c99',
          'default_library=both',
          'warning_level=0'])
```

Summary is just build details:

``` text
  Library Summary
    Optimization level: 0
    SIMD instructions : SSE2
    Numerical type    : floating-point
    FFT library       : OggVorbis FFT (built-in)
```

kcgen: Knocked out the entire Build Summary, as suggested.  Remaining summary is as follows:

``` text
  Library Summary
    Optimization level: 0
    SIMD instructions : SSE2
    Numerical type    : floating-point
    FFT library       : OggVorbis FFT (built-in)
```

kcgen:
Ok, ready for round three - hopefully ticks all the boxes this time.
Thanks for the detailed reviews @eli-schwartz and @dnicolodi .

kcgen:
I believe all feedback has been addressed.

@eli-schwartz , I know you wanted to fix upstream's use of `#define EXPORT`.  

Given this PR provides parity with upstream's build as-is (which also uses the same EXPORT define), can we merge this wrap, and then bump to rev -2 if/when upstream accepts the EXPORT improvements?

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,112 @@
...
+c_args = []
+cc = meson.get_compiler('c')
+host_cpu_family = host_machine.cpu_family()
+
+# Round up configuration data
+add_project_arguments('-DHAVE_CONFIG_H',
+                      language: 'c')
+cdata = configuration_data()
+
+# Set SIMD config data and flags
```
eli-schwartz, line 27: Interestingly, upstream has --enable-sse=[yes|no] and --enable-neon=[yes|no] options. If not passed on the command line, one is automatically selected based on the current system. (Basically a Meson combo with a "check" option.)

The default behavior is to check if it can be enabled, but it's possible to manually configure this off and generate really portable libraries that support being moved off the current machine and onto one that uh doesn't even support sse2 because that's incredibly realistic, right?

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,112 @@
...
+simd_type = 'None'
+if host_cpu_family.startswith('x86')
+  cdata.set('USE_SSE', true)
+  cdata.set('USE_SSE2', true)
+  c_args += ['-mfpmath=sse',
```
eli-schwartz, line 32: mfpmath is not set upstream, just -msse -msse2... dunno if this matters.

mfpmath=sse is also the default for x86_64 anyway. :)

subprojects/packagefiles/speexdsp/meson.build
```diff
@@ -0,0 +1,112 @@
...
+  c_args += ['-mfpu=neon-vfpv4',
+             '-mfloat-abi=hard',
+             '/arch:VFPv4']
+  simd_type = 'NEON (vfpv4)'
+
+elif host_cpu_family == 'aarch64'
+  cdata.set('USE_NEON', true)
+  c_args += '-mfpu=neon-fp-armv8'
+  simd_type = 'NEON (armv8)'
+endif
```
eli-schwartz, line 49: Upstream just uses `-march=armv7-a -mfpu=neon` "if neon is requested", but I'll confess to knowing practically nothing about the various ARM options.

Are these options more fine-tuned for specific unames?

subprojects/packagefiles/speexdsp/meson.build
kcgen: > upstream has --enable-sse=[yes|no] and --enable-neon=[yes|no] 

I went with a default-enabled bool, called "simd", because it's already architecture-specific at the compiler level.  In the rare case someone wants to ship binaries for pentium3 and lesser hardware, they can toggle this off.

eli-schwartz: Yup, makes perfect sense.

Having separate options seems rather pointless, and the autotools default is to enable it anyway (via checking) so your implementation seems very sane.

subprojects/packagefiles/speexdsp/meson.build
kcgen: Ok, let's tighten this up!

SSE and SSE2 are always available on x64-64 CPUs, so we don't need flags. So I'm now handing x86-64 on its own and merely toggle the `cdata` items on to let the code use explicit SSE intrinsic code paths:

For 32bit x86, we do need `-mfpmath=sse`, because it's the only way to both enable and use SSE extensions. Ref: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html.  

If cmake is missing this, then I suppose they're assuming the user is externally adding this flag.. otherwise no SSE for them :sweat_smile:

subprojects/packagefiles/speexdsp/meson.build
kcgen: `-march=armv7-a` and `march=armv8-a` are both fine.

These represent the generic Armv7.0-A (and 8.0-A) architectures without any incremental architecture extensions. (v7.0-A is basically `-march=x86` and 8.0-A is `-march=x86_64` of the ARM world... as basic as they get)

`-mfpu=neon` is fine, but there's a more generic -`march` extension, the `+simd` flag, that can be tacked onto `-march=armxyz+simd` to request SIMD. I like that it's a bit more generic and its name nicely matches the Meson option.

kcgen:
@eli-schwartz , thanks for the SIMD-related feedback yesterday.
I believe all are addressed now.

subprojects/packagefiles/speexdsp/meson.build
eli-schwartz: - `-msse -msse2` "allows" the build to use intrinsics.
- `-mfpmath=sse` I think just enables automatic conversion of codegen to prefer intrinsics. Either way if you manually code the intrinsics yourself, they will be used.

If the library requires SSE anyway, I guess there's no reason not to set -mfpmath=sse though, it's pure benefit. So that makes sense.

Thanks for the clarification and code comments.

kcgen:
@eli-schwartz , thanks for all the help with this. 
I will be referring back to this wrap for future contributions, as it's full of very helpful feedback.

PR #444, opened 2022-05-20
SDL2: provide option to disable SDL2main

### LINKED ISSUE #443: SDL2 forces dependents to use SDL_main on Windows
The SDL2 meson wrap doesn't allow the project that's including it to disable linking the `sdl2main` library. This is problematic for appliactions that don't want to export `WinMain` such as non-graphical apps (where `win_subsystem` is `console`)  or apps that simply want to use the same main function for all platforms without including SDL headers. It also creates linker errors for me that I can't seem to work around.

Perhaps there should probably be an option checked [where use_sdl2main is set](https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/sdl2/meson.build#L1045), because currently it's always true on supported platforms.

Jan200101:
closes #443


subprojects/packagefiles/sdl2/meson_options.txt
```diff
@@ -196,3 +196,9 @@ option(
     type : 'boolean',
     description : 'If possible, dlopen dependencies at runtime instead of linking (linux only)'
 )
+
+option(
+    'disable_sdl2main',
+    type : 'boolean',
+    description : 'Disable including the main function helper'
+)
```
dnicolodi, line 204: Maybe a `sdl2main` boolean option with a default value of `true` would be more in line with the Meson style recommendation of avoiding `enable-` (and by extension `disable-`) prefixed option names.

Jan200101: I'd argue that this would overload the usage of `sdl2main` to mean an option and the library.

I think doing it the same way SDL does it upstream is preferable to reduce friction.

eli-schwartz: I don't see how it's overloaded?

`sdl2main` is a meson.build variable, `-Dsdl2main` would be an option flag.

It's extremely disappointing that upstream has `sdl2.pc` but not `sdl2main.pc` but such is life...

Jan200101: not in the sense of it overloading any variables, but overloading terminology.

but I'm in the minority here, so I shortened it to `sdl2main` and default to `true`

Jan200101:
I changed it up and instead of putting all of SDL2main behind a build option only the linking is behind it.

So now a project can disable automatically linking against it but still do it manually when desired.

For the sake of doing the same as upstream I also removed the dummy check since it exists as a generic drop in main implementation

Jan200101:
force pushing for a minor inconsistency that is annoying me

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -1042,47 +1042,51 @@ sdl2 = library('sdl2',
...
+    sdl_link_whole += main_link_whole
+    sdl_c_args += main_c_args
+endif
+
+sdl2_dep = declare_dependency(
+    link_with : [sdl2] + sdl_link_with,
+    link_whole : sdl_link_whole,
+    include_directories : core_inc,
+    compile_args : c_args,
+)
```
eli-schwartz, line 1088: Perhaps you could just have the option toggle adding sdl2main_dep as a dependency of sdl2_dep? It would save some repetition here.

Jan200101: oh yeah, that should make this simpler.

subprojects/packagefiles/sdl2/meson_options.txt
```diff
@@ -196,3 +196,10 @@ option(
     type : 'boolean',
     description : 'If possible, dlopen dependencies at runtime instead of linking (linux only)'
 )
+
+option(
+    'sdl2main',
+    type : 'boolean',
+    value : 'true',
+    description : 'Disable including the main function helper'
```
dnicolodi, line 204: The description is now misleading. Maybe it should be something like "Inlude the main() function helper."

jpakkane: Options that are of type "disable x" are also confusing because they can't easily not lead to double negatives.

Project option names should not be prefixed with the project name. Meson adds that automatically, so this would actually be called `sdl2:sdl2main` when using from the command line. A better name would be something like `with_main`.

eli-schwartz: > Project option names should not be prefixed with the project name. Meson adds that automatically, so this would actually be called `sdl2:sdl2main` when using from the command line. A better name would be something like `with_main`.

Agreed, although in this case it's a bit ambiguous because the sdl2main option enables linking a static library called, literally, `sdl2main`. So I guess I can understand why people may think it could go either way.

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -1042,47 +1042,47 @@ sdl2 = library('sdl2',
...
 endif
 
+sdl2main_dep = declare_dependency(
+    link_with : main_link_with,
+    link_whole : main_link_whole,
+    include_directories : core_inc,
+    compile_args : main_c_args,
+)
+
+message('sdl2main: @0@'.format(get_option('sdl2main')))
```
eli-schwartz, line 1072: This is still using the old option name.

eli-schwartz (changes requested):

subprojects/packagefiles/sdl2/meson.build
Jan200101: yep

leftover from some testing, forgot to remove it

PR #445, opened 2022-05-30
Add wrap files for iir-1.9.2

kcgen:


subprojects/packagefiles/iir1/iir/meson.build
```diff
@@ -0,0 +1,41 @@
...
+  'Custom.cpp',
+  'PoleFilter.cpp',
+  'RBJ.cpp')
+
+cpp_args = []
+cxx = meson.get_compiler('cpp')
+host_cpu_family = host_machine.cpu_family()
+
+# Maybe disable exceptions
+if not get_option('exceptions')
```
eli-schwartz, line 16: Lovely! Upstream doesn't have an option and simply documents using this in README...

subprojects/packagefiles/iir1/iir/meson.build
```diff
@@ -0,0 +1,41 @@
...
+elif host_cpu_family == 'aarch64'
+  cpp_args += '-mfpu=neon-fp-armv8'
+endif
+# Note: MSVC by-default uses SSE2 instructions (unless overridden)
+# https://docs.microsoft.com/en-us/cpp/build/reference/arch-x86
+
+add_project_arguments(cxx.get_supported_arguments(cpp_args),
+                      language: 'cpp')
+
+lib = library('iir1',
```
eli-schwartz, line 36: The upstream library name is "iir", no 1 at the end.

There's also the cmake equivalent of Meson's `version: meson.project_version()`.

subprojects/packagefiles/iir1/test/meson.build
```diff
@@ -0,0 +1,15 @@
...
+  { 'name' : 'chebyshev2',  'should_fail' : false },
+  { 'name' : 'custom',      'should_fail' : false },
+  { 'name' : 'rbj',         'should_fail' : false },
+  { 'name' : 'state',       'should_fail' : false },
+]
+
+foreach t : tests
+  name = t.get('name')
+  exe = executable(name, name + '.cpp', dependencies : libiir1_dep)
+  test(name, exe, should_fail : t.get('should_fail'))
```
eli-schwartz, line 14: I haven't investigated, but upstream lists all these as ordinary cmake add_tests, can you clarify why some of them are should_fail?

eli-schwartz (changes requested):
Some comments on compatibility with the upstream build results.

There's also a demo directory which should maybe get built too.

releases.json
```diff
@@ -584,6 +584,14 @@
       "55.2-1"
     ]
   },
+  "iir1": {
+    "dependency_names": [
+      "iir1"
```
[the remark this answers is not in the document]
kcgen: Renamed.

subprojects/packagefiles/iir1/iir/meson.build
```diff
@@ -0,0 +1,41 @@
...
+host_cpu_family = host_machine.cpu_family()
+
+# Maybe disable exceptions
+if not get_option('exceptions')
+  cpp_args += '-DIIR1_NO_EXCEPTIONS'
+endif
+
+# Set commmonly available SIMD flags
+if host_cpu_family.startswith('x86')
+  cpp_args += ['-mfpmath=sse', '-msse2']
```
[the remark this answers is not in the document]
kcgen: Removed SIMD flags.
Added default warning Level 3.

subprojects/packagefiles/iir1/test/meson.build
kcgen: >  ordinary cmake add_tests

I use meson to avoid everything to do with cmake, so I'm not sure what these are.

If there's a more correct way of running a self test, in these cases, beyond the "test(...)" call, then I'm happy to be pointed to that and will adjust.

subprojects/packagefiles/iir1/iir/meson.build
kcgen: Renamed.
Added creation of simple pkgconfig and cmake records.

subprojects/packagefiles/iir1/iir/meson.build
kcgen: 👍

subprojects/packagefiles/iir1/test/meson.build
eli-schwartz: What I mean is I don't see anywhere in the cmake build description that says this command is supposed to return non-zero.

So I got a chance to try this wrap locally.

```
$ ninja -C builddir test
ninja: Entering directory `builddir'
[0/1] Running all tests.
1/7 badparam           UNEXPECTEDPASS  0.02s
>>> MALLOC_PERTURB_=230 /home/eschwartz/git/mesonbuild/wrapdb/builddir/subprojects/iir1-1.9.2/test/badparam
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
stderr:
Correct lowpass exception thrown for fc = fs/2: The cutoff frequency needs to be below the Nyquist frequency.
Correct highpass exception thrown for fc = fs/2: The cutoff frequency needs to be below the Nyquist frequency.
Correct bandstop exception thrown for fc = fs/2: The cutoff frequency needs to be below the Nyquist frequency.
Correct bandpass exception thrown for fc = fs/2: The cutoff frequency needs to be below the Nyquist frequency.
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

2/7 chebyshev2         OK              0.01s
3/7 custom             OK              0.01s
4/7 chebyshev1         OK              0.03s
5/7 rbj                OK              0.02s
6/7 state              OK              0.01s
7/7 butterworth        OK              0.44s

Summary of Failures:

1/7 badparam    UNEXPECTEDPASS  0.02s


Ok:                 6   
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    1   
Skipped:            0   
Timeout:            0   

Full log written to /home/eschwartz/git/mesonbuild/wrapdb/builddir/meson-logs/testlog.txt
FAILED: meson-test 
/home/eschwartz/.local/bin/meson test --no-rebuild --print-errorlogs
```

The should_fail is wrong, `test/badparam.cpp` uses the following pattern:
```cxx
        try {
                Iir::Butterworth::LowPass<3> f;
                f.setup (samplingrate, samplingrate / 2);
                assert_print(0,"No exception thrown by LowPass.");
        } catch ( const std::invalid_argument& e ) {
                fprintf(stderr,"Correct lowpass exception thrown for fc = fs/2: %s\n",e.what());
        }
```

> I use meson to avoid everything to do with cmake, so I'm not sure what and "ordinary cmake add_test" is versus other cmake non-ordinary or non-add_tests.

I use meson to avoid cmake too, but sadly I still have to know some cmake. :D I would expect that if a test is expected to return non-zero, Meson would have `should_fail: true` as a kwarg on the test, and cmake would have this sometime after the test to modify the test:
```cmake
set_tests_properties(testname PROPERTIES WILL_FAIL true)
```

subprojects/packagefiles/iir1/iir/meson.build
eli-schwartz: The library still needs to have a `version:` kwarg set.

(I would not bother with the cmake record, since cmake has several different files you need to install, and just the version file won't be enough to get things working. Luckily the pkg-config file is super simple and stands complete on its own!)

subprojects/packagefiles/iir1/test/meson.build
eli-schwartz: The reason this isn't picked up in the WrapDB CI is because the tests are disabled by default, and aren't re-enabled in ci_config.json

kcgen: Thanks for that correction; fixed.

subprojects/packagefiles/iir1/iir/meson.build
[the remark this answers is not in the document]
eli-schwartz: Unless I'm missing something, you added a default warning_level of 0 instead? Typo?

subprojects/packagefiles/iir/demo/meson.build
```diff
@@ -0,0 +1,9 @@
+demos = [
+  'iirdemo',
+  'ecg50hzfilt']
+
+foreach demo : demos
+  exe = executable(demo,
```
eli-schwartz, line 6: This assignment now does nothing, `exe` is never used.

Before your recent push, it was used to define a `test()`, and the removal of the test is correct since demos aren't necessarily supposed to be tested. (Also, the demos expect to be run in the working directory of the source code, ecg50hzfilt actually segfaults otherwise. Then it goes and writes data files to the source tree, which is *ugly* and definitely should not be triggered by a testsuite. People can manually run the demos if they want. :p)

tl;dr

Clean up from the test removal even more by removing this assignment. :)

subprojects/packagefiles/iir/meson.build
```diff
@@ -0,0 +1,20 @@
+project('iir', 'cpp',
+        version : '1.9.2',
+        license : 'MIT',
```
eli-schwartz, line 3: We can also add a `meson_version: '>=0.50.0'` kwarg to the `project()` function. Currently, the two most recent features this wrap uses are:

```
iir| * 0.46.0: {'pkgconfig.generate optional positional argument'}
iir| * 0.50.0: {'CMake Module'}
```

subprojects/packagefiles/iir/iir/meson.build
```diff
@@ -0,0 +1,32 @@
...
+libiir_dep = declare_dependency(include_directories : incdirs,
+                                link_with: lib)
+
+pkgconfig = import('pkgconfig')
+pkgconfig.generate(lib,
+                   description: 'IIR is an an infinite impulse response (IIR) filter library',
+                   version: meson.project_version())
+
+cmake = import('cmake')
+cmake.write_basic_package_version_file(name: meson.project_name(),
```
eli-schwartz, line 30: As I said buried in another comment, this version file isn't very useful on its own, because unlike pkg-config, you need several files at once for a full cmake dependency.

Also, note that this write_basic_package_version_file() method will cause building to fail with `ERROR: Unable to find cmake` if cmake is not installed, which is a bit unfortunate.

subprojects/packagefiles/iir/meson.build
eli-schwartz: (If you remove use of the cmake module, you can set it to 0.46.0 instead of 0.50.0.)

kcgen: Thanks for all the pointers @eli-schwartz ; will get these addressed.

subprojects/packagefiles/iir/iir/meson.build
kcgen: I'll delete the cmake config creation chunk.

You mentioned `pkgconfig` still needing a version kwarg, implying the above `version: meson.project_version()` is not the way to do it.

Because this is a cmake package, it doesn't have a `configure.ac` to lift out the prior `soversion` details like you helped me with in the speexdsp.

What's the easiest way to satisfy this?

subprojects/packagefiles/iir/demo/meson.build
kcgen: Done.

subprojects/packagefiles/iir1/iir/meson.build
kcgen: Added.

subprojects/packagefiles/iir/meson.build
kcgen: Added `meson_version: '>=0.46.0`'

subprojects/packagefiles/iir/iir/meson.build
kcgen: Removed cmake block.

eli-schwartz: No :) I said `library()` needs a version kwarg, to ensure that it gets created with the correct soname.

The upstream cmake configuration is very simple though, it sets cmake's `SOVERSION 1 VERSION ${PROJECT_VERSION}`

project version is 1.9.2 of course, so that maps to `version: meson.project_version()` in Meson, no need to set the soversion separately as Meson defaults to "the first part of version" if not set.

eli-schwartz:
Comparing the install tree from upstream to the one here, I noticed that we aren't installing:
- `Iir.h` to the default includedir
- iir/*.h to `subdir: 'iir'`

subprojects/packagefiles/iir/iir/meson.build
kcgen: > project version is 1.9.2 of course, so that maps to version: meson.project_version() in Meson

Thanks; now set:

``` python
lib = library('iir',
              sources,
              include_directories : incdirs,
              install: true,
              version: meson.project_version())
```

subprojects/packagefiles/iir/iir/meson.build
```diff
@@ -0,0 +1,46 @@
...
+  'Custom.h',
+  'Layout.h',
+  'MathSupplement.h',
+  'PoleFilter.h',
+  'RBJ.h',
+  'State.h',
+  'Types.h')
+
+install_headers(headers,
+                subdir: '.')
```
eli-schwartz, line 27: Like I said in my comment above, this needs to be `subdir: 'iir'`, because Iir.h (the main header) is an "include everything" file that then does:

```c
#include "iir/Common.h"
#include "iir/Biquad.h"
#include "iir/Cascade.h"
#include "iir/PoleFilter.h"
#include "iir/State.h"
#include "iir/Butterworth.h"
#include "iir/ChebyshevI.h"
#include "iir/ChebyshevII.h"
#include "iir/Custom.h"
#include "iir/RBJ.h"
```

So the component files need to be installed to a subdir next to Iir.h (mirroring the layout in the source tree, actually).

kcgen:
> Comparing the install tree from upstream to the one here, I noticed that we aren't installing:
> 
>     * `Iir.h` to the default includedir
> 
>     * iir/*.h to `subdir: 'iir'`

I'm now capturing the headers explicitly using files(..) and using install_headers, for both the top level header and the internal lib headers. Both installing to their local relative directory. 

I'm not sure how to check this though using the compile action though.

eli-schwartz:
> I'm now capturing the headers explicitly using files(..)

I'm not sure what "capturing the XXX explicitly" means, but `files('foo.h')` acts identical to the string type `'foo.h'` in all contexts that accept files, with one critical distinction:

files() can be used in a different directory, e.g. if you then `subdir('bar')` the file will act like the string `'../foo.h'` instead. With strings, you need to keep track of ../../ and all that junk -- with files() you do not.

It doesn't do anything other than help you define the input file, so you have to keep track of the install directory (e.g. `subdir: ` for install_headers) regardless.

kcgen:
Ok, so I don't need to use files() because I'm not making use of them across when changing the working path. I'll drop those uses and go with lists of strings.

eli-schwartz:
> Ok, so I don't need to use files() because I'm not making use of them across working path changes. I'll drop those uses and go with lists of strings.

Meson doesn't care, and it does not hurt except in the subjective opinionated sense that some people may think it is ugly, while other people insist on using it.

All I'm doing is clarifying that it does not exempt you from the `subdir: ` argument to install_headers().

subprojects/packagefiles/iir/iir/meson.build
kcgen: That's confusing. 

If non-install calls operate on files relative to the meson.build file, why would install calls not do the same?

It appears install actions drop this tree relationship awareness.

I'll add this subdirectory, none the less.

Thanks for explaining this.

eli-schwartz: For the same reason any other install calls, simply install directly to whichever install_dir they have.

executable installs to `get_option('bindir')`, even if it is in src/meson.build
library installs to `get_option('libdir')`, even if it is in lib/meson.build

So too, install_headers() installs to `get_option('includedir')`, even if it is in iir/meson.build

But install_headers() has a common requirement to install to "includedir, but a subdirectory of it", which is why it has a `subdir:` option in addition to `install_dir:`

We *cannot* automatically install to iir/, because often people have all their includes in include/meson.build, and it would be wrong to install them to `get_option('includedir') / 'include'`.

:)

kcgen:
> All I'm doing is clarifying that it does not exempt you from the subdir: argument to install_headers().

Indeed; I was hoping it's use would pin these headers into their respective relative paths (and relative to the  their meson.build files), but alas, that's not the case. Thanks for explaining this.

subprojects/packagefiles/iir/iir/meson.build
eli-schwartz: Also, can you please remove `subdir: '.'` from the other one? :) It does nothing.

eli-schwartz:
The thing you were expecting was probably this: https://github.com/mesonbuild/meson/pull/10299

subprojects/packagefiles/iir/iir/meson.build
kcgen: Done.

eli-schwartz:
You reminded me to merge it. :D

kcgen:
Thanks @eli-schwartz.
Next step is un-porting the iir sources and leveraging the wrap! 🥳

eli-schwartz:
Upstream iir has merged https://github.com/berndporr/iir1/pull/37 which means that a system copy can now be reliably detected too.

PR #450, opened 2022-05-31
libupnp: pkgconfig name fix + cleanups

neheb:
The first is important for other packages to find it.

The rest is cosmetic.


neheb:
@eli-schwartz so UPNP_VERSION gets defined wrong here. I think it's related to some libtool weirdness that you mentioned in a different PR. Can you take a look?

neheb:
I sort of fixed this. However now I see this:

```
/** The library version (string) e.g. "1.3.0" */
#define UPNP_VERSION_STRING ""

/** Major version of the library */
#define UPNP_VERSION_MAJOR

/** Minor version of the library */
#define UPNP_VERSION_MINOR
#ifndef UPNP_VERSION_MINOR
#       define UPNP_VERSION_MINOR 0
#endif

/** Patch version of the library */
#define UPNP_VERSION_PATCH
#ifndef UPNP_VERSION_PATCH
#       define UPNP_VERSION_PATCH 0
#endif
```

when built standalone, these get populated properly. When built as a subproject, they do not.

neheb:
yeah... seems like a meson bug.

neheb:
any chance of getting this merged?

PR #456, opened 2022-06-03
Add proxy-libintl wrap

xclaesse:
This is a dependency of GLib and can be used as fallback for gettext on
all platforms.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1323,6 +1323,14 @@
       "1.3.0-1"
     ]
   },
+  "proxy-libintl": {
+    "versions": [
+      "0.4-1"
+    ],
+    "dependency_names": [
+      "intl"
+    ]
+  },
   "pugixml": {
     "dependency_names": [
       "pugixml"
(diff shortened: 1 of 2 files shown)


eli-schwartz:
I stubbed this out too quite some time ago: bb1ff98d9603a9c187f4961711f67b610bd22abe

I never submitted it (or even set up a provides section, apparently), because as I wrote in the commit message, it seems like you're not "really" supposed to use it automatically as though it was a real intl, unless you really know you want it specifically?

xclaesse:
It can dlopen() the real libintl at runtime to not require it at build time. But it seems meson.build has `-DSTUB_ONLY` so it's all no-op functions. In practice glib (and all projects that depend on glib) use proxy-libintl as fallback to disable translation. If translation is needed then you need to real gettext which cannot be done as subproject.

bgilbert:
I have a Windows cross build that does indeed want to stub out libintl, and it'd be nice if this were available in WrapDB.

xref https://github.com/mesonbuild/wrapdb/issues/773

bgilbert:
Any chance of getting this in?  It's useful and seems harmless.

PR #459, opened 2022-06-04
actions: update to 22.04

neheb:
-latest is actually 20.04. 22.04 should be better as it has a newer GCC.

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   Ubuntu:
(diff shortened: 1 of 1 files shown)


eli-schwartz:
When is github planning to update this? The point of "latest" is supposed to be keeping it up to date in the long term, a short transition period when it "could" be updated slightly is probably not too terrible...

neheb:
No idea. Just saw this: https://github.com/actions/virtual-environments/commit/a814bcd5134b607cd09e188d99ac0e95ef33d8af

Maybe it would be better to test multiple versions.

PR #460, opened 2022-06-05
sdl2-mixer: try to fix on MSVC

neheb:


neheb:
```
 /home/runner/work/wrapdb/wrapdb/_build/../subprojects/SDL2-2.0.20/src/main/dummy/SDL_dummy_main.c:14: undefined reference to `SDL_main'
```

right...

neheb:
@Jan200101 any idea?

Jan200101:
issue occurs because its trying to link libSDL2.a against sdl2mixer, which results in it trying to resolve `SDL_main`, which normally is just `main` which SDL2 renamed using preprocessors.

But because this is a library there is no main it cannot resolve `SDL_main` and thus the error.

The solution in this case would be to simply link SDL2 without main for the library and SDL2 and SDLmain to playmus.

Here is a patch that does exactly that, uploaded as a txt because github
[mixer.patch.txt](https://github.com/mesonbuild/wrapdb/files/8840278/mixer.patch.txt)

neheb:
yeah,,,no.

```
sdl2_mixer| WARNING: Inconsistency: Subproject has overridden the dependency with another variable than 'sdl2main_dep'
```

neheb:
Fixed another way.

@Jan200101 the error in in SDL2, not in these packages. Note that all of them currently fail.

```
grep FAILED\: err.txt
FAILED: subprojects/box2d-2.4.1/unit-test/hello_world.p/hello_world.cpp.o
FAILED: subprojects/libfuse-fuse-2.9.9/util/ulockmgr_server.p/ulockmgr_server.c.o
FAILED: subprojects/libebml-release-1.4.2/liblibebml.a.p/src_EbmlString.cpp.o
FAILED: subprojects/libebml-release-1.4.2/liblibebml.a.p/src_EbmlUnicodeString.cpp.o
FAILED: subprojects/ffmpegthumbnailer-2.2.2/libffmpegthumbnailer.so.4.15.1.p/libffmpegthumbnailer_moviedecoder.cpp.o
FAILED: subprojects/SDL2_image-2.0.5/libsdl2image.so
FAILED: subprojects/SDL2_mixer-2.0.4/libsdl2mixer.so
FAILED: subprojects/SDL2_net-2.0.1/libsdl2_net.so
FAILED: subprojects/SDL2_ttf-2.0.12/libsdl2_ttf.so
```

Issue #455, opened 2022-06-03
CI failures


xclaesse:
Build All CI has been broken for a while, let's track here the failures:

- Linux:
  - libuv - https://github.com/mesonbuild/wrapdb/pull/367, `../../home/runner/work/wrapdb/wrapdb/subprojects/libuv-v1.44.1/include/uv/unix.h:136:9: error: unknown type name ‘pthread_rwlock_t’`

- MacOs:
  - sdl2 - `./../../../../../../Users/runner/work/wrapdb/wrapdb/subprojects/SDL2-2.0.20/src/core/linux/../../video/x11/SDL_x11sym.h:46:1: error: function cannot return function type 'int (Display *, Pixmap, Pixmap, XColor *, XColor *, unsigned int, unsigned int)' (aka 'int (struct _XDisplay *, unsigned long, unsigned long, XColor *, XColor *, unsigned int, unsigned int)')`

neheb:
@xclaesse an alpine linux CI will catch more failures like the first. That's been my experience with OpenWrt which also uses musl.

neheb:
https://github.com/mesonbuild/wrapdb/pull/459 should expose more.

neheb:
Hmm not too bad:

```
417:FAILED: subprojects/box2d-2.4.1/unit-test/hello_world.p/hello_world.cpp.o
3193:FAILED: subprojects/libfuse-fuse-2.9.9/util/ulockmgr_server.p/ulockmgr_server.c.o
8578:FAILED: subprojects/libebml-release-1.4.2/liblibebml.a.p/src_EbmlString.cpp.o
8609:FAILED: subprojects/libebml-release-1.4.2/liblibebml.a.p/src_EbmlUnicodeString.cpp.o
8867:FAILED: subprojects/ffmpegthumbnailer-2.2.2/libffmpegthumbnailer.so.4.15.1.p/libffmpegthumbnailer_moviedecoder.cpp.o
19130:FAILED: subprojects/SDL2_image-2.0.5/libsdl2image.so
19840:FAILED: subprojects/SDL2_mixer-2.0.4/libsdl2mixer.so
20485:FAILED: subprojects/SDL2_net-2.0.1/libsdl2_net.so
21244:FAILED: subprojects/SDL2_ttf-2.0.12/libsdl2_ttf.so
```

ninjaoflight:
#465 should fix the issue for hinnant-date in Windows.

xclaesse:
[...]

PR #465, opened 2022-06-05
fix hinnant-date on windows

ninjaoflight:
fix hinnant-date on windows when using the 'use_system_tzdb' on windows and display a warning instead, this should fix the issue for hinnant-date in #455 
- added as hinnant-date 3.0.1-2


subprojects/packagefiles/hinnant-date/meson.build
```diff
@@ -34,8 +34,13 @@ endif
...
 elif get_option('use_system_tzdb')
-  defines += ['-DAUTO_DOWNLOAD=0', '-DHAS_REMOTE_API=0']
-  tz_public_defines += ['-DUSE_OS_TZDB=1']
+  # only works if not linux
+  if host_machine.system() != 'windows'
+    defines += ['-DAUTO_DOWNLOAD=0', '-DHAS_REMOTE_API=0']
+    tz_public_defines += ['-DUSE_OS_TZDB=1']
+  endif
+  # display a warning but not fail
+  message('hinntant-date[WARNING]: use_system_tzdb is not supported on windows')
```
jpakkane, line 43: It is a bit confusing that the comment says "not linux" but the check is "not windows".

The warning message has a typo (hinntant-date) and it should be inside an `else` block so it is only printed when needed.

ninjaoflight: thanks!, I forgot to change the comment line while I was cheking it

PR #470, opened 2022-06-09
protobuf: update to 3.20.1

neheb:
Neeeded by the protobuf-c wrap.


neheb:
Something tells me something needs to be added under provide...

neheb:
@jpakkane this is ready to go. changed it to a version bump and added the ability to build standalone.

PR #472, opened 2022-06-13
libusb: add

neheb:

### DIFF
--- a/subprojects/packagefiles/libusb/meson.build
+++ b/subprojects/packagefiles/libusb/meson.build
@@ -1,0 +1,119 @@
+project('libusb', 'c', version: '1.0.26', meson_version: '>=0.60')
+
+add_project_arguments('-D_GNU_SOURCE', language: 'c')
+
+cc = meson.get_compiler('c')
+pconf = import('pkgconfig')
+
+thread_dep = dependency('threads')
+udev_dep = dependency('', required : false)
+macos_dep = dependency('', required: false)
+
+cdata = configuration_data()
+cdata.set('HAVE_SYSLOG', cc.has_function('syslog'))
+
+if not cdata.get('HAVE_SYSLOG') and get_option('system-log')
+  error('system logging requires syslog to be present')
+endif
+
+cdata.set('HAVE_CLOCK_GETTIME', cc.has_function('clock_gettime'))
+cdata.set('HAVE_EVENTFD', cc.has_function('eventfd'))
+cdata.set('HAVE_TIMERFD', cc.has_function('timerfd_create'))
+cdata.set('HAVE_PIPE2', cc.has_function('pipe2'))
+cdata.set('HAVE_PTHREAD_CONDATTR_SETCLOCK', cc.has_function('pthread_condattr_setclock', dependencies: thread_dep))
+cdata.set('HAVE_PTHREAD_SETNAME_NP', cc.has_function('pthread_setname_np', dependencies: thread_dep))
+cdata.set('HAVE_PTHREAD_THREADID_NP', cc.has_function('pthread_threadid_np', dependencies: thread_dep))
+
+cdata.set('HAVE_ASM_TYPES_H', cc.has_header('asm/types.h'))
+cdata.set('HAVE_NFDS_T', cc.has_header('nfds.h'))
+cdata.set('HAVE_STRING_H', cc.has_header('string.h'))
+cdata.set('HAVE_SYS_TIME_H', cc.has_header('sys/time.h'))
+
+cdata.set('HAVE_STRUCT_TIMESPEC', cc.has_header_symbol('time.h', 'struct timespec'))
+
+if cc.get_id() == 'msvc'
+  cdata.set('PRINTF_FORMAT(a, b)', '')
+else
+  cdata.set('PRINTF_FORMAT(a, b)', '__attribute__ ((__format__ (__printf__, a, b)))')
+endif
+
+cdata.set('ENABLE_LOGGING', get_option('logging'))
+cdata.set('ENABLE_DEBUG_LOGGING', get_option('debug-logging'))
+cdata.set('USE_SYSTEM_LOGGING_FACILITY', get_option('system-log'))
+
+incdirs = include_directories(['.', 'libusb', 'libusb/os'])
+install_headers('libusb/libusb.h', subdir: 'libusb-1.0')
+
+sources = files(
+  'libusb/core.c',
+  'libusb/descriptor.c',
+  'libusb/hotplug.c',
+  'libusb/io.c',
+  'libusb/strerror.c',
+  'libusb/sync.c',
+ )
+
+if host_machine.system() == 'windows'
+  cdata.set('PLATFORM_WINDOWS', true)
+  cdata.set('DEFAULT_VISIBILITY', '')
+  sources += files('libusb/os/events_windows.c', 'libusb/os/threads_windows.c')
+else
+  cdata.set('PLATFORM_POSIX', true)
+  cdata.set('DEFAULT_VISIBILITY', '__attribute__ ((visibility ("default")))')
+  sources += files('libusb/os/events_posix.c', 'libusb/os/threads_posix.c')
+endif
+
+if host_machine.system() == 'linux'
+  sources += 'libusb/os/linux_usbfs.c'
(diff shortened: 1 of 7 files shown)


subprojects/packagefiles/libusb/meson.build
```diff
@@ -0,0 +1,100 @@
...
+elif host_machine.system() == 'windows'
+  sources += files('libusb/os/windows_common.c', 'libusb/os/windows_usbdk.c', 'libusb/os/windows_winusb.c')
+elif host_machine.system() == 'darwin'
+  sources += files('libusb/os/darwin_usb.c')
+elif host_machine.system() == 'netbsd'
+  sources += files('libusb/os/netbsd_usb.c')
+elif host_machine.system() == 'openbsd'
+  sources += files('libusb/os/openbsd_usb.c')
+elif host_machine.system() == 'sunos'
+  sources += files('libusb/os/sunos_usb.c')
```
eli-schwartz, line 72: A lot of this can be simplified as:
```
elif host_system in ['darwin', 'netbsd', 'openbsd', 'sunos']
    sources += files('libusb/os/@0@_usb.c'.format(host_system))
else
```

neheb:
Good thing MSYS2 is not in the CI. Otherwise this would never pass.

subprojects/packagefiles/libusb/meson_options.txt
```diff
@@ -0,0 +1,18 @@
...
+  value: true,
+  description: 'use udev for device enumeration and hotplug support (recommended)',
+)
+
+option('debug_logging', type : 'boolean',
+  value: false,
+  description : 'use udev for device enumeration and hotplug support (recommended)',
+)
+
+option('system-log', type : 'boolean',
```
eli-schwartz, line 15: One uses underscores, the other uses a dash?

subprojects/packagefiles/libusb/meson.build
```diff
@@ -0,0 +1,115 @@
...
+
+if cc.get_id() == 'msvc'
+  cdata.set('PRINTF_FORMAT(a, b)', '')
+else
+  cdata.set('PRINTF_FORMAT(a, b)', '__attribute__ ((__format__ (__printf__, a, b)))')
+endif
+
+cdata.set('ENABLE_LOGGING',  get_option('logging'))
+cdata.set('ENABLE_DEBUG_LOGGING',  get_option('debug_logging'))
+cdata.set('USE_SYSTEM_LOGGING_FACILITY',  get_option('system-log'))
```
eli-schwartz, line 36: typo in the option, 'systemlog' -> 'system-log'.

Also upstream skips this on Windows and android, and checks for syslog.h, which is only part of POSIX XSI.

eli-schwartz: Oh, you fixed the typo already...

neheb:
```
 ../subprojects/libusb-1.0.26/examples/fxload.c(32): fatal error C1083: Cannot open include file: 'getopt.h': No such file or directory
```

I don't see the value in wasting time implementing examples and tests.

eli-schwartz:
Huh, interestingly, upstream does have `msvc/getopt/` presumably for this. Of course this is MSVC-specific, so it is only implemented in their manual VS solution files, not in the autotools build... :)

neheb:
```
 Starting test run: init_and_exit...
Failed to init libusb on iteration 0: -99
Failure (1)
Starting test run: get_device_list...
Failed to init libusb: -99
Failure (1)
Starting test run: many_device_lists...
Failed to init libusb: -99
Failure (1)
Starting test run: default_context_change...
Failed to init libusb: -99
Failure (1)
---
Ran 4 tests
Passed 0 tests
Failed 4 tests
Error in 0 tests
Skipped 0 tests
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Summary of Failures:
1/1 libusb_test FAIL            0.00s   exit status 1
Ok:                 0   
Expected Fail:      0   
Fail:               1   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   
```

Works locally.

neheb:
```
1/1 libusb_test FAIL            0.02s   (exit status 3221225781 or signal 3221225653 SIGinvalid)
```

Sigh this reminds me of the ogg failure. Actually I bet it's the same.

neheb:
```
 Starting test run: init_and_exit...
Failed to init libusb on iteration 0: -99
Failure (1)
Starting test run: get_device_list...
Failed to init libusb: -99
Failure (1)
Starting test run: many_device_lists...
Failed to init libusb: -99
Failure (1)
Starting test run: default_context_change...
Failed to init libusb: -99
Failure (1)
---
Ran 4 tests
Passed 0 tests
Failed 4 tests
Error in 0 tests
Skipped 0 tests
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Summary of Failures:
1/1 libusb_test FAIL            0.01s   exit status 1
Ok:                 0   
Expected Fail:      0   
Fail:               1   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   
```

weird thing is, it works fine locally under Fedora.

neheb:
Fixed finally. CI needs libudev-dev probably because the container can't access certain linux paths. Makes sense.

ping @eli-schwartz

eli-schwartz:
What's the rationale for the def file that is checked into git here? Seems odd to provide our own, especially since upstream has one at https://github.com/libusb/libusb/blob/master/libusb/libusb-1.0.def

I guess it's due to the mismatched dll filename? `name_prefix: 'lib'` is the default everywhere other than Windows due to MSVC conventions, but libusb doesn't follow the MSVC conventions, so we can just use the kwarg and force the point.

Also, since the library name has "1.0" in it, they don't embed the soname (which is "0") into the DLL.

So, I think you can use the canonical def file, and just add

```meson
name_prefix: 'lib',
soname: cc.get_argument_syntax() == 'msvc' ? '' : '0'
```

neheb:
doesn't work

```
 libusb| WARNING: Library target 'usb-1.0' has 'name_prefix' set. Compilers may not find it from its '-lusb-1.0' linker flag in the 'libusb-1.0.pc' pkg-config file.
```

neheb:
I should also note I'm trying to upstream this: https://github.com/libusb/libusb/pull/1162

neheb:
Same error:

```
1/1 libusb_test FAIL            0.01s   (exit status 3221225781 or signal 3221225653 SIGinvalid)
```

The def file is needed. Or some meson way to remove the first line...

eli-schwartz:
That is kind of a weird warning, it comes solely from pkgconfig.generate.

```meson
library('foo', name_prefix: 'lib')

library('libfoo', name_prefix: '')
```

I'd expect either one to have roughly the same effect. I suggested the former instinctively. The pkgconfig module special cases the latter and makes it skip the warning. I guess just use that. It's exactly what upstream libusb is trying to do after all. :)

neheb:
as discussed on IRC, can't do this warning free without the def file.

eli-schwartz:
Like I explained on IRC and in the comment above, you can. :)

```diff
diff --git a/subprojects/packagefiles/libusb/meson.build b/subprojects/packagefiles/libusb/meson.build
index adf2da2..dba9f09 100644
--- a/subprojects/packagefiles/libusb/meson.build
+++ b/subprojects/packagefiles/libusb/meson.build
@@ -91,13 +91,13 @@ cfile = configure_file(
   configuration: cdata,
 )
 
-libusb = library('usb-1.0',
+libusb = library('libusb-1.0',
   sources,
   dependencies: [ thread_dep, macos_dep, udev_dep ],
   include_directories: incdirs,
   version: '0.3.0',
   soversion: cc.get_argument_syntax() == 'msvc' ? '' : '0',
-  name_prefix: 'lib',
+  name_prefix: '',
   gnu_symbol_visibility: 'hidden',
   vs_module_defs: 'libusb/libusb-1.0.def',
   install: true,
```

neheb:
Alright then. pkgconfig files look similar still.

dcbaker:
Pass [] for the default value to name_prefix

eli-schwartz:
It seems to work quite well as-is though?

dcbaker:
I was just saying that if you want a default value [] will give it to you without the override warning

PR #474, opened 2022-06-14
icu: fix compilation under MSYS2

neheb:
Wrong argument syntax for these options.


subprojects/packagefiles/icu/meson.build
```diff
@@ -3,14 +3,16 @@ project('icu', 'c', 'cpp',
   meson_version: '>=0.55.0',
   default_options: 'cpp_std=c++17')
 
-if host_machine.system() == 'windows'
+cpp = meson.get_compiler('cpp')
+
+if cpp.get_argument_syntax() == 'msvc'
```
jpakkane, line 8: This seems a bit suspicious since it gates defines that are clearly based on the platform on the compiler syntax. Maybe it would be better to keep the `-D` flags as is and move `/utf-8` to its own `if` branch that checks the compiler.

PR #476, opened 2022-06-14
sdl2: link main with only msvc

neheb:
Fixes compilation under MSYS2.


neheb:
ping @Jan200101

PR #480, opened 2022-06-15
pcre2: update to 10.40

neheb:


neheb:
Adorable error on MSYS2...

```
Test 3: Locale-specific features (using 'french' locale)
** Locale test did not run successfully. The output did not match
   test3output, test3outputA or test3outputB.
   This may mean that there is a problem with the locale settings rather
   than a bug in PCRE2.
```

neheb:
fails with the regular MSYS2 pcre2 package. I'll just ignore.

There's also a RunGrepTest. Wonder if that should be added...

neheb:
Amazing..
sshot-17
.

neheb:
@jpakkane ready

ci_config.json
```diff
@@ -254,6 +254,12 @@
       "libboost-iostreams-dev"
     ]
   },
+  "pcre2": {
+    "_comment": "Tests are broken on Windows",
+    "build_options": [
+      "pcre2:test=false"
```
eli-schwartz, line 260: NACK

eli-schwartz (changes requested):

ci_config.json
eli-schwartz: > There's also a RunGrepTest. Wonder if that should be added...

You added another test, the test that was already there works, but the one you added fails on Windows. The solution to this isn't to then disable all tests for all platforms.

Simply mark this one as `should_fail: true` on that platform.

neheb: I think this should do it.

subprojects/packagefiles/pcre2/meson.build
```diff
@@ -160,13 +160,25 @@ if get_option('test') and not meson.is_cross_build() # wine wrappers are not bat
...
   test('RunTest', runtest,
+    should_fail: host_machine.system() == 'windows' and c_compiler.get_argument_syntax() != 'msvc',
+    env : ['srcdir=@0@'.format(meson.current_source_dir())],
+    workdir: meson.current_build_dir())
+
+  if c_compiler.get_argument_syntax() == 'msvc'
+    rungreptest = find_program('RunGrepTest.bat')
+  else
+    rungreptest = find_program('RunGrepTest')
+  endif
```
eli-schwartz, line 178: TBH I'd put that in the same if/else that finds `runtest`.

ci_config.json
eli-schwartz: Yep, thanks. :)

subprojects/packagefiles/pcre2/meson.build
neheb: you mean

```
if get_option('test') and not meson.is_cross_build()
```

?

eli-schwartz:
```meson
  if c_compiler.get_argument_syntax() == 'msvc'
    runtest = find_program('RunTest.bat')
  else
    runtest = find_program('RunTest')
  endif


  if c_compiler.get_argument_syntax() == 'msvc'
    rungreptest = find_program('RunGrepTest.bat')
  else
    rungreptest = find_program('RunGrepTest')
  endif
```

Why not just
```meson
  if c_compiler.get_argument_syntax() == 'msvc'
    runtest = find_program('RunTest.bat')
    rungreptest = find_program('RunGrepTest.bat')
  else
    runtest = find_program('RunTest')
    rungreptest = find_program('RunGrepTest')
  endif
```

neheb:
Anyway, I reported the test failures upstream.

eli-schwartz:
Great. For the sake of my future self, that seems to be https://github.com/PCRE2Project/pcre2/issues/127

PR #481, opened 2022-06-15
libkqueue: various cleanups

neheb:
Use cmake configure file directly instead of duplicating it.

Simplify and fix various configure checks.

Partially fixes MSYS2 compilation.


neheb:
@Jan200101

subprojects/packagefiles/libkqueue/meson.build
```diff
@@ -138,8 +124,9 @@ else
 endif
 
 configure_file(
-    input : 'config.h.meson',
+    input : 'config.h.in',
     output : 'config.h',
+    format : 'cmake',
```
Jan200101, line 129: I believe there had been a reason why I opted to chose to copy from upstream instead of using the cmake format, might have been that meson does not play well with `cmakedefine01` ?

I'll check on this as soon as I find time.

neheb:
Removed version.h.meson too. There's no point to this either.

eli-schwartz:
> Partially fixes MSYS2 compilation.

What's the precise fix here, as a matter of curiosity?

neheb:
The code has this section

```
#if HAVE_SYS_TYPES_H != 1
typedef int pid_t;
#endif
```

this should be ifdef but since it's not, it should be defined as #define HAVE_SYS_TYPES_H 0/1. With set(), it is not so.

neheb:
Actually the issue is different...

https://github.com/mheily/libkqueue/pull/140

subprojects/packagefiles/libkqueue/meson.build
Jan200101: I found no problem and the issue I've had in the past was issues with getting meson to substitute the `@VAR@`s within version.h

The only issue I did encounter was a problem when compiling with MinGW, but you've already fixed that one https://github.com/mheily/libkqueue/commit/f475d065c9a6a99f8a6c38e53abe0acebe184735

neheb:
@jpakkane this is ready.

PR #483, opened 2022-06-15
Fix / General fallback on find_library() for static/shared mismatch

markmaker:
# Description

Add a general fallback on `compiler.find_library()` when the wanted `static` preference is not found. 

The `"ERROR: C ... library 'ws2_32' not found"` does not only happen for MSVC compilers, but also when cross-compiling with MinGW.

The `compiler.get_id()` method does not work for MinGW, because it reports as generic 'gcc'. No other `compiler` property was found to make a clear distinction between Windows/Non-Windows. 

# Justification

Successfully cross-compile OpenSSL for Windows using MinGW.

# Caveats

The fallback now also happens on compilers other than MSVC. If strict static/shared preference enforcement is needed, another way must be found. I would need guidance to make it so.

### DIFF
--- a/subprojects/packagefiles/openssl/meson.build
+++ b/subprojects/packagefiles/openssl/meson.build
@@ -306,19 +306,16 @@ else
 endif
 
 foreach library: openssl_libraries
-  # MSVC fails with "ERROR: C static library 'ws2_32' not found" when `static` is specified at all
-  if compiler.get_id() == 'msvc'
-    dependencies += [
-      compiler.find_library(library),
-    ]
-  else
-    dependencies += [
-      compiler.find_library(
+  single_dep = compiler.find_library(
         library,
         static: get_option('default_library') == 'static',
-      ),
-    ]
+        required : false,
+      )
+  if not single_dep.found()
+    # MSVC and mingw fail with "ERROR: C ... library 'ws2_32' not found" when `static` is specified at all
+    single_dep = compiler.find_library(library)
   endif
+  dependencies += [ single_dep ]
 endforeach
 
 # We may need to add some defines for static builds


markmaker:
Sorry, I don't understand why checks fail:

`AssertionError: '71.1-1' not found in ['70.1-2', '70.1-1', '67.1-4', '67.1-3', '67.1-2', '67.1-1', '55.2-1'] : for icu`

neheb:
you need to rebase against latest master.

eli-schwartz:
That is a merge commit instead of a `git rebase origin/master`. It needs to be a rebase in order to be accepted here.

nazar-pc (commented):
I think strict static/shared separation is indeed needed, it is better for things to fail to compile and tell that result isn't a static library/executable, rather than silently using dynamic linking and failing at runtime when libraries are not found.

I think we can just have this behavior on Windows more generally: update check for MSVC compiler to check for Windows OS and I think it'll be what you want? It doesn't seem like fully static executables are generally a thing on Windows anyway.

eli-schwartz:
```
static: get_option('default_library') == 'static',
```

I think that this new feature from the unreleased changelog of Meson in git master, is what we actually want: https://mesonbuild.com/Release-notes-for-0-63-0.html#new-prefer_static-builtin-option

Keying off of whether openssl itself is being built statically is kind of a hack. ;) Maybe we can just drop the logic altogether?

nazar-pc:
Hm... I'm not so sure. The idea there was to include dependencies statically if library is built statically. On Windows though dependencies are neither static nor dynamic, so `static` property needs to be omitted for things to work at all (when referring to those system libraries at least).

eli-schwartz:
> The idea there was to include dependencies statically if library is built statically.

But this is wrong, because users might not want to do that. They might want to statically link the library, but dynamically link dependencies. Or they might want to build a "fat" openssl shared library. The new Meson option makes that able to be controlled separately.

nazar-pc:
Then I don't see how "preference" helps, there should be a way to have precise control and let things crash instead of producing unexpected results.

markmaker:
> That is a merge commit instead of a `git rebase origin/master`. It needs to be a rebase in order to be accepted here.

Like this?

xclaesse:
`static: get_option('default_library') == 'static'` is definitely wrong, one tried to make people stop doing that and `prefer_static` is an imperfect way of achieving that. Also it makes it inconsistent with all other meson projects that won't be doing that, so you might easily static link and dynamic link the same library at the same time.

neheb:
this is for openssl. Commit title should have that prefix.

kasper93:
I've actually reported and attempted to fix it locally before finding this PR. 

My solution, which should follow `prefer_static` well.
``` meson
foreach library: openssl_libraries
  lib = compiler.find_library(library,
                              static: get_option('prefer_static'),
                              required: not get_option('prefer_static'))
  if not lib.found() and get_option('prefer_static')
    lib = compiler.find_library(library)
  endif
  dependencies += [ lib ]
endforeach
```

But there seems to be meson bug, because setting `static` value to either fails or true fails detection. Looks like there is some caching issue where `static == None` is not the same as `static == false`? Strange...

For this code I get
``` meson
lib = compiler.find_library('ws2_32', required: false)
message(lib.found())
lib = compiler.find_library('ws2_32', static: false, required: false)
message(lib.found())
lib = compiler.find_library('ws2_32', static: true, required: false)
message(lib.found())
```

```
openssl| Library ws2_32 found: YES
openssl| Message: true
openssl| Library ws2_32 found: NO
openssl| Message: false
openssl| Library ws2_32 found: NO
openssl| Message: false
```

neheb:
I see no issue there. Meson decides to display the same booleans as YES/NO.

kasper93:
The issue is that library detection doesn't work at all if static keyword is used.

neheb:
Ah right. I remember a discussion on IRC about something similar.

qiracy:
Anyone figured out a fix?

PR #485, opened 2022-06-16
sdl2_mixer: get rid of sdl2 dependency hack

neheb:
In combination with the other sdl2 PR, fixes compilation under MSYS2.

ping @Jan200101


Jan200101:
:+1:
Works fine on Linux and MinGW

PR #500, opened 2022-06-25
fix ogg and vorbis under MinGW

neheb:
def files need changing.


jpakkane:
Hmmmm. I'm not actually sure if the update scripts can handle two commits that update two packages in the same merge request...

neheb:
Logs say both got built. Note that the latter depends on the former.

jpakkane:
Let\s live on the edge and see what happens.

PR #508, opened 2022-06-26
unit-system: update to 0.5.3

neheb:

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -16,7 +16,6 @@
     "libgpiod",
     "mocklibc",
     "termbox",
-    "unit-system",
     "zstd",
     "ludocode-mpack",
     "m4",
@@ -30,7 +29,6 @@
     "mocklibc",
     "rtaudio",
     "tinyxml2",
-    "unit-system",
     "zstd"
   ],
   "cexception": {
--- a/releases.json
+++ b/releases.json
@@ -2240,12 +2240,10 @@
   },
   "unit-system": {
     "dependency_names": [
-      "unit-system-full",
-      "unit-system-generic",
-      "unit-system-base",
-      "unit-system-common"
+      "unit-system"
     ],
     "versions": [
+      "0.6.1-1",
       "0.5.2-1",
       "0.5.0-1"
     ]
--- a/subprojects/unit-system.wrap
+++ b/subprojects/unit-system.wrap
@@ -1,9 +1,8 @@
 [wrap-file]
-directory = unit-system-0.5.2
-
-source_url = https://github.com/noah1510/unit-system/archive/refs/tags/v0.5.2.tar.gz
-source_filename = unit-system-0.5.2.tar.gz
-source_hash = 77758587bd1da89994cbc38c3c5f5513de77ff1c7df7334e7ab88db9854ee316
+directory = unit-system-0.6.1
+source_url = https://github.com/noah1510/unit-system/archive/refs/tags/v0.6.1.tar.gz
+source_filename = unit-system-0.6.1.tar.gz
+source_hash = d863c02bc2c99daf710e857ff21fc0ed915dab37af26f9b91a04d32a21e2768d
 
 [provide]
-dependency_names = unit-system-full, unit-system-generic, unit-system-base, unit-system-common
+unit-system = unit_system_dep


neheb:
I could probably fix this with a def fine. No idea how C++ functions are named though...

jpakkane:
In Debian at least it is possible to give C++ function names in unmangled form. If VS does not support that then the only reasonable way to make it work is with visibility attributes.

neheb:
a git grep of build_target indicates that this is just avoided by compiling static on MSVC.

neheb:
ping @noah1510 since you're active here, can you add dllexport to this package?

noah1510:
I copied the dllexport include from the imgui package into the latest commit. This should get it working right? If so I will create a new patch release.

neheb:
```
 LINK : warning LNK4001: no object files specified; libraries used
LINK : error LNK2001: unresolved external symbol _DllMainCRTStartup
subprojects\unit-system-0.5.3\generic\unit-system-generic-0.dll : fatal error LNK1120: 1 unresolved externals
```

No idea what's going on.

jpakkane:
The upstream build definition specifies a library file that has no sources, only headers. If this is a header-only library then it should not try to build a library, the `declare_dependency` only needs an `include_directories` in that case.

noah1510:
@jpakkane Thanks for the advice. Now the static compile works on windows.

Is there a way to automatically generate the def files? As long as there is none and I would have to add export macros to all of the functions of the headers, I will **not** support msvc if that is the case.

xclaesse:
Why is this adding a meson.build file? Upstream has one: https://github.com/noah1510/unit-system/blob/main/meson.build

neheb:
Intermediate step to trying to figure out MSVC compilation.

neheb:
> @jpakkane Thanks for the advice. Now the static compile works on windows.
> 
> Is there a way to automatically generate the def files? As long as there is none and I would have to add export macros to all of the functions of the headers, I will **not** support msvc if that is the case.

You can create a python script to do this. See https://github.com/libsndfile/libsndfile/blob/master/src/create_symbols_file.py

noah1510:
That would in theory be an option.
However for this library its kind of overkill since every symbol need to be exported.
To not miss a symbol I would automatically generate the symbol list from the linux so files and pray the list does not break anything.

Does the order of the symbols matter or is it fine if they are in whatever order they are in a so file by default?

noah1510:
I just uploaded a first set of def files to the git repo. Since I have no windows machine to test on right now, I can not integrate them into the build system files yet. Tomorrow I should get them working if everything goes according to plan.

noah1510:
I added the def files for the core modules. Now the shared compile works with msvc as long as the tests are disabled.

neheb:
```
2022-09-01T21:23:19.1822355Z [16/20] "link"  /MACHINE:x64 /OUT:subprojects/unit-system-0.5.3/generic/unit-system-generic-0.dll  "/nologo" "/release" "/nologo" "/DEBUG" "/PDB:subprojects\unit-system-0.5.3\generic\unit-system-generic-0.pdb" "/DLL" "/IMPLIB:subprojects\unit-system-0.5.3\generic\unit-system-generic.lib" "/MAP" "/MAPINFO:EXPORTS" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
2022-09-01T21:23:19.1824985Z FAILED: subprojects/unit-system-0.5.3/generic/unit-system-generic-0.dll subprojects/unit-system-0.5.3/generic/unit-system-generic-0.pdb 
2022-09-01T21:23:19.1827163Z "link"  /MACHINE:x64 /OUT:subprojects/unit-system-0.5.3/generic/unit-system-generic-0.dll  "/nologo" "/release" "/nologo" "/DEBUG" "/PDB:subprojects\unit-system-0.5.3\generic\unit-system-generic-0.pdb" "/DLL" "/IMPLIB:subprojects\unit-system-0.5.3\generic\unit-system-generic.lib" "/MAP" "/MAPINFO:EXPORTS" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
2022-09-01T21:23:19.1829089Z LINK : warning LNK4001: no object files specified; libraries used
2022-09-01T21:23:19.1829669Z LINK : error LNK2001: unresolved external symbol _DllMainCRTStartup
2022-09-01T21:23:19.1830476Z subprojects\unit-system-0.5.3\generic\unit-system-generic-0.dll : fatal error LNK1120: 1 unresolved externals
2022-09-01T21:23:19.1831194Z ninja: build stopped: subcommand failed.
```

@noah1510

neheb:
Never mind. Builds after adding another file. Waiting for new release so I can get rid of all of these files.

PR #509, opened 2022-06-26
lmdb: get rid of build_target

neheb:
Fix MSVC properly by providing a def file.


neheb:
ping @jpakkane this is ready.

jpakkane:
Anyone with review rights could do this, really. It would even be better to have more people on to spread the workload.

PR #510, opened 2022-06-26
liburing: update to 2.2

fischerling:
Simplify the test definitions and bump the required meson version to
use the filesystem module.

Reflect the upstream changes to the custom ./configure script in the
root meson build definition.

Signed-off-by: Florian Fischer <florian.fischer@muhq.space>


eli-schwartz:
Hmm, so you're just unconditionally linking to `-pthread` whether the test needs it or not?

fischerling:
> Hmm, so you're just unconditionally linking to `-pthread` whether the test needs it or not?

Yes this is what the upstream Makefile does to.
it was changed in https://github.com/axboe/liburing/commit/664bf782a6d78f701301ef6ebdd7cf63b2e3ee09

eli-schwartz (commented):
- https://github.com/axboe/liburing/commit/de214792d38700ba470a560cf3729a9bd62acb35 -- removes fomit-frame-pointer
- https://github.com/axboe/liburing/commit/2de98320d5b02951936fc0ab677dd01e4fb2a7a7 -- adds -fno-stack-protector

eli-schwartz:
While comparing the diff upstream, I realized that this wrap has never set the `version: ...` kwarg for the library. It's supposed to be the same as the project version (in this case, `'2.2'` with the soname auto-derived from the first component).

subprojects/packagefiles/liburing/test/meson.build
```diff
@@ -127,28 +151,30 @@ if has_array_bounds
...
-               dependencies : t[2],
+               dependencies : test_dependencies,
                install : true,
                install_dir : get_option('datadir') / 'liburing-test')
 
-    test(t[0],
+    test(test_name,
          runtests_sh,
-         args : t[0],
+         args : test_name,
```
eli-schwartz, line 171: The test harness will try to detect issues logged in dmesg, but relies on not having intermixed logs in that case. So I think this should probably have used `is_parallel: false` all along?

subprojects/packagefiles/liburing/meson.build
```diff
@@ -74,13 +88,19 @@ has_cxx = cpp.compiles(code, name: 'C++')
 has_ucontext = (cc.has_type('ucontext_t', prefix: '#include <ucontext.h>')
   and cc.has_function('makecontext', prefix: '#include <ucontext.h>'))
 
+has_memfd_create = cc.has_function('memfd_create',
+                                   args: ['-D_GNU_SOURCE'],
+                                   prefix: '#include <sys/mman.h>')
```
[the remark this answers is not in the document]
fischerling: But this commit is after the liburing-2.2 tag. Should we include the changes in our 2.2 wrap?

fischerling:
> While comparing the diff upstream, I realized that this wrap has never set the `version: ...` kwarg for the library. It's supposed to be the same as the project version (in this case, `'2.2'` with the soname auto-derived from the first component).

Thanks. The wrap now sets the library version and soversion derived from the project version.

I kept the changes relating the review in a separate commit which I would squash when the PR is ready.

subprojects/packagefiles/liburing/src/meson.build
```diff
@@ -11,6 +11,8 @@ liburing = library('uring',
                    c_args: '-DLIBURING_INTERNAL',
                    link_args: '-Wl,--version-script=' + meson.current_source_dir() + '/liburing.map',
                    link_depends: 'liburing.map',
+                   version: meson.project_version(),
+                   soversion: meson.project_version().split('.')[0],
```
eli-schwartz, line 15: Per the soversion documentation:
> If this is not specified, the first part of `version` is used instead (see below). For example, if `version` is `3.6.0` and `soversion` is not defined, it is set to `3`.

subprojects/packagefiles/liburing/test/meson.build
eli-schwartz: Since there are multiple test suites which I guess shouldn't all be run at the same time by a default `meson test`, perhaps we should use `add_test_setup()` to define a test profile which only runs the "once" tests, and declare it as the default?

eli-schwartz (changes requested):

subprojects/packagefiles/liburing/src/meson.build
fischerling: So you suggest dropping the `soversion` key word argument since we are reimplementing the default behavior?

eli-schwartz: Exactly.

subprojects/packagefiles/liburing/meson.build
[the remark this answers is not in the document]
eli-schwartz: Well, probably not. The newly added code is there in the release anyway, the define is needed to prevent conflicting definitions, but the newly added code was immediately dropped post-release, meh. :p

fischerling:
> Alright, that looks pretty good. Time to squash the fixup changes I think.
> 
> I'm thinking organization-wise it might make sense to have two commits:
> 
> * one for the cleanups
> * then one for the version bump and associated config.h / test list updates
> 
> What do you think?

Sounds good to me.
I have squashed and swapped the cleanup changes with the actual update.

PR #512, opened 2022-07-01
libusb: fix compilation with MinGW

neheb:
Upstream has a bug with this release where if clock_gettime is found,
compilation fails. Missing ifdef in the code. Fixed in latest master.

Also fixed a soversion issue with GNU ld + MinGW. Seems the MSVC
condition is not quite correct.


neheb:
For completeness, the upstream commit: https://github.com/libusb/libusb/commit/5691f303ad272a2514f68a96d6c18aeff62eb33d

PR #516, opened 2022-07-03
actions: add MSYS2 CI

neheb:

Probably suboptimal.

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -21,6 +21,34 @@ jobs:
         run: |
           ./tools/sanity_checks.py
 
+  MSYS2:
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    strategy:
+      matrix:
+        sys: [ MINGW64, MINGW32, UCRT64, CLANG32, CLANG64 ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: msys2/setup-msys2@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/sanity_checks.yml
```diff
@@ -21,6 +21,33 @@ jobs:
...
+        shell: msys2 {0}
+    strategy:
+      matrix:
+        sys: [ MINGW64, MINGW32, UCRT64, CLANG32, CLANG64 ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{matrix.sys}}
+          install: git base-devel
```
neheb, line 37: base-devel is interesting. It installs some autotools stuff but no cmake. Maybe should get rid of it.

.github/workflows/sanity_checks.yml
```diff
@@ -21,6 +21,33 @@ jobs:
...
           ./tools/sanity_checks.py
 
+  MSYS2:
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    strategy:
+      matrix:
+        sys: [ MINGW64, MINGW32, UCRT64, CLANG32, CLANG64 ]
```
neheb, line 31: MSYS is missing here. MSYS lacks many packages. Upstream does not want MSYS to be some alternative to cygwin. Maybe better to keep it out. Not sure.

neheb:
Let's not.

PR #517, opened 2022-07-04
mpdecimal: fix compilation on ARM

neheb:
ARM 32-bit does not have -m32 on GCC. It's only needed on 64-bit
platforms anyway.


subprojects/packagefiles/mpdecimal/meson.build
```diff
@@ -3,7 +3,7 @@ project('mpdecimal', 'c', 'cpp',
         license: 'BSD-2-Clause',
         default_options: [
           'c_std=c99',
-          'cpp_std=c++11',
+          'cpp_std=c++14',
```
dnicolodi, line 6: This hunk seems unrelated, or is c++14 required on ARM for some reason?

neheb: https://github.com/mesonbuild/wrapdb/actions/runs/2607518219

Needed for MSVC.

dnicolodi: I am sure that the CI compiled mpdecimal with `cpp_std=c++11` before. Did something change in Meson to make it fail when this setting is used with MSVC?

neheb: Yes.

subprojects _std settings recently started being respected. Before, they would inherit the _std settings of the host project which in this case happens to be c++17: https://github.com/mesonbuild/wrapdb/blob/master/meson.build

neheb:
This is ready to go.

subprojects/packagefiles/mpdecimal/meson.build
dcbaker: That warning should be non-fatal, this is a Meson bug

neheb:
ping @jpakkane

subprojects/packagefiles/mpdecimal/meson.build
skrah: The way in which the `mpdecimal` build machinery deals with this is to set `--std=c++11` for everything except `MSVC`.  `MSVC` automatically selects the lowest supported C++ version (currently `C++14`) if no standard is given.

`meson` could just omit the standard for `MSVC` if the selected standard is `<= C++14`.

PR #518, opened 2022-07-04
fmt: add override_dependency

neheb:
Allows spdlog to succeed in finding fmtlib. Useful when both are
included as subprojects.


subprojects/packagefiles/fmt/meson.build
```diff
@@ -27,6 +27,7 @@ fmt_dep = declare_dependency(
     compile_args : fmt_interface_cpp_args,
     link_with : fmt_lib
 )
+meson.override_dependency('fmt', fmt_dep)
```
eli-schwartz, line 30: This should be guarded with `if meson.version().version_compare('>=0.54.0')` to make sure it's only evaluated on versions of Meson that are new enough to have this method.

neheb: Fixed.

PR #521, opened 2022-07-05
libglvnd: add

neheb:
Upstream has a meson build. No need to write one.

tests require an X11 server. Can't figure out how to disable them. Hmmm...

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -216,6 +216,17 @@
       "windows": false
     }
   },
+  "libglvnd": {
+    "_comment": "Upstream explicitly only supports Unix-like OSes excluding macOS",
+    "build_on": {
+      "darwin": false,
+      "windows": false
+    },
+    "_comment": "GLX tests require an X server, which CI does not provide.",
+    "build_options": [
+      "libglvnd:glx=disabled"
+    ]
+  },
   "libjpeg-turbo": {
     "brew_packages": [
       "nasm"
--- a/releases.json
+++ b/releases.json
@@ -1048,6 +1048,17 @@
       "2.2.2-1"
     ]
   },
+  "libglvnd": {
+    "dependency_names": [
+      "egl",
+      "opengl",
+      "glesv1",
+      "glesv2"
+    ],
+    "versions": [
+      "1.6.0-1"
+    ]
+  },
   "libgpiod": {
     "versions": [
       "1.6.3-1"
--- a/subprojects/libglvnd.wrap
+++ b/subprojects/libglvnd.wrap
@@ -1,0 +1,11 @@
+[wrap-file]
+directory = libglvnd-v1.6.0
+source_url = https://gitlab.freedesktop.org/glvnd/libglvnd/-/archive/v1.6.0/libglvnd-v1.6.0.tar.bz2
+source_filename = libglvnd-1.6.0.tar.bz2
+source_hash = 2b22a72f9f35fe9ba630c2971f6cee6eefc0c8317e0eecc3c76134ed5f00bd8a
+
+[provide]
+egl = dep_egl
+opengl = dep_opengl
+glesv1 = dep_glesv1
+glesv2 = dep_glesv2


dcbaker:
MacOS and Windows are explicitly not supported (I wrote the upstream meson build files). the Right™ way to fix the tests is to have them check for an X connection and return 127 if they can't get one, but I'd review changes to add a `-Dtests` option or similar :)

releases.json
```diff
@@ -835,6 +835,17 @@
       "2.2.2-1"
     ]
   },
+  "libglvnd": {
+    "dependency_names": [
+      "gl",
+      "opengl",
+      "glesv1",
+      "glesv2"
```
dcbaker, line 843: You're missing `GLX` and `EGL` as well

neheb: Not following here:

```
mangix@ryzen ~/d/w/s/libglvnd-v1.4.0 (gl)> rg declare_de
src/util/meson.build
33:idep_glvnd_pthread = declare_dependency(
45:idep_app_error_check = declare_dependency(
56:idep_utils_misc = declare_dependency(
67:idep_trace = declare_dependency(
81:idep_winsys_dispatch = declare_dependency(
92:idep_cjson = declare_dependency(

src/GLdispatch/meson.build
50:idep_gldispatch = declare_dependency(

src/GLX/meson.build
55:idep_glx = declare_dependency(

src/GLdispatch/vnd-glapi/meson.build
93:  _dep = declare_dependency(
```

the last one is a loop listing ['gl', 'opengl', 'glesv1', 'glesv2']

dcbaker: Ooh, that's interesting, `idep_glx` should be publc, and we should have a dependency for `EGL` too (we as in libpkgconf)

dcbaker: And, those `idep_*` are short for `internaldep` those aren't for public consumption. I'll send a patch to fix upstream to be consumable

dcbaker:
https://gitlab.freedesktop.org/glvnd/libglvnd/-/merge_requests/269 Is needed to make this really possible

dcbaker:
My MR landed, so the *git* version of glvnd is wrappable, but the last released version isn't :/

neheb:
```
meson.build:12:6: ERROR: Requested variable "dep_gl" not found.
```

rofl.

neheb:
```
ld: unknown option: --version-script
```

Who to blame?

neheb:
I guess this is ready.

```
'-Wl,--version-script'
```

does not work with lld. Hence Darwin is not built.

releases.json
```diff
@@ -865,6 +865,17 @@
       "2.2.2-1"
     ]
   },
+  "libglvnd": {
+    "dependency_names": [
+      "egl",
+      "opengl",
+      "glesv1",
+      "glesv2"
```
dcbaker, line 873:
```suggestion
      "glesv2",
      "glx"
```

subprojects/libglvnd.wrap
```diff
@@ -0,0 +1,11 @@
...
+directory = libglvnd-v1.5.0
+source_url = https://gitlab.freedesktop.org/glvnd/libglvnd/-/archive/v1.5.0/libglvnd-v1.5.0.tar.bz2
+source_filename = libglvnd-1.5.0.tar.bz2
+source_hash = 148bf421fef62a3c2bb0a7264292fc76743b2bbcebe9995be4b63fe5f15c471f
+
+[provide]
+egl = dep_egl
+opengl = dep_opengl
+glesv1 = dep_glesv1
+glesv2 = dep_glesv2
```
dcbaker, line 11:
```suggestion
glesv2 = dep_glesv2
glx = dep_glx
```

dcbaker (changes requested):

ci_config.json
```diff
@@ -199,6 +199,17 @@
       "windows": false
     }
   },
+  "libglvnd": {
+    "_comment": "Windows needs pthreads. macOS/LLD does not support --version-script",
```
dcbaker, line 203:
```suggestion
    "_comment": "upstream explicitly only supports Unix-like OSes excluding macOS",
```

ci_config.json
```diff
@@ -199,6 +199,17 @@
...
+    "build_on": {
+      "darwin": false,
+      "windows": false
+    },
+    "_comment": "GLX tests require an X server, which CI does not provide.",
+    "build_options": [
+      "libglvnd:glx=disabled"
+    ]
```
dcbaker, line 211: glx is a pretty import feature of glvnd, I think we should install `xorg-dev` (there may be a few more dev files needed, but xorg-dev should be most of it) for build tests.

neheb: The tests require an X server running though.

dcbaker: Looking back through the comments it looks like the bigger problem is if the x server isn't running the tests fail... maybe update the comment to reflect that instead

eli-schwartz:
The version script argument would probably need special handling for Darwin, because they use a different format for the version script and also a different flag.

If upstream doesn't support that anyway then no big deal, we don't need to either. ;) But I agree with @dcbaker this is a matter of "what platforms upstream supports" rather than "these platforms are currently broken in CI due to missing XXX".

ci_config.json
neheb: Btw that linker failure can probably be done on Linux as well by using CC_LD=lld

ci_config.json
eli-schwartz: We can run tests with xvfb-run, maybe? Probably needs updates to sanity_checks.py though.

neheb: Upstream issues are beyond the scope of this PR.

eli-schwartz: How is that an upstream issue? The idea is to perform tests with `xvfb-run meson test -C builddir/`.

neheb: hrm? how would that be achieved?

neheb:
Updated CI comment.

Tried using lld on Linux. I can't get it to fail.

dcbaker:
more importantly, we *can't* support macOS or Windows. They have their own dispatch layers built into the OS that serve the same purpose as libglvnd :)

ci_config.json
dcbaker: macOS doesn't use lld by default though, it uses ld64, which netoriously doesn't support version scripts

I just tested on linux with gcc + lld and clang + lld and it works fine.

eli-schwartz: Right, this is absolutely a Darwin issue not an LLD one.

lld is supposed to be fairly compatible with bfd so it would be rather awkward if it was missing support for one of the most significant features used in practice... I'm hardly surprised they support it. :P

ci_config.json
eli-schwartz: That's why I said:
> Probably needs updates to sanity_checks.py though

neheb:
anything missing?

dcbaker:
It doesn't look like the glx dependency has been added

neheb:
It can't be as x is disabled.

dcbaker:
Putting glx in the [provides] would cause CI to fail?

neheb:
Yes

neheb:
@Jan200101 This doesn't seem to work right with sdl2:

```
subprojects/SDL2-2.24.0/include/SDL_opengles2.h:35:10: fatal error: 'GLES2/gl2platform.h' file not found
#include <GLES2/gl2platform.h>
```

Jan200101:
> @Jan200101 This doesn't seem to work right with sdl2:

Worked fine on my end using Fedora 36
```
$ meson build -Dwraps=libglvnd,sdl2
sdl2| Dependency glesv2 found: YES 1.5.0 (overridden)
```
(building also works without a hitch, but there is no worthwhile output to show from that)

Could you give more details on how to reproduce this?

neheb:
Anything else missing? Various PRs are dependent on this one being merged.

dcbaker:
We need to have glx in the provides section. it's not some minor little side thing, it's probably the second most used dependency in glvnd (after gl or opengl which are basically the same thing), if the CI can't handle that then we need to fix the CI

eli-schwartz:
I'm going to reiterate my previous point, that the only reason the CI doesn't enable glx (and thus provide it) is because "the tests need a running xorg server"... which is trivial to fix, simply run `xvfb-run meson test -C builddir`.

There doesn't seem to be any good reason not to test it afaict?

neheb:
Sounds like this PR can't proceed.

neheb:
> We need to have glx in the provides section. it's not some minor little side thing, it's probably the second most used dependency in glvnd (after gl or opengl which are basically the same thing), if the CI can't handle that then we need to fix the CI

My NAS also cannot handle it as it has no video outputs. It's unrealistic to expect an X server running.

eli-schwartz:
What does that have to do with whether it should be built in CI?

neheb:
How about merging as is and then someone updating this to add gl_dep and fix CI?

neheb:
@xclaesse added your multi dep name suggestion.

neheb:
sfml dependency is a lie. It needs glx.

neheb:
```
 glew| Dependency gl found: YES 1.5.0 (overridden)
glew| WARNING: Inconsistency: Subproject has overridden the dependency with another variable than 'dep_opengl'
```

@dcbaker know what this is about?

neheb:
@xclaesse the warning above is due to the multi name dependency. Sounds like a bug.

neheb:
```
 ERROR: test_releases (__main__.TestReleases) (step='check_meson_version', name='libglvnd')
  File "D:\a\wrapdb\wrapdb\tools\sanity_checks.py", line 133, in check_meson_version
    subproject, = [subproj for subproj in project_info["subprojects"] if subproj["name"] == name]
ValueError: not enough values to unpack (expected 1, got 0)
```

I've seen this with some PRs that update multiple packages.

neheb:
@dcbaker seems explicit OS checks are needed to make CI happy.

PR #523, opened 2022-07-07
Add Asio

brendanburkhart:
Simple wrap for standalone/non-Boost version of [Asio](https://github.com/chriskohlhoff/asio/) which is a header-only library.

### DIFF
--- a/subprojects/packagefiles/asio/meson.build
+++ b/subprojects/packagefiles/asio/meson.build
@@ -1,0 +1,19 @@
+project('asio',
+    'cpp',
+    version : '1.22.1',
+    license : 'Boost'
+)
+
+asio_deps = []
+
+# For windows builds, ASIO depends on ws2_32
+if target_machine.system() == 'windows'
+    compiler = meson.get_compiler('cpp')
+    ws2_32_dep = compiler.find_library('ws2_32', required: true)
+    asio_deps += ws2_32_dep
+endif
+
+asio_dep = declare_dependency(
+    include_directories : include_directories('asio/include'),
+    dependencies: asio_deps,
+)
--- a/subprojects/packagefiles/asio/meson.build
+++ b/subprojects/packagefiles/asio/meson.build
@@ -19,3 +19,5 @@ asio_dep = declare_dependency(
     include_directories : include_directories('include'),
     dependencies: asio_deps,
 )
+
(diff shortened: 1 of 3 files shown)


neheb:
You should add the tests.

eli-schwartz:
Also, include/asio/ and include/asio.hpp could be installed I guess...

Seems like -DASIO_STANDALONE is probably a good idea to guarantee when using. I don't know what -DASIO_SEPARATE_COMPILATION does, is there even an implementation file for it or something?

brendanburkhart:
Not sure if I understood what you meant by installed the header files, does what I did look right?

eli-schwartz:
-DASIO_STANDALONE should be a cpp_args on the dependency, so that it gets used any time someone tries to use the asio dependency.

add_project_arguments() only affects targets such as executable() or library() which are defined inside this subproject wrap.

subprojects/packagefiles/asio/meson.build
```diff
@@ -19,3 +19,5 @@ asio_dep = declare_dependency(
     include_directories : include_directories('include'),
     dependencies: asio_deps,
 )
+
+install_headers('include/asio.hpp', 'include/asio/*')
```
eli-schwartz, line 23:
```meson
install_headers('include/asio.hpp')
install_subdir('include/asio', install_dir: get_option('includedir'))
```

This is because install_headers() takes files, not directories. :(

brendanburkhart:
Sorry for all the pings 😅 I'm trying to run tools/sanity_checks.py for Windows & Ubuntu but it just says no work to do - I thought that was just since it's a header-only library but clearly the CI is actually doing something. Is there a better way to test locally?

neheb:
> Sorry for all the pings 😅 I'm trying to run tools/sanity_checks.py for Windows & Ubuntu but it just says no work to do - I thought that was just since it's a header-only library but clearly the CI is actually doing something. Is there a better way to test locally?

add tests and examples. that's how it can be tested.

jpakkane:
> meson.build:16:0: ERROR: declare_dependency got unknown keyword arguments "cpp_args"

Declare dependencies uses `compile_args` instead (basically because Pkg-Config does not support per-language compile args and we need to be compatible with that).

jpakkane:
Superseded by the referenced MR.

PR #528, opened 2022-07-10
openal-soft: update to 1.22.2

neheb:
Massive reorganization. Builds on Linux now.


neheb:
This is basically ready.

MSYS2 fails with

```
ld.lld: error: undefined symbol: WinMain
```

No idea why.

PR #531, opened 2022-07-12
libsndfile: fix compilation under Alpine Linux

neheb:
There's no python, only python3.


neheb:
Nonsensical that I have to use project_source_root() here...

subprojects/packagefiles/libsndfile/meson.build
```diff
@@ -104,11 +104,12 @@ sndfile = configure_file(
 
 incdirs = include_directories('.', 'src')
 
+pymod = import('python')
 def = custom_target(
   'def',
   output: 'sndfile.def',
   capture: true,
-  command: [ './src/create_symbols_file.py', 'win32', meson.project_version() ],
+  command: [ pymod.find_installation('python3'), meson.project_source_root() / 'src/create_symbols_file.py', 'win32', meson.project_version() ],
```
eli-schwartz, line 112: `pymod.find_installation('python3')` returns the same thing as `find_program('python3')`, except for the additional methods useful for building CPython extensions and installing python software to site-packages.

Also importing pymod is a mild performance penalty.

eli-schwartz (changes requested):

eli-schwartz:
> Nonsensical that I have to use project_source_root() here...

The second argument in a command isn't the program, so it doesn't automatically go through e.g. find_program, it's passed as a ***string literal*** to the actual program.

There are two solutions here:
- use meson.project_source_root()
- use `files('src/create_symbols_file.py')`

subprojects/packagefiles/libsndfile/meson.build
```diff
@@ -104,11 +104,12 @@ sndfile = configure_file(
 
 incdirs = include_directories('.', 'src')
 
+pymod = import('python')
 def = custom_target(
   'def',
   output: 'sndfile.def',
```
eli-schwartz, line 110: This could in theory just not be built on Linux, idk.

The bigger issue, is that you actually need to run it on Linux too, to create a GNU ld version script. It sets up symbol versioning e.g. `sf_close@@libsndfile.so.1.0`

eli-schwartz (changes requested):

PR #534, opened 2022-07-13
smoke-signal 1.0

jormundgand:


jpakkane:
It is the policy of WrapDB that only official releases can be added, not Git repos. Closing this, feel free to open a new one that uses a release, if one exists.

PR #535, opened 2022-07-14
libssh2: add

ilsme:


subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,196 @@
...
+
+    gcrypt_dep = dependency('gcrypt')
+
+    libssh2_deps += gcrypt_dep
+    extra_flags += ['-DLIBSSH2_GCRYPT']
+else
+    error('No crypto backend specified')
+endif
+
+libssh2_have_unistd_h = cc.has_header('unistd.h')
```
neheb, line 69: You could use a loop for this. example: https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/libsndfile/meson.build <-- 70

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,196 @@
...
+  add_project_arguments(['-DHAVE_ARPA_INET_H'], language: 'c')
+endif
+
+inc_dir = include_directories(['include', '.'])
+libssh2_lib = library('ssh2',
+    libssh2_sources,
+    include_directories: inc_dir,
+    dependencies: libssh2_deps,
+    c_args: extra_flags,
+    install: true
```
eli-schwartz, line 178: Upstream sets the libtool-compatible version `1:1:0` (current:revison:age, or C:R:A).

Translated to what the compiler is invoked with, this is:

(C - A).(A).(R)

`version: '1.0.1',` passed as a kwarg to library(). See also e.g. the comments in the libgpiod or libxmlpp wraps for discussion on how this works.

subprojects/packagefiles/libssh2/example/meson.build
```diff
@@ -0,0 +1,34 @@
...
+    'ssh2_agent.c',
+    'ssh2_echo.c',
+    'ssh2_exec.c',
+    'subsystem_netconf.c',
+    'tcpip-forward.c'
+]
+
+x11_dep = dependency('x11')
+
+if x11_dep.found()
```
eli-schwartz, line 28: Unless you pass `required: false` to the dependency, it will error out when not found anyway.

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,196 @@
...
+install_headers(
+    files('include/libssh2.h',
+          'include/libssh2_publickey.h',
+          'include/libssh2_sftp.h'
+    )
+)
+
+pkgconfig.generate(libssh2_lib)
+
+if get_option('examples').enabled()
```
eli-schwartz, line 194: Why is this a feature option? You only use it to see if it is enabled, so if it is "auto" you treat it like "false".

Instead of a tristate option, why not just use a boolean option?

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,196 @@
...
+)
+
+install_headers(
+    files('include/libssh2.h',
+          'include/libssh2_publickey.h',
+          'include/libssh2_sftp.h'
+    )
+)
+
+pkgconfig.generate(libssh2_lib)
```
eli-schwartz, line 192: This needs a name set. By default it uses the library name "ssh2", but it is supposed to create libssh2.pc instead.
Optionally you can set the url and description fields.

eli-schwartz (changes requested):

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,196 @@
...
+
+    openssl_dep = dependency('openssl', version : '>= 1.0.0')
+    libssh2_deps += openssl_dep
+
+    extra_flags += ['-DLIBSSH2_OPENSSL']
+elif get_option('crypto_backend') == 'embedtls'
+    with_embedtls = true
+    libssh2_sources += ['src/mbedtls.c']
+
+    embedtls_dep = dependency('embedtls')
```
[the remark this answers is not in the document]
neheb: good catch. I asked upstream to implement pkgconfig files. They refused.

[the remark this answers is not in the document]
eli-schwartz: No they did not???

You submitted the most recent of 4 attempts at implementing this. After a review comment in which someone said your PR causes the CI to fail on older versions of cmake, you closed the PR without further discussion. They still have a tracking ticket for implementing it.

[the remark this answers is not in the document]
neheb: Hrm I guess I remember wrong. Still weird that there are no cmake configuration files either.

subprojects/packagefiles/libssh2/meson.build
neheb: There's also the other pattern of setting name_prefix to ''.

subprojects/packagefiles/libssh2/meson.build
ilsme: Done

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,144 @@
...
+    soversion: '1.0.1',
+    version: meson.project_version()
+)
+libssh2_dep = declare_dependency(
+    include_directories: inc_dir,
+    link_with: libssh2_lib
+)
+
+install_headers(
+    files('include/libssh2.h',
```
neheb, line 134: files() is not needed here. install_headers is a variadic function.

subprojects/packagefiles/libssh2/example/meson.build
```diff
@@ -0,0 +1,74 @@
...
+    ['sys/time.h', 'HAVE_SYS_TIME_H'],
+    ['winsock2.h', 'HAVE_WINSOCK2_H'],
+]
+
+foreach h : check_example_headers
+    example_libssh2_config.set(h[1], cc.has_header(h[0]))
+endforeach
+
+check_funcs = [
+    ['strcasecmp', 'HAVE_STRCASECMP'],
```
neheb, line 45: No need to have a 2 dimentional array here.

subprojects/packagefiles/libssh2/example/meson.build
```diff
@@ -0,0 +1,74 @@
...
+    ['_stricmp', 'HAVE__STRICMP'],
+    ['snprintf', 'HAVE_SNPRINTF'],
+    ['_snprintf', 'HAVE__SNPRINTF']
+]
+
+foreach f : check_funcs
+    example_libssh2_config.set(f[1], cc.has_function(f[0]))
```
neheb, line 52: you can do

```
example_libssh2_config.set('HAVE_@0@'.format(f.underscorify().to_upper()), cc.has_function(f))
```

subprojects/packagefiles/libssh2/example/meson.build
```diff
@@ -0,0 +1,74 @@
...
+endforeach
+
+check_funcs = [
+    ['strcasecmp', 'HAVE_STRCASECMP'],
+    ['_stricmp', 'HAVE__STRICMP'],
+    ['snprintf', 'HAVE_SNPRINTF'],
+    ['_snprintf', 'HAVE__SNPRINTF']
+]
+
+foreach f : check_funcs
```
neheb, line 51: since these are only 4 functions, you can just do

```
 foreach f : ['strcasecmp', '_stricmp', 'snprintf', '_snprintf']
```

subprojects/packagefiles/libssh2/example/meson.build
```diff
@@ -0,0 +1,74 @@
...
+    ['sys/select.h', 'HAVE_SYS_SELECT_H'],
+    ['sys/socket.h', 'HAVE_SYS_SOCKET_H'],
+    ['netinet/in.h', 'HAVE_NETINET_IN_H'],
+    ['arpa/inet.h', 'HAVE_ARPA_INET_H'],
+    ['sys/time.h', 'HAVE_SYS_TIME_H'],
+    ['winsock2.h', 'HAVE_WINSOCK2_H'],
+]
+
+foreach h : check_example_headers
+    example_libssh2_config.set(h[1], cc.has_header(h[0]))
```
neheb, line 41: you could do

```
example_libssh2_config.set('HAVE_@0@'.format(h.underscorify().to_upper()), cc.has_header(h))
```

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,144 @@
...
+    ['sys/un.h', 'HAVE_SYS_UN_H'],
+    ['windows.h', 'HAVE_WINDOWS_H'],
+    ['ws2tcpip.h', 'HAVE_WS2TCPIP_H'],
+    ['winsock2.h', 'HAVE_WINSOCK2_H'],
+    ['ntdef.h', 'HAVE_NTDEF_H'],
+    ['ntstatus.h', 'HAVE_NTSTATUS_H']
+]
+
+foreach h : check_lib_headers
+    libssh2_config.set(h[1], cc.has_header(h[0]))
```
neheb, line 86: same as above.

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,144 @@
...
+endforeach
+
+if cc.has_header('fcntl.h')
+    libssh2_config.set('HAVE_O_NONBLOCK', cc.has_header_symbol('fcntl.h', 'O_NONBLOCK'))
+endif
+
+if cc.has_header('sys/ioctl.h')
```
neheb, line 93: here you can do

```
if libssh2_config.get('HAVE_SYS_IOCTL_H')
```

it's not clear to me that it's better but w/e.

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,144 @@
...
+)
+
+install_headers(
+    files('include/libssh2.h',
+          'include/libssh2_publickey.h',
+          'include/libssh2_sftp.h'
+    )
+)
+
+pkgconfig.generate(libssh2_lib, name: 'libssh2')
```
neheb, line 140: description would be nice. otherwise meson does funny things with it, especially when built as a subproject.

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,144 @@
...
+
+if get_option('examples')
```
neheb, line 142: there's another pattern that has emerged actually. the idea is enter the subdir and in example/meson.build do:

```
if not get_option('examples')
  subdir_done()
endif
```

I don't know which is better.

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,144 @@
...
+libssh2_sources += [libssh2_config_h]
+
+inc_dir = include_directories(['include', '.'])
+libssh2_lib = library('ssh2',
+    libssh2_sources,
+    include_directories: inc_dir,
+    dependencies: libssh2_deps,
+    c_args: extra_flags,
+    install: true,
+    soversion: '1.0.1',
```
neheb, line 125: 1.0.1 should be the version: , no?

Another thing to keep in mind is that on Windows, soversion should be empty. In other packages in wrapdb you see stuff like:

```
soversion = host_machine.system() == 'windows' ? '' : 1
```

subprojects/packagefiles/libssh2/example/meson.build
```diff
@@ -0,0 +1,67 @@
...
+endforeach
+
+example_libssh2_config_h = configure_file(
+    input: 'libssh2_config_cmake.h.in',
+    output: 'libssh2_config.h',
+    configuration: example_libssh2_config,
+    format: 'cmake',
+)
+
+x11_dep = dependency('x11', required: false)
```
neheb, line 55: wonder if it makes sense to make this a feature option.

subprojects/packagefiles/libssh2/meson.build
neheb: you can do 

```
libssh2_config.set('HAVE_@0@'.format(h.underscorify().to_upper()), cc.has_header(h))
```

subprojects/packagefiles/libssh2/meson.build
ilsme: Thanks for the info, I think in this particular case both patterns are equal so I would keep the first variant.

subprojects/packagefiles/libssh2/example/meson.build
ilsme: There is already the 'examples' option and library itself doesn't depend on x11, just this particular example, so I don't think it's a good idea to introduce separate option just for only one example.

jpakkane:
LGTM

subprojects/packagefiles/libssh2/example/meson.build
eli-schwartz: I guess the examples aren't installed anyway, so it's not a big deal if only some are built...

(An argument could be made to make it required when the examples are built, but eh.)

subprojects/packagefiles/libssh2/meson.build
```diff
@@ -0,0 +1,153 @@
...
+    libssh2_sources += ['src/mbedtls.c']
+
+    if dependency('mbedtls', required: false).found()
+        libssh2_deps += dependency('mbedtls')
+    else
+        libmbedtls = cc.find_library('mbedtls', has_headers: ['mbedtls/version.h'])
+        libmbedx509 = cc.find_library('mbedx509')
+        libmbedcrypto = cc.find_library('mbedcrypto')
+        libssh2_deps += [libmbedtls, libmbedx509, libmbedcrypto]
+    endif
```
eli-schwartz, line 53: But mbedtls can't ever exist, can it?

And even if it did, you should save the dependency() once, instead of looking it up a second time after checking if it got found the first time.

eli-schwartz (changes requested):

subprojects/packagefiles/libssh2/meson.build
eli-schwartz: This was changed above, but not here.

eli-schwartz (changes requested):

subprojects/packagefiles/libssh2/meson.build
eli-schwartz: Is this empty soname thing a matter of emptying it for specifically crafted .def files that assume it's empty, or a general Windows concept? I thought it was the former.

subprojects/packagefiles/libssh2/meson.build
ilsme: Hm, `compiler.find_library()` has `required: true` by default so no `mbedtls` in the system will cause the following error:
> Run-time dependency mbedtls found: NO (tried pkgconfig and cmake)

eli-schwartz: I'm not sure why that error is a problem? Don't we want that? If the mbedtls backend is selected, then its absence *should* be an error, no?

Also, it's still not clear to me how `dependency('mbedtls')` could exist since the upstream mbedtls project doesn't install any dependency exports.

ilsme: Oh, I get it now, this dependency('mbedtls') was added just in case the upstream will add dependency exports in future.

ilsme: @eli-schwartz so do you think it's a bad idea and I should remove it?

eli-schwartz: Well, it's a wasted lookup, produces confusing logs, and we can always add it if and when mbedtls upstream adds support.

ilsme: Ok, agree, removed it.

neheb:
Needs a rebase to latest master.

PR #541, opened 2022-07-22
m4: Add wrap

talisein:
The goal of this m4 wrap is to allow meson projects to easily use m4 on Windows. I intend to use it to be able to build gtkmm in maintainer mode on windows.

m4 uses Gnulib to fix up any platform's POSIX shortcomings, which in turn makes this somewhat of a monstrous wrap. In some cases I have closely emulated the autoconf checks, but in others I simply assumed windows was broken.

I've tested it in my own github actions, but we'll see how it does in wrapdb's CI...


talisein:
The ubuntu failure is from 198.sysval. This test has m4 execute a shell command "kill -9 $$" and expects no output on stderr. However, some shell environments end up printing "Killed". When I run this locally, and on my own github ubuntu runner, this test does not fail.

I can disable the test, but maybe someone has an idea of how to manipulate the shell environment so this message doesn't appear?

neheb:
stdout or stderr?

talisein:
> stdout or stderr?

stderr. As I have the test runner configured, it explains:
> Unexpected stderr:
> b"sh: line 1:  5940 Killed                  /bin/sh -c 'kill -9 $$'"

neheb:
2> /dev/null I guess. Alternatively run with /bin/sh or /bin/bash

neheb:
```
This test assumes kill is a shell builtin
```

Ubuntu default to dash instead of bash. I guess that's why it's failing in CI.

talisein:
Okay, I've just removed the test that depends on /bin/sh being bash. It doesn't seem worth the effort to figure out the true identity of /bin/sh

subprojects/packagefiles/m4/src/nononnull.h
```diff
@@ -0,0 +1,3 @@
+#ifndef _GL_ARG_NONNULL
+#define _GL_ARG_NONNULL(params)
+#endif
```
jpakkane, line 3: This should be a `.h.in` file that is configured with the copy only kwarg to stick with the requirement that we don't ship "active code" for lack of a better term.

PR #542, opened 2022-07-23
Update range-v3 version to 0.12.0-1

monwarez:


neheb:
Would be great to add tests if possible.

monwarez:
I added all test under `test/ ` except one that check for coroutine but that is enabled with a flags, which is not done in this meson wrap, so I am not sure that we need to add it at the moment

jpakkane:
Perfect is the enemy of the good. Those can be added later. Thanks.

PR #543, opened 2022-07-27
ftxui: add missing windows hack

ageh:
FTXUI supports a hack [1] for true color support on Windows that is set via a CMake flag. This patch adds the same flag to the Meson build file.

[1] https://github.com/ArthurSonzogni/FTXUI/issues/135


jpakkane:
You need to add a new entry to the release json file.

subprojects/packagefiles/ftxui/meson.build
```diff
@@ -9,7 +9,7 @@ if host_machine.system() == 'windows'
   if cpp.get_argument_syntax() == 'msvc'
```
neheb, line 9: I think this can be removed. meson adds /utf-8 by default AFAIK.

ageh:
> You need to add a new entry to the release json file.

Oh sorry, should be fixed now. Is there an easier way than making a second commit though?

jpakkane:
```
git commit -a --amend
git push --force
```

jpakkane:
Though to merge the two commits you have into one you need to do something like:

    git rebase --interactive HEAD~3

ageh:
This seems to work now?

Thank you for all the help @jpakkane , it's awesome that you take the time to help out on super small issues like this.

PR #545, opened 2022-07-29
cppzmq: new wrap

stephanlachnit:
This currently still requires libzmq, which I will add as wrap hopefully soon.

### DIFF
--- a/subprojects/packagefiles/cppzmq/meson.build
+++ b/subprojects/packagefiles/cppzmq/meson.build
@@ -1,0 +1,62 @@
+project('cppzmq', 'cpp',
+  version : '4.8.1',
+  license : 'MIT',
+  meson_version: '>=0.47.0',
+)
+
+libzmq_dep = dependency('libzmq')
+
+cppzmq_dep = declare_dependency(include_directories: include_directories('.'), dependencies: libzmq_dep)
+
+# examples
+thread_dep = dependency('threads')
+
+ex_pubsub_multithread_inproc_exe = executable('pubsub_multithread_inproc',
+  'examples/pubsub_multithread_inproc.cpp',
+  cpp_args: ['-include', 'thread'],  # fixed upstream in https://github.com/zeromq/cppzmq/pull/568
+  dependencies: [cppzmq_dep, thread_dep],
+  build_by_default: false)
+
+# don't test pubsub_multithread_inpro, it runs forever
+#test('ex_pubsub_multithread_inpro', ex_pubsub_multithread_inproc_exe)
+
+ex_hello_world_exe = executable('hello_world',
+  'examples/hello_world.cpp',
+  dependencies: [cppzmq_dep, thread_dep],
+  build_by_default: false)
+
+test('ex_hello_world', ex_hello_world_exe)
+
+ex_multipart_messages_exe = executable('multipart_messages',
+  'examples/multipart_messages.cpp',
+  dependencies: [cppzmq_dep, thread_dep],
+  build_by_default: false)
+
+test('ex_multipart_messages', ex_multipart_messages_exe)
+
+# unit tests
+catch2_dep = dependency('catch2', required: get_option('catch2'))
+if catch2_dep.found()
+  test_src = files([
+    'tests/buffer.cpp',
+    'tests/message.cpp',
+    'tests/context.cpp',
+    'tests/socket.cpp',
+    'tests/socket_ref.cpp',
+    'tests/poller.cpp',
+    'tests/active_poller.cpp',
+    'tests/multipart.cpp',
+    'tests/recv_multipart.cpp',
+    'tests/send_multipart.cpp',
+    'tests/codec_multipart.cpp',
+    'tests/monitor.cpp',
+    'tests/utilities.cpp'])
+  test_inc = include_directories('tests')
+
+  unit_tests_exe = executable('unit_tests', test_src,
+    dependencies: [cppzmq_dep, catch2_dep, thread_dep],
+    include_directories: [test_inc],
+    build_by_default: false)
+
+  test('unit_tests', unit_tests_exe)
+endif
--- a/ci_config.json
+++ b/ci_config.json
@@ -38,6 +38,21 @@
       "cpp_std=c++14"
     ]
   },
+  "cppzmq": {
+    "debian_packages": [
+      "libzmq3-dev"
+    ],
+    "brew_packages": [
+      "zeromq"
+    ],
+    "build_options": [
+      "cppzmq:catch2=enabled"
+    ],
+    "_comment": "- libzmq not yet available as wrap",
+    "build_on": {
+      "windows": false
+    }
+  },
   "cpr": {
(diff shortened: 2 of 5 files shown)


stephanlachnit:
@eli-schwartz @jpakkane can we merge this already? I've made some progress on libzmq (https://github.com/stephanlachnit/libzmq/tree/p-meson), but it selecting the correct option will take some more time. Doesn't block cppzmq though as it works fine with system libzmq.

eli-schwartz:
Sorry, I forgot to look at it because I assumed that you were going to be updating the PR to also wrap zeromq.

I think you should install zeromq for macOS via `brew_packages`. Additionally, how about building and running the tests/ directory?

stephanlachnit:
> I think you should install zeromq for macOS via brew_packages.

Good point

> Additionally, how about building and running the tests/ directory?

Build the tests, could also add the examples. Looks like the CI doesn't take catch2 from wrapdb, how do I do this?

Is there some kind of preferred way to enable testing in meson? I think when using cppzmq in a subproject, the tests and examples should not be build by default.

eli-schwartz:
Just add a meson_options.txt with a tests feature option, and use that as the required kwarg for catch2. If it is required, then it will be picked up via wraps, wrap fallbacks don't automatically get used for optional dependencies.

(There's a kwarg to allow fallback anyway.)

stephanlachnit:
Ah right, I see. But then the feature would have to be enabled by default for this to work, right?

A bit offtopic, but maybe meson should add a builtin tests bool? That could be disabled for subprojects by default, but e.g. here in the CI enabled via cli.

neheb:
That wouldn't make much sense. Note that you still have to run ninja tests or meson tests to get them to run. A simple compile does that but does not run or install them.

subprojects/packagefiles/cppzmq/meson.build
```diff
@@ -1,10 +1,39 @@
...
 )
 
 libzmq_dep = dependency('libzmq')
 
 cppzmq_inc = include_directories('.')
 
 cppzmq_dep = declare_dependency(include_directories: cppzmq_inc, dependencies: libzmq_dep)
+
+# tests
+catch2_dep = dependency('catch2', required: false)
```
neheb, line 13: Should probably use a feature option here to allow disabling the tests.

stephanlachnit:
> That wouldn't make much sense.

I disagree, tests should only be compiled if wished by the user (developer vs user). Sure, enabling them by default is a good default, but having a default way to not compile tests is something useful. It might not make a difference here, but if you have a giant testsuite containing maybe hundreds of files, it can take quite a while to compile.
Lots of CMake projects have it, and even some meson projects (e.g. https://github.com/libratbag/piper/blob/7f9834cc5901f790a596bae61e3eb81e63d7c2bd/meson_options.txt#L1-L4).

> Note that you still have to run ninja tests or meson tests to get them to run. A simple compile does that but does not run or install them.

Sure, but I have my own tests defined, I don't care about tests from the subproject - possible test failures have nothing to do with my project.

eli-schwartz:
> I disagree, tests should only be compiled if wished by the user (developer vs user). Sure, enabling them by default is a good default, but having a default way to not compile tests is something useful.

Actually, this shouldn't be an option, and enabling to compile them by default is a pointless waste.

It's just a bug in Meson that currently, `build_by_default: false` is ignored for testsuite executables.

What should happen is you mark every test executable to not build by default, but it's fine to unconditionally define their rules. Running `ninja test` should build the testsuite executables, but `ninja all` shouldn't build them.

This works fine in most cases, though if the tests require additional dependencies then it's probably desirable to have an option for that.

stephanlachnit:
> What should happen is you mark every test executable to not build by default, but it's fine to unconditionally define their rules. Running ninja test should build the testsuite executables, but ninja all shouldn't build them.

Ah thanks! This seems like a good idea - but there is still the subproject problem. But this should be fixable with an option for `meson test`, i.e. `meson test --with-subprojects`.

eli-schwartz:
```
meson test --suite mainprojectname
```

stephanlachnit:
> meson test --suite mainprojectname

Thanks!

stephanlachnit:
Added tests and examples with `build_by_default: false`, so from my side good to merge now.

> Just add a meson_options.txt with a tests feature option, and use that as the required kwarg for catch2. If it is required, then it will be picked up via wraps, wrap fallbacks don't automatically get used for optional dependencies.

Did that, it still is not picked up by the CI when left on auto.

subprojects/packagefiles/cppzmq/meson.build
```diff
@@ -0,0 +1,62 @@
+project('cppzmq', 'cpp',
+  version : '4.8.1',
+  license : 'MIT',
+)
+
+libzmq_dep = dependency('libzmq')
+
+cppzmq_inc = include_directories('.')
+
+cppzmq_dep = declare_dependency(include_directories: cppzmq_inc, dependencies: libzmq_dep)
```
eli-schwartz, line 10: Can we just inline the inc here?

subprojects/packagefiles/cppzmq/meson.build
```diff
@@ -0,0 +1,62 @@
...
+
+# examples
+thread_dep = dependency('threads')
+
+ex_pubsub_multithread_inproc_exe = executable('pubsub_multithread_inproc',
+  'examples/pubsub_multithread_inproc.cpp',
```
eli-schwartz, line 16: I was wondering why I could not build this until I saw https://github.com/zeromq/cppzmq/pull/568

An absolutely filthy hack: 

```suggestion
  'examples/pubsub_multithread_inproc.cpp',
  # fixed upstream in https://github.com/zeromq/cppzmq/pull/568
  cpp_args: ['-include', 'thread'],
```

subprojects/packagefiles/cppzmq/meson.build
```diff
@@ -0,0 +1,62 @@
+project('cppzmq', 'cpp',
+  version : '4.8.1',
+  license : 'MIT',
```
eli-schwartz, line 3:
```meson
meson_version: '>=0.47.0'
```

eli-schwartz (commented):
The ci_config.json can pass `-Dcppzmq:catch2=required` as build_options for this wrap. This will guarantee fallback happens.

It's also possible to use `dependency(..., allow_fallback: true)` but that bumps the minimum version of Meson from 0.47.0 to 0.56.0

subprojects/packagefiles/cppzmq/meson.build
stephanlachnit: Ah thanks, that's easier than adding patch files.

stephanlachnit:
> The ci_config.json can pass `-Dcppzmq:catch2=required` as build_options for this wrap. This will guarantee fallback happens.

`cppzmq:catch2=enabled` but thanks, done.

stephanlachnit:
@eli-schwartz any more objections or can this be merged?

PR #547, opened 2022-08-05
Use `c_std=gnu99` in SpeexDSP for inline ASM

kcgen:
Building via wrap on aarch64 currently fails with: `"error: ‘asm’ undeclared (first use in this function)"` due to the use of the more restrictive C99 standard.

This commit switches to the `gnu99` standard, which permits and builds the inline assembly.


kcgen:
:partying_face: thank you for all your work on Meson, @eli-schwartz !

PR #328, opened 2022-03-15
WIP: libjpeg-turbo: Add simd support for x86

xclaesse:

PR #548, opened 2022-08-05
libjpeg-turbo: update to 2.1.3, add x86 and x86_64 SIMD support

skullernet:


neheb:
/me grabs popcorn. https://github.com/mesonbuild/wrapdb/pull/328

skullernet:
Didn't realize there's already a PR for this, sorry. But since that PR depends on functionality not yet available in Meson, maybe this one can still be accepted as temporary solution?

subprojects/packagefiles/libjpeg-turbo/simd/meson.build
```diff
@@ -0,0 +1,81 @@
...
+  'i386/jfdctint-avx2.asm', 'i386/jfdctint-mmx.asm', 'i386/jfdctint-sse2.asm',
+  'i386/jidctflt-3dn.asm', 'i386/jidctflt-sse2.asm', 'i386/jidctflt-sse.asm',
+  'i386/jidctfst-mmx.asm', 'i386/jidctfst-sse2.asm', 'i386/jidctint-avx2.asm',
+  'i386/jidctint-mmx.asm', 'i386/jidctint-sse2.asm', 'i386/jidctred-mmx.asm',
+  'i386/jidctred-sse2.asm', 'i386/jquant-3dn.asm', 'i386/jquantf-sse2.asm',
+  'i386/jquanti-avx2.asm', 'i386/jquanti-sse2.asm', 'i386/jquant-mmx.asm',
+  'i386/jquant-sse.asm', 'i386/jsimdcpu.asm',
+]
+
+simd_opt = get_option('simd').require(host_cpu == 'x86' or host_cpu == 'x86_64',
```
neheb, line 36: host_cpu should be changed to use host_machine.cpu_family()

On 32-bit x86, it's returning i686 for me.

neheb:
builds with MSYS2 (all backends) with the noted change.

subprojects/packagefiles/libjpeg-turbo/simd/meson.build
eli-schwartz, line 36: You'll need to raise the minimum version of Meson, currently unspecified, to an extremely recent version to use .require() -- or rather, to signify to people using it that they need that recent a version.

skullernet: Not sure why I wrote `cpu()` in first place, fixed.

skullernet: Rewrote without use of `require()`.

neheb:
need to rebase against current master.

PR #550, opened 2022-08-06
speexdsp: build NEON with clang on Debian armhf

neheb:
On GCC, __thumb__ evaluates to true and skips this section (correct as
GCC refuses to build non thumb code). Under clang, __thumb__ does not
evaluate and thus builds this code. Needs some adjustment as +simd is
not suppored with clang.

and yes, I have tried removing the __thumb__ check and build with GCC. Doesn't work.

edit: my specific board runs NEON code just fine. It's just Debian's binutils and GCC that are broken.

ping @kcgen


kcgen:
Even better; thanks for the fix-up @neheb !

neheb:
This could probably be fixed upstream, but I have no idea how to write code for THUMB, which does support NEON.

PR #551, opened 2022-08-06
flac: fix compilation on non x86 and clang

neheb:
clang's cpuid header errors with "this header is for x86 only". Make it
conditional.


neheb:
the flac code also only uses __cpuid for x86. ping @jpakkane

dcbaker:
Use check _header instead of has_header, which will catch this

dcbaker:
Actually, this sort is bs is exactly why check_header exists :/

neheb:
sounds like has_header should be deprecated.

dcbaker:
Has_header is faster, but I never use it because iof this concern. We should at least have a big warning in the docs about it

neheb:
```
~/d/wrapdb (master)> git grep "\.check_header"
subprojects/packagefiles/cmocka/meson.build:    if compiler.check_header(h)
subprojects/packagefiles/sdl2/meson.build:    if cc.check_header(h, dependencies : core_deps)
```

not widely used it seems.

dcbaker:
I’m honestly tempted to add a check keyword to has_header that defaults to false and note that in the future it will change to true

neheb:
ping @jpakkane

PR #552, opened 2022-08-06
xz: fix compilation with non x86 and clang

neheb:
clang's immintrin.h fails on non x86 and clang.


neheb:
ping @jpakkane

PR #556, opened 2022-08-07
Add ImGuizo

tmayoff:
An extension to ImGui

Also running sanity_checks on my fork clones and tests all the wraps not just the new one, any ideas why

### DIFF
--- a/subprojects/packagefiles/imguizmo/meson.build
+++ b/subprojects/packagefiles/imguizmo/meson.build
@@ -1,0 +1,13 @@
+project('ImGuizmo', 'cpp', version: '1.83')
+
+sources = files(
+        'GraphEditor.cpp',
+        'ImCurveEdit.cpp',
+        'ImGuizmo.cpp',
+        'ImSequencer.cpp')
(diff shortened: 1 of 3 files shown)


subprojects/packagefiles/imguizmo/meson.build
```diff
@@ -0,0 +1,13 @@
...
+sources = files(
+        'GraphEditor.cpp',
+        'ImCurveEdit.cpp',
+        'ImGuizmo.cpp',
+        'ImSequencer.cpp')
+
+imgui = dependency('imgui')
+
+imguizmo_inc = include_directories('.')
+imguizmo_lib = library('ImGuizmo', sources, include_directories: imguizmo_inc, dependencies: imgui)
```
jpakkane, line 12: Library names are typically all lowercase unless upstream explicitly chooses to use something else (in this case they don't even build a library file FWICT).

jpakkane:
You need to add an entry to the releases json file.

Please also squash your commits so there is only one.

PR #558, opened 2022-08-08
ftxui: fix compilation under Cygwin

neheb:
gnu++ standard needed for fileno definition.


jpakkane:
It does not seem to like VS. I'm not sure if we need to support Cygwin here as the number of people who use ftxui on that platform is roughly zero. :D

PR #559, opened 2022-08-09
imgui wrap windows improvements

noah1510:
This pull request has two fixes for the imgui wrap for windows.

As described in a comment about ImGui_ImplWin32_EnableAlphaCompositing linking with '-ldwmapi' is required for non MSVC.

Additionally instead of having libdl as requirement for opengl, GL is used as the opengl dependency as it is described in the meson dependency documentation.

If you want some more changes or have other questions please go ahead and write a comment 😄


subprojects/packagefiles/imgui/meson.build
```diff
@@ -53,10 +53,10 @@ if metal_dep.found()
     sources += 'backends/imgui_impl_metal.mm'
     dependencies += metal_dep
 endif
-libdl_dep = cpp.find_library('dl', required: get_option('opengl'))
-if libdl_dep.found()
+libgl_dep = dependency('GL', required: get_option('opengl'))
```
neheb, line 56: why is GL capitalized?

noah1510: It is capitalized in the [meson manual](https://mesonbuild.com/Dependencies.html#gl), so I also do that. This way it compiles on windows and on linux, so i did not try it in small letters.

stephanlachnit: You should be able to lower-case it (see e.g. Boost, CUDA, AppleFrameworks in the manual)

stephanlachnit (approved):
LGTM, I think this should be tagged as a new patch release in [releases.json](https://github.com/mesonbuild/wrapdb/blob/a2c2a2d62d422651b3d8f9a4ec75851a53c049ec/releases.json#L673)

noah1510:
I created the patch release. Anything else that needs to be done for the merge request?

neheb:
As far as case sensitivity, I wonder what the case is for macOS. macOS defaults to case-insensitivity but if turning it on, I wonder if something like dependency('appleframeworks', modules : [ 'CoreFoundation', 'IOKit', 'Security' ]) would fail.

PR #564, opened 2022-08-12
`mt32emu`: update to 2.7.0

kcgen:
Suggest reviewing commit-by-commit, as the second commit is a whitespace-only reformat.


neheb:
hrm what muon settings did you use? I personally use

```
indent_by='  '
space_array=false
wide_colon=false
```

kcgen:
Just defaults, @neheb.  Thanks for these, I'll re-run it.

PR #565, opened 2022-08-16
lz4 1.9.4

jcfj:
(Seems like lz4's contrib directory is not protected by semver. Still, I'd prefer to use the files from upstream as is.)

PR #566, opened 2022-08-16
librdkafka 1.9.2

jcaesar:


jcfj:
It's a bit weird that [this check](https://github.com/mesonbuild/wrapdb/pull/565#issuecomment-1216241094) failed because I changed nothing about the `meson.build` before it failed. I guess it's a new check?

PR #570, opened 2022-08-20
SDL2: update to 2.24.0

Jan200101:
supersedes #415


neheb:
looks like

```
add_project_arguments('-fobjc-weak', language: 'objcpp')
```

is needed.

Jan200101:
Gotten it to compile on all platforms now.

Have not had the time to test it yet, will try as soon as possible.

After this PR I'll look into hooking up the SDL tests, having done multiple SDL updates now there are way too many minor things that change that can easily be missed so its better to ensure everything is as it should be.

neheb:
Bah wish I looked closer before merging. Upstream has https://github.com/libsdl-org/SDL/blob/main/include/SDL_config.h.cmake , which can be used instead of a custom meson config file.

PR #572, opened 2022-08-22
check: use cmake configuration file

neheb:


neheb:
```
#define timer_t ${timer_t}
```

wonder why meson transforms it like this...

eli-schwartz:
> wonder why meson transforms it like this...

Because meson.build sets `TIMER_T` which isn't used, and not `timer_t` which is used. Meson simply skips processing the line altogether, there's nothing to modify there.

neheb:
timer_t is set to int...

PR #574, opened 2022-08-22
SDL2: compile unit tests, switch to cmake config.h

Jan200101:
- switches to the cmake
- add unit testing capabilities
- add all platform independent tests

related to #570


subprojects/packagefiles/sdl2/meson_options.txt
```diff
@@ -1,3 +1,16 @@
...
+    'test',
+    type : 'boolean',
+    value : 'false',
+    description : 'build tests'
+)
+
+option(
+    'run_test',
+    type : 'boolean',
+    value : 'false',
```
xclaesse, line 11: AFAIK it is uncommon to have separate option to run tests, is it really needed? I would at least make it `true` by default.

Jan200101: It is, but I would rather not run these by default
the tests create windows, do audio playback, etc. and do not automatically terminate.

xclaesse: By default you already don't even build them, that seems good enough, no?

Jan200101: I guess?

I find compiling the tests more useful than actually running them because it shows if anything is missing, but I can merge the two and keep tests off by default

xclaesse: That's true. Maybe set `tests` to true by default, and `run_tests` false? Just small nitpicking anyway ;P

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -1100,3 +1102,9 @@ sdl2_dep = declare_dependency(
 if meson.version().version_compare('>=0.54')
     meson.override_dependency('sdl2', sdl2_dep)
 endif
+
+# MSVC does not correctly resolve the symbols when linking dynamically
+# works fine when linking statically or MinGW
+if get_option('test') and (cc.get_id() != 'msvc' or get_option('default_library') == 'static')
```
Jan200101, line 1108: I hate this "fix" but I couldn't find a better way of doing this without explicitly creating a shared and static library

xclaesse: What error do you get? This hack smell like hiding the symptoms of an important root cause that needs fixing.

Jan200101: Same as the [previous CI run](https://github.com/mesonbuild/wrapdb/runs/7958764685?check_suite_focus=true), `unresolved external symbol`

After doing slightly more digging I found that my comment is incorrect and MSVC does resolve symbols correctly, but the symbols for the test functions are never `dllexport`'d and thus the tests need to be statically linked.
https://github.com/libsdl-org/SDL/blob/cd79e9c4e25ffd15010953b15ae8cbc25b7e7a90/include/SDL_test_font.h#L93

This is not an issue on MinGW because it exports everything by default (afaik)

xclaesse: Ok, seems cmake requires static too: https://github.com/libsdl-org/SDL/blob/main/test/CMakeLists.txt#L4 ?

I think it `test` option is enabled and it's not a static build it should error() instead of silently skipping tests. Shouldn't we always require static build regardless of the compiler? If it works with GCC it means we are not handling symbol visibility, maybe that's a bug too, no?

subprojects/packagefiles/sdl2/meson_options.txt
eli-schwartz: There's a pretty simple solution, you can use `add_test_setup(..., is_default: true)` in combination with setting suites for these tests, to only run a subset of tests with the default `meson test`.

subprojects/packagefiles/sdl2/meson.build
xclaesse: I can confirm that's a bug in our gcc/clang builds, we should add `gnu_symbol_visibility: 'hidden'` because they have `DECLSPEC` macro that does the right thing to mark public APIs.

tristan957: Can't you avoid this logic by doing an internal static dependency? https://github.com/lz4/lz4/blob/dev/contrib/meson/meson/lib/meson.build#L47

eli-schwartz: Or the slightly more elaborate version I did here: https://github.com/facebook/zstd/commit/6548ec7440712eb531f4148ed0568cf90fa1e523

Jan200101: > Ok, seems cmake requires static too: https://github.com/libsdl-org/SDL/blob/main/test/CMakeLists.txt#L4 ?

Yes, but cmake also requires developers to explicitly define different static and shared libraries which is exactly why `SDL2-static`.

Meson goes down the route of allowing developers to either specify static or shared libraries or allow the consumer to specify what they need using `default_library`


> I can confirm that's a bug in our gcc/clang builds, we should add `gnu_symbol_visibility: 'hidden'` because they have `DECLSPEC` macro that does the right thing to mark public APIs.

Are you talking about gcc and clang on Windows?
I don't consider them bugged because they exhibit the same behavior as on other platforms, its just that the Windows Native tooling made the decision hide symbols


> Can't you avoid this logic by doing an internal static dependency? https://github.com/lz4/lz4/blob/dev/contrib/meson/meson/lib/meson.build#L47

Yes, thats how I initially solved this with my cmake brain.
And there are multiple ways of doing exactly that, be it extracting the objects out of the library or linking whole.

But then the question comes up for me again if this is really the right way, having a static internal library that is always build regardless of what the consumer decides on.

Patching is out of the question so maybe its worth talking to upstream to see if they are interested in giving the test methods a DECLSPEC?
I don't think they are suppose to be publicly linked against so I doubt they will do such a thing

cmake provides the ability to create an "object" library that you can then directly reuse for other libraries, perhaps it might be worthwhile adding a `object_library()` or similar to meson?

Maybe this should just be accepted as one of those MSVC things that needs to be worked around like `default_library=both` or the sdl2main linking that work fine on gcc and clang.

---

I'll leave the details of it all up to you all, I mostly care about getting the wrap feature complete and comparable to the upstream CMakeLists.txt

xclaesse: > But then the question comes up for me again if this is really the right way, having a static internal library that is always build regardless of what the consumer decides on.

Yes, that's the right way of doing it, the intermediary static library won't be installed, it's only used for linking unit tests.

> Are you talking about gcc and clang on Windows?
> I don't consider them bugged because they exhibit the same behavior as on other platforms, its just that the Windows Native tooling made the decision hide symbols

With gcc/clang on Linux currently exposes all symbols, including test symbols, that's a bug. cmake sets visibility to hidden by default too: https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L650.

PR #578, opened 2022-08-23
pcre: use cmake config files instead of meson

neheb:
Less error prone.


subprojects/packagefiles/pcre/meson.build
```diff
@@ -80,24 +149,22 @@ char_tables = custom_target('chartables',
...
+  link_with: pcre8,
+)
 
 pcretest = executable('pcretest',
   'pcretest.c', 'pcre_printint.c',
-  c_args : compile_args,
-  link_with : [pcre8, pcre_posix],
-  include_directories : conf_inc,
+  c_args: compile_args,
+  dependencies: libedit_dep,
```
xclaesse, line 166: Does it build if libedit_dep is not found?

subprojects/packagefiles/pcre/meson.build
```diff
@@ -80,24 +149,22 @@ char_tables = custom_target('chartables',
   output : 'pcre_chartables.c',
   command : [dftables, '@OUTPUT@'])
 
-pcre8 = build_target('pcre',
+pcre8 = library('pcre',
   pcre8_sources, char_tables,
-  c_args : compile_args,
-  target_type : libtype,
-  include_directories : conf_inc)
+  c_args: compile_args,
```
xclaesse, line 154: Shouldn't you pass all deps you lookup above?

subprojects/packagefiles/pcre/meson.build
neheb: It does yes.

subprojects/packagefiles/pcre/meson.build
neheb: Yeah probably. The interesting thing is that these dependencies seem to be used not by the library but utilities that it comes with.

xclaesse: Do we build those utilities? If not, then we don't need to lookup for those deps at all.

neheb: pcretest is built, yes.

I guess since they're all optional they can be removed.

neheb: Yeah:

```
~/p/pcre-8.45 (master)> git grep SUPPORT_LIBZ *.c
pcregrep.c:#ifdef SUPPORT_LIBZ
pcregrep.c:#ifdef SUPPORT_LIBZ
pcregrep.c:#if defined SUPPORT_LIBZ || defined SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBZ
pcregrep.c:#ifdef SUPPORT_LIBZ
pcregrep.c:#ifdef SUPPORT_LIBZ
pcregrep.c:#ifdef SUPPORT_LIBZ
pcregrep.c:#if defined SUPPORT_LIBZ || defined SUPPORT_LIBBZ2
pcregrep.c:#if defined SUPPORT_LIBZ || defined SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBZ
pcregrep.c:#ifdef SUPPORT_LIBZ
~/p/pcre-8.45 (master)> git grep SUPPORT_LIBB *.c
pcregrep.c:#ifdef SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBBZ2
pcregrep.c:#if defined SUPPORT_LIBZ || defined SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBBZ2
pcregrep.c:#if defined SUPPORT_LIBZ || defined SUPPORT_LIBBZ2
pcregrep.c:#if defined SUPPORT_LIBZ || defined SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBBZ2
pcregrep.c:#ifdef SUPPORT_LIBBZ2
~/p/pcre-8.45 (master)> git grep SUPPORT_LIBE *.c
pcretest.c:#if defined(SUPPORT_LIBREADLINE) || defined(SUPPORT_LIBEDIT)
pcretest.c:#if defined(SUPPORT_LIBREADLINE) || defined(SUPPORT_LIBEDIT)
pcretest.c:#if defined(SUPPORT_LIBREADLINE) || defined(SUPPORT_LIBEDIT)
mangix@TrapNAS ~/p/pcre-8.45 (master)> git grep SUPPORT_LIBR *.c
pcretest.c:#if defined(SUPPORT_LIBREADLINE) || defined(SUPPORT_LIBEDIT)
pcretest.c:#if defined(SUPPORT_LIBREADLINE)
pcretest.c:#if defined(SUPPORT_LIBREADLINE) || defined(SUPPORT_LIBEDIT)
pcretest.c:#if defined(SUPPORT_LIBREADLINE) || defined(SUPPORT_LIBEDIT)
```

It's all pcretest and pcregrep. The former is built but not the latter.

xclaesse: ok good

subprojects/packagefiles/pcre/meson_options.txt
```diff
@@ -1 +1,7 @@
-option('shared_lib', type : 'boolean', value : false)
```
eli-schwartz, line 1: Not strictly part of the stated purpose of this PR.

On the other hand, thank you! :D :D :D

neheb: Yeah that's probably just a cmake option to deal with cmake's inability to have both library types. At least not easily.

xclaesse:
@neheb I don't understand why, but this PR breaks pcre symbol visibility on Windows, all wraps that depend on pcre fails because of this.

```
2022-08-28T00:42:51.5733561Z [207/1248] "link" @subprojects/glib-2.72.2/glib/glib-2.0-0.dll.rsp
2022-08-28T00:42:51.5735299Z FAILED: subprojects/glib-2.72.2/glib/glib-2.0-0.dll subprojects/glib-2.72.2/glib/glib-2.0-0.pdb 
2022-08-28T00:42:51.5815967Z "link" @subprojects/glib-2.72.2/glib/glib-2.0-0.dll.rsp
2022-08-28T00:42:51.5821853Z    Creating library subprojects\glib-2.72.2\glib\glib-2.0.lib and object subprojects\glib-2.72.2\glib\glib-2.0.exp
2022-08-28T00:42:51.5824645Z D__a_wrapdb_wrapdb_subprojects_glib-2.72.2_glib_gregex.c.obj : error LNK2001: unresolved external symbol pcre_free```

neheb:
I'll look into it.

eli-schwartz:
This PR has a nonsense change unrelated to the commit message.

That change should not have passed review, it got rid of a highly intentional Windows workaround.

xclaesse:
> it got rid of a highly intentional Windows workaround

Could you please elaborate on what was intentional?

xclaesse:
I tried to debug myself and I haven't found why dll are not working. I'm wondering if shared lib ever worked even upstream. We were building static libraries by default before this PR, that could be why it ever worked.

eli-schwartz:
Literally that.

Dlls didn't work on Windows, the previous version of this wrap acknowledged that and forced a static library.

This PR ceased that, and it did so silently, without stating in the commit message that it was doing so. This is wrong for several reasons, but the main reason that matters today is that by hiding what it did, it failed to draw attention to a reviewable change, that change did not in fact get reviewed, and that change created a regression.

xclaesse:
Ok, I did not know that whole story. I assumed since it was using build_target() is was just decade old code. Thanks for the clarification.

xclaesse:
Side note, glib master has been switched to pcre2, that will be part of next stable release 2.74. So this pcre1 wrap will become less important to me at least.

xclaesse:
https://github.com/mesonbuild/wrapdb/pull/608

eli-schwartz:
build_target isn't deprecated, it's highly useful in modern code, as long as that code needs to set the library type based on something like the platform instead of or in addition to user options.

I knew I'd seen it used that way before, but mea culpa, pcre wasn't one of them.

It's still a suspicious change of the previous wrap version -- before, it defaulted to static, now it defaults to shared.

But it's not obvious why, like it's obvious elsewhere.

PR #579, opened 2022-08-23
libtiff: update to 4.4.0

neheb:
Less error prone than custom ones.


neheb:
Cleaned it up massively and added support for various libraries.

PR #581, opened 2022-08-24
Add dependabot for checking action updates

tristan957:
Dependabot will submit PRs to update GitHub Actions. This is the same
config as the meson repo.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,7 @@
+version: 2
+
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly


eli-schwartz:
The Meson repo does not do this. Actually, I will quote my reply when someone submitted a PR to add dependabot to the meson repo:

> Maybe when dependabot stops spamming forks and refusing to group updates together, etc. Personally, I find it all rather gross.
> 
> Supposedly renovate solves both of these problems. The OpenSSF scorecard even seems to subtly suggest a preference for renovate.
> 
> Also nice: renovate is open source and reduces vendor lockin. This is both nice, in the sense that OSS projects should support OSS solutions, and also probably the reason that they actually manage to improve the state of affairs vs. dependabot.

tristan957:
Not gonna argue

PR #582, opened 2022-08-24
CI updates

tristan957:
- Fixes a missing required fields
- Adds concurrency

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -9,6 +9,7 @@ on:
   schedule:
     # Run every Sunday at midnight
     - cron: '0 0 * * 0'
+
 jobs:
   Ubuntu:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -5,6 +5,10 @@ on:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
```
eli-schwartz, line 10: I think the release workflow, specifically, should probably work the same way that the meson repo's documentation workflow `website.yml` works:

```yml
concurrency:
  group: mesonbuild.com
  # We do NOT want `cancel-in-progress` here since only one website job
  # should run at a time to avoid upload race conditions.
```

.github/workflows/build_all.yml
```diff
@@ -9,6 +9,11 @@ on:
   schedule:
     # Run every Sunday at midnight
     - cron: '0 0 * * 0'
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
```
eli-schwartz, line 15: build_all runs either on a schedule or manually, unlike sanity_checks which is run on PRs and can be iterated.

I *think* (but am not sure) that setting concurrency would prevent manually triggering this job twice in quick succession, with different values for `fatal_warnings`. It might be better to just drop concurrency from this one job.

PR #583, opened 2022-08-24
liblzma: fix dep name

neheb:
There's no liblzma_dep anywhere.

@eli-schwartz surprised this was not a hard error.


eli-schwartz:
Apparently it was simply broken since the original implementation in #162

Making it a hard error would require somehow checking the variables exported by a meson.build, hmm...

eli-schwartz:
I tried to implement this in #584 and it seems to work, what do you think?

neheb:
Broken since initial PR. Incredible. I'm assuming nobody has ever used it?

eli-schwartz:
Well, that or they just used the `fallback:` kwarg...

jpakkane:
[I have used it personally](https://github.com/jpakkane/parzip/blob/master/meson.build). Dunno why it did not work, maybe I had not tried that particular version on non-Linux.

eli-schwartz:
That does fail to build with `--wrap-mode=forcefallback`.

Also it doesn't use lzma at all on windows, dunno why, so that may have narrowed down the scope of the effects.

neheb:
I noticed this issue with libtiff when I was testing the dependencies. It refused to work because of wrong name.

My Debian install does not have liblzma-dev installed. Just zlib-dev.

PR #584, opened 2022-08-24
sanity checks: verify that provide varnames are in fact provided

eli-schwartz:


neheb (approved):

xclaesse:
I'm late to the game, but what about instead adding `depnames` and `prognames`:
```
foreach dep : depnames
  d = dependency(dep)
  if not d.found() or d.type_name != 'internal'
    error()
  endif
endforeach
```

That way it tests also the case where `[provide]` has `dependency_names` and the subproject needs to override them.

xclaesse:
@eli-schwartz https://github.com/mesonbuild/wrapdb/pull/585

eli-schwartz:
Because I didn't think of type_name as a way to be sure it wasn't resolved from the system.

And I didn't think much into it, simply because I figured it was unlikely to mess up override_dependency, unlike variable names. The former is already basically sanity-checked by the override method itself.

PR #586, opened 2022-08-26
cairo: add

neheb:


neheb:
wow...

```
subprojects/cairo-1.17.6/meson.build:83:10: ERROR: File meson-cc-tests/check-unused-result.c does not exist.
```

neheb:
```
 cairo| subprojects\cairo-1.17.4\meson.build:873: DEPRECATION: Library png16 was passed to the "libraries" keyword argument of a previous call to generate() method instead of first positional argument. Adding cairo-png to "Requires" field, but this is a deprecated behaviour that will change in a future version of Meson. Please report the issue if this warning cannot be avoided in your case.
```

Among macOS issues.

xclaesse:
cairo has upstream Meson build system, why are you adding another one here?

xclaesse:
Note that I'm all for adding the wrap, but without a patch overlay. cairo is an important dependency for so many projects, it's good to have in wrapdb.

neheb:
There's already a PR for it.

PR #587, opened 2022-08-26
libtiff: don't check for jbig function

neheb:
Throws 'ERROR: Dependencies must be external dependencies' when building
jbig as a subproject. Just check if it's found.


eli-schwartz:
The funny thing is that this is *literally* impossible to ever resolve as anything other than a subproject wrap. #579 was broken, it should not have made this a dependency() lookup -- just use `cc.find_library()`. CI only passed because it was an optional dependency that wasn't found.

neheb:
would cc.find_library() find the subproject?

eli-schwartz:
It absolutely does not matter, in order for the wrap to do its fundamental job of, well, building the feature, it MUST (in the rfc 2119 sense) detect that feature's availability in the upstream-compatible manner.

There are no ifs, ands, or buts. We need to fix that by using cc.find_library, and only after that do we get to debate whether and how to make subprojects ***also*** work.

It cannot use `dependency()`s that do not work.

We are not going to build an ecosystem of stuff that only works if used with vendored Meson subprojects.

jpakkane:
Why would `dependency` not work? The Debian package at least [has a .pc file](https://packages.debian.org/bullseye/amd64/libjbig2dec0-dev/filelist), though it is called `jbig2dec`.

eli-schwartz:
Because [jbigkit](https://packages.debian.org/source/bullseye/jbigkit) (filelist: [libjbig-dev](https://packages.debian.org/bullseye/arm64/libjbig-dev/filelist)) and [jbig2dec](https://packages.debian.org/source/bullseye/jbig2dec) (filelist: [libjbig2dec0-dev](https://packages.debian.org/bullseye/amd64/libjbig2dec0-dev/filelist)) are different projects

neheb:
Switched to find_library.

neheb:
anything else?

PR #588, opened 2022-08-26
mm-common: add

neheb:

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1271,6 +1271,17 @@
       "3.0.6-1"
     ]
   },
+  "mm-common": {
+    "dependency_names": [
+      "mm-common-libstdc++"
+    ],
+    "program_names": [
+      "mm-common-get"
+    ],
+    "versions": [
+      "1.0.4-1"
+    ]
+  },
   "mocklibc": {
     "versions": [
       "1.0-2"
--- a/subprojects/mm-common.wrap
+++ b/subprojects/mm-common.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = mm-common-1.0.4
+source_url = https://download.gnome.org/sources/mm-common/1.0/mm-common-1.0.4.tar.xz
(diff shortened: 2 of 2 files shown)


eli-schwartz:
What do we actually need this for? It's always used as a subproject() iirc, and is only relevant when building in "maintainer mode" (a concept copied from autotools conventions into the gtkmm community's Meson ports).

neheb:
```
 meson.build:19: WARNING: Project targets '>=0.55.0' but uses feature introduced in '0.58.0': meson.global_build_root.
```

Unrelated error.

neheb:
Oh this is used by libxml++.

eli-schwartz:
> Oh this is used by libxml++.

No it isn't, not unless you build libxml++ from a git checkout of libxml++.

Also they have a wrap file for this already.

neheb:
Yeah they do. What happens with wrap_mode=forcefallback is that their mm-common.wrap file gets copied. This is wrong behavior I think.

eli-schwartz:
> their mm-common.wrap file gets copied. This is wrong behavior I think.

If you don't think Meson should create wrap-redirect files, that can be discussed in https://github.com/mesonbuild/meson/pull/9423

I don't think it's accurate to claim that the behavior is "wrong" from a wrapdb sense, nor is the solution to add it to the wrapdb.

neheb:
```
 AssertionError: Lists differ: ['abs[1249 chars]', 'miniz', 'minizip-ng', 'mm-common', 'mockli[1066 chars]std'] != ['abs[1249 chars]', 'mm-common', 'miniz', 'minizip-ng', 'mockli[1066 chars]std']

First differing element 109:
'miniz'
'mm-common'

Diff is 2972 characters long. Set self.maxDiff to None to see it.
```

I don't get it.

eli-schwartz:
libxml++ works perfectly fine in CI, and it also works perfectly fine when installed into a project and built locally.

PR #590, opened 2022-08-26
meson: update minimum version

neheb:
needed for global_build_root.


xclaesse:
Thanks, that was my mistake :)

PR #591, opened 2022-08-26
libsndfile: try to fix Ubuntu compilation.

neheb:
Instead of looking for the header, check the package in pkgconfig.
Matches what is done in CMake.


eli-schwartz:
Huh, fascinating. How was this ever *supposed* to work then... the pkg-config file offers a shared library to link to.

neheb:
I have no idea. The so called sndio package is for OpenBSD according to the comments. So CI wouldn't have caught it.

Except some package provides sndio.h, which is not the right one.

PR #592, opened 2022-08-26
libxml2: try to fix compilaiton with newer meson

neheb:


xclaesse:
I'm not against this change, but it should have been fixed in Meson master already. https://github.com/mesonbuild/meson/pull/10717

PR #593, opened 2022-08-26
emilk-loguru: use empty soversion for windows

neheb:
As is done elsewhere.


neheb:
ping @xclaesse

PR #594, opened 2022-08-26
expat: use cmake configuration file

neheb:


neheb:
```
  #cmakedefine XML_DEV_URANDOM
```

meson fails to substitute it. Why do I feel I reported this before?

neheb:
@guijan how is libbsd-overlay supposed to work? I'm getting.

```
ccache cc -Isubprojects/expat-2.4.8/libexpat.so.1.6.11.p -Isubprojects/expat-2.4.8 -I../subprojects/expat-2.4.8 -I../subprojects/expat-2.4.8/lib -fvisibility=hidden -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -O0 -g -DXML_ENABLE_VISIBILITY=1 -DHAVE_EXPAT_CONFIG_H -fno-strict-aliasing -DXML_DEV_URANDOM -fPIC -isystem /usr/include/bsd -DLIBBSD_OVERLAY -MD -MQ subprojects/expat-2.4.8/libexpat.so.1.6.11.p/lib_xmlparse.c.o -MF subprojects/expat-2.4.8/libexpat.so.1.6.11.p/lib_xmlparse.c.o.d -o subprojects/expat-2.4.8/libexpat.so.1.6.11.p/lib_xmlparse.c.o -c ../subprojects/expat-2.4.8/lib/xmlparse.c
In file included from ../subprojects/expat-2.4.8/lib/xmlparse.c:116:
/usr/include/bsd/stdlib.h:30:25: error: no include path in which to search for stdlib.h
   30 | #include_next <stdlib.h>
```

guijan:
When switching from libbsd to libbsd-overlay, remove all the lines that include headers from the `bsd/` subdirectory in the system include path, and include the equivalent standard headers if they're not included already.

In expat's case, these lines need to be removed: 
https://github.com/libexpat/libexpat/blob/3bab6c09bbe8bf42d84b81563ddbcf4cca4be838/expat/lib/xmlparse.c#L114-L117

I tested compiling expat tag R_2_4_8 with this PR's current meson.build and meson_options.txt with the change above, use_libbsd enabled, and the `#cmakedefine` fix on Hyperbola Linux and it succeeded.

guijan:
Yes, that would also work.

neheb:
sounds like another PR for upstream :).

eli-schwartz:
Huh, why?

neheb:
to avoid this libbsd mess.

neheb:
@eli-schwartz just read your earlier comment.

I don't think cc.has_function works without adding a dependency to libbsd. Same with has_header_symbol, which uses regular system includes instead of the libbsd overlay.

Actually speaking of the former, it can't be used with libobsd as a subproject. Need to switch to the latter.

eli-schwartz:
But the point is that HAVE_LIBBSD says "the bsd functions are in a namespaced header space".
HAVE_FUNCTIONNAME says "I have the functions".

If you're using an overlay, then the former is false. libbsd itself also provides an overlay mechanism, and offers a choice.

neheb:
libobsd (available as a wrap here) is overlay only.

On that note, even has_header_symbol throws

```
subprojects/expat-2.4.8/meson.build:43:7: ERROR: Dependencies must be external dependencies
```

when using libobsd as a subproject. strange.

neheb:
I took out the cmake config commit. Now it's just a cleanup one.

neheb:
@xclaesse rebased.

ci_config.json
```diff
@@ -22,7 +22,6 @@
     "termbox",
     "unit-system",
     "zstd",
-    "libexif",
```
xclaesse, line 25: This PR is about expat, no?

subprojects/packagefiles/expat/meson.build
```diff
@@ -153,23 +130,16 @@ _files = [
...
-  extra_kwargs += {'vs_module_defs' : 'lib/libexpat.def'}
-endif
-
 libexpat = library(
   'expat',
   _files,
   include_directories : include_directories('lib'),
   dependencies : dep_libbsd,
-  version : '1.6.11',
+  version : '1.8.8',
```
xclaesse, line 138: Better define that at the top of the file, so we are less likely to forget when bumping version in project().

neheb: Better solution would be to derive it from the project version. Not too sure how to do so. @eli-schwartz knows more I think.

eli-schwartz: The library version is libtool `9:8:8`, which is mathematically expressed as 1.8.8, it has no basis in the project version.

Upstream just uses code comments to remember to bump it in every file whenever editing it in one.

neheb: This was fixed.

ci_config.json
neheb: Fixed.

neheb:
anything else?

xclaesse:
Fine with me

subprojects/packagefiles/expat/meson.build
```diff
@@ -24,44 +24,27 @@ project(
...
-endif
-
-if cc.has_header_symbol(header, 'arc4random_buf')
-  config.set('HAVE_ARC4RANDOM_BUF', true)
-elif cc.has_header_symbol(header, 'arc4random')
-  config.set('HAVE_ARC4RANDOM', true)
-endif
+dep_libbsd = dependency('libbsd-overlay', required : get_option('use_libbsd'), allow_fallback: false)
+config.set('HAVE_ARC4RANDOM_BUF', cc.has_header_symbol('stdlib.h', 'arc4random_buf', dependencies: dep_libbsd))
+config.set('HAVE_ARC4RANDOM', cc.has_header_symbol('stdlib.h', 'arc4random', dependencies: dep_libbsd))
```
eli-schwartz, line 40: Can't we just allow the dependency fallback, but if we do fallback then check if dep_libbsd.type_name() == 'internal' and unconditionally set these without running has_header_symbol at all?

eli-schwartz (changes requested):

subprojects/packagefiles/expat/meson.build
neheb: sounds like quite the hack :)

neheb: Added this suggestion.

PR #596, opened 2022-08-26
Fix build of cmocka

xclaesse:
The main library was set to be installed by not built by default, which
makes no sense and is not passing wrapdb CI any more.


subprojects/packagefiles/cmocka/src/meson.build
```diff
@@ -26,9 +26,8 @@ libcmocka = library(
   vs_module_defs : 'cmocka.def',
   soversion : host_machine.system() != 'windows' ? lib_version.split('.')[0] : '',
   version : lib_version,
-  install : get_option('buildtype') != 'static',
+  install : true,
   override_options : _overrides,
-  build_by_default: false,
```
eli-schwartz, line 31: This should have only removed `build_by_default:`

xclaesse: `install : get_option('buildtype') != 'static',` is completely wrong and we should absolutely stop doing that.

eli-schwartz: Not only is it generally correct...

... this is a testing tool. It's even more correct for this particular project.

xclaesse: The don't install it at all. But installing only if shared makes absolutely no sense. You should stop encouraging people of doing that.

In this particular case, since it does not install any header files either, that lib should probably not being installed at all. That I can agree.

PR #598, opened 2022-08-26
yajl: try to fix compilation with MSVC

neheb:


subprojects/packagefiles/yajl/meson.build
```diff
@@ -1,4 +1,9 @@
 project('YetAnotherJSONParser', 'c', version: '2.1.0')
+
+if get_option('buildtype') != 'static'
```
xclaesse, line 3: Do you mean `default_library`?

neheb: Yep. Turns out there are 12 more places in here with the same mistake.

neheb: Fixed.

neheb:
Anything else?

xclaesse:
@neheb sorry I'm late at reviewing this. Could you update the PR to remove it from broken_windows list please? Other than that LGTM.

neheb:
Done.

PR #599, opened 2022-08-26
utf8proc: update to 2.7.0

neheb:
Try to fix MSVC compilation.

The library had no c files....


subprojects/packagefiles/utf8proc/meson.build
```diff
@@ -1,22 +1,19 @@
...
-    utf8proc_args += '-DUTF8PROC_STATIC'
+if get_option('default_library') != 'static' and host_machine.system() == 'windows'
+  add_project_arguments('-DUTF8PROC_EXPORTS', language: 'c')
 endif
 
 utf8proc_lib = library('utf8proc',
-    include_directories: include_directories('.'),
-    c_args: ['-DUTF8PROC_EXPORTS']
+  'utf8proc.c',
+  include_directories: include_directories('.'),
```
jpakkane, line 13: This is never needed. The current source and build dirs are in include dirs by default.

PR #601, opened 2022-08-28
glew: clean up

neheb:


neheb:
Requires: https://github.com/mesonbuild/wrapdb/pull/521

neheb:
```
 ERROR: test_releases (__main__.TestReleases) (step='check_meson_version', name='libglvnd')
  File "D:\a\wrapdb\wrapdb\tools\sanity_checks.py", line 130, in check_meson_version
    subproject, = [subproj for subproj in project_info["subprojects"] if subproj["name"] == name]
ValueError: not enough values to unpack (expected 1, got 0)
```

sounds like an issue where I have two packages in this PR.

subprojects/packagefiles/glew/meson.build
```diff
@@ -23,45 +23,45 @@ project(
...
 )
 
 cc = meson.get_compiler('c')
 host_system = host_machine.system()
 
-user_args = []
+dep_gl = dependency('gl', required: false)
+if not dep_gl.found()
+  dep_gl = dependency('opengl')
+endif
```
xclaesse, line 35: That's why we have `ependency('gl', 'opengl')` now :)

neheb: Hmm?

xclaesse: We can give multiple names to `dependency()` and it will return the first found. But that requires fairly new Meson version, I think @eli-schwartz wants to keep wrapdb compatible with older Meson version. So let's ignore this.

neheb:
There are plenty of wraps that require 0.60 or newer. What's the minimum version?

neheb:
Let's just fold into https://github.com/mesonbuild/wrapdb/pull/521

PR #603, opened 2022-08-28
SDL2: fixes

Jan200101:
- define a test library to be used when compiling the tests (talked about in #574)
- Fix dependencies not outwardly being defined (reported in #521)

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -408,9 +408,7 @@
       "libxt-dev",
       "libxv-dev",
       "libxxf86vm-dev",
-      "libgles2-mesa-dev",
-      "pkg-config",
-      "wayland-protocols"
+      "libgles2-mesa-dev"
     ]
   },
   "sds": {
--- a/ci_config.json
+++ b/ci_config.json
@@ -40,8 +40,7 @@
     "sdl2_image",
     "tinyxml2",
     "unit-system",
-    "zstd",
-    "sdl2"
+    "zstd"
   ],
   "cexception": {
     "build_options": [
@@ -409,6 +408,14 @@
       "libxv-dev",
       "libxxf86vm-dev",
       "libgles2-mesa-dev"
+    ],
+    "brew_packages": [
+      "libx11",
+      "libxcursor",
+      "libext",
+      "libxi",
+      "libxkbcommon",
+      "libxrandr"
     ]
   },
   "sds": {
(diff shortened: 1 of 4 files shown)


neheb:
So libglvnd as in that PR does not provide a gl deep but an opengl one. Would it make sense to fall back to opengl?

Jan200101:
> So libglvnd as in that PR does not provide a gl dep but an opengl one. Would it make sense to fall back to opengl?

there is technically a difference between libgl and libopengl, but SDL does not care about the details, so its best to rewrite it to accept both

neheb:
Another thing. I see in ci_config.json there's pkg-config and wayland-protocols. The former seems redundant and the latter has a wrap.

eli-schwartz:
While you're at it, I discovered that this wrap doesn't compile on my machine if I have libunwind installed -- it probes for headers and doesn't link to the pkg-config dependency or even a find_library().

Jan200101:
the libunwind problem has been fixed in the same way upstream does it.

Lets defer merging this for a bit so that more eyes can look at this and give feedback and so I can find a MacOS machine for #455

Jan200101:
rebased onto master to get the CI PR in
enabled basic BSD support
hopefully fix the MacOS CI? I have no idea why the build_all workflow wasn't working since the PR checks always worked fine and I couldn't reproduce this in a VM so fingers crossed?

If anything else needs doing throw it

neheb:
@xclaesse what do you think?

Jan200101:
> enabled basic BSD support
Turns out, FreeBSD cannot compile the tests correctly
so I need to figure that out

eli-schwartz:
The CI for this never ran, because it didn't update the sdl2 version. Is this actually known to work? Was it intentionally merged as a source-only update without releasing a new version?

Jan200101:
> The CI for this never ran, because it didn't update the sdl2 version.

If this is in regards to the "fix Darwin CI" commit, this was to reference #455.
CI in previous PRs always worked on MacOS and only the build_all solution failed.
I couldn't reproduce this in a MacOS VM but for some reason the Runner corrupt packages preinstalled or had some kind of configuration that wasn't reproducable on my end

> Is this actually known to work?
- Linux yes
- Windows No idea
- (Free)BSD No
- MacOS No idea

eli-schwartz:
It's actually in regard to "I thought this was still a work in progress PR and was surprised to see it merged".

PR #605, opened 2022-08-28
imgui-sfml: accept opengl as a dependency

neheb:


neheb:
Let's just fold into https://github.com/mesonbuild/wrapdb/pull/521

Issue #595, opened 2022-08-26
Error during build with libwebsockets v4.0.21


lorenzo-arena:
I'm trying to build a project with libwebsockets using the latest v4.0.21 wrap; I added the dependency in my *meson.build* file using

```
libwebsockets_proj = subproject('libwebsockets', default_options: ['warning_level=0', 'werror=false'])
libwebsockets_dep = libwebsockets_proj.get_variable('libwebsockets_dep')
```

However, when trying to build I get this error:

```
../subprojects/libwebsockets-v4.0.21/include/libwebsockets.h:42:10: fatal error: lws_config.h: No such file or directory
   42 | #include "lws_config.h"
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

I can see that the header file is being correctly generated during the build:

```
~ find . -iname "lws_config*"
./subprojects/libwebsockets-v4.0.21/lws_config_private.h
./subprojects/libwebsockets-v4.0.21/lws_config.h
```

neheb:
Needs to be added in declare_dependency.

neheb:
https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/libwebsockets/meson.build#L347 needs '.' included.

PR #607, opened 2022-08-29
libwebsockets: add missing include path for lws_config.h

lorenzo-arena:
This fixes https://github.com/mesonbuild/wrapdb/issues/595; an additional include is necessary to be able to find the generated `lws_config.h` file.


neheb:
Needs a bump in releases.json

lorenzo-arena:
I did a fixup on the original commit

PR #608, opened 2022-08-29
pcre: Force building static library only

xclaesse:
DLLs are broken on Windows.


xclaesse:
CC @neheb @eli-schwartz

eli-schwartz:
NACK.

Please use the same code that existed prior to the recent PR. Or I can fix it later today.

xclaesse:
You mean restoring the option we had? Was that really useful?

eli-schwartz:
Hold on, I may be confusing this with another wrap that had a similar but not identical workaround for the same reason.

xclaesse:
TBH, I think you're right that we tried to make 2 changes at once and it should have been splitted. If someone cares about fixing DLLs I'm happy to accept PRs. In the meatime this PR should fix the problem by making static libs by default again.



> Please use the same code that existed prior to the recent PR

Done.

eli-schwartz:
Sorry, I misremembered slightly, and while the build_target trick is what I meant, keying off of a custom option wasn't.

Instead, we should key off of this:
https://github.com/mesonbuild/wrapdb/blob/97bac4a0292ef0f342574e43381f12d0916a5e49/subprojects/packagefiles/protobuf/meson.build#L5
or this:
https://github.com/mesonbuild/wrapdb/blob/97bac4a0292ef0f342574e43381f12d0916a5e49/subprojects/packagefiles/liblangtag/meson.build#L86-L94

Notice that we only force it to static on platforms we don't expect it to work shared. Otherwise, we set the libtype to "library" and have it respect -Ddefault_library.

The previous code here could actually have been replaced by `default_options: ['default_library=static']` but it's best to guarantee correctness on Windows and respect preferences elsewhere.

neheb:
The issue here seems to be pcre not exporting 1 or more symbols. Can this be worked around with a def file?

eli-schwartz:
Upstream doesn't have one, and I'm not convinced we should be adding lots of our own all over the place instead.

subprojects/packagefiles/pcre/meson_options.txt
```diff
@@ -1,3 +1,5 @@
+option('shared_lib', type : 'boolean', value : false)
```
neheb, line 1: why?

neheb:
CMake build:

```
$ strings libpcre.dll | grep pcre_free
pcre_free
pcre_free_study
pcre_free_substring
pcre_free_substring_list
.refptr._pcre_free
_pcre_free
_pcre_free_substring_list
_pcre_free_substring
_pcre_free_study
```

MSYS2 build
```
$ strings libpcre.dll | grep pcre_free
pcre_free
pcre_free_study
pcre_free_substring
pcre_free_substring_list
pcre_free_substring_list
pcre_free_substring
pcre_free
pcre_free_study
.refptr._pcre_free
_pcre_free
_pcre_free_substring_list
_pcre_free_substring
_pcre_free_study
```

hmmmmmmm.

neheb:
The versiondata config is completely wrong.

subprojects/packagefiles/pcre/meson_options.txt
xclaesse: Just reinstating what we had before.

xclaesse: Ok, removed again, doing what @eli-schwartz suggested above.

xclaesse:
> The issue here seems to be pcre not exporting 1 or more symbols. Can this be worked around with a def file?

Upstream does have the dllexport macros, so in theory it should just work without a def file... but I did not understand why it's broken.

> host_machine.system() == 'windows' ? 'static_library' : 'library' 

Nice trick, doing something like that now.

neheb:
I think I know why it's broken. HAVE_CONFIG_H is not being defined...

xclaesse:
> I think I know why it's broken. HAVE_CONFIG_H is not being defined...

It should: `compile_args = ['-DHAVE_CONFIG_H']`

xclaesse:
oh, but there are C++ files too it seems.

neheb:
Yeah that's not it. I can't reproduce with MSYS2 either.

neheb:
If I had to guess, pcre_globals.c is the issue:

```
#if defined _MSC_VER || defined  __SYMBIAN32__
static void* LocalPcreMalloc(size_t aSize)
  {
  return malloc(aSize);
  }
static void LocalPcreFree(void* aPtr)
  {
  free(aPtr);
  }
PCRE_EXP_DATA_DEFN void *(*PUBL(malloc))(size_t) = LocalPcreMalloc;
PCRE_EXP_DATA_DEFN void  (*PUBL(free))(void *) = LocalPcreFree;
PCRE_EXP_DATA_DEFN void *(*PUBL(stack_malloc))(size_t) = LocalPcreMalloc;
PCRE_EXP_DATA_DEFN void  (*PUBL(stack_free))(void *) = LocalPcreFree;
PCRE_EXP_DATA_DEFN int   (*PUBL(callout))(PUBL(callout_block) *) = NULL;
PCRE_EXP_DATA_DEFN int   (*PUBL(stack_guard))(void) = NULL;

#elif !defined VPCOMPAT
PCRE_EXP_DATA_DEFN void *(*PUBL(malloc))(size_t) = malloc;
PCRE_EXP_DATA_DEFN void  (*PUBL(free))(void *) = free;
PCRE_EXP_DATA_DEFN void *(*PUBL(stack_malloc))(size_t) = malloc;
PCRE_EXP_DATA_DEFN void  (*PUBL(stack_free))(void *) = free;
PCRE_EXP_DATA_DEFN int   (*PUBL(callout))(PUBL(callout_block) *) = NULL;
PCRE_EXP_DATA_DEFN int   (*PUBL(stack_guard))(void) = NULL;
#endif
```

sounds like some bug with earlier MSVC versions.

subprojects/packagefiles/pcre/meson.build
```diff
@@ -65,14 +65,23 @@ foreach t : [ 'long long', 'unsigned long long' ]
   cdata.set('HAVE_@0@'.format(t.underscorify().to_upper()), cc.has_type(t))
 endforeach
 
+if host_machine.system() == 'windows'
```
neheb, line 68: I'd change this to get_argument_syntax() == 'msvc'. MinGW works fine.

neheb:
https://github.com/microsoft/vcpkg/pull/25671/commits/5a76c5681a3f0530e5ffc578834efb52f9e07188 is supposedly a fix for shared compilation. Can't say I see how.

subprojects/packagefiles/pcre/meson.build
xclaesse: done

xclaesse:
Let's merge this fix for now we can always try to fix msvc shared lib later.

PR #609, opened 2022-08-30
add: indicators and tabulate, update: doctest

rtgiskard:


neheb:
You need entries in releases.json

xclaesse:
Please split into 1 PR per wrap, I don't think the CI will be able to create new releases when there is multiple wrap updates.

subprojects/packagefiles/indicators/meson.build
```diff
@@ -0,0 +1,16 @@
...
+    'cpp',
+    version : '2.2',
+    license : 'MIT'
+)
+
+install_headers([
+    'single_include/indicators/indicators.hpp',
+], subdir: 'indicators')
+
+includes = include_directories(['include'])
```
xclaesse, line 12: This should be `single_include` too I think, otherwise projects that uses indicators as subproject won't include same header than when using installed indicators.

subprojects/packagefiles/tabulate/meson.build
```diff
@@ -0,0 +1,16 @@
...
+    'cpp',
+    version : '1.4',
+    license : 'MIT'
+)
+
+install_headers([
+    'single_include/tabulate/tabulate.hpp',
+], subdir: 'tabulate')
+
+includes = include_directories(['include'])
```
xclaesse, line 12: likewise

releases.json
```diff
@@ -253,6 +253,7 @@
       "doctest"
     ],
     "versions": [
+      "2.4.9-1",
```
xclaesse, line 256: Needs to add indicators and tabulate in this file too

rtgiskard:
Sorry, fixed the above issue, reopen

PR #611, opened 2022-08-31
ci: Blacklist remaining failing wraps in build-all

xclaesse:
Most Windows failures are because of symbol visibility, the CI now catch
that bug because it tries to install .lib file that does not get created
when the DLL has no public API.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -9,24 +9,42 @@
   "broken_linux": [
     "arduinocore-avr",
     "openh264",
-    "protobuf-c", "zstd"
+    "protobuf-c",
+    "zstd"
   ],
   "broken_windows": [
     "arduinocore-avr",
     "google-woff2",
     "libgpiod",
-    "mocklibc", "openh264",
-    "protobuf-c", "termbox","unit-system",
-    "zstd"
+    "mocklibc",
+    "openh264",
+    "protobuf-c",
+    "termbox",
+    "unit-system",
+    "zstd",
+    "emilk-loguru",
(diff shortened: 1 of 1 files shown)


ci_config.json
```diff
@@ -9,24 +9,42 @@
...
+    "termbox",
+    "unit-system",
+    "zstd",
+    "emilk-loguru",
+    "libexif",
+    "ludocode-mpack",
+    "m4",
+    "miniz",
+    "tinyply",
+    "yajl"
```
neheb, line 31: This has a PR for it

ci_config.json
```diff
@@ -9,24 +9,42 @@
...
-    "protobuf-c", "termbox","unit-system",
-    "zstd"
+    "mocklibc",
+    "openh264",
+    "protobuf-c",
+    "termbox",
+    "unit-system",
+    "zstd",
+    "emilk-loguru",
+    "libexif",
```
neheb, line 26: This has a PR for it.

ci_config.json
```diff
@@ -9,24 +9,42 @@
...
-    "mocklibc", "openh264",
-    "protobuf-c", "termbox","unit-system",
-    "zstd"
+    "mocklibc",
+    "openh264",
+    "protobuf-c",
+    "termbox",
+    "unit-system",
+    "zstd",
+    "emilk-loguru",
```
neheb, line 25: This has a PR for it.

neheb:
These should be sorted alphabetically I think.

PR #613, opened 2022-09-02
fuse: clean up meson build

neheb:


eli-schwartz:
IIRC this was mainly copied from the fuse3 branch as a backport for people stuck with old versions of fuse. Not sure how worth it it is to rewrite that.

I'm especially extremely skeptical of using extremely recent features of Meson such as the new custom dependencies. The main purpose of *old* versions of fuse is probably for people on old systems with old Meson, and either way this doesn't even declare the minimum required version of Meson.

neheb:
Makes sense. There are libfuse3 typos in here.

The dependency stuff doesn't seem like a big deal to me. Iconv is not required. Maybe dl is.

xclaesse:
Maybe update the wrap to upstream fuse3, those who are stuck with old fuse2 can still use old wraps.

neheb:
fuse 2 and 3 use different pkgconfig names. I don't see why they should be mutually exclusive.

xclaesse:
> fuse 2 and 3 use different pkgconfig names. I don't see why they should be mutually exclusive.

oh, should be different wraps then, indeed.

neheb:
@eli-schwartz added fallbacks for older meson versions.

PR #615, opened 2022-09-03
lz4: try to fix MSVC compilation

neheb:


subprojects/packagefiles/lz4/meson.build
```diff
@@ -383,4 +383,15 @@ project(
   meson_version: '>=0.49.0'
 )
 
+if host_machine.system() == 'windows' and get_option('default_library') != 'static'
+  add_project_arguments('-DLZ4_DLL_EXPORT=1', language: 'c')
```
anarazel, line 387: I suspect this might cause problems for the binaries inside the wrap - if DLZ4_DLL_EXPORT is set, they will see declspec(dllexport) instead of declspec(dllimport).  Not sure this can be fixed without patching the contents of the lz4 meson.build.

neheb: I tested this under MSYS2. Program compiles and functions fine.

anarazel: That probably doesn't guarantee it working with msvc, as gcc is a lot more lenient about exporting/importing symbols. But I did just test it with msvc by patching the meson.build with the changes here, and it indeed seems to fix it.

subprojects/packagefiles/lz4/meson.build
```diff
@@ -383,4 +383,15 @@ project(
   meson_version: '>=0.49.0'
 )
 
+if host_machine.system() == 'windows' and get_option('default_library') != 'static'
```
anarazel, line 386: Probably worth adding a comment about why this is here?

neheb: Added.

subprojects/packagefiles/lz4/meson.build
neheb: Great.

PR #618, opened 2022-09-03
libXinerama: add

neheb:

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -317,6 +317,12 @@
       "libudev-dev"
     ]
   },
+  "libxinerama": {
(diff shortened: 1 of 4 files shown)


jpakkane:
Not to be a bit of a drag, but is this really something people actually need? On platforms that have X, this should always be provided by the system and on others (basically Windows?) nobody uses X so it is not really needed.

neheb:
From my point of view, wraps are for fallbacks when the actual package is missing. Note that in ci_config.json these have to be installed.

neheb:
Replaced with https://github.com/mesonbuild/wrapdb/pull/626

PR #620, opened 2022-09-06
tinysfm: add wrap

stephanlachnit:
for https://github.com/digint/tinyfsm

### DIFF
--- a/subprojects/packagefiles/tinyfsm/meson.build
+++ b/subprojects/packagefiles/tinyfsm/meson.build
@@ -1,0 +1,56 @@
+project('tinyfsm', 'cpp',
+  version: '0.3.3')
+
+tinyfsm_dep = declare_dependency(include_directories: include_directories('include'))
+
+# build examples as compile tests
+build_examples = get_option('build_examples')
+
+# elevator example
+ex_elevator_exe = executable('elevator',
+  ['examples/elevator/elevator.cpp', 'examples/elevator/main.cpp', 'examples/elevator/motor.cpp'],
+  include_directories: 'examples/elevator',
+  dependencies: tinyfsm_dep,
+  build_by_default: build_examples)
+
+# debugging_switch example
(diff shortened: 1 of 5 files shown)


stephanlachnit:
```
Found ninja-1.10.2 at C:\ProgramData\Chocolatey\bin\ninja.EXE
WARNING: msvc does not support C++11; attempting best effort; setting the standard to C++14

ERROR: Fatal warnings enabled, aborting

A full log can be found at D:\a\wrapdb\wrapdb\_build\meson-logs\meson-log.txt
```

Proposed solution?

subprojects/packagefiles/tinyfsm/meson.build
```diff
@@ -0,0 +1,56 @@
+project('tinyfsm', 'cpp',
+  version: '0.3.3',
+  default_options: ['cpp_std=c++11'])
+
+tinyfsm_dep = declare_dependency(include_directories: include_directories('include'))
+
+# build examples as compile tests
```
eli-schwartz, line 7: If these are `build_by_default: false` then they won't be built in CI. :D I wonder what the solution there is, hmm...

If they can be run successfully as a smoketest, then they could be added as `test()`s and then `ninja test` will build them I guess.

neheb:
C++14

eli-schwartz (commented):
The fatal warning would be fixed by https://github.com/mesonbuild/meson/pull/10582

Although you could just set it to c++14 I suppose.

subprojects/packagefiles/tinyfsm/meson.build
stephanlachnit: They are interactive, so no. I guess a `build_examples` option (disabled by default) is the best option here...

stephanlachnit: Added a `build_examples` option

stephanlachnit:
@eli-schwartz good to merge now?

PR #621, opened 2022-09-07
SDL2: speedbump

Jan200101:
Quick follow up to patch things up left open by #603

- BSD support removed again
  the libary builds just fine but I cannot get the tests to compile against the SDL libary.
  Upon digging further the upstream CMakeLists.txt is also broken on FreeBSD and they instead rely on the old-school configure to get things working.
  2.24.0 has (seemingly?) introduced BSD specific bugs that wouldn't make it work either way
- Tag release
  finally


eli-schwartz:
CI fails because a brew package in the previous PR was misspelled.

jpakkane:
LGTM

neheb:
@Jan200101 some of the CI packages got added to wrapdb recently. Please remove them.

Jan200101:
Hmm
the fact libxcursor is available as a wrap could explain why the CI was failing to begin with.
I'll have to look into it

neheb:
libxcursor wrap is a recent addition.

jpakkane:
This is the one case where I would advise against using wrap dependencies. SDL 2 is a bit special and it is expected to link against "system libs" or at least that is how 99% of all users are going to use it. Thus it would make sense for us to test the most common case.

Jan200101:
I think it would be cool to move all dependencies to wraps if possible

but as shown with libxcursor this can very well cause hard to hard to debug issues that only occur when certain wraps are combined.

The CI should probably continue depending on system libraries to ensure it all works as expected (at least the slice that the CI compiles), the other wraps should practically be a drop-in solution if they work correctly

neheb:
Hmmm? This compiles fine when subprojects are used. My post was about Wayland.

The macOS failures with build all are unrelated.

jpakkane:
All green now, so merging. If there are still issues to fix, please file new MRs.

PR #623, opened 2022-09-07
freetype: Provide both freetype and freetype2 names

xclaesse:
Harfbuzz lookup the freetype2 dependency like that:
```
freetype_dep = dependency(cpp.get_argument_syntax() == 'msvc' ?
'freetype' : 'freetype2', ...)
```
The reason is "freetype" is the CMake name. This causes meson to not lookup for freetype2 fallback when using MSVC.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -421,9 +421,11 @@
   },
   "freetype2": {
     "dependency_names": [
-      "freetype2"
+      "freetype2",
+      "freetype"
     ],
     "versions": [
+      "2.12.1-2",
       "2.12.1-1",
       "2.12.0-1",
       "2.11.1-1",
--- a/subprojects/freetype2.wrap
+++ b/subprojects/freetype2.wrap
@@ -6,3 +6,4 @@ source_hash = 4766f20157cc4cf0cd292f80bf917f92d1c439b243ac3018debf6b9140c41a7f
 
 [provide]
 freetype2 = freetype_dep
+freetype = freetype_dep


xclaesse:
oh, they did the same with ICU: https://github.com/harfbuzz/harfbuzz/blob/main/meson.build#L96.

Made another similar PR: https://github.com/mesonbuild/wrapdb/pull/625

xclaesse:
Note that those issues can be solved by doing `dependency('freetype', 'freetype2')` since Meson 0.60, but they depend on older version.

eli-schwartz:
IMO this is wrong and should be reverted. It now claims to provide a fallacious name that must not be used for portability reasons.

The correct solution is for badly behaved projects that refuse to use pkg-config and don't support recent meson versions, to manually specify a fallback: kwarg.

xclaesse:
I don't see what's wrong with providing both names the project uses, why is pkgconfig more right than cmake? One could even argument that freetype upstream is primarily a CMake project, so the pkgconfig name is wrong...

eli-schwartz:
The name comes from CMake's own vendored FindFreetype.cmake, it's not distributed by the freetype project at all.

eli-schwartz:
I would also be intimidated to claim:

> One could even argument that freetype upstream is primarily a CMake project, so the pkgconfig name is wrong...

Considering that the install instructions for freetype2 talk about running autogen.sh and `./configure`, and note ahead of time:

> For builds  with `cmake` please check file `CMakeLists.txt';  this
> is a contributed file not directly supported by the FreeType team.

xclaesse:
> The name comes from CMake's own vendored FindFreetype.cmake, it's not distributed by the freetype project at all.

Ok, I was wrong indeed. IMHO it does not hurt to use both names, the project really is "freetype" (git repo, website).

eli-schwartz:
In this case it's probably harmless, but I'm still not required to like it. :p

Hopefully users of the wrap will use the pkg-config name rather than the extra wrap provides name in their `dependency()` lookups. ;)

PR #625, opened 2022-09-07
icu: Also provide ICU name

xclaesse:
That's the CMake name that harfbuzz lookup when using MSVC.

### DIFF
--- a/subprojects/icu.wrap
+++ b/subprojects/icu.wrap
@@ -6,6 +6,7 @@ source_hash = 67a7e6e51f61faf1306b6935333e13b2c48abd8da6d2f46ce6adca24b1e21ebf
 patch_directory = icu
 
 [provide]
+ICU = icuuc_dep
 icu-uc = icuuc_dep
 icu-io = icuio_dep
 icu-i18n = icui18n_dep
--- a/releases.json
+++ b/releases.json
@@ -656,6 +656,7 @@
   },
   "icu": {
     "dependency_names": [
+      "ICU",
       "icu-uc",
       "icu-io",
       "icu-i18n"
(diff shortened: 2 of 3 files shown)


xclaesse:
See how harfbuzz lookup for ICU: https://github.com/harfbuzz/harfbuzz/blob/main/meson.build#L96

xclaesse:
CI fails because of https://github.com/mesonbuild/meson/pull/10789

eli-schwartz:
I don't like this change, nor the other one. The correct solution is for harfbuzz to manually specify an old style fallback: kwarg.

subprojects/icu.wrap
```diff
@@ -6,6 +6,7 @@ source_hash = 67a7e6e51f61faf1306b6935333e13b2c48abd8da6d2f46ce6adca24b1e21ebf
 patch_directory = icu
 
 [provide]
+ICU = icuuc_dep
```
eli-schwartz, line 9: This is outright broken, the cmake.org published FindICU.cmake module (not officially part of the icu project) exports ICU::uc, as well as ICU::io, ICU::i18n, so in order to use cmake lookups you need to pass a `modules:` kwarg, and there's a fairly good chance the user didn't want icu-uc specifically.

eli-schwartz (changes requested):

subprojects/icu.wrap
xclaesse: Hmm, they indeed pass a component: https://github.com/harfbuzz/harfbuzz/blob/main/meson.build#L96. Our `[provide]` mechanism cannot deal with components, so I guess there is no solution here :/

PR #630, opened 2022-09-09
rdkafka: improve MinGW support

neheb:
Added regcomp and libdl support.


eli-schwartz (commented):
@jcaesar more random breakage, I think.

subprojects/packagefiles/rdkafka/meson.build
```diff
@@ -1,14 +1,17 @@
-project('rdkafka', 'c', 'cpp', version: '1.9.2', default_options: ['cpp_std=c++17', 'c_std=c11'])
+project(
+  'rdkafka',
+  'c',
+  'cpp',
+  version: '1.9.2',
+  default_options: ['cpp_std=c++17', 'c_std=c11', 'warning_level=0'],
+  meson_version: '>=0.57.0'
```
eli-schwartz, line 7: AFAICT the most recent feature it uses is from 0.47.0, not 57. Typo?

neheb:
```
foreach c : conf.keys()
```

0.57.0

subprojects/packagefiles/rdkafka/meson.build
```diff
@@ -201,7 +208,7 @@ endif
 rdkafka = library('rdkafka', srcc, install: true, dependencies: rdkafka_deps)
 rdkafka_dep = declare_dependency(
   link_with: rdkafka,
-  include_directories: include_directories('include-workaround-meson'),
+  include_directories: '.',
```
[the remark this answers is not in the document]
neheb: hrm unfortunate.

subprojects/packagefiles/rdkafka/meson.build
eli-schwartz: Hmm, that code simply isn't used at all except on Windows? What's the issue here, why isn't this applicable elsewhere, and is there an upstream bug report for it?

neheb: yeah it's a windows only thing. There's the comment:

```
#upstream is broken and all config.h entries must be passed as compiler flags
```

What's actually happening for Windows is that all of the config options are included with a static win32_config.h. And to support dynamically setting options, it passes everything as compiler flags. No idea why this decision was made.

subprojects/packagefiles/rdkafka/meson.build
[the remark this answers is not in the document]
jcaesar: Is `include_directories: '.'` necessary for anything?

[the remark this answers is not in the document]
eli-schwartz: No, because all it does is allow you to `#include <src/........>` and compile against it using this declare_dependency.

Which is broken because it doesn't provide compatibility with `#include <librdkafka/.......>`

PR #631, opened 2022-09-09
sanity checks: verify that a wrap is actually unsupported

eli-schwartz:
Historically, we have a setting "linux_only" for wraps which are software that targets Linux APIs and doesn't intend to compile elsewhere. After adding more CI coverage, we changed this to "build_on" as an operating system key. The intention is that we don't need to fail CI for software that is performing within its expected parameters.

Unfortunately, this has been repeatedly misused for wraps that are currently broken and need to be fixed, which defeats the point of having CI at all. And these PRs get merged when they shouldn't be despite review objections.

Stop skipping these wraps. We should verify that they actually fail, and why. This has a couple benefits:

- if a wrap is disabled because it is missing dependencies, that's not ideal, but tolerable. Still, disabling the CI for it means that if the dependency becomes satisfiable, we still skip CI. Now, we can raise an error that it is satisfiable and can configure properly, and needs to be unblocked in ci_config.json

- if a wrap is disabled because the software isn't coded for a platform, mysterious configure failures are a bad user experience, it's far better to error out with `error('foobar is not supported on Windows')`

- if a wrap is disabled because, say, it compiles successfully on Windows but produces broken DLLS that have no symbol table or implib, then something went badly wrong and that's not a good reason to disable it

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -261,10 +261,11 @@ class TestReleases(unittest.TestCase):
         system = platform.system().lower()
         ci = self.ci_config.get(name, {})
         # kept for backwards compatibility
+        expect_working = True
         if ci.get('linux_only', False) and not is_linux():
-            return
+            expect_working = False
         elif not ci.get('build_on', {}).get(system, True):
-            return
+            expect_working = False
 
         options = ['-Dpython.install_env=auto', f'-Dwraps={name}']
         options.append('-Ddepnames={}'.format(','.join(deps or [])))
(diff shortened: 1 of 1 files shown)


neheb:
this needs work. When faced with a failing package, the error is:

```
wrap failed to configure due to bugs in the wrap, rather than due to being unsupported
```

when getting rid of

```
error('libfuse does not support Windows.\n' + 'Take a look at http://www.secfs.net/winfsp/ instead')
```

and some other stuff, this is thrown:

```
Exception: Wrap successfully configured but was expected to fail
```

xclaesse:
> these PRs get merged when they shouldn't be despite review objections.

And then you merge it with no review...

eli-schwartz:
@neheb yeah, the libfuse case is interesting, because it's multiline. Hopefully fixed in #635

Likewise https://gist.github.com/neheb/824cc909162f7b53e88d531e69e152fa is because test_build_all is too forgettable :o and we can just handle exceptions as errors. subtests usually give us that for free, but test_build_all doesn't use those.

PR #632, opened 2022-09-10
htslib: update to 1.16

neheb:


neheb:
```
ERROR: test_releases (__main__.TestReleases) (step='check_new_release', name='htslib')
  File "D:\a\wrapdb\wrapdb\tools\sanity_checks.py", line 213, in test_releases
  File "D:\a\wrapdb\wrapdb\tools\sanity_checks.py", line 296, in check_new_release
    raise Exception('Wrap successfully configured but was expected to fail')
Exception: Wrap successfully configured but was expected to fail
```

@eli-schwartz

neheb:
```
cc = meson.get_compiler('c')
if cc.get_argument_syntax() == 'msvc'
  error('MSVC is unsupported. Only MinGW.')
endif
```

even though I added that, MinGW will fail with the above error.

neheb:
ping @jpakkane

PR #634, opened 2022-09-11
libssh: add mingw support

neheb:


neheb:
ehhh ended up adding MSVC support too.

ping @ilsme

jpakkane:
LGTM but it would be nice to get an actual usage report.

neheb:
Should be the same. I didn't touch any include_directories.

Only real question is whether or not to have an explicit wincng option. It's Windows only anyways.

neheb:
further examination reveals libssh2 only uses openssl's libcrypto. libssl is not used.

Wonder if dependency('openssl') should be removed in that case...

neheb:
my CI test reveals

```
libssh2| Run-time dependency libgcrypt found: YES 1.10.1
```

is broken.

neheb:
CI currently tests three backends: openssl, libgcrypt, and wincng. Cool.

mbedtls was tested locally to work as well.

neheb:
ping @jpakkane

PR #641, opened 2022-09-14
flac: update to 3.4.0

neheb:
Add flac and metaflac binary.


neheb:
Tested on ARM32 as well. Compiles.

PR #585, opened 2022-08-24
Check that programs and dependencies are effectively provided

xclaesse:

PR #642, opened 2022-09-18
Update glib to 2.74.0

xclaesse:


xclaesse:
Hmm, the CI issue is interesting: `WARNING: Subproject 'glib' did not override 'gio-windows-2.0' dependency and no variable name specified`. When building on Linux we don't have it of course... That's a regression caused by https://github.com/mesonbuild/wrapdb/pull/585.

eli-schwartz:
So huh, that actually worked in my original check which only validated `foo = foo_dep` but not dependency_names?

Also I wonder how this works out for situations where people use dependency('platform-specific', required: false, allow_fallback: true) and just let it be not-found on other platforms. At least the logging would be suboptimal, trying to remember if the entire subproject gets marked as bad...

No idea what the solution is there. Platform specific dependencies make my head hurt. :(

xclaesse:
> So huh, that actually worked in my original check which only validated `foo = foo_dep` but not dependency_names?

In this specific case, yes. But we could still have platform specific deps with name=var syntax.

> Also I wonder how this works out for situations where people use dependency('platform-specific', required: false, allow_fallback: true) and just let it be not-found on other platforms.

They would get a well deserved warning, unless the project does `meson.override_dependency('platform-specific', dependency('', required: false))` in which case it will return not-found properly.

> No idea what the solution is there. Platform specific dependencies make my head hurt. :(

I see potential solutions:
- Do you have platform specific (or any optional) dependencies in `[provide]` section. That's what I do in this PR but it's a bit unfortunate because it really is just a WrapDB specific issue.
- Have a dedicated `optional_dependency_names` field in wrap files that Meson would always return not-found if they are not overridden. This could be used for example for `cairo-gobject` that is only overriden in the case cairo is built with glib support.
- Add an override in WrapDB ci_config.json that tells some deps should not be checked, because it is mostly a wrapdb ci problem that should be dealt at the wrapdb level.

xclaesse:
Ohhh, this glib 2.74 release brings another interesting issue: when glib is a subproject (like in wrapdb CI) Meson git clone gvdb sub-subproject instead of using the one already included into glib release tarball (it's a git submodule).

xclaesse:
glib can simply delete gvdb.wrap and that fix that problem: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2906.

eli-schwartz:
That doesn't make sense, meson is supposed to defer to a local copy of it exists.

xclaesse:
> That doesn't make sense, meson is supposed to defer to a local copy of it exists.

Meson checks for local copy in wrapdb/subprojects/gvdb and not wrapdb/subprojects/glib-2.74/subprojects/gvdb. That's a Meson bug that we can luckily work around by just deleting the useless wrap file.

eli-schwartz:
Why does it only have that issue when a wrap file exists? That's bizarre.

Shouldn't this be creating a wrap-redirect?

xclaesse:
@eli-schwartz see https://github.com/mesonbuild/meson/pull/10828.

neheb:
Is this ready?

xclaesse:
> Is this ready?

I'm a bit unsure what to do here. The problem is when glib is a subproject (which is the main purpose of WrapDB) it will git clone gvdb instead using the copy already included in glib release tarball.

That's a Meson bug that has been fixed in master, not released yet. I made a [workaround in GLib](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2906) but it's not merged yet.

Question, should we wait for Meson 0.64? Wait for glib 2.74.1 with the workaround? Go ahead now since it works even if it does a useless git clone?

eli-schwartz:
It works, even if meson git master has a performance optimization.

PR #644, opened 2022-09-19
libtirpc: add

neheb:


subprojects/packagefiles/libtirpc/meson.build
```diff
@@ -0,0 +1,193 @@
+project('libtirpc', 'c', version: '1.3.3')
```
jpakkane, line 1: We should probably add the `license` key to all things in WrapDB. Especially new entries. Based on the project page this one is BSD.

neheb: Good point.

neheb: @eli-schwartz should a license check be added to sanity_checks.py?

The reference says license should be an SPDX identifier.

PR #649, opened 2022-09-21
README: update the release section

dabrain34:
Add a small description on how to release a
new version of a wrap.


eli-schwartz:
We should unify this with "add new wrap" somehow so that the same information isn't listed twice.

dabrain34:
Up to my recent experience, it has been a long time that I contributed to the openjp2 wrap. And I just wanted to find back the right information from the README (which I didnt) on how to create a new release from an existing wrap. So I guess for the next time I would appreciate to just find the section.

Issue #103, opened 2021-06-08
Remove vulkan-validationlayers


stephanlachnit:
If you rework wrapdb, please remove vulkan-validationlayers (at least from `releases.json`). See https://github.com/mesonbuild/vulkan-validationlayers/issues/3.

PR #661, opened 2022-09-24
vulkan-{headers,validationlayers}: update to 1.3.229

Kelvium:
I've also added `[provide]` section in .wraps.


neheb:
I think the actual provide should be Vulkan-Headers based on https://github.com/KhronosGroup/Vulkan-Headers/pull/207

I know @eli-schwartz isn't a big fan of adding provide to packages that don't export a pkgconfig or cmake target file.

eli-schwartz:
The problem is that it literally doesn't work in any way shape or form for any users anywhere, because if there isn't *some* form of dependency lookup provided by the actual actual software, then people ***cannot*** use `dependency()` under any circumstances, and therefore as a result they cannot get it from a wrap fallback either.

neheb:
Nvm the comment says

```
searched for by find_package(VulkanHeaders)
```

jpakkane:
Since the linked issue seems to be stuck, how about dropping the provide entry from this, merging it and then adding the provide once upstream makes a decision?

Kelvium:
> Since the linked issue seems to be stuck, how about dropping the provide entry from this, merging it and then adding the provide once upstream makes a decision?

Okay :)

stephanlachnit:
Please do not update vulkan-validationlayers, it is essentially unusable. See https://github.com/mesonbuild/wrapdb/issues/103 and https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2331#issuecomment-731376649

PR #662, opened 2022-09-25
Update tomlplusplus to 3.2.0

Mabi19:
A little late, since I kept forgetting to do this for a while.


jpakkane:
No worries, thanks.

PR #667, opened 2022-09-30
libebml + libmatroska updates

neheb:


neheb:
hrm I thought the MSVC failure would be reproducible with MinGW if I kept all the declspec stuff the same. Guess not. That trick seems to work everywhere else though.

PR #669, opened 2022-10-03
libssh2: fix build with build all

neheb:
For some reason, there's no OpenSSL and it ends up finding mbedTLS for some reason. Unfortunately, it seems it's not compatible with mbedtls 3.0+. For now, just work around by making it the last tested backend.


neheb:
```
Package mbedtls-devel-2.28.0-3.fc37.x86_64 is already installed.
```

Fascinating. No wonder it works for me.

PR #671, opened 2022-10-04
imgui: update to 1.88

neheb:
Fixed compilation under MinGW.


neheb:
ping @stephanlachnit

ci_config.json
```diff
@@ -159,14 +159,6 @@
       "gettext"
     ]
   },
-  "imgui": {
-    "debian_packages": [
```
stephanlachnit, line 163: Why are these removed?

neheb: Not needed.

stephanlachnit: Ehm, yes but they provide a useful test* for the CI? Please add them back...

* i.e. that the dependency lookup and compilation of the files still work

neheb: IMO It's more useful to test other wraps.

Maybe edit CI to also test --wrap-mode=forcefallback

stephanlachnit: Ah, fair, glfw and sdl2 are also available as wrap. allegro5 is not though.

PR #675, opened 2022-10-04
Update cpp-httplib to 0.11.2

tmayoff:


releases.json
```diff
@@ -2248,4 +2249,4 @@
       "1.3.3-1"
     ]
   }
-}
+}
```
neheb, line 2252: no newline?

neheb:
Seems fine. Maybe squash all into one commit.

PR #678, opened 2022-10-05
workflows: install pkgconfiglite for Windows

neheb:
Prevents packages marked false for windows from failing.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -49,6 +49,7 @@ jobs:
       - name: Install packages
         run: |
           pip install git+https://github.com/mesonbuild/meson
+          choco install -y pkgconfiglite
 
       - uses: ilammy/msvc-dev-cmd@v1
       - name: Sanity Checks
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -40,6 +40,7 @@ jobs:
       - name: Install packages
         run: |
           pip install --pre meson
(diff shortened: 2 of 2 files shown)


xclaesse:
What does this fix exactly?

neheb:
Fixes this from throwing on Windows builds: https://github.com/mesonbuild/wrapdb/blob/master/tools/sanity_checks.py#L339

xclaesse:
Can you elaborate a bit how it fixes that? I fail to understand why pkg-config is needed on Windows at all, it will never find any dependency, does it?

neheb:
Example of such a failure: https://github.com/mesonbuild/wrapdb/actions/runs/3178610216/jobs/5180284621

xclaesse:
```
    "_comment": "libmicrohttpd and libcurl are not available with MSVC CI",
    "build_on": {
      "windows": false
    }
```

It is expected to not build on Windows, installing pkgconfiglite should not make any difference. If it does make a difference that's a bug we should investigate the root cause.

xclaesse:
Pretty sure that's another issue with https://github.com/mesonbuild/wrapdb/pull/631, Meson is right in failing with `ERROR: Pkg-config binary for machine 1 not found. Giving up.` but that PR does not parse correctly such error message.

@eli-schwartz could you check that please?
@neheb could you remove the couple `pkgconfiglite` you added into ci_config.json already?

eli-schwartz:
Check this PR ^^

PR #679, opened 2022-10-06
json-c: fix SIZEOF_SSIZE_T

emersion:
On my setup `SIZEOF_SSIZE_T` ends up being negative:

    #define SIZEOF_SSIZE_T -1

This happens because `<stdint.h>` is not enough to pull the
definition of `ssize_t`.

Instead, use `<sys/types.h>` as per the [POSIX spec][1].

While at it, also fix the same issue for `size_t`, which was pulled
in by chance. Switch to `<stddef.h>` which is ISO C.

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html


jpakkane:
The issue should probably go away by rebasing.

subprojects/packagefiles/json-c/config/meson.build
```diff
@@ -100,7 +100,14 @@ else
   check_types += 'ssize_t'
 endif
 foreach type : check_types
-  size = cc.sizeof(type, prefix: '#include <stdint.h>')
+  if type == 'size_t'
```
neheb, line 103: This whole section seems like it can be simplified by removing the loop and doing everything manually.

neheb: Something like

```
conf.set('SIZEOF_INT', cc.sizeof(int))
conf.set('SIZEOF_INT64_T', cc.sizeof(int64_t, prefix: '#include <stdint.h>'))
conf.set('SIZEOF_LONG', cc.sizeof(long))
conf.set('SIZEOF_LONG_LONG', cc.sizeof(long long))
conf.set('SIZEOF_SIZE_T', cc.sizeof(size_t, prefix: '#include <stddef.h>'))
sst = cc.get_argument_syntax() == 'msvc' ? '#include <BaseTsd.h>'  : '#include <sys/types.h>'
conf.set('SIZEOF_SSIZE_T', cc.sizeof('SSIZE_T', prefix: sst))
```

emersion: I've simplified, but kept the loop.

emersion:
Thanks!

PR #687, opened 2022-10-08
libccp4c: update to 8.0.0

neheb:


neheb:
ping @taw10

neheb:
MinGW fails because of missing header: https://github.com/ndevenish/libccp4/blob/main/ccp4/cmtzlib.c#L26 . Unfortunate.

taw10:
Thanks @neheb!  I did a quick test, and the updated wrap works for me.  I've uploaded the 8.0.0 tarball to make the fallback mirror URL work (it is not done automatically).  Neat that it now compiles on MSVC!  I'll do some more thorough testing shortly - perhaps some of the silencing of compiler warnings are not needed any more.

About MinGW: strange that there's a problem, because MinGW is supposed to be supported upstream (MSVC in contrast not).  Perhaps the policy changed.

neheb:
it's a silly oversight by the developer. stdint.h should be included unconditionally.

PR #688, opened 2022-10-08
flatbuffers: update to 2.0.8

neheb:


subprojects/packagefiles/flatbuffers/meson.build
```diff
@@ -1,25 +1,15 @@
...
 endif
 
 if cpp.get_argument_syntax() == 'gcc'
-    cpp_args += [
-        '-Wall',
-        '-pedantic',
-        '-Werror',
-        '-Wextra'
-    ]
+  cpp_args += ['-Wall', '-pedantic', '-Werror', '-Wextra']
```
jpakkane, line 12: These really should not be here. Meson has builtins for all of these and libraries should not enable `-Werror` unconditionally because it breaks downstream users that have e.g. newer compilers that have more checks that what the developers had available.

neheb: Yeah probably. I didn’t bother removing them. Will do so.

neheb: Fixed.

PR #689, opened 2022-10-09
fluidsynth: update to 2.3.0

neheb:


neheb:
@eli-schwartz so I tried cross compiling and got this:

```
fluidsynth| Run-time dependency OpenMP for c found: YES 4.5
```

libgomp is missing for mingw, which results in the error:

```
x86_64-w64-mingw32-g++: fatal error: cannot read spec file 'libgomp.spec': No such file or directory
```

PR #693, opened 2022-10-11
sdl2_net: update to 2.2.0

neheb:
ping @Jan200101


subprojects/packagefiles/sdl2_net/meson.build
```diff
@@ -1,35 +1,39 @@
...
+)
 
 sdl2_net_dep = declare_dependency(
-    link_with : [sdl2_net],
-    dependencies: plat_libs,
-    version : meson.project_version(),
-    include_directories : '.')
+  link_with: sdl2_net,
+  dependencies: plat_libs,
+  include_directories: '.',
```
xclaesse, line 38: you never need `'.'`

xclaesse:
Other than small nitpicking, LGTM

subprojects/packagefiles/sdl2_net/meson.build
neheb: it is for declare_dependency()

xclaesse: Ah right, indeed.

PR #695, opened 2022-10-11
cppzmq: fix catch2 dependency

neheb:
Not compatible with version 3.


neheb:
@xclaesse this is a build all fix.

PR #699, opened 2022-10-15
Feat/glad

Inerska:
Hey just included glad, that's the basic for OpenGL projects development imo.

**Glad** to created my first contribution in this incredible build system.

The tests passed locally


releases.json
```diff
@@ -491,6 +491,14 @@
       "1.14.1-1"
     ]
   },
+    "glad": {
+        "dependency_names": [
+            "glad"
+        ],
+        "versions": [
+            "v0.1.36"
```
xclaesse, line 499: You need to append `-1` to the version, that's the wrapdb revision number.

subprojects/packagefiles/glad/meson.build
```diff
@@ -0,0 +1,9 @@
+project('glad',
+    'c',
+    version: 'v0.1.36',
+    license: 'MIT'
+)
+
+glad_dep = declare_dependency(
+    include_directories: include_directories('.')
+)
```
xclaesse, line 9: From a quick look at the glad project, you need to actually generate all headers. Their cmake needs to be converted to Meson.

Inerska:
I fixed the version but I guess I won't be able to set up the instructions to extract the header files alone

PR #702, opened 2022-10-16
zlib: update to 1.2.13

neheb:


subprojects/packagefiles/zlib/meson.build
```diff
@@ -16,9 +16,9 @@ if cc.get_argument_syntax() == 'msvc'
 else
   # Don't spam consumers of this wrap with these warnings
   compile_args += cc.get_supported_arguments('-Wno-implicit-fallthrough', '-Wno-implicit-function-declaration')
-  if cc.get_id() == 'gcc' and host_machine.system() != 'windows'
-    vflag = '-Wl,--version-script,@0@/zlib.map'.format(meson.current_source_dir())
-    link_args += [vflag]
+  gnu_sym_ldflag = '-Wl,--version-script,@0@/zlib.map'.format(meson.current_source_dir())
+  if get_option('buildtype') != 'static' and cc.links('', name: '-Wl,--version-script', args: ['-shared', gnu_sym_ldflag])
```
eli-schwartz, line 20: You don't need to check the buildtype, link_args will simply do nothing whatsoever if the linker is never getting run?

eli-schwartz: Also what buildtype is this supposed to be, presumably default_library?

neheb: copy/paste from libsndfile. looks like that's wrong too...

neheb:
@jpakkane this is ready.

neheb:
wonder if it's better to use https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerget_supported_link_arguments actually...

PR #704, opened 2022-10-16
bdwgc: update to 8.2.2

neheb:


subprojects/packagefiles/bdwgc/meson.build
```diff
@@ -1,42 +1,48 @@
...
+  'obj_map.c',
+  'os_dep.c',
+  'ptr_chck.c',
+  'reclaim.c',
+  'specific.c',
+  'typd_mlc.c',
+)
+
+if host_machine.system() == 'sparc64'
+  gc_src += files('sparc_mach_dep.S')
```
neheb, line 36: I don't seem to be able to include this.

neheb: nvm got it.

PR #707, opened 2022-10-19
libdrm: update to 2.4.114

neheb:

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -169,6 +169,18 @@
       "darwin": false,
       "windows": false
     },
+    "_comment": "- Upstream has no idea how to use feature options",
+    "build_options": [
+      "libdrm:amdgpu=enabled",
+      "libdrm:etnaviv=enabled",
+      "libdrm:exynos=enabled",
+      "libdrm:freedreno=enabled",
+      "libdrm:intel=enabled",
+      "libdrm:nouveau=enabled",
+      "libdrm:omap=enabled",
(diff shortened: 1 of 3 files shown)


neheb:
MSVC and Darwin so called failures are fixed in latest git. Does not fail under Build all.

neheb:
this is ready.

jpakkane:
Is CI fixed now so that rebasing makes this go green?

ci_config.json
```diff
@@ -169,6 +169,18 @@
       "darwin": false,
       "windows": false
     },
+    "_comment": "- Upstream has no idea how to use feature options",
```
xclaesse, line 172: Could you elaborate a bit what's the issue? Did you report upstream?

xclaesse: IMHO, you should leave all options to `auto`, report upstream issue because it seems it disables some parts that should work by default. In ci-config.json we have that new section to tell some of deps provided are optional:
```
    "skip_dependency_check": [
      "libdrm_etnaviv",
      ...
    ]
```

neheb: Makes no sense. the .wrap file specifies a bunch of dependencies that must be enabled.

neheb:
CI passes now.

ci_config.json
xclaesse: At least report issue upstream. Saying here they don't understand feature option and do nothing about it is not helping anyone.

PR #710, opened 2022-10-21
uchardet: add wrap

Jan200101:
CI might or might not fail.

Tests are a bit ugly but upstream does it via globbing so there is little choice.
FS module is needed to check if a specific language has a specific charset unless someone decides to hardcode every single lang and charset combination.

Upstream decides to enable SSE2 specific options if possible to optimize the code.


subprojects/uchardet.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = uchardet-0.0.7
+source_url = https://github.com/freedesktop/uchardet/archive/refs/tags/v0.0.7.tar.gz
+source_filename = uchardet-0.0.7.tar.gz
+source_hash = 561db71ffe3b090da48cd17f441bb7976694db55edd9cc16a6bd6ffcb9e66e8f
+patch_directory = uchardet
+
+[provide]
+uchardet = uchardet_dep
+
```
neheb, line 10: extra whitespace. this is no longer needed/added.

Jan200101:
Hmmm... what is the best way to handle the VS error?

subprojects/packagefiles/uchardet/meson.build
```diff
@@ -0,0 +1,30 @@
...
+  meson_version: '>=0.53.0',
+  default_options: ['cpp_std=c++11'],
+)
+
+base_cpp_args = []
+
+cxx = meson.get_compiler('cpp')
+host_cpu_family = host_machine.cpu_family()
+
+use_sse2 = get_option('use_sse2').enabled() or get_option('use_sse2').auto()
```
neheb, line 15: maybe

not get_option('use_sse2').disabled()

subprojects/packagefiles/uchardet/src/meson.build
```diff
@@ -0,0 +1,72 @@
...
+  'nsSBCSGroupProber.cpp',
+  'nsSBCharSetProber.cpp',
+  'nsSJISProber.cpp',
+  'nsUTF8Prober.cpp',
+  'nsLatin1Prober.cpp',
+  'nsUniversalDetector.cpp',
+  'uchardet.cpp',
+)
+
+uchardet_inc = [
```
neheb, line 53: erm. include_directories is a variadic function. is there a need to wrap it in an array?

subprojects/packagefiles/uchardet/src/meson.build
```diff
@@ -0,0 +1,72 @@
...
+
+uchardet_args = [
+  '-DVERSION="@0@"'.format(meson.project_version()),
+  '-Wall',
+]
+
+uchardet = library(
+  'uchardet',
+  uchardet_sources,
+  include_directories: uchardet_inc,
```
neheb, line 65: '.' is implicit in library.

subprojects/packagefiles/uchardet/src/meson.build
```diff
@@ -0,0 +1,72 @@
...
+  'nsLatin1Prober.cpp',
+  'nsUniversalDetector.cpp',
+  'uchardet.cpp',
+)
+
+uchardet_inc = [
+  include_directories('.')
+]
+
+uchardet_args = [
```
neheb, line 57: wonder if add_project_arguments would be better.

subprojects/packagefiles/uchardet/test/meson.build
```diff
@@ -0,0 +1,69 @@
+
+test_executable = executable(
+    'uchardet-tests',
+    files('uchardet-tests.c'),
```
neheb, line 4: no need for files().

subprojects/packagefiles/uchardet/src/meson.build
Jan200101: it could be argued that these args are only relevant for the library at hand and no other C++ code.

neheb:
@Jan200101 either set it to C++14 or omit it. Doing the latter inherits the main meson.build standard, which is C++17.

subprojects/packagefiles/uchardet/meson.build
```diff
@@ -0,0 +1,30 @@
...
+)
+
+base_cpp_args = []
+
+cxx = meson.get_compiler('cpp')
+host_cpu_family = host_machine.cpu_family()
+
+use_sse2 = not get_option('use_sse2').disabled()
+
+if host_cpu_family == 'x86_64' or host_cpu_family == 'x86'
```
neheb, line 17: Now that I look at this again, you can probably get rid of this if. has_arguments() already checks if it can be used.

subprojects/packagefiles/uchardet/src/meson.build
neheb: right. Although I do remember projects like glew where some refactoring work resulted in the specified c_args not being applied.

Anyway, up to you.

Jan200101:
ugh
MSVC is being as much as a pain as always (how many people even use msvc with meson?)
I'll fix it when I get the time

neheb:
yeah. probably need to build it as static.

subprojects/packagefiles/uchardet/meson.build
Jan200101: I think its worth preserving this.

`-ffloat-store` seems to be an optimization that only makes sense on x86 and x86_64 where the float register gives excess precision when the precide definition of IEEE floats is required.

PR #715, opened 2022-10-23
icu: update to 72.1

neheb:


srl295:
Do you use the sources.txt file to source the source file lists? `icu/icu4c/source/common/sources.txt` etc?

neheb:
no.

srl295:
> no.

that might make the build files more flexible, and not need to change each release.

BTW, @neheb i have some suggestions over at #950

PR #718, opened 2022-10-25
Add wrap files for qarchive-2.0.2

tastytea:


tastytea:
ah, forgot to package libarchive first. 🥴

eli-schwartz:
You can install that in ci_config.json for Linux and macOS via deb/brew packages.

For Windows you can mark it as build_on -> windows -> false.

Technically no software ever supports Windows on account of dependencies :p but the WrapDB CI will verify that it builds correctly if libarchive gets added.

There's support for Choco packages for Windows too, iirc, but I doubt that works at all for libraries. It's been used for wraps that depend on the nasm program...

neheb:
Choco will not work. vcpkg is a solution but it needs to be added.

Or a libarchive wrap…

subprojects/packagefiles/qarchive/meson.build
```diff
@@ -0,0 +1,69 @@
...
+  'include/qarchiveextractor.hpp',
+  'include/qarchivecompressor_p.hpp',
+  'include/qarchivecompressor.hpp',
+  'include/qarchivediskextractor.hpp',
+  'include/qarchivememoryextractor.hpp',
+  'include/qarchivediskcompressor.hpp',
+  'include/qarchivememorycompressor.hpp',
+  'include/qarchive_global.hpp')
+
+include_dirs = ['include', '.']
```
neheb, line 50: I’d use include_directories instead of an array.

subprojects/packagefiles/qarchive/meson.build
```diff
@@ -0,0 +1,69 @@
+project(
+  'QArchive',
+  'cpp',
+  version: '2.1.1',
+  license: 'BSD-3-Clause',
+  default_options: ['cpp_std=c++11'],
+  meson_version: '>=0.59.0') # because of compile_moc
```
neheb, line 7: Hmm? I remember @eli-schwartz complaining about this before.

subprojects/packagefiles/qarchive/meson.build
eli-schwartz: It makes no difference unless you access the variable from another directory's `meson.build` file.

neheb: Sure. But it’s good practice

subprojects/packagefiles/qarchive/meson.build
eli-schwartz: Right. It's possible to support versions of Meson older than 0.59.0 by using `qt_mod.preprocess` instead of the `compile_moc` method.

There are plenty of people working on machines that have a system Meson older than 0.59.0, supporting those is a nice-to-have and doesn't really cost anything.

tastytea: Is there a minimum version you're targeting? Looks like `qt5.preprocess()` will work too and is older.

eli-schwartz: There's no hard and fast rule. qt5.preprocess has existed since Meson 0.22, i.e. since the stone age, so it's going to work basically regardless of version.

https://repology.org/project/meson/versions lists the accessibility of Meson in various places. Of particular note is:
- Debian oldstable, has Meson 0.49.2
- Debian stable, has Meson 0.56.2
- Ubuntu 20.04 LTS, has Meson 0.53.2

The more places it works, the better. On the flip side, sometimes you really, really need features from the latest version of Meson.

Where to find the balance is a tough question, and the answer tends to be different depending on the project you ask. Since the main point of the WrapDB is to be consumed by other projects, I like to be conservative about it, and be compatible with as many versions as I can, unless it subjectively feels too hard or complex to support. Because we don't know what the compat policy is of every project that wants to install this wrap file.

In the case of "use one older function instead of a newer function", it seems pretty easy to support.

tastytea:
added libarchive to the CI, used `include_directories`, `qt5.preprocess` instead of `qt5.compile_moc`.

subprojects/packagefiles/qarchive/meson.build
eli-schwartz: Note that even without qt.compile_moc, you still use the following:
```
qarchive| * 0.47.0: {'dict'}
qarchive| * 0.49.0: {'configuration_data dictionary'}
qarchive| * 0.54.0: {'meson.override_dependency'}
```

override_dependency is not really crucial, since the `[provide]` section also ensures mapping the dependency, although use of the override does perform a consistency check.
- You could drop it.
- You could guard it inside `if meson.version().version_compare('>=0.54.0')`
- You could just require `meson_version: '>=0.54.0'`

Attempting to support Meson less than 0.49.0 is probably a waste of time.

tastytea:
I don't know what i should do about `AssertionError: Not permitted files found: subprojects/packagefiles/qarchive/other/meson/config.h.in`. The original is a [file with #cmakedefine](https://github.com/antony-jr/QArchive/blob/master/other/cmake/config.h.in) in it and `config.h` is needed…

eli-schwartz:
- Use `*.h.meson`, and that way you don't even need to hide it away in its own directory.
- Use configure_file with `format: 'cmake@'`, assuming it works. (Note that Meson's support for `cmake@` is a filthy half-baked hack, but it often works.)

neheb:
The latter is nice as it avoids maintaining a separate config file.

neheb:
You should add back the minimum meson version. Typically 0.49.0

subprojects/packagefiles/qarchive/meson.build
```diff
@@ -0,0 +1,79 @@
...
+  license: 'BSD-3-Clause',
+  default_options: ['cpp_std=c++11'],
+  meson_version: '>=0.49.0')
+
+qt5 = import('qt5')
+
+if (build_machine.system() == 'darwin')
+  # brew installs into this path
+  libarchive_dep = dependency('libarchive',
+    default_options: 'pkg_config_path=/usr/local/opt/libarchive/lib/pkgconfig')
```
eli-schwartz, line 14: Huh, is this not set up by brew already?

neheb: This should be added to ci_config.json , not here.

neheb: Actually I don’t know if libarchive even has a pkgconfig file…

tastytea: It has at least on Linux…

tastytea:
The libarchive in Ubuntu focal is slightly too old. 🙁

neheb:
qt5 is keg only so it needs similar treatment.

neheb:
strange that libarchive is keg only. Anyway, there are two solutions to the Ubuntu error:

1: Update the Ubuntu version used in CI. This would probably get denied.

2: Create a libarchive wrap. No idea if you want to spend time on that.

Maybe a third would be to use an older version of qarchive…

subprojects/packagefiles/qarchive/meson.build
```diff
@@ -0,0 +1,72 @@
+project(
+  'QArchive',
+  'cpp',
+  version: '2.1.1',
+  license: 'BSD-3-Clause',
+  default_options: ['cpp_std=c++11'],
+  meson_version: '>=0.49.0')
```
neheb, line 7: It’s recommended to use , and ) on a new line. That’s how muon’s meson formatter does it. Here and elsewhere.

neheb:
Hmm this supports qt6 too. If you install that on brew, you can avoid the pkgconfig path. It’s also nice as qt5 is tested on Linux and qt6 on macOS.

tastytea:
> 1: Update the Ubuntu version used in CI. This would probably get denied.
> 
> 2: Create a libarchive wrap. No idea if you want to spend time on that.
> 
> Maybe a third would be to use an older version of qarchive…

Looks like 2.0.2 could work, but that's not ideal… I'll try making a libarchive wrap and then incorporate the other suggestions.

PR #719, opened 2022-10-25
cmark-gfm: add

hdoc:
Per @eli-schwartz's suggestion in https://github.com/hdoc/hdoc/pull/25, we'd like to contribute a Meson wrap for the [cmark-gfm](https://github.com/github/cmark-gfm) package.

This is a C library that implements parsing of Markdown files, and includes some extensions that GitHub (the company) has made. It's built using CMake, and this PR adds Meson support as a wrap.

### DIFF
--- a/subprojects/cmark-gfm.wrap
+++ b/subprojects/cmark-gfm.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = cmark-gfm-0.29.0.gfm.6
+source_url = https://github.com/github/cmark-gfm/archive/refs/tags/0.29.0.gfm.6.tar.gz
+source_filename = cmark-gfm-0.29.0.gfm.6.tar.gz
+source_hash = b17d86164c0822b5db3818780b44cb130ff30e9c6ec6a64f199b6d142684dba0
+patch_directory = cmark-gfm
+
+[provide]
+libcmark-gfm = cmark_gfm_dep
--- a/subprojects/packagefiles/cmark-gfm/cmark-gfm-extensions_export.h.meson
+++ b/subprojects/packagefiles/cmark-gfm/cmark-gfm-extensions_export.h.meson
@@ -1,0 +1,38 @@
+#ifndef CMARK_GFM_EXTENSIONS_EXPORT_H
+#define CMARK_GFM_EXTENSIONS_EXPORT_H
+
+/* Is this an exclusively static build */
+#if @CMARK_GFM_STATIC_DEFINE@ && ! defined CMARK_GFM_STATIC_DEFINE
+#  define CMARK_GFM_STATIC_DEFINE
+#endif
+
+/*
+ * Here is the complicated part. Windows is special -- you cannot just define
+ * entry points unconditionally.
+ * */
+#if defined _WIN32 || defined __CYGWIN__
+   /* When building static libraries, avoid marking public ones */
+#  if defined CMARK_GFM_STATIC_DEFINE
+#    define CMARK_GFM_EXTENSIONS_EXPORT
+   /* We are building this library */
+#  elif defined libcmark_gfm_EXPORTS
+#    define CMARK_GFM_EXTENSIONS_EXPORT __declspec(dllexport)
+   /* We are using this library */
+#  else
+#    define CMARK_GFM_EXTENSIONS_EXPORT __declspec(dllimport)
+#  endif
+
+/* On to the easy part. GCC and lookalikes such as clang just work */
+#elif defined __GNUC__ && __GNUC__ >= 4
+#  define CMARK_GFM_EXTENSIONS_EXPORT __attribute__((visibility("default")))
+
+/* Older solaris support, why not */
+#elif defined __SUNPRO_C && __SUNPRO_C >= 0x550
+#  define CMARK_GFM_EXTENSIONS_EXPORT __global
+
+/* All else failed, and we don't know about this compiler. Be conservative. */
+#else
+#  define CMARK_GFM_EXTENSIONS_EXPORT
+#endif
+
+#endif /* CMARK_GFM_EXTENSIONS_EXPORT_H */
--- a/subprojects/packagefiles/cmark-gfm/cmark-gfm_export.h.meson
+++ b/subprojects/packagefiles/cmark-gfm/cmark-gfm_export.h.meson
@@ -1,0 +1,38 @@
+#ifndef CMARK_GFM_EXPORT_H
+#define CMARK_GFM_EXPORT_H
+
+/* Is this an exclusively static build */
+#if @CMARK_GFM_STATIC_DEFINE@ && ! defined CMARK_GFM_STATIC_DEFINE
+#  define CMARK_GFM_STATIC_DEFINE
+#endif
+
+/*
+ * Here is the complicated part. Windows is special -- you cannot just define
+ * entry points unconditionally.
+ * */
+#if defined _WIN32 || defined __CYGWIN__
+   /* When building static libraries, avoid marking public ones */
+#  if defined CMARK_GFM_STATIC_DEFINE
+#    define CMARK_GFM_EXPORT
+   /* We are building this library */
+#  elif defined libcmark_gfm_EXPORTS
+#    define CMARK_GFM_EXPORT __declspec(dllexport)
+   /* We are using this library */
+#  else
+#    define CMARK_GFM_EXPORT __declspec(dllimport)
+#  endif
+
+/* On to the easy part. GCC and lookalikes such as clang just work */
+#elif defined __GNUC__ && __GNUC__ >= 4
+#  define CMARK_GFM_EXPORT __attribute__((visibility("default")))
+
+/* Older solaris support, why not */
+#elif defined __SUNPRO_C && __SUNPRO_C >= 0x550
+#  define CMARK_GFM_EXPORT __global
+
+/* All else failed, and we don't know about this compiler. Be conservative. */
+#else
+#  define CMARK_GFM_EXPORT
+#endif
+
+#endif /* CMARK_GFM_EXPORT_H */
--- a/subprojects/packagefiles/cmark-gfm/meson.build
+++ b/subprojects/packagefiles/cmark-gfm/meson.build
@@ -1,0 +1,141 @@
+project(
+    'cmark-gfm',
+    'c',
+    version: '0.29.0.gfm.6',
+    default_options: ['warning_level=3'],
+    meson_version: '>=0.49.0',
+    license: 'BSD-2-Clause',
+)
+
+cc = meson.get_compiler('c')
+has_stdbool_header = cc.has_header('stdbool.h')
+has_builtin_expect_code = 'int main() { __builtin_expect(0,0); return 0; }'
+has_builtin_expect = cc.links(
+    has_builtin_expect_code,
+    name: 'has __builtin_expect',
+)
+has_attribute_code = '''
+  int f(void) __attribute__ (());
+  int main() { return 0; }
+'''
+has_attribute = cc.compiles(has_attribute_code, name: 'has __attribute__')
+
+cdata = configuration_data(
+    {
+        'HAVE_STDBOOL_H': has_stdbool_header,
+        'HAVE___BUILTIN_EXPECT': has_builtin_expect,
+        'HAVE___ATTRIBUTE__': has_attribute,
+    },
+)
+
+configure_file(
+    format: 'cmake@',
+    input: 'src/config.h.in',
+    output: 'config.h',
+    configuration: cdata,
+)
+
+varr = meson.project_version().split('.')
+vdata = configuration_data(
+    {
+        'PROJECT_VERSION_MAJOR': varr[0],
+        'PROJECT_VERSION_MINOR': varr[1],
+        'PROJECT_VERSION_PATCH': varr[2],
+        'PROJECT_VERSION_GFM': varr[4],
+    },
+)
+
+configure_file(
+    format: 'cmake@',
+    input: 'src/cmark-gfm_version.h.in',
+    output: 'cmark-gfm_version.h',
+    configuration: vdata,
+)
+
+edata = configuration_data(
+    {
+        'CMARK_GFM_STATIC_DEFINE': get_option('default_library') == 'static',
+    },
+)
+
+configure_file(
+    input: 'cmark-gfm_export.h.meson',
+    output: 'cmark-gfm_export.h',
+    configuration: edata,
+)
+
+configure_file(
+    input: 'cmark-gfm-extensions_export.h.meson',
+    output: 'cmark-gfm-extensions_export.h',
+    configuration: edata,
+)
+
+# Required to avoid linker warnings on Windows.
+add_project_arguments('-Dlibcmark_gfm_EXPORTS', language: 'c')
+
+cmark_gfm_src = files(
+    'src/arena.c',
+    'src/blocks.c',
+    'src/buffer.c',
+    'src/cmark.c',
+    'src/cmark_ctype.c',
+    'src/commonmark.c',
+    'src/footnotes.c',
+    'src/houdini_href_e.c',
+    'src/houdini_html_e.c',
+    'src/houdini_html_u.c',
+    'src/html.c',
+    'src/inlines.c',
+    'src/iterator.c',
+    'src/latex.c',
+    'src/linked_list.c',
+    'src/man.c',
+    'src/map.c',
+    'src/node.c',
+    'src/plaintext.c',
+    'src/plugin.c',
+    'src/references.c',
+    'src/registry.c',
+    'src/render.c',
+    'src/scanners.c',
+    'src/syntax_extension.c',
+    'src/utf8.c',
+    'src/xml.c',
+)
+
+cmark_gfm_extensions_src = files(
+    'extensions/autolink.c',
+    'extensions/core-extensions.c',
+    'extensions/ext_scanners.c',
+    'extensions/ext_scanners.h',
+    'extensions/strikethrough.c',
+    'extensions/table.c',
+    'extensions/tagfilter.c',
+    'extensions/tasklist.c',
+)
+
+# cmark-gfm has many unused parameter warnings, so we disable that warning
+# on compilers that support it.
+add_project_arguments(
+    cc.get_supported_arguments('-Wno-unused-parameter'),
+    language: 'c',
+)
+
+includes = include_directories('.', 'src', 'extensions')
+cmark_gfm_lib = library(
+    'cmark-gfm',
+    cmark_gfm_src,
+    include_directories: includes,
+)
+
+cmark_gfm_extensions_lib = library(
+    'cmark-gfm-extensions',
+    cmark_gfm_extensions_src,
+    include_directories: includes,
+    link_with: cmark_gfm_lib,
+)
+
+cmark_gfm_dep = declare_dependency(
+    link_with: [cmark_gfm_lib, cmark_gfm_extensions_lib],
+    include_directories: includes,
+)
--- a/releases.json
+++ b/releases.json
@@ -168,6 +168,14 @@
       "1.9.1-1"
     ]
   },
+  "cmark-gfm": {
+    "dependency_names": [
+      "libcmark-gfm"
+    ],
+    "versions": [
+      "0.29.0.gfm.6-1"
+    ]
+  },
   "cmock": {
     "dependency_names": [
       "cmock"


subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -0,0 +1,73 @@
...
+})
+
+configure_file(
+  format: 'cmake@',
+  input: 'src/cmark-gfm_version.h.in',
+  output: 'cmark-gfm_version.h',
+  configuration: vdata)
+
+src = [
+  'src/cmark.c',
```
neheb, line 30: Should use files() here. Allows muon's FMT to alphabetize the entries.

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -0,0 +1,73 @@
...
+  'extensions/ext_scanners.c',
+  'extensions/ext_scanners.h',
+  'extensions/strikethrough.c',
+  'extensions/table.c',
+  'extensions/tagfilter.c',
+  'extensions/tasklist.c',
+]
+
+# Many unused parameters warnings when these warnings are enabled.
+add_project_arguments('-fvisibility=hidden', language: 'c')
```
neheb, line 68: IIRC this one can be done with gnu_symbol_visibility.

subprojects/packagefiles/cmark-gfm/meson.build
hdoc: Fixed, thank you.

eli-schwartz:
> Unfortunately, the upstream CMake config uses the generate_export_config CMake functionality (used here, here, and here) that doesn't seem to be supported by Meson. These lines generate the cmark-gfm-extensions_export.h and cmark-gfm_export.h files. Since I couldn't find a way to get Meson to generate these files, I've (temporarily) included them with the patch.

Because CMake is so amazingly wonderful, it generates a statically coded boilerplate file that really is trivial to check into git. It's pretty annoying that projects rely on a weird CMake function instead.

But the really bad part of this is that CMake generates a different file depending on your OS. This should just include support for all OSes via compiler check macros. Basically:

- if compiler supports GNUC 4 or greater, set up attribute visibility
- if compiler has the _MSC_VER define, check 8 or 10 different conditions before deciding whether to set up dllimport, dllexport, or nothing
- if all else fails, define macro as empty so it's a no-op

The CMake generated file just ritualizes this with project-specified scope names, including a bit of gunk nobody actually uses in their CMake projects.

subprojects/packagefiles/cmark-gfm/meson.build
eli-schwartz: Does the symbol table of the shared library expose a bunch of private symbols?

Given there's an export file, I imagine the source code annotates the functions which are public, so specifying `gnu_symbol_visibility: 'hidden'` is beneficial and cleans the symbol table while also allowing the compiler to generate slightly more optimized code.

hdoc: Given that this is being built as a static library I don't think visibility actually affects anything. If it were a shared library then the visibility attribute would have a purpose.

hdoc:
I agree that the CMake export functionality is extremely annoying. If we restrict this wrap to solely build cmark-gfm as a static library then we can drastically simplify the export files but ignoring all of the visibility `#defines`. That would leave the "deprecated" `#defines`. For those, I suggest we manually check if the compiler supports the `deprecated` attribute with some code like this:
```
cc = meson.get_compiler ('c')
if (cc.has_function_attribute('deprecated'))
  # Set the CMARK_GFM_DEPRECATED define to whatever
else
  # Don't set any defines
endif
```

That would then get written to the `*_export.h` files which would be generate at configure-time, perhaps using a `generator`.

Does this sound reasonable?

eli-schwartz:
> That would leave the "deprecated" `#defines`. For those, I suggest we manually check if the compiler supports the `deprecated` attribute with some code like this:

That's just random goop which CMake creates, and cmark-gfm doesn't use.

Anyway, I went and ported cmark to use a simpler header that Meson can use too. It would need to be merged into cmark, then merged again into cmark-gfm, then released, before this wrap could take advantage of it... but this wrap could still include the header itself in the meantime.

eli-schwartz:
Here's a much better template to use.

```
$ cat src/cmark-gfm_export.h.in 
#ifndef CMARK_GFM_EXPORT_H
#define CMARK_GFM_EXPORT_H

/* Is this an exclusively static build */
#if @CMARK_GFM_STATIC_DEFINE@ && ! defined CMARK_GFM_STATIC_DEFINE
#cmakedefine CMARK_GFM_STATIC_DEFINE
#endif

/*
 * Here is the complicated part. Windows is special -- you cannot just define
 * entry points unconditionally.
 * */
#if defined _WIN32 || defined __CYGWIN__
   /* When building static libraries, avoid marking public ones */
#  if defined CMARK_GFM_STATIC_DEFINE
#    define CMARK_GFM_EXPORT
   /* We are building this library */
#  elif defined libcmark_gfm_EXPORTS
#    define CMARK_GFM_EXPORT __declspec(dllexport)
   /* We are using this library */
#  else
#    define CMARK_GFM_EXPORT __declspec(dllimport)
#  endif

/* On to the easy part. GCC and lookalikes such as clang just work */
#elif defined __GNUC__ && __GNUC__ >= 4
#  define CMARK_GFM_EXPORT __attribute__((visibility("default")))

/* Older solaris support, why not */
#elif defined __SUNPRO_C && __SUNPRO_C >= 0x550
#  define CMARK_GFM_EXPORT __global

/* All else failed, and we don't know about this compiler. Be conservative. */
#else
#  define CMARK_GFM_EXPORT
#endif

#endif /* CMARK_GFM_EXPORT_H */
```

I also removed the pointless duplication of headers... but that needs patches to cmark-gfm.

hdoc:
Nice, thanks for making the PR upstream. Let me know if there is anything I can do to help. In the meantime I suppose we will let this PR sit on ice until upstream gets fixed, unless you have a better idea...

eli-schwartz:
Simply name the headers `*.h.meson` instead of `*.h.in`. :)

subprojects/packagefiles/cmark-gfm/cmark-gfm-extensions_export.h.meson
```diff
@@ -0,0 +1,38 @@
...
+/* Older solaris support, why not */
+#elif defined __SUNPRO_C && __SUNPRO_C >= 0x550
+#  define CMARK_GFM_EXTENSIONS_EXPORT __global
+
+/* All else failed, and we don't know about this compiler. Be conservative. */
+#else
+#  define CMARK_GFM_EXTENSIONS_EXPORT
+#endif
+
+#endif /* CMARK_GFM_EXTENSIONS_EXPORT_H */
```
neheb, line 38: No newline. Wonder what text editor does this...

subprojects/packagefiles/cmark-gfm/cmark-gfm_export.h.meson
```diff
@@ -0,0 +1,38 @@
...
+/* Older solaris support, why not */
+#elif defined __SUNPRO_C && __SUNPRO_C >= 0x550
+#  define CMARK_GFM_EXPORT __global
+
+/* All else failed, and we don't know about this compiler. Be conservative. */
+#else
+#  define CMARK_GFM_EXPORT
+#endif
+
+#endif /* CMARK_GFM_EXPORT_H */
```
neheb, line 38: No newline.

eli-schwartz:
These commits should all be squashed into one commit, probably...

subprojects/packagefiles/cmark-gfm/cmark-gfm_export.h.meson
hdoc: Fixed.

subprojects/packagefiles/cmark-gfm/cmark-gfm-extensions_export.h.meson
hdoc: Fixed.

hdoc:
@eli-schwartz thanks, I implemented your suggestion.

Commits are all squashed and the wrap builds and passes sanity check locally. I think the remaining items are to run CI on this to ensure it's clean, and then merge if everyone find it satisfactory.

hdoc:
Looks like I made a mistake when I formatted the file, one moment while I fix it...

hdoc:
Should be good to go now.

neheb:
```
cl : Command line error D8021 : invalid numeric argument '/Wno-unused-parameter'
```

hdoc:
Saw that, fixing it now.

hdoc:
@neheb fixed, though only for compilers that support `-Wno-unused-parameter`. Not sure how to make it work on MSVC, we don't have much experience with it.

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -94,7 +94,10 @@ src = files(
 )
 
 # Many unused parameters warnings when this warning is enabled.
-add_project_arguments('-Wno-unused-parameter', language: 'c')
+cc = meson.get_compiler('c')
+if (cc.has_argument('-Wno-unused-parameter'))
```
eli-schwartz, line 98: `if (expression)` is equivalent to `if (expression)`, the grouping isn't doing anything here.

Also note there is:

```meson
add_project_arguments(cc.get_supported_arguments('-Wno-unused-parameter'), language: 'c')
```

which means you don't need to repeat the argument name.

hdoc: Thanks, the extra brackets were an honest mistake. Incorporated your feedback on the latest commit.

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -0,0 +1,106 @@
...
+    default_options: ['warning_level=3'],
+)
+
+cdata = configuration_data(
+    {
+        'HAVE_STDBOOL_H': true,
+        'HAVE___BUILTIN_EXPECT': true,
+        'HAVE___ATTRIBUTE__': true,
+    },
+)
```
eli-schwartz, line 14: In the cmake build, these are defined as:
```cmake
CHECK_INCLUDE_FILE(stdbool.h HAVE_STDBOOL_H)
CHECK_C_SOURCE_COMPILES(
  "int main() { __builtin_expect(0,0); return 0; }"
  HAVE___BUILTIN_EXPECT)
CHECK_C_SOURCE_COMPILES("
  int f(void) __attribute__ (());
  int main() { return 0; }
" HAVE___ATTRIBUTE__)
```

I am not sure it makes sense to hardcode these...

neheb:
since there's dllexport functionality now, probably makes sense to use library() instead of static_library().

subprojects/packagefiles/cmark-gfm/meson.build
hdoc: Fixed.

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -0,0 +1,106 @@
...
+    input: 'src/cmark-gfm_version.h.in',
+    output: 'cmark-gfm_version.h',
+    configuration: vdata,
+)
+
+edata = configuration_data(
+    {
+        'CMARK_GFM_STATIC_DEFINE': true,
+    },
+)
```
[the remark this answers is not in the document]
hdoc: Never mind, I figured it out. Fixing it now.

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -0,0 +1,121 @@
+project(
+    'cmark-gfm',
+    'c',
+    version: '0.29.0.gfm.6',
```
eli-schwartz, line 4: Can you add `meson_version: '>=0.49.0',`? That's the minimum required version of Meson based on the features this uses -- always better to be explicit IMO.

eli-schwartz (commented):
Could be rebased again. :)

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -0,0 +1,121 @@
...
+has_builtin_expect_code = 'int main() { __builtin_expect(0,0); return 0; }'
+has_builtin_expect = cc.links(
+    has_builtin_expect_code,
+    name: 'has __builtin_expect',
+)
+has_attribute_code = '''
+  int f(void) __attribute__ (());
+  int main() { return 0; }
+'''
+has_attribute = cc.links(has_attribute_code, name: 'has __attribute__')
```
eli-schwartz, line 19: I suspect it's probably enough to check if these `cc.compiles()`, without also running the linker.

subprojects/packagefiles/cmark-gfm/meson.build
hdoc: Done.

subprojects/packagefiles/cmark-gfm/meson.build
hdoc: Agreed, fixed.

hdoc:
Rebased it and incorporated all feedback so far.

eli-schwartz:
On MSVC:

```
cmark-gfm| Checking if "has __builtin_expect" compiles: YES
cmark-gfm| Checking if "has __attribute__" compiles: NO
```

Apparently a compile test is NOT good enough for that one. :) Makes sense in retrospect, it's not syntactically wrong.

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -0,0 +1,123 @@
...
+    'extensions/autolink.c',
+    'extensions/core-extensions.c',
+    'extensions/ext_scanners.c',
+    'extensions/ext_scanners.h',
+    'extensions/strikethrough.c',
+    'extensions/table.c',
+    'extensions/tagfilter.c',
+    'extensions/tasklist.c',
+    'src/arena.c',
+    'src/blocks.c',
```
eli-schwartz, line 83: Shouldn't these be defined as two different libraries?

subprojects/packagefiles/cmark-gfm/cmark-gfm-extensions_export.h.meson
```diff
@@ -0,0 +1,38 @@
...
+/*
+ * Here is the complicated part. Windows is special -- you cannot just define
+ * entry points unconditionally.
+ * */
+#if defined _WIN32 || defined __CYGWIN__
+   /* When building static libraries, avoid marking public ones */
+#  if defined CMARK_GFM_STATIC_DEFINE
+#    define CMARK_GFM_EXTENSIONS_EXPORT
+   /* We are building this library */
+#  elif defined libcmark_gfm_EXPORTS
```
eli-schwartz, line 18: Ah right, there's some MSVC warnings about inconsistent linkage because these defines are not passed at the time of building.

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -0,0 +1,123 @@
...
+configure_file(
+    format: 'cmake@',
+    input: 'src/cmark-gfm_version.h.in',
+    output: 'cmark-gfm_version.h',
+    configuration: vdata,
+)
+
+edata = configuration_data(
+    {
+        'CMARK_GFM_STATIC_DEFINE': get_option('default_library') == 'static',
```
eli-schwartz, line 57: (Keeping in mind this is only needed for Windows, i.e. when `_WIN32 || __CYGWIN__` is defined, in the dllexport code branch.)

We should probably also add this define as a cflag for the declare_dependency and the pkg-config file, if it ends up being true and needed on Windows for linkage.

subprojects/packagefiles/cmark-gfm/meson.build
hdoc: Yes upstream defines them as two separate libraries so I'll make a follow-up commit to match that.

subprojects/packagefiles/cmark-gfm/meson.build
hdoc: Looks like [upstream defines them as cflags](https://github.com/github/cmark-gfm/blob/master/extensions/CMakeLists.txt#L54-L56) too, I've made a commit to match that behavior. As for the pkg-config, I'm not sure I follow. Do we need to define one? Likewise the Windows linker errors are a mystery to me.

eli-schwartz: I was going to say they were needed as compile_args in the declare_dependency, but...

I realized this is likely not true. For static builds, we should be covered by:
```
/* Is this an exclusively static build */
#if @CMARK_GFM_STATIC_DEFINE@ && ! defined CMARK_GFM_STATIC_DEFINE
#  define CMARK_GFM_STATIC_DEFINE
#endif
```

subprojects/packagefiles/cmark-gfm/meson.build
```diff
@@ -108,16 +100,43 @@ src = files(
...
+cmark_gfm_extensions_includes = include_directories('.', 'src', 'extensions')
+cmark_gfm_extensions_lib = library(
+    'cmark-gfm-extensions',
+    cmark_gfm_extensions_src,
+    include_directories: cmark_gfm_extensions_includes,
+    link_with: cmark_gfm_lib,
+)
+cmark_gfm_extensions_dep = declare_dependency(
+    link_with: cmark_gfm_extensions_lib,
+    include_directories: cmark_gfm_extensions_includes,
```
eli-schwartz, line 149: I don't think we need a separate dependency for this. When installing with cmake globally, it installs a single pkg-config file "libcmark-gfm.pc" that links to `-lcmark-gfm -lcmark-gfm-extensions`.

subprojects/cmark-gfm.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = cmark-gfm-0.29.0.gfm.6
+source_url = https://github.com/github/cmark-gfm/archive/refs/tags/0.29.0.gfm.6.tar.gz
+source_filename = cmark-gfm-0.29.0.gfm.6.tar.gz
+source_hash = b17d86164c0822b5db3818780b44cb130ff30e9c6ec6a64f199b6d142684dba0
+patch_directory = cmark-gfm
+
+[provide]
+cmark-gfm = cmark_gfm_dep
```
eli-schwartz, line 9: I think this should be called "libcmark-gfm", as that is the name of the pkg-config file this dependency corresponds to.

hdoc: Fixed.

subprojects/packagefiles/cmark-gfm/meson.build
hdoc: Fixed.

subprojects/packagefiles/cmark-gfm/meson.build
hdoc: Do we revert https://github.com/mesonbuild/wrapdb/pull/719/commits/36fd79c33555afa7e3c0b76637ea0cd515bf1b4a then?

eli-schwartz: Indeed I believe that is not necessary either.

hdoc: Reverted it.

hdoc:
Thoughts on these Windows linker warnings?

neheb:
This should be squashed.

As far as the file, you could probably avoid all that mess by manipulating cpp_args. See my recent PRs doing just that.

eli-schwartz:
You cannot manipulate cpp_args in order to solve the issue of the upstream main header attempts to `#include` a file that cmake generates containing the defines.

Also in general command line length bloat can cause bugs where e.g. Windows just errors out because too long, but sometimes it is regrettably unavoidable...

neheb:
Sure you can. Just empty it out. That's what I did for libebml.

eli-schwartz:
Why painfully bend over backwards to install an empty file, when if you are just going to hardcode a list of static defines you can just dump them as a set of configuration_data() values into the exact same configure_file()?

neheb:
I’d rather not. Note that it’s dllexport when building the library and dllimport when using it.

hdoc:
I've squashed the series of commits just now.

I would also prefer to use the same semantics as upstream does by generating the full `*_export.h` files instead of dumping an empty file and using `cpp_args`.

I believe the last outstanding issue is the Windows linker warnings. I do not have a Windows machine to test with (other than CI), so I would appreciate any help (a patch is especially welcome) in solving that issue.

Thank you.

subprojects/packagefiles/cmark-gfm/cmark-gfm-extensions_export.h.meson
hdoc: I believe the most recent commit should fix this issue by adding `-Dlibcmark_gfm_EXPORTS` as a `cpp_arg`. Let's see if it passes CI.

eli-schwartz: Right. CMake automatically does that by default just in case you want to use generated export headers (and in our case, we are using templated export headers) and I didn't think to check that that was implemented here.

hdoc:
Linker warnings on Windows are fixed. I'm squashing the PR and I think it is ready to be merged in now barring any final review comments.

jpakkane:
Issues seem to be fixed and tests are green so merging. Further minor issues can be fixed in subsequent MRs.

eli-schwartz:
Yup, agreed.

PR #720, opened 2022-10-25
Add libffi 3.4.4

xclaesse:
This is a rebase of
https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi onto 3.4.4 release.

It has been submitted upstream:
https://github.com/libffi/libffi/pull/746


xclaesse:
This is WIP because is requires Meson 0.64.0 release for MASM support.

xclaesse:
CC @nirbheek @tp-m

subprojects/packagefiles/libffi/meson-scripts/test-unwind-section.py
```diff
@@ -0,0 +1,36 @@
...
+        extern void foo();
+        int main(){foo();}
+    ''')
+
+args = sys.argv[1:]
+subprocess.check_call(args + ['-Wa,--fatal-warnings', '-c', infile1, '-o', outfile1],
+                      stdout=subprocess.PIPE, universal_newlines=True)
+subprocess.check_call(args + [infile2, outfile1], stdout=subprocess.PIPE, universal_newlines=True)
+
+sys.exit(0)
```
neheb, line 36: Missing end newline, here and elsewhere

xclaesse: fixed.

nirbheek:
What is the policy regarding adding wraps to wrapdb that also have the same meson port upstream? Maybe we should wait and see if libffi merges the meson build files? It's looking quite likely that it will be merged IMO.

jpakkane:
I don't think there is one but I'd say in this case it is a matter of how long does it take until upstream makes a release with Meson files. If it is expected to take long (weeks) then I'd say just merge this. It will get fixed on next upload anyway.

xclaesse:
I'll let a bit of time for upstream reviews, it's annoying to update in both sides anyway.

bgilbert:
This looks a bit stalled upstream.  Would it make sense to merge it here meanwhile?

PR #721, opened 2022-10-25
sdl2: try to make build fail

neheb:


neheb:
ping @Jan200101

neheb:
the reason Build all is failing is because some packages install qt, which includes dbus.

Jan200101:
Thank you for finding a way to reproduce this.

I have tracked this down to https://github.com/mesonbuild/wrapdb/blob/b5fcb4f408d331b8c512fefe6862d802d1b8c3a6/subprojects/packagefiles/sdl2/meson.build#L348-L351

through which we end up compiling

https://github.com/mesonbuild/wrapdb/blob/b5fcb4f408d331b8c512fefe6862d802d1b8c3a6/subprojects/packagefiles/sdl2/src/core/meson.build#L23-L27

and then include

https://github.com/libsdl-org/SDL/blob/b085c18251bc0298b1a78a621025978702ca07f7/src/core/linux/SDL_fcitx.c#L31-L33

which then goes down a short import chain before trying to use an undefined `Cursor`, which X.h should have provided.

Brew appears to be using the official xlib and X.h has the correct definition.
SDL_x11dyn.h has an import to Xlib.h which includes X.h, so that is fine too.

Could it be that MacOS ships with incomplete X11 headers?
I guess a workaround for now is simply disabling x11 on MacOS.

neheb:
Or disable dbus on macOS.

jpakkane:
Or both of them, really.

Jan200101:
Upstream entirely ignores X11 on MacOS, so lets do the same here.

PR #722, opened 2022-10-27
[RFC]treewide: run through dos2unix

neheb:


jpakkane:
The test script should be changed to check for Dos newlines, otherwise this is going to break again in the future.

xclaesse:
> The test script should be changed to check for Dos newlines, otherwise this is going to break again in the future.

Isn't the git attribute going to rewrite in UNIX eol and utf8 automatically when committing? If not, what does those attributes?

PR #723, opened 2022-10-27
SpeexDSP: Add a compiler-check for Neon on 32-bit Arm hosts

kcgen:
This check will fail on pre-Arm v7 hosts (like the Raspberry Pi 1) that don't have Neon support.


neheb:
Why not just use check_header on arm_neon?

kcgen:
> Why not just use check_header on arm_neon?


The presence of a header (and even the ability to compile its routines) is often possible regardless of the host processor's ability to run those instructions. 

For example, my Skylake system doesn't support the AVX2 instruction sets, however GCC provides and headers and even compiles the code:

```
g++ -march=skylake-avx512 -Ofast avx2.cpp
```

But the real test comes at runtime:

```
[06:12:22@server avx] ./a.out
Illegal instruction
[06:12:27@server avx] echo $?
132
```

In this case with Neon, we want to limit the compiler's own abilities to emit those instructions if the host doesn't support them (like on the Pi 1); but we do want those optimizations on newer Arm hosts. 

Unfortunately GCC doesn't have thorough host introspection on Arm, otherwise we could use a flag like `-march=native` that on x86-64 figures out which sse/avx/avx2/.. instructions the host supports.

(I should clarify that one purpose of this check is about limiting the compiler's own use of SIMD  instructions when optimizing bog-standard C code).

PR #736, opened 2022-10-31
Add orocos-kdl

tmayoff:


releases.json
```diff
@@ -2325,4 +2333,4 @@
       "1.3.3-1"
     ]
   }
-}
+}
```
neheb, line 2336: What?

neheb:
What's with these missing newlines? What editor is this?

tmayoff:
using VSCode, will look into changing the behaviour

neheb:
That is indeed bad behavior. Wonder if there's a way to force git to add newlines...

tmayoff:
VSCode does have a setting to add a newline at the end of the file, since I have external formatters for my other projects I'll enable it

subprojects/packagefiles/orocos-kdl/subprojects/eigen.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = eigen-3.4.0
+source_url = https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.tar.bz2
+source_filename = eigen-3.4.0.tar.bz2
+source_hash = b4c198460eba6f28d34894e3a5710998818515104d6e74e5cc331ce31e46e626
+patch_directory = eigen
+
+[provide]
+eigen3 = eigen_dep
+
```
neheb, line 10: Not necessary empty line.

neheb:
ending newlines are a unix thing. various tools work better with ending newlines. GitHub notices the problem in its webview. I think dos2unix also adds them.

subprojects/packagefiles/orocos-kdl/subprojects/eigen.wrap
tmayoff: The sanity checks are actually complaining about this file, am I not allowed to include it? I'd like whatever project I include orcos-kdl in to automatically download eigen3 if it doesn't already exist on the system

neheb:
```
AssertionError: Not permitted files found: subprojects/packagefiles/orocos-kdl/subprojects/eigen.wrap
```

tmayoff:
Might be missing something how would I add eigen3 as a subproject to this subproject as part of the wrap? I feel like I've done it before

neheb:
You don’t. If an eigen3 subproject exists, it gets used.

tmayoff:
I think I screwed around with the history of this branch a bit too much tried squashing commits and it didn't work, might just reopen this on a new branch and close this PR

Issue #708, opened 2022-10-20
expat cmake config file case sensitivity issue if size_t not in sys/types.h


chrishecker:
If the sys/types.h header doesn't define size_t, this config.set runs in the meson.build in the expat wrap:

`if not cc.has_header_symbol('sys/types.h', 'size_t')
  config.set('size_t', 'unsigned')
endif
`
The line in the expat_config.h.cmake file looks like this:

`/* Define to `unsigned' if <sys/types.h> does not define. */
#cmakedefine size_t @SIZE_T@
`

From debugging mesonlib/universal.py a bit, it seems like these two symbols have to be the same case.  If you specify case matching the first cmakedefine param, then you get a blank #define because it can't match the upper case SIZE_T in the confis (which in this case defines size_t to nothing which makes for all sorts of fun), if you specify upper case SIZE_T then it doesn't do the substitution at all and you undef size_t.  Doesn't seem like there's a way to fix this without modifying the cmake file, maybe in the patch file to make the second var lowercase.  Or, is cmake case insensitive and meson should match size_t to SIZE_T for the substitution?

Chris

neheb:
It's supposed to be define, not cmakedefine. Should be fixed upstream I guess.

chrishecker:
Hmm, I tried using an older expat from the wrapdb and it worked and the unsigned was written into the config...its still got cmakedefine and the all caps `@SIZE_T@` but the meson.build is slightly different with

```
if not (config.get('HAVE_SYS_TYPES_H', false) and
        cc.has_header_symbol('sys/types.h', 'size_t'))
  config.set('size_t', 'unsigned')
endif
```
which at first glance seems like it'd be equivalent but I haven't thought about it too much or looked into it, was just surprised it worked...oh, I see, the old meson.build ignores the expat_config.h.cmake file completely, and just dumps all the config variables directly, so it doesn't try to replace it in the cmake file, so it doesn't hit the case bug.

Why should the .cmake line be a define?  Doesn't it need variable substitution so needs to be a cmakedefine?

I think the question is whether cmake is case insensitive for replacements or not.  If it is, then this is a meson bug, if it's not, then the wrap meson.build needs to be modified to have a lowercase `@size_t@`

Chris

eli-schwartz:
> oh, I see, the old meson.build ignores the expat_config.h.cmake file completely, and just dumps all the config variables directly, so it doesn't try to replace it in the cmake file, so it doesn't hit the case bug.

[...]

PR #738, opened 2022-11-02
sdl2: update to 2.24.2

Jan200101:


Jan200101:
I am happy to announce that the MacOS failure is not in my control

neheb:
This does not build with muon. Can you take a look?

neheb:
That's fine. The other errors are enough.

Jan200101:
I'll see if I can fix the cmakedefine implementation in muon

neheb:
If you're looking into fixing cmake config file handling, I'll point to https://github.com/mesonbuild/wrapdb/issues/708 as well.

Anyway, beyond the scope of this PR.

neheb:
Tested locally. Works fine. Merging.

Jan200101:
I've been told, and have tested, that this is not a problem with muon being more wrong but being more strict.

It follows the behavior of meson, which does not handle `#cmakedefine`s where the name and the value are not the same.

Thus my hands are tied.

Jan200101:
https://github.com/mesonbuild/meson/issues/11006

PR #743, opened 2022-11-08
check: use gnu_symbol_visibility = hidden

neheb:
Also fix exports.


neheb:
@jpakkane CI passes now.

subprojects/packagefiles/check/src/meson.build
```diff
@@ -1,12 +1,11 @@
-if get_option('default_library') == 'shared' and cc.get_id() == 'msvc'
-  ck_dll_exp_interface = '-DCK_DLL_EXP=__declspec(dllimport)'
-  ck_dll_exp_private = '-DCK_DLL_EXP=__declspec(dllexport)'
+if host_machine.system() == 'windows'
+  ck_dll_exp_interface = '-DCK_DLL_EXP=@0@'.format(get_option('default_library') != 'static' ? '__declspec(dllimport)' : '""')
+  ck_dll_exp_private = '-DCK_DLL_EXP=@0@'.format(get_option('default_library') != 'static' ? '__declspec(dllexport)' : '""')
 else
-  ck_dll_exp_interface = '-DCK_DLL_EXP='
-  ck_dll_exp_private = '-DCK_DLL_EXP='
+  ck_dll_exp_interface = '-DCK_DLL_EXP=__attribute__((visibility("default")))'
```
eli-schwartz, line 5: The pkg-config / declare_dependency value actually doesn't need or want to be defined, simply set it `= []` as a no-op to avoid uselessly confusing command lines.

eli-schwartz (changes requested):

subprojects/packagefiles/check/src/meson.build
neheb: Not sure I understand the confusing command line comment.

neheb: A quick edit of such a thing looks even more confusing:

```
--- a/subprojects/packagefiles/check/src/meson.build
+++ b/subprojects/packagefiles/check/src/meson.build
@@ -1,8 +1,8 @@
-  ck_dll_exp_interface = '-DCK_DLL_EXP=@0@'.format(get_option('default_library') != 'static' ? '__declspec(dllimport)' : '""')
-  ck_dll_exp_private = '-DCK_DLL_EXP=@0@'.format(get_option('default_library') != 'static' ? '__declspec(dllexport)' : '""')
+  ck_dll_exp_interface = get_option('default_library') != 'static' ? '__declspec(dllimport)' : '""')
+  ck_dll_exp_private = get_option('default_library') != 'static' ? '__declspec(dllexport)' : '""')
 else
-  ck_dll_exp_interface = '-DCK_DLL_EXP=__attribute__((visibility("default")))'
+  ck_dll_exp_interface = ''
   ck_dll_exp_private = '-DCK_DLL_EXP=__attribute__((visibility("default")))'
 endif
 
@@ -35,7 +35,7 @@ checklib = library(
   'check',
   sources,
   gnu_symbol_visibility: 'hidden',
-  c_args: ['-DHAVE_CONFIG_H=1', ck_dll_exp_private],
+  c_args: ['-DHAVE_CONFIG_H=1', '-DCK_DLL_EXP=@0@'.format(ck_dll_exp_private)],
   include_directories: [confinc],
   dependencies: [libm, compatlib_dep],
 )
@@ -43,5 +43,5 @@ checklib = library(
 check_dep = declare_dependency(
   include_directories: '.',
   link_with: checklib,
-  compile_args: ck_dll_exp_interface,
+  compile_args: '-DCK_DLL_EXP=@0@'.format(ck_dll_exp_interface),
 )

```

eli-schwartz: But that's also not what I suggested so I don't know why that bothers you.

neheb: I think this is what you meant now?

eli-schwartz: Yes, exactly. :)

neheb:
Now that I tested it further, I don't think this is the right way to go.

tests only get built when not built as a subproject and when they do, a bunch of undefined references pop up.

When I get rid of gnu_symbol_visibility and the subproject check, they build and run just fine with sanity_checks,py. Which indicates the check should have been get_argument_syntax == 'msvc' all along...

Sounds like the tests run internal non public APIs honestly...

eli-schwartz:
> tests only get built when not built as a subproject and when they do, a bunch of undefined references pop up.

This should be fixed to use an option defaulting to false, which we can set to true in CI.

neheb:
enabled tests when possible.

tests require libregex when built with MinGW. Added check for that.

neheb:
Alright. I got tests working on MinGW, even when cross compiling.

@jpakkane this is ready.

PR #748, opened 2022-11-10
Add fallback urls for all of Gnome's download.gnome.org-hosted packages

kcgen:
Ever since the Gnome project migrated to GitLab, their release downloads have suffered periodic outages. The worst one (so far) was around this time last year, which lasted roughly a week.  Unfortunately these outages [continue](https://github.com/dosbox-staging/dosbox-staging/actions/runs/3438962655/jobs/5735879668).

```
fluidsynth| Looking for a fallback subproject for the dependency 
glib-2.0 because: fluidsynth| Building fallback subproject with 
default_library=static fluidsynth| Downloading glib source from

https://download.gnome.org/sources/glib/2.74/glib-2.74.0.tar.xz
 fluidsynth| <urlopen error [Errno 54] Connection reset by peer>

fluidsynth| WARNING: failed to download with error: could not get 
https://download.gnome.org/sources/glib/2.74/glib-2.74.0.tar.xz is 
the internet available?. Trying after a delay...
fluidsynth| <urlopen error [Errno 61] Connection refused>

... <repeats>

fluidsynth| A fallback URL could be specified using 
source_fallback_url key in the wrap file Message: You can disable 
this dependency with: -Duse_fluidsynth=false
```

During these outages, CI jobs that leverage gnome-based wraps fail, just as is happening right now. This requires manually re-checking Gnome's GitLab website and then manually restarting the CI jobs.

The fallbck host (ftp.acc.umu.se) is provided by one of Sweden's largest national universities: Umeå University. It's also used by the Linux From Scratch (LFS) project for source packages. See LFS's glib page, as an example:

https://www.linuxfromscratch.org/blfs/view/svn/general/glib2.html


kcgen:
I can't figure out why the VisualStudio build is failing. Found a couple errors in the meson.log's:

```
   Command line:  cl D:\a\wrapdb\wrapdb\_build\meson-private\tmppo4vyzrr\testfile.cpp /FeD:\a\wrapdb\wrapdb\_build\meson-private\tmppo4vyzrr\output.exe /nologo /showIncludes /utf-8 /Zc:__cplusplus /MDd /nologo /showIncludes /utf-8 /Zc:__cplusplus /Od /Oi- zlib1zlib.lib /link /nologo /release /nologo 
  
  Code:
   int main(void) { return 0; }
  
  Compiler stdout:
   testfile.cpp
  LINK : fatal error LNK1181: cannot open input file 'zlib1zlib.lib'
  
  Compiler stderr:
   
  Running compile:
  Working directory:  D:\a\wrapdb\wrapdb\_build\meson-private\tmp0_p0tuda
  Command line:  cl D:\a\wrapdb\wrapdb\_build\meson-private\tmp0_p0tuda\testfile.cpp /nologo /showIncludes /utf-8 /Zc:__cplusplus /EP /nologo /showIncludes /utf-8 /Zc:__cplusplus /EP /Od /Oi- 
  
  Code:
   
          #ifdef __has_include
           #if !__has_include("zlib.h")
            #error "Header 'zlib.h' could not be found"
           #endif
          #else
           #include <zlib.h>
          #endif
  Compiler stdout:
   
  
          
           
            
  Compiler stderr:
   testfile.cpp
  D:\a\wrapdb\wrapdb\_build\meson-private\tmp0_p0tuda\testfile.cpp(4): fatal error C1189: #error:  "Header 'zlib.h' could not be found"
  
  Run-time dependency zlib found: NO (tried pkgconfig, cmake and system)
  Running compile:
  Working directory:  D:\a\wrapdb\wrapdb\_build\meson-private\tmpcp6sozhn
  Command line:  cl D:\a\wrapdb\wrapdb\_build\meson-private\tmpcp6sozhn\testfile.cpp /FeD:\a\wrapdb\wrapdb\_build\meson-private\tmpcp6sozhn\output.exe /nologo /showIncludes /utf-8 /Zc:__cplusplus /MDd /nologo /showIncludes /utf-8 /Zc:__cplusplus /Od /Oi- ws2_32.lib /link /nologo /release /nologo 
  
  Code:
   int main(void) { return 0; }
  
  Compiler stdout:
   testfile.cpp
  
  Compiler stderr:
   
  Library ws2_32 found: YES
  Using subprojects\libxml++-5.0.2\subprojects\libxml2_cmake.wrap
  
  subprojects\libxml++-5.0.2\meson.build:179:4: ERROR: Git command failed: ['C:\\Program Files\\Git\\bin\\git.EXE', '-c', 'advice.detachedHead=false', 'clone', '--depth', '1', '--branch', 'master', 'https://gitlab.gnome.org/GNOME/libxml2.git', 'libxml2_cmake']
  The Meson build system
  Version: 0.64.0
```

kcgen:
Unfortunately this PR only solves part of the problem. We still have single points of failure when a wrap needs `git`-repo access to a given project:

neheb:
In OpenWrt there's a big mirror list: https://github.com/openwrt/openwrt/blob/master/scripts/download.pl#L312

Not sure if that can be adapted here.

kcgen:
> In OpenWrt there's a big mirror list: https://github.com/openwrt/openwrt/blob/master/scripts/download.pl#L312
> 
> Not sure if that can be adapted here.

@neheb - by "adapted", do you mean that the Meson WrapDB backend should/could be updated to support a generalized `TEMPLATE` -> `LIST_OF_MIRRORS` approach, similar to OpenWrt?

But specifically looking at this PR, the first mirror in OpenWrt's list is `download.gnome.org` - the source of ongoing problems.

The mirror i'm suggesting as a fallback is actually the 3rd item on OpenWrt's list :-)

``` perl
push @mirrors, "https://download.gnome.org/sources/$1";
push @mirrors, "https://mirror.csclub.uwaterloo.ca/gnome/sources/$1";
push @mirrors, "http://ftp.acc.umu.se/pub/GNOME/sources/$1";
push @mirrors, "http://ftp.kaist.ac.kr/gnome/sources/$1";
push @mirrors, "http://www.mirrorservice.org/sites/ftp.gnome.org/pub/GNOME/sources/$1";
push @mirrors, "http://mirror.internode.on.net/pub/gnome/sources/$1";
push @mirrors, "http://ftp.belnet.be/ftp.gnome.org/sources/$1";
push @mirrors, "ftp://ftp.cse.buffalo.edu/pub/Gnome/sources/$1";
push @mirrors, "ftp://ftp.nara.wide.ad.jp/pub/X11/GNOME/sources/$1";
push @mirrors, "https://mirrors.ustc.edu.cn/gnome/sources/$1";
```

neheb:
Right. Multiple mirrors are always a good idea.

kcgen:
This PR only adds fallback URLs, so I don't understand why CI is failing on `glib`. 

I'll split this PR up by wrap, and try each individually.

kcgen:
Split into separate PRs - all passing (unbelievable :man_facepalming: :laughing:):



Let's run with it; closing this PR down.

xclaesse:
One thing I've been thinking about is we could make WrapDB do its own mirror. Shouldn't be too hard to upload the original tarball from CI when tagging a release in WrapDB. Not sure this would cause excessive storage usage for our github account? Dunno if there is even a limit.

kcgen:
GitHub discourages large blobs (> 50 MB size warning, blocked at 100 MB) and recommends users avoid storing dependencies inside their repos; conveniently suggesting using a package manager instead :laughing:

Given wrapdb _is a package manager_, I think we're an exception.

If we look at other package managers hosted on GitHube:
 - https://github.com/freebsd/freebsd-ports
 - https://github.com/microsoft/vcpkg/tree/master/ports
 - https://github.com/macports/macports-ports

Although they are massive (surely 100-fold beyond wrapdb), none of them store package sources in the repo itself.  If they tried, they would quickly go /way/ beyond GitHub recommend repo size limit (~1 to 5 GB). 

Meson's unique approach of having [per-package repos](https://github.com/mesonbuild/zlib) could be used to hold the packages' versioned sources and surely not get close to these gigabyte limits for many many years - and by then, perhaps some reasonable pruning logic could take care of it (and we'd still have 3rd party hosts to rely on for "deep history"). 

https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github

eli-schwartz:
No one is suggesting to `git commit` a tarball. This is not an option anyway, it's terrible and horrible for the health of your git repo and GitHub is simply imitating the community sentiment when they say not to do it.

Even if GitHub allowed it, I wouldn't. No, using per package repos doesn't make a difference here, still not happening.

...

GitHub wants you to use attachments added to the Releases page. This is a GitHub-like approach to the classic "ftp directory" of downloads, it is basically what a package manager uses for binary artifacts, and it supports files of basically any size, and does it well.

We already *use* that. For the WrapDB patches.

eli-schwartz:
No one is suggesting to `git commit` a tarball. This is not an option anyway, it's terrible and horrible for the health of your git repo and GitHub is simply imitating the community sentiment when they say not to do it.

Even if GitHub allowed it, I wouldn't. No, using per package repos doesn't make a difference here, still not happening.

...

GitHub wants you to use attachments added to the Releases page. This is a GitHub-like approach to the classic "ftp directory" of downloads, it is basically what a package manager uses for binary artifacts, and it supports files of basically any size, and does it well.

We already *use* that. For the WrapDB patches.

kcgen:
> Even if GitHub allowed it, I wouldn't. No, using per package repos doesn't make a difference here, still not happening.

I definitely share that sentiment about carrying blobs artifacts in a source-repos.  I would assume the per-package repos would become Git-LFS buckets holding the versioned releases, specifically supported by the adjacent wrapdb project.

> GitHub wants you to use attachments added to the Releases page. This is a GitHub-like approach to the classic "ftp directory" of downloads

Even better - and these files are "best effort" anyway, if GitHub one day starts to roll off the oldest files (or the least-accessed files) when a project uses too much space.

PR #749, opened 2022-11-11
[No-op] Re-check the existing glib wrap in CI

kcgen:
This simply checks the existing `glib` wrap in CI, because I suspect it's currently broken.


eli-schwartz:
Since you have Actions enabled in your fork, you shouldn't need to create a PR just to check the CI.

kcgen:
> Since you have Actions enabled in your fork, you shouldn't need to create a PR just to check the CI.

Thanks for that, @eli-schwartz  - Indeed; I went looking for the 'rerun failed' here, but didn't realize my own local fork was running them as well (which makes perfect sense). Apologies for the churn here.

PR #755, opened 2022-11-11
taglib: Fix typo in taglib file name.

szanni:
`taglig` to `taglib`.


neheb:
well that's embarrassing. is the name even used for anything?

szanni:
> well that's embarrassing. is the name even used for anything?

It is the name of the file saved in `subprojects/packagecache` - so not really important. It just caught my eye while reviewing.

I am more concerned about the checks failing. Is this a common occurrence? Or could this be due to the transitive `zlib` dependency?

neheb:
@szanni it's failing because wrapdb's CI now uses meson 0.64. It just needs meson_version to be bumped to 0.50.0.

eli-schwartz:
It's also possible to be backwards-compatible and use

```diff
-taglib_includes = [
+taglib_includes = include_directories(
```

Depending on whether you are concerned about people whose Meson version is >= 0.47 but less than 0.50. For example, Debian buster (oldstable) has 0.49.

neheb:
New minimum is 0.49.0 because of / string paths: https://github.com/mesonbuild/wrapdb/pull/776

neheb:
I keep asking this, but should a minimum version of meson be mandated here? Say 0.49.0 as you mentioned.

eli-schwartz:
Kind of yes and kind of no. The thing is that if it's possible to support Debian oldstable, I think it's a good idea to do it. But if you *need* a newer version (for example to use the fs module, among other things) then that's fine.

neheb:
Sounds like optional nonfatal CI should be added then.

neheb:
This was merged with https://github.com/mesonbuild/wrapdb/commit/7b61ba2707eb3c2ccca7385d91f8c7d05e8d6b86

PR #756, opened 2022-11-12
imgui-sfml: add missing define for exports

neheb:
Run through muon's fmt.


neheb:
@jpakkane passes now. issue was not with this PR.

PR #760, opened 2022-11-12
htslib: simplify tests

neheb:
Massive size reduction by using a loop.


neheb:
I wonder if I can use the strategy used here to also run tests on check and Windows.

PR #765, opened 2022-11-13
pugixml: fix exports under Windows

neheb:

### DIFF
--- a/subprojects/packagefiles/pugixml/meson.build
+++ b/subprojects/packagefiles/pugixml/meson.build
@@ -8,17 +8,14 @@ project(
 
 sources = files('src/pugixml.cpp')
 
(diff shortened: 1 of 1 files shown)


neheb:
@eli-schwartz so I'm getting

```
pugixml| WARNING: Library target 'pugixml' has 'name_prefix' set. Compilers may not find it from its '-lpugixml' linker flag in the 'pugixml.pc' pkg-config file.
pugixml| WARNING: Library target 'pugixml' has 'name_prefix' set. Compilers may not find it from its '-lpugixml' linker flag in the 'pugixml-uninstalled.pc' pkg-config file.
```

I did this because the rc file lists only pugixml. In retrospect, this seems to be a mistake. Thoughts?

PR #768, opened 2022-11-13
qarchive: build with C++17

neheb:
It seems newer qt6 requires C++17. Errors with:

error: "Qt requires a C++17 compiler"


neheb:
```
WARNING: Project targets '>=0.49.0' but uses feature introduced in '0.57.0': module qt6.
```

oh boy. I think I know how this wasn't caught.

neheb:
ping @tastytea

neheb:
I'm also seeing this failure:

https://gist.github.com/neheb/8e3d9e8334c09bf09eecf90a1b33f984

No idea what the issue is.

tastytea:
>     qt_dep = dependency('qt6', modules: ['Core'], required: get_option('qt6'))
>     if qt_dep.found() and meson.version().version_compare('>=0.57.0')

On older merson versions, that would use Qt5 even if Qt6 is explicitly requested. I think it should fail instead if `qt6` is `enabled`.

neheb:
I think this is what you meant.

neheb:
OTOH this kills the point for using features. qt6=disabled will still find and use qt6 after this change.

neheb:
macOS build problem is fixed with https://github.com/antony-jr/QArchive/pull/60

tastytea:
> I think this is what you meant.

No, what if someone uses meson 0.50 but has `qt6` set to `enabled`? `qt_dep.found()` would be true but `meson.version().version_compare('>=0.57.0')` would be false, so `qt_dep` ends up being overwritten with Qt5, right?

I was thinking something like this:

```meson
qt_dep = dependency('qt6', modules: ['Core'], required: get_option('qt6'))
if qt_dep.found()
  if meson.version().version_compare('>=0.57.0')
    qt = import('qt6')
  else
    error('Qt6 support needs meson 0.57.0 or higher')
  endif
else
  qt_dep = dependency('qt5', modules: ['Core'])
  qt = import('qt5')
endif
```

neheb:
That error triggers on qt6=auto though.

tastytea:
How about this?

```meson
qt_dep = dependency('qt6', modules: ['Core'], required: get_option('qt6'))
if qt_dep.found() and meson.version().version_compare('>=0.57.0')
  qt = import('qt6')
elif qt_dep.found() and get_option('qt6').enabled()
  error('Qt6 support needs meson 0.57.0 or higher')
else
  qt_dep = dependency('qt5', modules: ['Core'])
  qt = import('qt5')
endif
```

neheb:
Actually maybe I'm wrong...

https://mesonbuild.com/Dependencies.html#qt4-qt5 doesn't mention qt6. So I wonder if meson < 0.57 can even find the qt6 dependency...

tastytea:
Wouldn't meson find it via CMake? And the [tarball](https://download.qt.io/official_releases/qt/6.4/6.4.0/submodules/qtbase-everywhere-src-6.4.0.tar.xz) does include `PkgConfigLibrary.pc.in`, so it seems to install pkg-config files too, normally…

neheb:
Good question. My experience with CMake dependencies is that they're case sensitive. If I had to guess, it might find dependency('Qt6') but not dependency('qt6'). The latter is special meson magic.

neheb:
Extra testing indicates this needs work.

QARCHIVE_STATIC is set to on by default, even when building as a shared library. QARCHIVE_BUILD is not set anywhere. There are also inline dllimport warnings for some reason.

antony-jr:
@neheb I think I missed this in the documentation but QArchive builds static by default and you can build it as a shared library by passing ```-DBUILD_SHARED_LIBS=ON``` (or **off** which is the default value), you **should not use** ```QARCHIVE_STATIC``` at all like used here -> https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/qarchive/meson.build#L25

TL;DR Use ```-DBUILD_SHARED_LIBS=ON``` to get shared libs(and don't set QARCHIVE_STATIC manually) for QArchive.

neheb:
Hmm? What's the purpose of QARCHIVE_STATIC then?

neheb:
oh this is the older PR. Newer one is here: https://github.com/mesonbuild/wrapdb/pull/772

antony-jr:
> Hmm? What's the purpose of QARCHIVE_STATIC then?

Internal purpose mostly, if QARCHIVE_STATIC is set then we will not import or export in case of dlls. QARCHIVE_BUILD is only set during the build of the library itself which is used to know if we need to use export, when this is not defined we simply use import, this is only done with shared libs. So we need to know after build if it's a static or shared build (because we will need to import in case of a shared build in Windows), to do that we use a config.h(which stores the build time value of QARCHIVE_STATIC) to make this all magical.

**So all you need to do is just simply use the** ```BUILD_SHARED_LIBS``` **option by cmake** -> https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

antony-jr:
*(PS: QArchive can include meson build system, upstream it if you have time )*

@neheb oh, ok so this is a bit tricky but the logic is,

config.h is generated by the build system (CMake or meson), this defines the build time value of QARCHIVE_STATIC 
(defines or does not define)

```
#include "config.h"
```

**config.h.in** (used to generate the config.h file) (**it would be nice to replace this with a meson equivalent**)
```
#ifndef CONFIG_H_QARCHIVE_INCLUDED
#define CONFIG_H_QARCHIVE_INCLUDED
#cmakedefine QARCHIVE_STATIC
#endif
```


---

In the header files, the global include file has these conditions..

If it's static we don't need to export anything for any platform

```
#ifdef QARCHIVE_STATIC
# define QARCHIVE_EXPORT
#else
```

else it's a shared build, QARCHIVE_BUILD is expected to be only set during the build of lib
if we are building the library now and it's not static, then we need to export 
in the sources

```
# ifdef QARCHIVE_BUILD
#  define QARCHIVE_EXPORT Q_DECL_EXPORT
# else
```
else we are not building the lib and the lib was not built static, some project is using the library,
so we need to import into that project

```
#  define QARCHIVE_EXPORT Q_DECL_IMPORT
# endif // QARCHIVE_BUILD

#endif // QARCHIVE_STATIC
```

neheb:
> _(PS: QArchive can include meson build system, upstream it if you have time )_
current meson support in here is only for wraps. There's no support for installing it with headers, pkgconfig and whatnot. I don't remember if meson even supports generating those .cmake files that get installed.
> 
> @neheb oh, ok so this is a bit tricky but the logic is,
> 
> config.h is generated by the build system (CMake or meson), this defines the build time value of QARCHIVE_STATIC (defines or does not define)
> 
> ```
> #include "config.h"
> ```
> 
> **config.h.in** (used to generate the config.h file) (**it would be nice to replace this with a meson equivalent**)
> 
> ```
> #ifndef CONFIG_H_QARCHIVE_INCLUDED
> #define CONFIG_H_QARCHIVE_INCLUDED
> #cmakedefine QARCHIVE_STATIC
> #endif
> ```
> 
> In the header files, the global include file has these conditions..
> 
> If it's static we don't need to export anything for any platform
> 
> ```
> #ifdef QARCHIVE_STATIC
> # define QARCHIVE_EXPORT
> #else
> ```
> 
> else it's a shared build, QARCHIVE_BUILD is expected to be only set during the build of lib if we are building the library now and it's not static, then we need to export in the sources
> 
> ```
> # ifdef QARCHIVE_BUILD
> #  define QARCHIVE_EXPORT Q_DECL_EXPORT
> # else
> ```
> 
> else we are not building the lib and the lib was not built static, some project is using the library, so we need to import into that project
> 
> ```
> #  define QARCHIVE_EXPORT Q_DECL_IMPORT
> # endif // QARCHIVE_BUILD
> 
> #endif // QARCHIVE_STATIC
> ```

That's exactly the logic in that PR. QARCHIVE_BUILD is only passed to library(). QARCHIVE_STATIC is defined when default_library==static.

PR #780, opened 2022-11-17
liblzma: update to 5.2.8

neheb:


neheb:
So the tarball has a beta CMake build that goes not use config.h. Curious if it's a good idea to do so here.

OTOH, the config.h file used here needs updating anyway.

neheb:
Updated the config.h file. meson generated one is close enough.

neheb:
ping @jpakkane this is ready

PR #782, opened 2022-11-17
icu: Fixes for consistency, Windows and Android support

amyspark:
👋 

This PR's general goal is to make the libicu wrap be consistent with the autotools-based toolchain.

New features:

- libicudata can now be built out of the box
- proper library naming following the official toolchain's convention
- inclusion of resource files (for DLL metadata)
- Android support

Bugfixes:

- inclusion of SOVERSION to work around Windows having a stunted copy of libicu
- use consistent library type and the corresponding linking for MSVC
- a fix for a compatibility warning with Meson 0.49
- a fix for the wrong check being done for ELF format

There are a couple code fixes that are needed upstream for this to work, affecting the MinGW support. I'll refer to the corresponding PR once I submit it.


amyspark:
See the following PRs for associated bug fixes: unicode-org/icu#2252, unicode-org/icu#2253, unicode-org/icu#2254

amyspark:
Going to add the pkg-config entries later today or tomorrow, please stay tuned.

subprojects/packagefiles/icu/meson.build
```diff
@@ -34,6 +34,20 @@ if get_option('default_library') == 'static'
   add_project_arguments('-DU_STATIC_IMPLEMENTATION', language: 'c')
   add_project_arguments('-DU_STATIC_IMPLEMENTATION', language: 'cpp')
   usage_args = ['-DU_STATIC_IMPLEMENTATION']
+  if host_machine.system() == 'windows' and cpp.get_argument_syntax() == 'msvc'
```
neheb, line 44: is it even possible for argument syntax to be msvc on non windows?

subprojects/packagefiles/icu/source/data/meson.build
```diff
@@ -10,7 +10,14 @@ endif
 
 if host_machine.system() == 'windows'
   if cpp.get_argument_syntax() == 'msvc'
```
neheb, line 12: Same as previous.

amyspark:
@neheb thanks for the review! I will send the corrections later re: MSVC.

I just added Android support, that will enable us at Krita to use this wrap for all our supported OSes. Please check if that's OK with you, as it's a fair bit of duplication (mostly because there's no way to QEMU run an Android executable.)

subprojects/packagefiles/icu/source/data/meson.build
amyspark: Here it's because I'm checking for MSVC vs MinGW. I've moved the check to the GCC/Clang clause, leaving this one for MSVC only.

subprojects/packagefiles/icu/meson.build
amyspark: No, I'll fix it.

amyspark:
I did not tag a new version of the wrap, will it be done on your end?

neheb:
I seem to have overlooked that. Which means CI didn't build anything...

https://github.com/mesonbuild/wrapdb/pull/784

PR #784, opened 2022-11-19
icu: bump release

neheb:


neheb:
```
subprojects/icu/source/data/icudt72.c(1953260): fatal error C1060: compiler is out of heap space
``

quite the error.`

**Comment by amyspark** [comment]
> quite the error.

That's a definite bug in libicu itself. They seem to rely on custom .obj creation for MSVC and haven't tested the .c compilation on small boxes (mine has 16GB). I'll try and see what I can do on my end.

**Comment by amyspark** [comment]
FTR, the custom .obj logic they employ for Windows is only valid for MSVC and willing to bet it's for amd64 only.

**Comment by amyspark** [comment]
Bug reports filed with Microsoft: [MASM](https://developercommunity.visualstudio.com/t/Severe-performance-degradation-in-MASM-w/10207114), [compiler](https://developercommunity.visualstudio.com/t/Severe-compiler-performance-degradation/10207115)

**Comment by amyspark** [comment]
Bugs filed regarding Meson support of NASM: https://github.com/mesonbuild/meson/issues/11082, https://github.com/mesonbuild/meson/issues/11083

The only usable workaround I found, that would maintain cross platform compatibility and be performant, is to add a NASM generator upstream and use that as the compiler.

Given that wraps cannot add functionality via source patches, @neheb, could you remove libicudata as a buildable target if the compiler is MSVC?

**Comment by eli-schwartz** [comment]
According to your MSVC bug report I'm guessing this is another example of "embed binary data into a C sources file by using extremely big strings" or usually its regional equivalent, extremely big char arrays.

This is the state of the industry and is known and expected to take ludicrous amounts of RAM, particularly with MSVC. There's a bunch of tricks for cheating and making it work fast by using e.g. custom ELF sections via direct use of ld, but portably doing it has historically always required biting the bullet and taking ludicrous amounts of RAM to process your big char arrays.

It rarely does cross people's minds to "just don't do that" and use data files. :p Although ICU does support data files, but then the application has to know it's built against a dependency like that and explicitly calculate the data file path -- the library doesn't know how to do that *and* be relocatable.

It's all very sad. The solution is in C23, though: https://thephd.dev/finally-embed-in-c23

The chances of portable software relying on C23 of course lie somewhere in between "you wish" and "lol for even asking". :(

In the meantime use mingw on the same machine for fantastic speedups, because *wow* is MSVC bad at big arrays even for compilers (that all suck at them).

...

There's no real solution, custom object files generators and linker tricks are the only actually performant solution and upstream has to handle making that robust themselves.

We can solve this by supporting it on:
- non-MSVC
- MSVC + amd64

**Comment by amyspark** [comment]
@eli-schwartz for the Krita use we're planning, we will use it with MinGW Clang for release, or with the amd64 version of MSVC for my personal use. Clang's already handled, and MSVC is usable in my box with the C file.

I've already made a small patch for us and upstream to enable NASM assembly generation, but I presume that will still leave aarch64 MSVC in the cold. For that reason, I filed two bugs with MSVC:

- https://developercommunity.visualstudio.com/t/Severe-performance-degradation-in-MASM-w/10207114 (MASM)
- https://developercommunity.visualstudio.com/t/Severe-compiler-performance-degradation/10207115 (CL)

File: subprojects/packagefiles/icu/source/data/meson.build [hunk 1]
```diff
@@ -60,19 +60,21 @@ if host_machine.system() == 'windows'
   )
 endif
 
-icudata = library(
+if cpp.get_argument_syntax() != 'msvc'
```
amyspark, line 63: This one is incorrect, clang-cl can perfectly build icudata.

neheb:
I have no experience with clang-cl so 🤷

amyspark (approved):

neheb:
@amyspark seems MSYS2 doesn't like this:

```
[1/3] Generating subprojects/icu/source/data/icudata_asm with a custom command (wrapped by meson to set PATH)
FAILED: subprojects/icu/source/data/icudt72.S
"C:/Users/Mangix/scoop/apps/msys2/2022-09-04/clang32/bin/meson" "--internal" "exe" "--unpickle" "C:/Users/Mangix/scoop/apps/msys2/2022-09-04/home/Mangix/devstuff/wrapdb/icu/meson-private/meson_exe_genccode_b4ef5a1c3daf3713a39c48d9561eca9c289397c7.dat"
while executing ['C:/Users/Mangix/scoop/apps/msys2/2022-09-04/home/Mangix/devstuff/wrapdb/icu/subprojects/icu/source/tools/genccode/genccode.exe', '-d', 'subprojects/icu/source/data', '-e', 'icudt72', '-f', 'icudt72', '../subprojects/icu/source/data/in/icudt72l.dat', '-a', 'gcc-mingw64']
--- stdout ---

--- stderr ---

```

neheb:
```
./genccode.exe
C:/Users/Mangix/scoop/apps/msys2/2022-09-04/home/Mangix/devstuff/wrapdb/icu/subprojects/icu/source/tools/genccode/genccode.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```

neheb:
So I tried to do this:

```
--- a/subprojects/packagefiles/icu/source/common/meson.build
+++ b/subprojects/packagefiles/icu/source/common/meson.build
@@ -200,7 +200,7 @@ endif

 icuuc_name = '@0@icuuc@1@'.format(library_prefix, library_suffix)

-common_lib = library(
+common_lib = static_library(
   icuuc_name,
   sources,
   include_directories: incdir,
@@ -208,14 +208,13 @@ common_lib = library(
   cpp_args: '-DU_COMMON_IMPLEMENTATION',
   link_with: stubdata_lib,
   dependencies: dl_dep,
-  version: U_ICU_VERSION,
   install: true,
 )

 if meson.can_run_host_binaries()
   common_native_lib = common_lib
 else
-  common_native_lib = library(
+  common_native_lib = static_library(
     '@0@-native'.format(icuuc_name),
     sources,
     include_directories: incdir,
@@ -223,7 +222,6 @@ else
     cpp_args: '-DU_COMMON_IMPLEMENTATION',
     link_with: stubdata_native_lib,
     dependencies: dl_native_dep,
-    version: U_ICU_VERSION,
     native: true,
   )
 endif
diff --git a/subprojects/packagefiles/icu/source/tools/toolutil/meson.build b/subprojects/packagefiles/icu/source/tools/toolutil/meson.build
index ca2ccc8..7e0102b 100644
--- a/subprojects/packagefiles/icu/source/tools/toolutil/meson.build
+++ b/subprojects/packagefiles/icu/source/tools/toolutil/meson.build
@@ -33,7 +33,7 @@ if have_elf_h
   toolutil_cpp_args += ['-DU_HAVE_ELF_H']
 endif

-toolutil_lib = library(
+toolutil_lib = static_library(
   'toolutil',
   sources,
   c_args: toolutil_c_args,
@@ -44,7 +44,7 @@ toolutil_lib = library(
 if meson.can_run_host_binaries()
   toolutil_native_lib = toolutil_lib
 else
-  toolutil_native_lib = library(
+  toolutil_native_lib = static_library(
     'toolutil-native',
     sources,
     c_args: toolutil_c_args,
```

Unfortunately, this gives me duplicate symbol errors.

neheb:
I got further with -Ddefault_library=static

```
Mangix@DESKTOP-VIRSTQA ~/d/w/icu (master) [4]> ./subprojects/icu/source/tools/genccode/genccode.exe -d subprojects/icu/source/data -e icudt72 -f icudt72 ../s
ubprojects/icu/source/data/in/icudt72l.dat --skip-dll-export -a gcc-mingw64
generating assembly code for ../subprojects/icu/source/data/in/icudt72l.dat
fish: Job 1, './subprojects/icu/source/tools/…' terminated by signal SIGSEGV (Address boundary error)
```

amyspark:
@neheb The crashes you're getting are known. However, from the ["Adding new projects to WrapDB"](https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#changes-to-original-source) doc page:

> It should especially be noted that there must not be any patches to functionality. 

>  All such changes must be submitted to upstream. 

That's why, in the first post here, I added the relevant merge requests. For the avoidance of doubt, here are the bug reports I've sent:

- https://unicode-org.atlassian.net/browse/ICU-22203 (unable to override the entry point with the C generator, even the docs and flags so accept it)
- https://unicode-org.atlassian.net/browse/ICU-22204 (crash with generating the C code for MinGW)
- https://unicode-org.atlassian.net/browse/ICU-22205 (MASM generator being 32-bit only; unused here)
- https://unicode-org.atlassian.net/browse/ICU-22209 (MASM generator generating .S instead of .asm/.masm; unused here)
- https://unicode-org.atlassian.net/browse/ICU-22210 (working around MSVC being really slow, by using NASM)
- https://developercommunity.visualstudio.com/t/Severe-performance-degradation-in-MASM-w/10207114
- https://developercommunity.visualstudio.com/t/Severe-compiler-performance-degradation/10207115

And the fixes are here:
- https://github.com/unicode-org/icu/pull/2252 (ICU-22203)
- https://github.com/unicode-org/icu/pull/2253 (ICU-22204)
- https://github.com/unicode-org/icu/pull/2254 (ICU-22205)
- https://github.com/unicode-org/icu/pull/2263 (ICU-22210)

PR #789, opened 2022-11-22
sdl2: update to 2.26.0

Jan200101:
released in the past 24 hours
https://github.com/libsdl-org/SDL/releases/tag/release-2.26.0

CI is expected to fail


Jan200101:
despite announcing it, I did not intend to forget to update releases.json

Jan200101:
Compiled locally against Fedora 37 and Windows using [MXE](https://mxe.cc/), ran some tests as well as Lite-XL and it appears to work flawlessly (and the MSVC & MacOS CI passes which is a miracle).

Some more eyes should look over this because this is the bare minimum to make things work flawlessly on my end, and I might have dropped the ball on some architecture or OS I do not use.

neheb:
Hmm I don't see any dllexport or def file for pulseaudio. I do know meson is broken with the latter for clang on MSYS2. See https://github.com/mesonbuild/meson/issues/10782

Biswa96:
> The issue might be broken pulseaudio under the latter honestly...

My apology, I have no idea about that topic 🙇

neheb:
> > The issue might be broken pulseaudio under the latter honestly...
> 
> My apology, I have no idea about that topic 🙇

maybe @mehdichinoune knows

As far as I can tell, https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-pulseaudio/0002-meson-check-version-script.patch is a clue.

GNU ld supports it but LLVM LLD does not.

the map-file includes pa_strerror.

Biswa96:
We can add DEF file with pulseaudio in upstream. But I do not know if that issue is related to this SDL2 change.

neheb:
It is not.

adding a def file will not work because of https://github.com/mesonbuild/meson/issues/10782

Biswa96:
Adding a DEF file in pulseaudio worked with gcc and with clang. See the required changes in this merge request https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/761

neheb:
Weird. Wonder why I ran into the libusb issue. Maybe lld quirk relating to functions being exported twice.

PR #791, opened 2022-11-24
lame: Small updates

amyspark:
Hi,

These are small quality-of-life updates for LAME wrap users on Windows.


amyspark:
[This sanity check failure](https://github.com/mesonbuild/wrapdb/actions/runs/3540961426/jobs/5944637823#step:5:58) means it expects all dependencies to be enabled by default?

jpakkane:
Seems like it. You'd need to change it to `required: false` and do the juggling required to build it without mpg123, assuming it really is an optional dependency. If it truly is mandatory then it needs to be added to WrapDB too.

neheb:
It's just the meson_option.txt options that need to be flipped. Alternatively, do so in CI_config.json

amyspark:
It can be built without it, but it will save headaches if we follow @neheb's suggestion. CI's green now!

neheb:
I cleaned it up further: https://github.com/neheb/wrapdb/commits/fluid

Would be great to add here.

neheb:
Hmmm I don't see dllimport anywhere. There's also this issue:

```
libsndfile| Library mp3lame found: NO
```

AFAIK there's no pkgconfig file that comes with LAME.

version change seems wrong too:

```
> find /usr | grep mp3lame
/usr/lib64/libmp3lame.so.0
/usr/lib64/libmp3lame.so.0.0.0
```

eli-schwartz:
>  lame: Install pkg-config file 

This does not match upstream, which means that other projects ***cannot*** (are not allowed to) use it, because it will then fail to detect that lame is installed on the system. Same reason I *still* think it was wrong to add a provide section. :D 

This needs to be solved in https://sourceforge.net/p/lame/feature-requests/42/

neheb:
But without the provide section the wrap cannot be used at all. Which means no point in having it...

In the case of libsndfile, I'm changing find_library() to dependency() to get it to use the wrap.

The last comment of that thread is interesting. Seems Debian maintainers want a pkgconfig file.

I'm still curious why missing dllimport results in proper linking. Something tells me dllexport is not being applied and MinGW just compiles file.

eli-schwartz:
> But without the provide section the wrap cannot be used at all. Which means no point in having it...

Sure it can be used, via subproject('foo').get_variable() in an if/else.

You already do need to use if/else, because you need to first check with cc.find_library() to check if it is installed to the system and only after that fails can you go look for a subproject.

eli-schwartz:
> The last comment of that thread is interesting. Seems Debian maintainers want a pkgconfig file.

Debian maintainers want lots of things.

Debian maintainers have, for ***decades***, been modifying and patching software with whatever changes they believe make software better for Debian's own use. It's not a concern of theirs if that makes it more difficult for ***other*** upstream projects to know whether their software works on non-Debian systems.

Most other issues I have with Debian, I acknowledge to be personal preference or just griping about bugs. This particular issue, I actually believe makes Debian a dishonest member of the open source community and a distro that people should refuse to use and encourage others to not use. It's very frustrating.

neheb:
I guess MinGW supports mixing dllexport and __attribute__((visibility))...

This is an interesting upstream change: https://sourceforge.net/p/lame/svn/6485/

amyspark:
@eli-schwartz @neheb,

> Hmmm I don't see dllimport anywhere. There's also this issue:

I've not changed anything regarding dllimport; in any case, it's set in https://github.com/mesonbuild/wrapdb/blob/3eeff0ec57a580f8238f0db8ed64b52f86cc6a7a/subprojects/packagefiles/lame/meson.build#L140-L147

>  AFAIK there's no pkgconfig file that comes with LAME.
> 
> This does not match upstream, which means that other projects cannot (are not allowed to) use it, because it will then fail to detect that lame is installed on the system. Same reason I still think it was wrong to add a provide section. :D

I've added a pkg-config file as per https://lists.gnu.org/archive/html/guix-patches/2020-12/msg00359.html and https://sourceforge.net/p/lame/mailman/message/9923548/, to aid in looking stuff up. If that's bothersome, please let me know and I'll remove the commit.

> edit: this wrap doesn't even work:

Is that caused by my changes or has the wrap been always broken? If the latter, is it something that I should address here?

> Most other issues I have with Debian, I acknowledge to be personal preference or just griping about bugs. This particular issue, I actually believe makes Debian a dishonest member of the open source community and a distro that people should refuse to use and encourage others to not use. It's very frustrating.

This PR is about Meson, enhancing the usability and error-proof-ness of one of its wraps in Windows. I am not involved with Debian in any capacity, nor I am an user of their software, so I am incapable and unwilling to make any judgments as to the design decisions of the Debian distribution.

 > I guess MinGW supports mixing dllexport and attribute((visibility))...

`__declspec(dllexport)` is the Windows (read: MSVC) extended attribute syntax, and `attribute((visibility))` is the GCC/Clang version. It's cumbersome, but makes sense that both would be supported.

amyspark:
I've removed the pkg-config entry and the soversion change for compatibility, as the [Ubuntu package](https://packages.ubuntu.com/jammy/amd64/libmp3lame0/filelist) doesn't show any.

neheb:
@amyspark that's dllexport, not import. In any case, it works only because MinGW supports mixing dllexport and attribute((visibility)). Not really relevant to these changes.

the wrap not working has probably always been the case. I'd just pull the changes in my branch.

amyspark:
Closing in favour of #799.

PR #792, opened 2022-11-24
sdl2_mixer: Define version

Charadon:
sdl2_mixer was missing a definition of it's version, so I added one.


eli-schwartz:
Can you please also update the build revision (the part after the `-` in the versions entry) in releases.json -- that will cause a new version of the wrap itself to be released.

Charadon:
That good?

eli-schwartz (approved):
Thanks!

PR #798, opened 2022-11-26
Add libtheora wrap

amyspark:
👋 

This PR is a proposal to add libtheora to the WrapDB. I tested this with the following compilers:

- Windows: MSVC x64, MSYS CLANG64, MSYS UCRT64
- macOS: Big Sur 11.7.1 + Xcode 13.2.1

Additionally, I've verified that the dependency is correctly recognised by GStreamer's FFmpeg wrap, which we are testing for a broader release at Krita: https://invent.kde.org/graphics/krita/-/merge_requests/1599.

Discussion
---

- Are the arm2gnu and fig2pdf wrappers OK?
  - Both the Perl script and fig2pdf output to stdout, so I needed them to capture the data and output to the correct files.
  - The latter unfortunately cannot be used with a `generator` because a list of GeneratedFiles cannot be used as `dependencies` for the `custom_target` that drives the spec generator.
- Added a dependency on TeX Live's `latexmk` script. plus a few copy steps. to ensure it will drive pdflatex for us.
  -  Upstream drives pdflatex and bibtex manually, which is prone to error...

These two possibly warrant further analysis:

- Added a Doxyfile wrapper to patch the OUTPUT_DIRECTORY to match Meson's Doxygen test's output folder.
- Added a .rc file wrapper for Windows, to massage the existing ones into a form that's compatible with both MSVC and GCC.


amyspark:
Downgraded the wrap version to 1.1.1 as it's the latest one officially released.

I'd based mine (and the one we'll ship) on top of the latest branch, which doesn't seem to have worked well.

amyspark:
> Downgraded the wrap version to 1.1.1 as it's the latest one officially released.
> 
> I'd based mine (and the one we'll ship) on top of the latest branch, which doesn't seem to have worked well.

Ok, there are two big issues here. Upstream expect us to ship three libraries: theora, theoradec, and theoraenc. However, all official tarballs somehow lack the exports file and include header for it.

Moreover, does Meson not allow Git wraps, even though it's a supported method?

neheb:
@amyspark for missing exports there are 2 solutions: replace the header file as is done with lame.h or include a custom def file with meson and use vs_module_args. I've done the latter with multiple wraps.

amyspark:
Gone with the latter, I've copied the module definition file that's shipped with current master. It seems there are unexpected test failures with MSVC, I'll check what's up with my desktop.

subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -0,0 +1,204 @@
...
+    'idct.c',
+    'info.c',
+    'internal.c',
+    'quant.c',
+    'state.c',
+)
+
+if asm
+    if conf.get('CPU_x86_64', false)
+        if cc.get_id() != 'msvc'
```
neheb, line 43: here and elsewhere, I think it makes more sense to use get_argument_syntax instead of get_id.

subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -0,0 +1,204 @@
...
+        input: [encoder_rc, '../win32/xmingw32/libtheoraenc-all.rc'],
+        output: ['libtheoraenc.rc']
+    )
+
+    encoder_res += windows.compile_resources(libtheoraenc_rc)
+endif
+
+if host_machine.system() == 'darwin'
+    version_flag = '-Wl,-exported_symbols_list,@0@'
+    theoradec_version_arg = join_paths(meson.current_source_dir(), 'theoradec.exp')
```
neheb, line 142: an alternative to join_paths is to use the / operator.

subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -0,0 +1,204 @@
+encoder_sources = files(
+    'apiwrapper.c',
```
neheb, line 2: Typical meson style is to use 2 spaces for indentation. My muon fmt config does that as well.

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,87 @@
...
+)
+
+conf = configuration_data()
+
+compile_args = []
+
+cc = meson.get_compiler('c')
+
+if cc.get_argument_syntax() != 'msvc'
+    compile_args += cc.get_supported_arguments(['-Wall', '-Wno-parentheses'])
```
neheb, line 36: No need for []. get_supported_arguments is a variadic function.

subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -0,0 +1,204 @@
...
+    version: meson.project_version()
+)
+
+libtheoraenc = library(
+    'theoraenc',
+    encoder_sources + encoder_res,
+    dependencies: [m_dep, ogg_dep, config_dep],
+    include_directories: incdir,
+    vs_module_defs: '../win32/xmingw32/libtheoraenc-all.def',
+    link_args: cc.get_supported_link_arguments([
```
neheb, line 176: No need for []. get_supported_link_arguments is a variadic function.

subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -0,0 +1,204 @@
...
+    theora_version_arg = '@0@/Version_script'.format(meson.current_source_dir())
+endif
+
+libtheoradec = library(
+    'theoradec',
+    decoder_sources + decoder_res,
+    dependencies: [m_dep, ogg_dep, config_dep],
+    include_directories: incdir,
+    vs_module_defs: '../win32/xmingw32/libtheoradec-all.def',
+    link_args: cc.get_supported_link_arguments([
```
neheb, line 158: No need for [] in variadic function.

subprojects/packagefiles/theora/lib/meson.build
amyspark: It's the incorrect semantics. `get_argument_syntax` covers Clang/CL, while the one that can't do inline assembly is MSVC itself.

subprojects/packagefiles/theora/lib/meson.build
amyspark: Much cleaner, thanks!

subprojects/packagefiles/theora/lib/meson.build
neheb: ah OK.

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,88 @@
...
+  default_options: ['c_std=c11']
+)
+
+conf = configuration_data()
+
+compile_args = []
+
+cc = meson.get_compiler('c')
+
+if cc.get_argument_syntax() != 'msvc'
```
neheb, line 36: I've had arguments about this before. I think the core developers are in favor of this approach.

OTOH, get_supported_arguments checks that the arguments can be actually used, which implies it's safe to use without cc.get_argument_syntax() != 'msvc'

subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -0,0 +1,205 @@
+encoder_sources = files(
```
neheb, line 1: muon's fmt automatically sorts files like this. Probably should do the same.

subprojects/packagefiles/theora/meson.build
amyspark: Those checks are an obvious fail on GCC/Clang/ICC, so they'll just pile up noise as they error out. Hence why I put them behind a conditional clause.

subprojects/packagefiles/theora/lib/meson.build
amyspark: Took me a bit, but it's done.

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,85 @@
+project(
+  'theora',
+  'c',
+  'cpp',
+  version: '1.1.1',
+  meson_version: '>= 0.59.0',
```
neheb, line 6: This should really be 0.49.0. Compatibility with Debian oldstable is recommended though not required.

Most usages of new meson functionality seems to revolve around using feature options as booleans.

neheb: yeah...that's exactly what's being done.

amyspark: I'm using the following new features:

```
theora| WARNING: Project specifies a minimum meson_version '>= 0.49.0' but uses features which were added in newer versions:
theora| * 0.50.0: {'install arg in configure_file', 'not_found_message arg in dependency'}
theora| * 0.56.0: {'meson.project_source_root'}
theora| * 0.59.0: {'feature_option.disable_auto_if()', 'feature_option.allowed()'}
```

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,269 @@
...
+)
+
+conf = configuration_data()
+
+enable_valgrind = get_option('valgrind-testing').allowed()
+
+test_env = []
+
+if enable_valgrind
+  valgrind_exe = find_program('valgrind')
```
eli-schwartz, line 18: The usual point of a feature option is to do:

```
valgrind_exe = find_program('valgrind', required: get_option('valgrind-testing'))
```

The current state of this instead causes `-Dvalgrind-testing=auto` to force it to enabled...

subprojects/packagefiles/theora/meson.build
neheb: The first and second can be worked around easily.

The last one is interesting. I cloned this branch locally and saw most usages were of the form:

```
dependency(x: required: y.enabled())
```

The thing is, if y is disabled, the dependency will be set to not found, even if realistically found.

@eli-schwartz might have more input.

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,269 @@
...
+
+  conf.set('HAVE_VALGRIND', valgrind_exe.found())
```
eli-schwartz, line 20: Since valgrind_exe discovered with `required: <default>`, it is mandatory and this code cannot ever be false -- the build would simply error out. Unless you use my above suggestion.

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,269 @@
...
+      '--leak-check=yes',
+      '--show-reachable=yes',
+      '--num-callers=100',
+    ]
+
+    message(
+      'Tests will be run under: @0@'.format(env.get('VALGRIND_ENVIRONMENT')),
+    )
+  else
+    message('Valgrind not found')
```
eli-schwartz, line 34: Redundant, Meson logs whether or not it's found.

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,269 @@
...
+  endif
+else
+  message('Valgrind testing not enabled')
```
eli-schwartz, line 37: Redundant, meson logs "Program valgrind skipped: feature valgrind-testing disabled". :p

eli-schwartz (changes requested):

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,269 @@
...
+  doxygen_exe = find_program(
+    'doxygen',
+    required: doc.enabled(),
+    disabler: true,
+  )
+  if not doxygen_exe.found()
+    message('*** doxygen not found, API documentation will not be built')
+  endif
+  doc.disable_auto_if(not doxygen_exe.found())
+endif
```
eli-schwartz, line 189: Dunno what all this is. When `-Ddoc=disabled`, this evaluates as `required: false`, doxygen_exe is searched for nevertheless, and then you cast doc from auto to disabled, returning a new immutable object that you then garbage collect.

subprojects/packagefiles/theora/doc/meson.build
```diff
@@ -0,0 +1,47 @@
+static_docs = files(
+  'color.html',
+  'draft-ietf-avt-rtp-theora-00.txt',
+  'draft-ietf-avt-rtp-theora-00.xml',
+  'vp3-format.txt',
+)
+
+install_data(static_docs)
+
+if doc.allowed()
```
eli-schwartz, line 10: Then this checks if doc is enabled or auto, and in the auto case, doxygen can be not-found without the build erroring, but doc.allowed() is still true. It's allowed, even though it didn't get *found*.

Instead, use doxygen_exe.found(), which reports not just whether the build options *allow* it, but whether the build environment *is capable* of it.

eli-schwartz (changes requested):

subprojects/packagefiles/theora/examples/meson.build
```diff
@@ -0,0 +1,95 @@
+theora_libs = [libtheora]
+theoradec_libs = [libtheoradec]
+theoraenc_libs = [libtheoraenc, libtheoradec]
+
+compat_dep = dependency('compat', required: host_machine.system() == 'openbsd')
+
+if cc.has_header_symbol('unistd.h', 'getopt_long')
+  getopt_dep = disabler()
```
eli-schwartz, line 8: A disabler acts like a viral NaN and causes all the later executable()s that try linking to it, to *also* become disabler objects and not be built. Is that the intention here?

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,269 @@
...
+else
+  message('Valgrind testing not enabled')
+endif
+
+compile_args = []
+
+cc = meson.get_compiler('c')
+
+if cc.get_argument_syntax() != 'msvc'
+  compile_args += cc.get_supported_arguments('-Wall', '-Wno-parentheses')
```
eli-schwartz, line 45: I don't see where this variable is used, hmm...

(Also, -Wall is handled by `-Dwarning_level=1`, the default warning level.)

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,269 @@
...
+endif
+
+m_dep = cc.find_library('m', required: false)
+
+if get_option('collect-metrics').enabled()
+  add_project_arguments('-DOC_COLLECT_METRICS', language: 'c')
+endif
```
eli-schwartz, line 52: This should be a boolean, it's either true or false and it's confusing if `-Dcollect-metrics=auto` and `-Dcollect-metrics=enabled` do exactly the same and both mean "unconditionally enable metrics".

eli-schwartz (changes requested):

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,269 @@
...
+
+cairo_dep = dependency('cairo', required: false)
+
+if telemetry.allowed()
+  if cairo_dep.found()
+    conf.set('HAVE_CAIRO', true)
+  else
+    message('libcairo not found -- not compiling telemetry output support ')
+  endif
+endif
```
eli-schwartz, line 116: Seems like this should all just be:

```meson
cairo_dep = dependency('cairo', required: get_option('telemetry'))
if cairo_dep.found()
    conf.set('HAVE_CAIRO', true)
endif
```

subprojects/packagefiles/theora/tests/meson.build
```diff
@@ -0,0 +1,109 @@
...
+
+tests += {'granulepos_theora': granulepos_theora}
+
+foreach name, exe : tests
+  if conf.get('HAVE_VALGRIND', false)
+    test(name, valgrind_exe, args: test_env + exe)
+  else
+    test(name, exe)
+  endif
+endforeach
```
eli-schwartz, line 109: I wonder if maybe this wouldn't be better off as:

```meson
if valgrind_exe.found()
    add_test_setup('valgrind', exe_wrapper: [valgrind] + test_env)
endif
```

You could then run the entire test suite with `meson test --setup=valgrind` or even pass `is_default: true` to the test setup definition.

...


While I'm on that topic, why is HAVE_VALGRIND a config.h header define? It's certainly not used in the codebase according to my code search. Maybe this doesn't even need an option in meson_options.txt -- simply look for it if there, and if so, define a test setup but not by default.

subprojects/packagefiles/theora/meson.build
amyspark: Ported from https://gitlab.xiph.org/xiph/theora/-/blob/master/configure.ac#L166, but I'll remove it.

subprojects/packagefiles/theora/meson.build
amyspark: `--enable-doc` in ./configure, here it'd be `-Ddoc=enabled`.

subprojects/packagefiles/theora/examples/meson.build
amyspark: No, this should be a dummy `declare_dependency`.

subprojects/packagefiles/theora/meson.build
amyspark: Hm, these should be passed on to the libraries in lib, must've slipped when I rebased the patches for 1.1.x...

subprojects/packagefiles/theora/meson.build
amyspark: I'm using feature flags because I get Meson's boolean interpretation always mixed up.

e.g. your change would imply we'd have `-Dcollect-metrics=true` and also `-Dtelemetry=enabled`, but Meson won't recognise `enabled` for its obvious meaning and error out if it's assigned to a boolean.

subprojects/packagefiles/theora/tests/meson.build
amyspark: > While I'm on that topic, why is HAVE_VALGRIND a config.h header define?

It's a carry-over from my FFmpeg wrap inspiration, tossing everything into the configure object. If Meson has test setups (TIL!) then I should be able to simplify all this.

subprojects/packagefiles/theora/meson.build
eli-schwartz: There's no need to emulate upstream's handling of whether or not to error out vs. disable a feature in configure.ac -- that's a bug in their configure.ac if it ignores the user's instructions. :)

But if you were going to copy configure.ac that closely, you'd use meson boolean options for everything.

subprojects/packagefiles/theora/meson.build
amyspark: > it's confusing if -Dcollect-metrics=auto and -Dcollect-metrics=enabled do exactly the same

As per https://mesonbuild.com/Reference-manual_returned_feature.html#featureenabled `feature.enabled()` returns false if it's set to `auto`. Did you mean `feature.allowed()`?

eli-schwartz: Err, probably.

You could equally say that if you use `if get_option('collect-metrics').enabled()` then passing:
- `-Dcollect-metrics=auto`, and 
- `-Dcollect-metrics=disabled`

do exactly the same thing. The general point remains the same, that this isn't much of a tristate option if you merge the meaning of two states.

subprojects/packagefiles/theora/meson.build
```diff
@@ -0,0 +1,224 @@
...
+        test_tex_packages,
+        '--pdflatex',
+        pdflatex,
+        pkg,
+        check: spec.enabled(),
+      )
+      if pkg_test.returncode() != 0
+        tex_pkg_ok = false
+        break
+      endif
```
eli-schwartz, line 210: But over here, you've finished doing some manual checks for tex packages, and discovered that the spec cannot be built even though the programs are available. What if spec was enabled, instead of auto? This should probably be:

```meson
if pkg_test.returncode() != 0
  if spec.enabled()
    error('Format Specification is required, but cannot be built due to missing tex packages')
  else
    tex_pkg_ok = false
    break
  endif
endif

Comment (amyspark) [reply subprojects/packagefiles/theora/meson.build:210]: The `run_command` above was designed to [error out](https://mesonbuild.com/Reference-manual_functions.html#run_command), as I'm using `check: spec.enabled()`.

e.g. with -Dspec=auto 
```
Program doc/spec/check-needed-tex-packages.py found: YES (C:\Program Files\Python310\python.exe E:\wrapdb\subprojects\libtheora-1.1.1\doc/spec/check-needed-tex-packages.py)
Program doc/spec/test-tex-packages.py found: YES (C:\Program Files\Python310\python.exe E:\wrapdb\subprojects\libtheora-1.1.1\doc/spec/test-tex-packages.py)
Message: Checking for packages in E:/wrapdb/subprojects/libtheora-1.1.1/doc/spec/spec.tex...
Message: for Tex package latexsym
Message: *** Format Specification will not built.
```

and with `-Dspec=enabled`
```
Message: Checking for packages in E:/wrapdb/subprojects/libtheora-1.1.1/doc/spec/spec.tex...
Message: for Tex package latexsym

meson.build:197:6: ERROR: Command "C:\Program Files\Python310\python.exe E:\wrapdb\subprojects\libtheora-1.1.1\doc/spec/test-tex-packages.py --pdflatex D:\texlive\2022\bin\win32\pdflatex.EXE 
latexsym" failed with status 1.

A full log can be found at E:\wrapdb\subprojects\libtheora-1.1.1\build\meson-logs\meson-log.txt
```

Your alternative does make the error more obvious, so I'll take it.

Comment (eli-schwartz) [reply subprojects/packagefiles/theora/meson.build]: Right. I missed that indeed. But the error message is superior anyway. :D

File: subprojects/packagefiles/theora/meson.build [hunk 22]
```diff
@@ -0,0 +1,57 @@
...
+
+m_dep = cc.find_library('m', required: false)
+
+if get_option('collect-metrics')
+  add_project_arguments('-DOC_COLLECT_METRICS', language: 'c')
+endif
+
+ogg_dep = dependency('ogg', version: '>= 1.1', required: true)
+
+asm = not get_option('asm').disabled()
```
neheb, line 110: Probably replace not disabled with auto or enabled.

subprojects/packagefiles/theora/meson.build
```diff
@@ -75,4 +119,8 @@ subdir('include')
 
 subdir('lib')
 
+if not get_option('examples').disabled()
```
neheb, line 136: Same here.

amyspark:
@neheb @eli-schwartz changes pushed, please check if it's okay for merging?

jpakkane:
All reported issues seem to have been fixed and CI is green. Merging, any remaining issues can be dealt with in followups.

xclaesse:
@amyspark I gave this a quick try to build GStreamer, but I got that error:
```
cc  -o subprojects/gst-plugins-base/tests/check/pipelines_theoraenc subprojects/gst-plugins-base/tests/check/pipelines_theoraenc.p/pipelines_theoraenc.c.o -Wl,--as-needed -Wl,--no-undefined -Wl,-Bsymbolic-functions -Wl,-z,nodelete '-Wl,-rpath,$ORIGIN/../../../gstreamer/gst:$ORIGIN/../../../gstreamer/libs/gst/base:$ORIGIN/../../../gstreamer/libs/gst/net:$ORIGIN/../../../gstreamer/libs/gst/check:$ORIGIN/../../gst-libs/gst/audio:$ORIGIN/../../gst-libs/gst/tag:$ORIGIN/../../../orc/orc:$ORIGIN/../../gst-libs/gst/video:$ORIGIN/../../gst-libs/gst/pbutils:$ORIGIN/../../gst-libs/gst/rtp:$ORIGIN/../../gst-libs/gst/rtsp:$ORIGIN/../../gst-libs/gst/allocators:$ORIGIN/../../gst-libs/gst/app:$ORIGIN/../../gst-libs/gst/fft:$ORIGIN/../../gst-libs/gst/riff:$ORIGIN/../../gst-libs/gst/sdp:$ORIGIN/../../../libtheora-1.1.1/lib' -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gstreamer/gst -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gstreamer/libs/gst/base -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gstreamer/libs/gst/net -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gstreamer/libs/gst/check -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/audio -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/tag -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/orc/orc -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/video -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/pbutils -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/rtp -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/rtsp -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/allocators -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/app -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/fft -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/riff -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/gst-plugins-base/gst-libs/gst/sdp -Wl,-rpath-link,/home/xclaesse/programmation/gstreamer/builddir/subprojects/libtheora-1.1.1/lib -Wl,--start-group subprojects/gstreamer/gst/libgstreamer-1.0.so.0.2102.0 subprojects/gstreamer/libs/gst/base/libgstbase-1.0.so.0.2102.0 subprojects/gstreamer/libs/gst/net/libgstnet-1.0.so.0.2102.0 subprojects/gstreamer/libs/gst/check/libgstcheck-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/audio/libgstaudio-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/tag/libgsttag-1.0.so.0.2102.0 subprojects/orc/orc/liborc-0.4.so.0.33.0 subprojects/gst-plugins-base/gst-libs/gst/video/libgstvideo-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/pbutils/libgstpbutils-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/rtp/libgstrtp-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/rtsp/libgstrtsp-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/allocators/libgstallocators-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/app/libgstapp-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/fft/libgstfft-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/riff/libgstriff-1.0.so.0.2102.0 subprojects/gst-plugins-base/gst-libs/gst/sdp/libgstsdp-1.0.so.0.2102.0 subprojects/libtheora-1.1.1/lib/libtheoraenc.so -lm /usr/lib/x86_64-linux-gnu/libglib-2.0.so /usr/lib/x86_64-linux-gnu/libgobject-2.0.so /usr/lib/x86_64-linux-gnu/libgmodule-2.0.so -pthread /usr/lib/x86_64-linux-gnu/libz.so /usr/lib/x86_64-linux-gnu/libgio-2.0.so /usr/lib/x86_64-linux-gnu/libpng16.so /usr/lib/x86_64-linux-gnu/libogg.so -Wl,--end-group
/usr/bin/ld: subprojects/gst-plugins-base/tests/check/pipelines_theoraenc.p/pipelines_theoraenc.c.o: in function `check_libtheora':
/home/xclaesse/programmation/gstreamer/builddir/../subprojects/gst-plugins-base/tests/check/pipelines/theoraenc.c:63: undefined reference to `theora_version_number'
```

neheb:
hmm so the def file uses CRLF line endings. Wonder why https://github.com/mesonbuild/wrapdb/blob/master/.gitattributes did not modify it on commit...

PR #799, opened 2022-11-26
lame: fix wrap and add fixes from other PR

neheb:
Added https://github.com/mesonbuild/wrapdb/pull/791 and actually fix wrap to work as a subproject.

ping @amyspark


neheb:
Funny. frontend binaries compile now.

releases.json
```diff
@@ -887,6 +887,7 @@
       "mpglib"
     ],
     "versions": [
+      "3.100-6",
```
amyspark, line 890: You'll need to reorder this one to the front (and maybe co-authored-by it?).

subprojects/packagefiles/lame/meson_options.txt
```diff
@@ -1,4 +1,6 @@
-option('decoder', type : 'boolean', value : false,
-       description : 'Build the decoder')
-option('tools', type : 'boolean', value : false,
+option('decoder', type : 'boolean', value : true,
+       description : 'Build the mpg123 decoder')
+option('iconv', type : 'feature',
+       description : 'Build tools with iconv support')
+option('tools', type : 'feature',
```
amyspark, line 5: Is the default value `auto` here? If so, you could make `get_option('decoder').disable_auto_if(not curses_dep.found())` and then use the returned value with `decoder.allowed()` for clarity.

amyspark (approved):
Just those two nitpicks, otherwise LGTM. Closing mine.

releases.json
neheb: Not sure what you mean.

subprojects/packagefiles/lame/meson_options.txt
neheb: feature options default to auto yes. Not sure what decoder has to do with tools. It probably has to remain enabled as there is a dependency listed in the wrap file.

releases.json
amyspark: After the version bump come 10 commits introducing further changes. These should all come *before* the bump?

imagen

neheb: Fixed.

PR #800, opened 2022-11-27
Add nowide

Stepland:
Hi !

This is my very first time contributing a wrap, sorry if the PR is not as polished as it could be, I'm really not that knowledgeable in build systems in general.

I'm looking for feedback on some specific points : 

**wrap name**

[nowide](https://github.com/boostorg/nowide) is apparently part of boost by default, but this wrap downloads the non-boost "standalone" variant specifically. Should it be called `nowide-standalone` then ?

**required meson version**

I have it set to `>=0.60.3` which seems a bit restrictive since I don't even know if I'm using any recent features.

**override_dependency**

To be honest I don't even really understand what this does but I saw that some of the other wraps don't use it, should I remove it ?

Thanks in advance for you help.

### DIFF
--- a/subprojects/packagefiles/nowide/meson.build
+++ b/subprojects/packagefiles/nowide/meson.build
@@ -20,10 +20,14 @@ subdir('src')  # fills in 'sources'
 
 includes = include_directories('include')
 
+install_subdir('include/nowide', install_dir: get_option('includedir'))
+
 nowide_lib = library(
     'nowide',
     sources,
     include_directories: includes,
+    version: meson.project_version(),
+    install: true
 )
 
 nowide_dep = declare_dependency(
--- a/subprojects/packagefiles/nowide/meson.build
+++ b/subprojects/packagefiles/nowide/meson.build
@@ -4,7 +4,7 @@ project(
     version: '11.2.0',
     license: 'BSL-1.0',
     default_options : ['cpp_std=c++14'],
-    meson_version : '>=0.37.0',
+    meson_version : '>=0.53.0',
 )
 
 compile_args = []
@@ -31,3 +31,7 @@ nowide_dep = declare_dependency(
     include_directories: includes,
     link_with: nowide_lib,
 )
+
+if get_option('tests')
+    subdir('test')
+endif
--- a/subprojects/packagefiles/nowide/meson.build
+++ b/subprojects/packagefiles/nowide/meson.build
@@ -1,0 +1,33 @@
+project(
+    'nowide',
+    'cpp',
+    version: '11.2.0',
+    license: 'BSL-1.0',
+    default_options : ['cpp_std=c++14'],
+    meson_version : '>=0.37.0',
+)
+
+compile_args = []
+
+if get_option('default_library') != 'static'
+    add_project_arguments('-DNOWIDE_DYN_LINK', language: 'cpp')
+    compile_args += ['-DNOWIDE_DYN_LINK']
+endif
+
+sources = []
+
+subdir('src')  # fills in 'sources'
+
+includes = include_directories('include')
+
+nowide_lib = library(
+    'nowide',
+    sources,
+    include_directories: includes,
+)
+
+nowide_dep = declare_dependency(
+    compile_args: compile_args,
+    include_directories: includes,
+    link_with: nowide_lib,
+)
--- a/ci_config.json
+++ b/ci_config.json
@@ -333,6 +333,11 @@
       "mpdecimal:tests=true"
     ]
   },
+  "nowide": {
+    "build_options": [
+      "nowide:tests=true"
+    ]
+  },
   "openh264": {
     "brew_packages": [
       "nasm"
--- a/releases.json
+++ b/releases.json
@@ -1602,6 +1602,14 @@
       "1.3.0-1"
     ]
   },
+  "nowide": {
+    "dependency_names": [
+      "nowide"
+    ],
+    "versions": [
+      "11.2.0-1"
+    ]
+  },
   "ogg": {
     "dependency_names": [
       "ogg"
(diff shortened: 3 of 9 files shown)


eli-schwartz:
> wrap name

What does upstream call it? Their CMake files are a bit confusingly indirect and I can't tell just by reading it whether the project is called boost_nowide or nowide. Nor can I tell what their CMake `find_package()` export name is!

> required meson version

If you set the minimum required version to 0.37.0 meson will print a warning for each feature newer than that which you use. That's the version where we first started tracking the version of features, more or less.

> override_dependency

You don't really need that, because the `[provide]` section of the wrap also handles this.

eli-schwartz:
I would also set up the test/ directory so that the CI can verify the library is built correctly. Probably also a good idea to mark the library and headers for installation.

Stepland:
Thanks for the quick response !

Hmm, for the name, it looks like it's just called `nowide` on both the default and the `standalone` branch (because of [this line ?](https://github.com/boostorg/nowide/blob/cdcd7043c6ac0b1b67dd91aa50888bd17e3d0159/CMakeLists.txt#L68))

> I would also set up the test/ directory so that the CI can verify the library is built correctly. Probably also a good idea to mark the library and headers for installation.

Will do !

eli-schwartz:
> Hmm, for the name, it looks like it's just called `nowide` on both the default and the `standalone` branch (because of [this line ?](https://github.com/boostorg/nowide/blob/cdcd7043c6ac0b1b67dd91aa50888bd17e3d0159/CMakeLists.txt#L68))

Excellent, that means there's nothing to fix.

(We need to match the cmake name, because `dependency('nowide')` can check to see if a `nowideConfig.cmake` exists usable via cmake `find_package(nowide)`.)

Stepland:
Ok I added support for ... most of the tests ? There were two strange things :

eli-schwartz:
A couple of the files show "No newline at end of file", could you please fix those up with `git commit --amend && git push --force` to fix the second commit in place? :)



> One test (`test_iostream`) has a specific part that can be skipped with a build option because apparently it can fail in some environments ?

This was implemented upstream in commit:
```
 CI: Skip flaky test-part on affected Appveyor images

Creating console input events fails sporadically on recent Appveyor images when using CMake/CTest.
Add an option to disable that and use it for the recent Appveyor image
```

Sounds overly specific to both Appveyor and cmake, so I would not worry about it.



> Another test (`test_iostream_interactive`) uses a sub-cmake file to pass some text to a compiled test executable via stdin and make sure the same text appears _somewhere_ in stdout, I don't know how to do that in meson so I just made a python script that's (hopefully) equivalent. To me this looks _almost_ possible to do in pure meson but not quite

In both meson and cmake, you cannot mock up the stdin of a test and need a wrapper script. In cmake's case, that's a cmake script, in Meson's case, that's your choice of scripting environments but it's simpler to write a cross platform python one than to write a posix sh script and a Windows batch file and alternate between the two based on platform. :D :D

Stepland:
> A couple of the files show "No newline at end of file", could you please fix those up with `git commit --amend && git push --force` to fix the second commit in place? :)

Done !

> This was implemented upstream in commit:
> 
> ```
>  CI: Skip flaky test-part on affected Appveyor images
> 
> Creating console input events fails sporadically on recent Appveyor images when using CMake/CTest.
> Add an option to disable that and use it for the recent Appveyor image
> ```
> 
> Sounds overly specific to both Appveyor and cmake, so I would not worry about it.

Hmm okay

> it's simpler to write a cross platform python one than to write a posix sh script and a Windows batch file and alternate between the two based on platform. :D :D

Yep that exactly what I was trying to avoid !

Stepland:
Aaand I forgot to install the lib and headers :facepalm:

eli-schwartz:
CI failure indicates that dllexport isn't correctly set up for Windows. After a bit of poking around, I believe the issue is that you need to define `-DNOWIDE_DYN_LINK` both via `add_project_arguments()` and in the declared dependency for other projects to consume, if `get_option('default_library') != 'static'`. This sets up both dllexport/dllimport for Windows, and `gnu_symbol_visibility: 'inlineshidden'` prep for GNU compilers.

Stepland:
Oh,  didn't know about that :thinking:, well I've just added it by mimicking another wrap that does something similar

I still need to add lib and headers installation ...

eli-schwartz:
Well, the library build now runs and passes all tests. :)

Stepland:
yay !

I've just added lib and headers installation but I don't really know what it does, how do you test that this works ?

Stepland:
forgot the ending newlines again ...

subprojects/packagefiles/nowide/meson.build
```diff
@@ -20,10 +20,14 @@ subdir('src')  # fills in 'sources'
 
 includes = include_directories('include')
 
+subdir('include/nowide')  # calls install_headers()
```
eli-schwartz, line 23: You could actually simplify this with:
```meson
install_subdir('include/nowide', install_dir: get_option('includedir'))
```

Since the entire directory gets installed as-is.

Stepland: TIL ...

eli-schwartz: Since Meson 0.63.0 -- but this is pretty bleeding edge -- you can also do `install_headers('list', 'of', 'headers/in/subdirs', preserve_path: true)` to name each installed file but not have to duplicate relative path information.

But really, install_subdir is fine for common cases like this where there's a well-formed include directory.

Stepland:
Wow, thanks a ton for the quick merge ! :tada:

eli-schwartz:
No, thank *you* for contributing it. :p

PR #806, opened 2022-12-04
Update OpenSSL to 3.0.7

nazar-pc:
Node.js dropped 2 variants of PPC platform, but I don't think too many people will cry because of it.


nazar-pc:
Hm... something has changed in this repo, generator is not supposed to be running on non-Linux, corresponding keys are `false` in `ci_config.json` :thinking:

eli-schwartz:
Can you add an explicit check whether the generator has previously been run (and the wrap is running with a release-bundled snapshot of generated files) and error out with `error('running the generator is not supported except on Linux')`?

eli-schwartz:
> Alternatively preparation can run on Linux and then compilation can continue on macOS/Windows.

If that's something you're interested in investigating, then by all means :D however currently the WrapDB simply wants to make sure that wraps which are marked as only for a specific OS due to e.g. missing dependencies start getting checked once those deps are added to the WrapDB. Anonymous failures are assumed to be broken -- errors that have the words "not supported" or "unsupported" etc. tell the CI "yeah this is expected to error out".

nazar-pc:
Well, that didn't work either :shrug:

eli-schwartz:
Did you try the explicit wording I suggested?

nazar-pc:
Of course not :upside_down_face:  

Appears to be good now :+1:

subprojects/openssl.wrap
```diff
@@ -1,8 +1,8 @@
 [wrap-file]
-directory = openssl-3.0.2
-source_url = https://www.openssl.org/source/openssl-3.0.2.tar.gz
-source_filename = openssl-3.0.2.tar.gz
-source_hash = 98e91ccead4d4756ae3c9cde5e09191a8e586d9f4d50838e7ec09d6411dfdb63
+directory = openssl-3.0.7
+source_url = https://www.openssl.org/source/openssl-3.0.7.tar.gz
+source_filename = openssl-3.0.72.tar.gz
```
nazar-pc, line 4: Hm... I made a typo here, but it still works fine. I'm a bit confused :thinking:

eli-schwartz: It will simply use a weird cache filename in `subprojects/packagecache/`, but the file's remote location still matches `source_url = ` and the contents of the file still match `directory = `.

So it all works.

PR #802, opened 2022-11-28
rdkafka: use openssl as dependency

neheb:
Should fix MSVC compilation.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

PR #807, opened 2022-12-06
libxml2: update build to latest CMake

neheb:
Minizes the diff between upstream generated config.h and meson one.


neheb:
@eli-schwartz any idea why test fails? Some CI update introduced it.

eli-schwartz:
```
  1/1 dict FAIL            0.03s   (exit status 3221225785 or signal 3221225657 SIGinvalid)
```

This doesn't seem like a CI-specific problem, the actual binary itself seems to be crashing or something.

neheb:
So the third prior build all has

```
2022-11-20T02:45:36.4269331Z libxml2| Run-time dependency liblzma found: NO (tried pkgconfig and cmake)
2022-11-20T02:45:36.4269957Z libxml2| Run-time dependency zlib found: NO (tried pkgconfig, cmake and system)
```

this one has

```
2022-12-06T06:03:50.7337591Z libxml2| Run-time dependency liblzma found: YES 5.2.4
2022-12-06T06:03:50.7338016Z libxml2| Run-time dependency zlib found: YES 1.2.11
```

My guess is some CI update, where meson is finding some cmake dependency and failing to add the proper paths. See https://github.com/mesonbuild/wrapdb/pull/802

neheb:
I'll mention I can't reproduce this with MinGW.

eli-schwartz:
That is an interesting insight, and I suspect the problem is in fact Strawberry Perl.

We blacklist Strawberry pkg-config because Strawberry Perl's entire concept is broken, basically. But cmake might find it on its own -- and since 3.25 it allows finding `libfoo.a` style libraries, but does so using its usual "look in every lib subdirectory off of PATH", which finds way too many environments...

(Hardcoding the pkg-config method is wrong. We should be able to find with "system" too. But this shows what the CI problem is. The actual solution is probably rm -rf the entire Strawberry Perl installation.)

neheb:
@eli-schwartz will you be pushing a change fixing this?

neheb:
updated. Better supports MinGW now.

@eli-schwartz ping.

PR #809, opened 2022-12-06
Add websocketpp

nirbheek:
header-only c++ dependency

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -2343,6 +2343,14 @@
       "1.24-1"
     ]
   },
+  "websocketpp": {
+    "dependency_names": [
+      "websocketpp"
+    ],
+    "versions": [
+      "0.8.2-1"
+    ]
+  },
   "wren": {
     "dependency_names": [
       "wren"
--- a/subprojects/packagefiles/websocketpp/meson.build
+++ b/subprojects/packagefiles/websocketpp/meson.build
@@ -1,0 +1,5 @@
+project('websocketpp', 'cpp', version: '0.8.2')
+
+websocketpp_dep = declare_dependency(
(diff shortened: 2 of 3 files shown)


eli-schwartz:
The wrap should be added to releases.json.

What about adding the examples/tests behind an option (and using it in CI)?

nirbheek:
> The wrap should be added to releases.json.

Thanks, I knew I was forgetting something :)

> What about adding the examples/tests behind an option (and using it in CI)?

What's the point of doing that? It's just a fancy `include_directories:`. This sort of thing should probably be handled by the wrap system automatically via special `[provide]` syntax even.

eli-schwartz:
For example, people using the wrap could then test that websocketpp itself doesn't have bugs on their system or compiler.

nirbheek:
I looked at that, and it's more work than I am willing to spend on this right now. I'm happy to let someone else complete it if they want to.

releases.json
```diff
@@ -2343,6 +2343,14 @@
       "1.24-1"
     ]
   },
+  "websocketpp": {
+    "dependency_names": [
+      "websocketpp"
+    ],
+    "versions": [
+      "0.8.2-0"
```
eli-schwartz, line 2351:
```suggestion
      "0.8.2-1"
```

It's the first published wrap for this websocketpp release.

nirbheek: Ah, fdk-aac and a couple of other wraps use `-0`. Fixed.

eli-schwartz: It's a very wishy-washy thing to begin with. I just tend to view "0" as meaning "locally testing this, will bump for publication", so it looks off. :D

PR #810, opened 2022-12-07
theora: Add "missing" dependency on theoradec to theoraenc

amyspark:
:wave:

This is a PR to fix a "bug" in the theora wrap. 

When I modeled the pkg-config files, I didn't notice that theoraenc "depends" on theoradec too through its pkg-config module. This bogus dependency is actually relied on by any consumer of theoraenc to access, at the very least, the `theora_version_string` and `th_version_string` functions, which are otherwise invisible as specified by the corresponding version scripts/module definition files.

See https://github.com/mesonbuild/wrapdb/pull/798#issuecomment-1339834903

cc @xclaesse


subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -199,14 +199,19 @@ libtheoraenc = library(
...
+# version scripts/.def files.
 pkg.generate(
   libtheoraenc,
+  requires: libtheoradec,
   description: 'Theora video codec (encoder)',
   version: meson.project_version(),
 )
 
 theoraenc_dep = declare_dependency(
-  link_with: libtheoraenc,
```
neheb, line 209: AFAIK, an alternative to this would be to add libtheoradec_dep to dependencies.

jpakkane: If the include directories are the same, then this is simpler.

xclaesse: They have the same incdir, so I think this is fine like this.

PR #811, opened 2022-12-07
flatbuffers: update to 22.11.23

jmillan:


releases.json
```diff
@@ -417,6 +417,7 @@
       "flathash"
     ],
     "versions": [
+      "22.11.23",
```
jpakkane, line 420: This should be `22.11.23-1`.

jmillan: Thanks, done.

jmillan:
Thank you 💯

PR #814, opened 2022-12-08
WIP: gmp

Stepland:
I'm creating this wip PR so I can ask for help.

`gmp` is a project with a *huge* autotools script. Since this PR is my very first attempt at porting an autotools project, I'm really unsure what I should keep or not. The dedicated wiki page on porting autotools projects didn't fully answer the questions I have.

A *very* large portion of `gmp`'s configure script is dedicated to setting flags and options of all sorts for a dauntingly long list of compilers and supported hosts : 

- default compiler flags
- sets of flags to be tested and used if working
- available compilers for the host
- available abis for each compiler
- search path for assembly implementations of the "mpn" functions
- tests to be performed for given compiler + abi combinations
- some precompiled things related with profiling ? (`speed_cyclecounter_obj`, `cyclecounter_size`)
- etc ...

This list is not even exhaustive, you can have a look at the top of `configure.py` to see everything that's currently handled by my port.

1. Are there things in this (non-exhaustive) list we don't need to port over to meson ?
2. The original script relies **heavily** on pattern matching on the `host` and `host_cpu` values. Since i don't know how to do that in meson, I resorted to porting that part to python because and use [`fnmatch()`](https://docs.python.org/3/library/fnmatch.html?highlight=fnmatch#module-fnmatch). Isn't it a problem for a meson wrap to rely on python ?

### DIFF
--- a/subprojects/packagefiles/gmp/meson.build
+++ b/subprojects/packagefiles/gmp/meson.build
@@ -1,0 +1,315 @@
+project(
+  'gmp',
+  'c',
+  'cpp',
+  version : '6.2.1',
+)
+
+conf = configuration_data()
+
+conf.set('WANT_ASSERT', get_option('assert').to_int())
+conf.set('WANT_FFT', get_option('fft').to_int())
+conf.set('WANT_OLD_FFT_FULL', get_option('old-fft-full').to_int())
+
+profiling = get_option('profiling')
+if profiling != 'no'
+  conf.set('WANT_PROFILING_' + profiling.to_upper(), 1)
+endif
+fomit_frame_pointer = []
+if profiling == 'gprof':
+  fomit_frame_pointer += ['-fomit-frame-pointer']
+endif
+
+fake_cpuid = get_option('fake-cpuid')
+conf.set('WANT_FAKE_CPUID', fake_cpuid.to_int())
+
+assembly = get_option('assembly')
+fat = get_option('fat')
+
+if fat and not assembly
+  error('when doing a fat build, disabling assembly will not work')
+endif
+
+if fake_cpuid and not fat
+  error('enabling fake-cpuid requires enabling fat')
+endif
+
+host_machine_cpu = host_machine.cpu()
+conf.set('HAVE_HOST_CPU_' + host_machine.cpu(), 1)
+conf.set('HAVE_HOST_CPU_FAMILY_' + host_machine.cpu_family(), 1)
+
+abi_list = ['standard']
+cc_list = ['gcc', 'cc']
+gcc_cflags = ['-O2', '-pedantic']
+gcc_64_cflags = ['-O2', '-pedantic']
+cc_cflags = ['-O']
+cc_64_cflags = ['-O']
+SPEED_CYCLECOUNTER_OBJ = []
+cyclecounter_size = 2
+mpn_search_path = []
+
+conf.set('HAVE_HOST_CPU_FAMILY_power', 0)
+conf.set('HAVE_HOST_CPU_FAMILY_powerpc', 0)
+
+variables = {
+  'alpha': {
+    'mpn_search_path': []
+  }
+}
+
+
+check_headers = [
+  'alloca.h',
+  'dlfcn.h',
+  'fcntl.h',
+  'float.h',
+  'inttypes.h',
+  'invent.h',
+  'langinfo.h',
+  'locale.h',
+  'machine/hal/sysinfo.h',
+  'memory.h',
+  'nl/types.h',
+  'stdint.h',
+  'stdlib.h',
+  'strings.h',
+  'string.h',
+  'sys/attributes.h',
+  'sys/iograph.h',
+  'sys/mman.h',
+  'sys/param.h',
+  'sys/processor.h',
+  'sys/pstat.h',
+  'sys/resource.h',
+  'sys/stat.h',
+  'sys/sysctl.h',
+  'sys/sysinfo.h',
+  'sys/syssgi.h',
+  'sys/systemcfg.h',
+  'sys/times.h',
+  'sys/time.h',
+  'sys/types.h',
+  'unistd.h',
+]
+
+cc = meson.get_compiler('c')
+
+foreach h : check_headers
+  if cc.has_header(h)
+    conf.set('HAVE_' + h.underscorify().to_upper(), 1)
+  endif
+endforeach
+
+check_functions = [
+  ['HAVE_ALARM', 'alarm', '#include<unistd.h>'],
+  ['HAVE_ALLOCA', 'alloca', '#include<alloca.h>'],
+# check token ['HAVE_ATTRIBUTE_CONST']
+# check token ['HAVE_ATTRIBUTE_MALLOC']
+# check token ['HAVE_ATTRIBUTE_MODE']
+# check token ['HAVE_ATTRIBUTE_NORETURN']
+# check token ['HAVE_ATTR_GET']
+# check token ['HAVE_CALLING_CONVENTIONS']
+  ['HAVE_CLOCK', 'clock', '#include<time.h>'],
+  ['HAVE_CLOCK_GETTIME', 'clock_gettime', '#include<time.h>'],
+# check token ['HAVE_CPUTIME']
+# check token ['HAVE_DECL_FGETC']
+# check token ['HAVE_DECL_FSCANF']
+# check token ['HAVE_DECL_OPTARG']
+# check token ['HAVE_DECL_SYS_ERRLIST']
+# check token ['HAVE_DECL_SYS_NERR']
+# check token ['HAVE_DECL_UNGETC']
+# check token ['HAVE_DECL_VFPRINTF']
+# check token ['HAVE_DOUBLE_IEEE_BIG_ENDIAN']
+# check token ['HAVE_DOUBLE_IEEE_LITTLE_ENDIAN']
+# check token ['HAVE_DOUBLE_IEEE_LITTLE_SWAPPED']
+# check token ['HAVE_DOUBLE_VAX_D']
+# check token ['HAVE_DOUBLE_VAX_G']
+# check token ['HAVE_DOUBLE_CRAY_CFP']
+  ['HAVE_GETPAGESIZE', 'getpagesize', '#include<unistd.h>'],
+# check token ['HAVE_GETRUSAGE']
+# check token ['HAVE_GETSYSINFO']
+  ['HAVE_GETTIMEOFDAY', 'gettimeofday', '#include<sys/time.h>'],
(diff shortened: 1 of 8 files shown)


jpakkane:
The unofficial motto of WrapDB is "never let perfect be the enemy of good". Typically you'd need to support x86_64 on Linux and preferably also Windows and macOS. After that there are nice-to-haves like arm support but it's not typically mandatory for merging. Fine tuning and polish can be added in subsequent MRs.

> Isn't it a problem for a meson wrap to rely on python ?

No. Meson is implemented in Python so it is always available.

eli-schwartz:
Default compiler flags -- stuff like optimization or debug symbols is handled already by meson, so you can skip that. Lists of warning flags can probably be ignored in favor of just deferring support for warnings, that's mostly just something project developers care about.

subprojects/packagefiles/gmp/meson.build
```diff
@@ -0,0 +1,298 @@
+project(
+  'gmp',
+  'c',
+  'cpp',
+  version : '6.2.1',
+)
+
+conf = configuration_data()
+
+conf.set('WANT_ASSERT', get_option('assert').to_int())
```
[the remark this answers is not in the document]
neheb: Autotools projects typically do if x then set(‘’,1)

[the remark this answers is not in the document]
Stepland: okay !

subprojects/packagefiles/gmp/meson.build
```diff
@@ -0,0 +1,548 @@
...
+  ['vfprintf', 'stdio.h'],
+  ['sys_errlist', 'errno.h'],
+  ['sys_nerr', 'errno.h'],
+]
+
+foreach f : check_functions
+  if cc.has_header_symbol(f.get(1), f.get(0))
+    conf.set('HAVE_DECL_' + f.get(0).to_upper(), 1)
+  endif
+endforeach
```
Stepland, line 269: really unsure about this bit

Stepland: In configure.ac it's this :

```m4
# Reasons for testing:
#   optarg - not declared in mingw
#   fgetc, fscanf, ungetc, vfprintf - not declared in SunOS 4
#   sys_errlist, sys_nerr - not declared in SunOS 4
#
# optarg should be in unistd.h and the rest in stdio.h, both of which are
# in the autoconf default includes.
#
# sys_errlist and sys_nerr are supposed to be in <errno.h> on SunOS according
# to the man page (but aren't), in glibc they're in stdio.h.
#
AC_CHECK_DECLS([fgetc, fscanf, optarg, ungetc, vfprintf])
AC_CHECK_DECLS([sys_errlist, sys_nerr], , ,
[#include <stdio.h>
#include <errno.h>])
```

Stepland: woops and there's even a typo

eli-schwartz: Can you clarify what about it you aren't sure about?

subprojects/packagefiles/gmp/meson.build
```diff
@@ -0,0 +1,548 @@
+project(
+  'gmp',
+  'c',
+  version : '6.2.1',
+)
+
+add_languages('cpp', required: get_option('cxx'))
```
Stepland, line 7: `cxx` was the original name in the configure script, I'm tempted to rename it to cpp, opinions on this ?

Stepland:
> Default compiler flags -- stuff like optimization or debug symbols is handled already by meson, so you can skip that. Lists of warning flags can probably be ignored in favor of just deferring support for warnings, that's mostly just something project developers care about.

Well that's fantastic news ! It looks like only flags set by the script for `x86` are combinations of `-march=...` and `-mtune=...`.
I think this makes the python script irrelevant so I removed it.

subprojects/packagefiles/gmp/meson.build
Stepland: I'm looking at what `AC_CHECK_DECLS` does according to [this webpage](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Generic-Declarations.html) and i'm not sure it's really equivalent of `cc.has_header_symbol()`

eli-schwartz:
Uh. Looking at configure.ac, I notice `gcc_cflags_arch` which basically... defines the default -march for your host CPU that is already used anyway, and then manually passes it?

It's basically completely reimplementing the entire concept of a cross compiler.
- When you natively compile, you just run `$CC` and -march defaults to the march-ness of the CPU that GCC was compiled for, and which you are running on.
- When you cross compile, you run the cross compiler `$CC`, and -march defaults to the march-ness of the CPU that the cross compiler cross compiles *for*.

This is idk, attempting to support cross compiling by hacking a native compiler with -march flags in order emit code as if you used a cross compiler? Totally strange. This is absolutely not the job of a build description -- that goes inside the https://mesonbuild.com/Cross-compilation.html toolchain file.

Definitely drop all that. :D

Stepland:
This is harder than expected
- For a start, I was unaware, but *most* of the code of GMP is actually just CPU-specific assembly, not C, not sure how to deal with that
- the coupling with Autotools is **tight**. It's using m4 templates to create assembly files ... I have no idea what to do with these. Should I just resort to their "generic" C fallback for everything ?
- for fat binary setups, the configure script is generating a *whole* header file on its own, without using a .in file, I don't think meson's templating capabilities are powerful enough for that yet, so I think I'll be leaving fat builds out for now
- the configure script is *constantly* parsing gmp's sources with sed to check for formatted comments that define ABI compatibility, which I think is *really* weird

jpakkane:
>  Should I just resort to their "generic" C fallback for everything

That's a reasonable start. Getting things working is the first step. People can then add polishings they need. (I also suspect that most of the asm is either for obsolete platforms or no longer necessary with  modern compilers.)

eli-schwartz:
- Meson does have asm language support recently. The GNU compiler has handled its own variant out of the box for a long time, simply add the relevant .s files as sources if the compiler is gcc/clang.
- m4 isn't autotools coupling. :D It's a general-purpose templating language. One of the very earlier ones! If there's one big indictment against it, it's the fact that it's hard to get ahold of on Windows. And for that, Meson has a neat secret: someone made a wrap for it: #541
- you can include a `config.in.meson` as part of the wrap
- bizarre and painful, very sad -- I'm not sure what the options are there other than reimplementing that in python for portability, or hardcoding tons of info in meson.build. If it's just for ABI compatibility, does that mean it's not needed for static library builds? That could be an option for the initial "get it to work" first step -- simply implement static builds and note that shared builds are not available yet.

Stepland:
Well since m4 and sed look like they are only in use when compiling with the assembly files, I'll just focus on trying to get the C fallback implementation to work first.

Stepland:
> you can include a config.in.meson as part of the wrap

There already is one I believe, I meant that, if asked for a fat build, the configure script will create a completely new header file from scratch (`fat.h`) by just `echo >>`ing several things to it (without using a .in template)

eli-schwartz:
I meant you can use a template in order to properly handle the complexity of those echoes, without running afoul of the limitations inherent in using a more structured configure_file() approach.

Stepland:
I need some help for these  `__declspec()` problems :sweat: 

Here's the bit that seems to define the macro in question in `gmp-h.in` :

```c
#if defined (__GNUC__)
#define __GMP_DECLSPEC_EXPORT  __declspec(__dllexport__)
#define __GMP_DECLSPEC_IMPORT  __declspec(__dllimport__)
#endif
#if defined (_MSC_VER) || defined (__BORLANDC__)
#define __GMP_DECLSPEC_EXPORT  __declspec(dllexport)
#define __GMP_DECLSPEC_IMPORT  __declspec(dllimport)
#endif
#ifdef __WATCOMC__
#define __GMP_DECLSPEC_EXPORT  __export
#define __GMP_DECLSPEC_IMPORT  __import
#endif
#ifdef __IBMC__
#define __GMP_DECLSPEC_EXPORT  _Export
#define __GMP_DECLSPEC_IMPORT  _Import
#endif

#if __GMP_LIBGMP_DLL
#ifdef __GMP_WITHIN_GMP
/* compiling to go into a DLL libgmp */
#define __GMP_DECLSPEC  __GMP_DECLSPEC_EXPORT
#else
/* compiling to go into an application which will link to a DLL libgmp */
#define __GMP_DECLSPEC  __GMP_DECLSPEC_IMPORT
#endif
#else
/* all other cases */
#define __GMP_DECLSPEC
#endif
```
Everything I could find online mention that `__declspec` is a *Microsoft* extension ... so I don't really understand why it's used when __GNUC__ is defined ...

eli-schwartz:
It's used when a mingw GCC compiles DLL files targeting Windows.

The odd thing is that you're not supposed to check for `__GNUC__`, you're supposed to check for the win32 defines that are available in both MSVC and GCC. The way they do it here means that gmp is broken if you define the `__GMP_LIBGMP_DLL` macro unconditionally when targeting shared libraries. And on regular Unix GCC they should be using the gnu visibility attribute.

It's a mess, but just make sure to only use the activation macros on Windows.

Stepland:
Oh ... I've just now understood what the configure script meant by *multi-function files* (in case it's not a usual term : these are .c files you have to compile multiple times with different macro definitions to give you different functions)

What's the "correct" way to deal with these ?

eli-schwartz:
You could compile them into multiple static libraries, one for each macro. And then link all the static libraries together.

(Why do they do this though.)

Stepland:
Not really sure what's happening with the double definition errors on msvc, I'll check back later. But also finally some progress ! yay :tada:

Stepland:
Ok bear with me while I test out some things with the CI pipeline because I cannot for the *life* of me figure out what's wrong with my msvc setup on my windows VM I'm trying to do local tests on ...

Stepland:
I give up, it's just too hard for me.

I've been trying to compare preprocessed outputs of msvc and gcc for days, but I don't even know what I'm looking for ?

gcc is perfectly fine with having duplicate decls and inline defs :
```c
mp_limb_t __gmpn_add_1 (mp_ptr, mp_srcptr, mp_size_t, mp_limb_t) ;

extern __inline__ __attribute__ ((__gnu_inline__))
mp_limb_t
__gmpn_add_1 (/* ... */)
{
/* ... */
}
```

in basically every file.

While msvc complains about having *just*
```c
__inline
mp_limb_t
__gmpn_add_1 (mp_ptr __gmp_dst, mp_srcptr __gmp_src, mp_size_t __gmp_size, mp_limb_t __gmp_n) 
{
/* ... */
}
```

I really don't know how to dig further on this.

I'm out of motivation for the time being.

If some kind soul wants to pick up where I left off they're very welcome

PR #816, opened 2022-12-10
add sqlitecpp 3.2.0-1

ninjaoflight:
adds [SQLiteCpp](https://github.com/SRombauts/SQLiteCpp) version 3.2.0 to the list of wraps

- tested environments
  - Windows 11 22H2
     - `clang++ 15.0.5` (static compilation only)
     - `ninja 1.11.1`
     - `meson 0.64.1`
  - Arch Linux
    - `gcc 12.2.0`
    - `ninja 1.11.0`
    - `meson 0.64.1`

> note: if there is any issue regarding windows static build please see: SRombauts/SQLiteCpp#380 or just set `CXX` and `CXX_LD` environment variables to `clang++` and `lld-link`

If there is anything else required just let me know


neheb:
wrap-git is disallowed here.

ninjaoflight:
thanks for the notice,
I will wait until there is a new release of SQLiteCpp so I can update this PR to use wrap-file instead, which should not take much time

eli-schwartz:
Please also take a look at the macOS compile failure.

ninjaoflight:
> Please also take a look at the macOS compile failure.

sure!, I will open an issue on SQLiteCpp in order to check it out

subprojects/sqlitecpp.wrap
```diff
@@ -1,6 +1,9 @@
...
-depth = 1
+[wrap-file]
+directory = SQLiteCpp-3.2.1
+source_url = https://github.com/SRombauts/SQLiteCpp/archive/refs/tags/3.2.1.zip
+source_filename = sqlitecpp-3.2.1.zip
+source_hash = 584f7c36b6b169c10c890240ea5fb36318d0ea5713c908f08201a95a93642d21
+patch_directory = sqlitecpp
+
 [provide]
 sqlitecpp = sqlitecpp_dep
```
neheb, line 9: Should have a newline

subprojects/packagefiles/sqlitecpp/meson.build
```diff
@@ -0,0 +1,298 @@
...
+# sqlite3 support
+sqlite3_dep = dependency(
+                'sqlite3',
+                fallback: ['sqlite3', 'sqlite3_dep']
+)
+
+sqlitecpp_incl = [ 
+    include_directories('include')
+]
+sqlitecpp_srcs = [
```
neheb, line 42: should use files()

subprojects/packagefiles/sqlitecpp/meson.build
```diff
@@ -0,0 +1,298 @@
...
+    'src/Backup.cpp',
+    'src/Column.cpp',
+    'src/Database.cpp',
+    'src/Exception.cpp',
+    'src/Savepoint.cpp',
+    'src/Statement.cpp',
+    'src/Transaction.cpp',
+]
+sqlitecpp_args = [
+    '-Wall',
```
neheb, line 52: meson already adds Wall.

subprojects/packagefiles/sqlitecpp/meson.build
```diff
@@ -0,0 +1,298 @@
...
+sqlitecpp_deps = [
+    sqlite3_dep,
+    thread_dep,
+]
+## used to override the default sqlitecpp options like cpp standard
+sqlitecpp_opts = []
+
+## tests
+
+sqlitecpp_test_srcs = [
```
neheb, line 64: files()

subprojects/packagefiles/sqlitecpp/meson.build
```diff
@@ -0,0 +1,298 @@
...
+]
+sqlitecpp_sample2_srcs = [
+    'examples/example2/src/main.cpp',
+]
+
+# if not using MSVC we need to add this compiler arguments
+# for a list of MSVC supported arguments please check:
+# https://docs.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-alphabetically?view=msvc-170
+if not (host_machine.system() == 'windows' and cxx.get_id() == 'msvc')
+    sqlitecpp_args += [
```
neheb, line 90: could simplify with cc.get_supported_arguments.

eli-schwartz:
```
subprojects/SQLiteCpp-3.2.1/meson.build:170:8: ERROR: Unknown function "warming".
```

That's very much a blocking issue...

subprojects/packagefiles/sqlitecpp/meson.build
```diff
@@ -0,0 +1,298 @@
...
+        warming('You can disable this warning by setting SQLITE_OMIT_LOAD_EXTENSION to false')
+        sqlitecpp_args += ['-DSQLITE_OMIT_LOAD_EXTENSION']
+    endif
+endif
+
+
+
+if unix_like
+    sqlitecpp_args += [
+        '-DfPIC',
```
neheb, line 181: this is already default.

subprojects/packagefiles/sqlitecpp/meson.build
ninjaoflight: done

subprojects/sqlitecpp.wrap
ninjaoflight: done

subprojects/packagefiles/sqlitecpp/meson.build
ninjaoflight: done

subprojects/packagefiles/sqlitecpp/meson.build
ninjaoflight: done

subprojects/packagefiles/sqlitecpp/meson.build
eli-schwartz: The flip side is that hardcoding it based on get_id() is faster because you don't need to run the compiler each time to check which ones are supported.

eli-schwartz:
I thought there was an upstream meson.build file, though?

eli-schwartz:
These commits should all be squashed together using `git rebase -i upstream/master`. Currently they introduce issues in one commit and fix them in another commit.

subprojects/packagefiles/sqlitecpp/meson.build
ninjaoflight: > I thought there was an upstream meson.build file, though?

yes, however 3.2.1 was already launched today with a fix for `std::filesystem` so this time those patches should be made on this release and the next one removed, also it would be good to add a CI on that repo for meson as testing all the environments manually is hard for me

ninjaoflight:
> I thought there was an upstream meson.build file, though?

yes, however 3.2.1 was already launched today with a fix for `std::filesystem` so this time those patches should be made on this release and the next one removed, also it would be good to add a CI on that repo for meson as testing all the environments manually is hard for me

> These commits should all be squashed together using git rebase -i upstream/master. Currently they introduce issues in one commit and fix them in another commit.

sure, will make the fixes and then squash them

subprojects/packagefiles/sqlitecpp/meson.build
eli-schwartz: w.r.t. `-DfPIC`, this isn't the default anywhere -- largely because it seems to have been a typo. This defines a macro called `#define fPIC`, rather than passing the `-fPIC` option, which is an `-f` (Options for Code Generation Conventions) grouped option enabling PIC code generation.

Admittedly, `-fPIC` is indeed part of Meson, and you even get a warning if you try to pass it manually.

ninjaoflight: solved

subprojects/packagefiles/sqlitecpp/meson.build
ninjaoflight: I agree, it will be a faster option if the parameters are known in advance, as there are few arguments you can simply use this option, if in the future there were more arguments, perhaps it would be a better option to check if MSVC is used, looking on the bright side we can remove additional verification that is used only for MSVC.

Honestly I don´t know much about the meson implementation and if those values are cached for an specific compiler/setup as it could reduce setup times if they are checked across projects, but I think that is more dependent on implementation than project configuration, maybe a linter should be helpful for this cases(and for new users), as for the moment I think I will set `cxx.get_supported_arguments` as the meson.build in the project is not a 1-1 copy to the CMakeLists.txt anyways

ninjaoflight: done

eli-schwartz: It would only be cached within a project so that checking the same option again already knows if it's supported.

ninjaoflight:
as for the tests on MacOS I did them on a borrowed machine with the following specs:
- C++ `compiler: c++ (clang 14.0.0 "Apple clang version 14.0.0 (clang-1400.0.29.202)")`
- C++ linker: `c++ ld64 820.1`
- Ninja: `ninja 1.11.1.git.kitware.jobserver-1`
- Meson: `meson version 0.64.1`
- Python: `Python 3.11.1`
as I don´t have access to a MacOS I do those fixes blindly, but I will try to check if I can get access to another machine.

please let me know if I left another fix missing

ninjaoflight:
seems to be failing due to:
```
sqlitecpp| subprojects/SQLiteCpp-3.2.1/meson.build:164: WARNING: Detected bundled SQLite3 in OSX, but it does not support load extension
```
is there any issue if I change it to a message()?
as it adds the required flag anyways

ninjaoflight:
checked the last error, seems to be an edge case related to SRombauts/SQLiteCpp#394 basically on MacOS 11 in GitHub images `SQLITE_OPEN_NOFOLLOW` is not defined, will add it as a patch here also as is not a breaking change on the project itself

PR #817, opened 2022-12-11
Add [Boost::ext].DI

ArchieSW:
Add [di](https://github.com/boost-ext/di) library from boost-ext.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -62,6 +62,14 @@
       "7.6.8-1"
     ]
   },
+  "boost_di": {
+    "dependency_names": [
+      "boost_di"
+    ],
+    "versions": [
+      "1.2.0-1"
+    ]
+  },
   "box2d": {
     "dependency_names": [
       "box2d"
--- a/ci_config.json
+++ b/ci_config.json
@@ -31,6 +31,13 @@
     "tinyxml2",
     "zstd"
   ],
+  "boost_di": {
+    "_comment": "- DI: C++14 Dependency Injection Library",
+    "build_options": [
+      "boost_di:BOOST_DI_OPT_BUILD_TESTS=true",
+      "boost_di:BOOST_DI_OPT_BUILD_EXAMPLES=true"
+    ]
+  },
   "cexception": {
     "build_options": [
       "cexception:werror=false"
--- a/subprojects/boost_di.wrap
+++ b/subprojects/boost_di.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = di-1.2.0
+
+source_url = https://github.com/boost-ext/di/archive/refs/tags/v1.2.0.tar.gz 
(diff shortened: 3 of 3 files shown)


releases.json
```diff
@@ -2433,5 +2433,13 @@
       "1.3.3-2",
       "1.3.3-1"
     ]
+  },
+  "boost-ext-di": {
+    "dependency_names": [
+      "di"
```
eli-schwartz, line 2439: The meson.build produces a pkg-config file for `boost_di`.

From a quick look, the CMake files don't appear to install anything under any name at all.

ArchieSW: I seem to have confused `dependency_names` and `program_names`. The wrap file provides the `di` dependency, so it should have been in `program_names`. Correctly?

eli-schwartz: No. Program names is for when the project creates an `executable()` that can then be used as a command in e.g. a custom_target. :)

The question I have here really is just about what the "proper" name is. do we call the dependency "di"? Do we call it "boost_di"?

ArchieSW: Now I understand the question, thx for the explanation :)
I think since `di` is not part of `boost` yet, it's better to name the dependency `boost-ext-di`. Should I change the name in the files?

eli-schwartz: The name in the WrapDB should always match whatever upstream uses.

ArchieSW: Yes, now I understand. In the `meson.build` file in the project, the name is `boost_di`.

Boost.di [meson.build](https://github.com/boost-ext/di/blob/cpp14/meson.build) file:
```Meson
project('boost_di', 'cpp', 
  version: '1.1.0',
  license: 'boost',
  default_options : ['warning_level=3', 'cpp_std=c++14']
)

boost_di_dep = declare_dependency(
  include_directories: include_directories('include', 'extension/include')
)

if not meson.is_subproject()
  install_headers('include/boost/di.hpp', subdir: 'di')
  
  pkgc = import('pkgconfig')
  pkgc.generate(name: 'boost_di',
    version: meson.project_version(),
    description: 'C++14 Dependency Injection Library'
  )
endif
```

I'll fix it now.

eli-schwartz:
Could you squash these changes together since the later ones fix issues in the first commit?

You can do this with `git rebase -i upstream/master`. Change the second, third and fourth "pick" lines to "fixup". Then force push the result.

ArchieSW:
> Could you squash these changes together since the later ones fix issues in the first commit?
> 
> You can do this with `git rebase -i upstream/master`. Change the second, third and fourth "pick" lines to "fixup". Then force push the result.

ArchieSW:
> Could you squash these changes together since the later ones fix issues in the first commit?
> 
> You can do this with `git rebase -i upstream/master`. Change the second, third and fourth "pick" lines to "fixup". Then force push the result.

I hope I did well...

PR #819, opened 2022-12-14
xkbcommon: add

neheb:

ping @bluetech

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -566,6 +566,24 @@
       "libxml2-dev"
     ]
   },
+  "libxkbcommon": {
+    "build_options": [
+      "libxkbcommon:enable-docs=false",
+      "libxkbcommon:enable-tools=false",
+      "libxkbcommon:enable-wayland=false",
+      "libxkbcommon:enable-x11=false",
+      "libxkbcommon:enable-xkbregistry=false"
+    ],
+    "alpine_packages": [
+      "bison"
+    ],
+    "brew_packages": [
+      "bison"
+    ],
+    "choco_packages": [
+      "winflexbison"
+    ]
+  },
   "libxrandr": {
     "_comment": "X11 needs a POSIX system",
     "build_on": {
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -121,6 +121,7 @@ jobs:
         with:
           msystem: ${{matrix.platform}}
           install: >-
+            bison
             dos2unix
             git
           pacboy: >-
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -107,6 +107,7 @@ jobs:
         with:
           msystem: ${{matrix.platform}}
           install: >-
+            bison
             dos2unix
             git
           pacboy: >-
--- a/releases.json
+++ b/releases.json
@@ -1974,6 +1974,14 @@
       "1.1.4-1"
     ]
   },
+  "libxkbcommon": {
+    "dependency_names": [
+      "xkbcommon"
+    ],
+    "versions": [
+      "1.7.0-1"
+    ]
+  },
   "libxml2": {
     "dependency_names": [
       "libxml-2.0"
--- a/subprojects/libxkbcommon.wrap
+++ b/subprojects/libxkbcommon.wrap
@@ -1,0 +1,8 @@
+[wrap-file]
+directory = libxkbcommon-1.7.0
+source_url = https://xkbcommon.org/download/libxkbcommon-1.7.0.tar.xz
+source_filename = libxkbcommon-1.7.0.tar.xz
+source_hash = 65782f0a10a4b455af9c6baab7040e2f537520caa2ec2092805cdfd36863b247
+
+[provide]
+xkbcommon = dep_libxkbcommon


bluetech:
I pushed a commit to xkbcommon which fixes the `Consider using the built-in option for sanitizers instead of using "-fsanitize-undefined-trap-on-error".` meson warning, so that should be fine in the next release.

Not sure about the test failures, sorry. Seems like it's missing the test data files? If it's something that can be improved on the xkbcommon side, let me know.

neheb:
I'm using the regular tarbal not the GitHub one. Maybe tests should be made optional.

bluetech:
The test data is included in the tarball. It is produced by meson and I believe `meson dist` makes sure that tests pass when using it.

dcbaker:
I have patches that would fix that `-fsantize` thing. That really is a meson issue, and I should get those rebased and merged...

bluetech:
I think the path problems are caused by xkbcommon using `meson.source_root()`/`meson.build_root()` instead of the `current` variants. This will be fixed in the next xkbcommon release.

neheb:
```
[1/114] Generating from 'subprojects/libxkbcommon-1.5.0/src/xkbcomp/parser.y'..
FAILED: subprojects/libxkbcommon-1.5.0/libxkbcommon.0.dylib.p/parser.c subprojects/libxkbcommon-1.5.0/libxkbcommon.0.dylib.p/parser.h 
/usr/bin/bison --defines=subprojects/libxkbcommon-1.5.0/libxkbcommon.0.dylib.p/parser.h -o subprojects/libxkbcommon-1.5.0/libxkbcommon.0.dylib.p/parser.c -p _xkbcommon_ ../subprojects/libxkbcommon-1.5.0/src/xkbcomp/parser.y
../subprojects/libxkbcommon-1.5.0/src/xkbcomp/parser.y:230.5-9: syntax error, unexpected type, expecting string or identifier
[2/114] Generating from 'subprojects/libxkbcommon-1.5.0/src/xkbcomp/parser.y'..
FAILED: subprojects/libxkbcommon-1.5.0/libxkbcommon-test-internal.a.p/parser.c subprojects/libxkbcommon-1.5.0/libxkbcommon-test-internal.a.p/parser.h 
/usr/bin/bison --defines=subprojects/libxkbcommon-1.5.0/libxkbcommon-test-internal.a.p/parser.h -o subprojects/libxkbcommon-1.5.0/libxkbcommon-test-internal.a.p/parser.c -p _xkbcommon_ ../subprojects/libxkbcommon-1.5.0/src/xkbcomp/parser.y
../subprojects/libxkbcommon-1.5.0/src/xkbcomp/parser.y:230.5-9: syntax error, unexpected type, expecting string or identifier
```

hmmmm

neheb:
@bluetech MSYS2 has some questionable warnings. Also errors on termios.h inclusion. The format warnings are AFAIK from having an outdated MINGW define.

bluetech:
I'll try to fix these when I get a chance.

The termios.h includes are already protected, but only for MSVC compiler. This was fixed recently in https://github.com/xkbcommon/libxkbcommon/commit/bd79a960f596574e9c2ea59347f9afac469ef532, will be in the next release. I hope it will work for msys2.

The printf format ones I can probably fix with some casts (no `%z` support what year are we?!?).

The macos bison failure I'm not sure about. From a quick look I can't see where it gets its bison, do you have an idea?

neheb:
From macOS itself.

The MSYS2 no %z is caused by MINGW_STRICT_ANSI.

bluetech:
> From macOS itself.

Looks like macos bison is version 2.3, from 2006... And we seem to need... 2.3a, one version later :)

I've added a version check to xkbcommon now. For the macos build to succeed a newer bison is needed. byacc is also supported.

> The MSYS2 no %z is caused by MINGW_STRICT_ANSI.

Hmm `%z` is C99 so I wonder why it doesn't like it?

xkbcommon currently uses `c_std=c11`. Maybe `c_std=gnu11` will fix this, do you mind trying?

neheb:
MINGW_STRICT_ANSI uses the older msvcrt printf AFAIK which does not support %z. By default it works fine with %z. There's some define in the code messing this up.

bluetech:
> Ah no the problem is different. It's ATTR_PRINTF that's wrong. See https://github.com/Exiv2/exiv2/blob/main/src/image_int.hpp#L15 for a proper way to handle.

Thanks for the hint, added this [here](https://github.com/xkbcommon/libxkbcommon/commit/5a5ab3e8806c2b2a5006d53b75e79b6c90fe810e), hopefully will fix these warnings.

neheb:
@bluetech seems there are issues regarding _MSC_VER vs _WIN32 in the testing code.

neheb:
```
libxkbcommon| Program bison win_bison found: NO found 2.3 but need: '>= 2.3a' (/usr/bin/bison)
libxkbcommon| Program byacc found: NO

subprojects/libxkbcommon-1.7.0/meson.build:175:8: ERROR: Problem encountered: Could not find a compatible YACC program (bison or byacc)
```

@eli-schwartz bison is keg only., Any way to override the search path?

benoit-pierre:
This work:
```diff
 ci_config.json         | 3 +++
 tools/sanity_checks.py | 1 +
 2 files changed, 4 insertions(+)

diff --git c/ci_config.json w/ci_config.json
index 7d02be0..eec2df8 100644
--- c/ci_config.json
+++ w/ci_config.json
@@ -571,6 +571,9 @@
     "alpine_packages": [
       "bison"
     ],
+    "brew_packages": [
+      "bison"
+    ],
     "choco_packages": [
       "winflexbison"
     ],
diff --git c/tools/sanity_checks.py w/tools/sanity_checks.py
index aec8655..41e6144 100755
--- c/tools/sanity_checks.py
+++ w/tools/sanity_checks.py
@@ -373,6 +373,7 @@ class TestReleases(unittest.TestCase):
             else:
                 s = ', '.join(brew_packages)
                 print(f'The following packages could be required: {s}')
+            subprocess.check_call(['brew', 'link', '--force'] + brew_packages)
         elif choco_packages and is_windows():
             if is_ci():
                 subprocess.check_call(['choco', 'install', '-y'] + choco_packages)
```

But should probably be amended to only `brew link …` on CI.

benoit-pierre:
Correction, it works for amd64, but x86_64, not sure why…

benoit-pierre:
This work in both cases:
```diff
 ci_config.json         | 3 +++
 tools/sanity_checks.py | 5 +++++
 2 files changed, 8 insertions(+)

diff --git c/ci_config.json w/ci_config.json
index 7d02be0..eec2df8 100644
--- c/ci_config.json
+++ w/ci_config.json
@@ -571,6 +571,9 @@
     "alpine_packages": [
       "bison"
     ],
+    "brew_packages": [
+      "bison"
+    ],
     "choco_packages": [
       "winflexbison"
     ],
diff --git c/tools/sanity_checks.py w/tools/sanity_checks.py
index aec8655..fa28049 100755
--- c/tools/sanity_checks.py
+++ w/tools/sanity_checks.py
@@ -370,6 +370,11 @@ class TestReleases(unittest.TestCase):
         elif brew_packages and is_macos():
             if is_ci():
                 subprocess.check_call(['brew', 'install'] + brew_packages)
+                # Ensure binaries from keg-only formulas are available (e.g. bison).
+                for prefix in subprocess.check_output(['brew', '--prefix'] + brew_packages).decode().split('\n'):
+                    bindir = Path(prefix) / 'bin'
+                    if bindir.exists():
+                        meson_env['PATH'] = str(bindir) + ':' + meson_env['PATH']
             else:
                 s = ', '.join(brew_packages)
                 print(f'The following packages could be required: {s}')
```

neheb:
@benoit-pierre fold into https://github.com/mesonbuild/wrapdb/pull/1522 ?

benoit-pierre:
> @benoit-pierre fold into #1522 ?

Done.

neheb:
@benoit-pierre no dice.

ci_config.json
```diff
@@ -566,6 +566,24 @@
...
+    ],
+    "alpine_packages": [
+      "bison"
+    ],
+    "choco_packages": [
+      "winflexbison"
+    ],
+    "darwin_packages": [
+      "bison"
+    ]
```
benoit-pierre, line 585:
```suggestion
    "brew_packages": [
      "bison"
    ]
```
;)

PR #826, opened 2022-12-17
github: update build_all to sanity_checks

neheb:

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -69,8 +69,8 @@ jobs:
       - name: Install packages
         run: |
           brew update
-          brew install ninja python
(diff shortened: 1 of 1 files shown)


.github/workflows/build_all.yml
```diff
@@ -69,8 +69,8 @@ jobs:
       - name: Install packages
         run: |
           brew update
-          brew install ninja python
-          python3 -m pip install git+https://github.com/mesonbuild/meson
+          brew install ninja
+          python3 -m pip install --pre meson
```
eli-schwartz, line 73: Absolutely not!!!

The purpose of this was extremely clear. We want to test from meson git master, it's the *entire* point of the build_all check.

neheb: Should be one or the other globally I think.

PR #832, opened 2022-12-18
gdbm: consistency updates

neheb:
Don't build on Windows and move build closer to upstream.


neheb:
@jpakkane this is ready

PR #833, opened 2022-12-18
Roll back some CI changes

eli-schwartz:
The recent changes to `.github/workflows/` mixed some harmless/good changes together with some that aren't so great. Specifically:
- we absolutely positively want to use meson from git master in the build_all CI, because the entire point of that CI is to run regression tests against bleeding-edge future versions of Meson and catch issues before release.
- macOS is broken. It's broken on multiple levels, and one of them caused CI to go red, but the real solution is the same one I PR'ed to the Meson repository, which includes a more thorough fix. I probably forgot to sync that over.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -66,8 +66,8 @@ jobs:
       - name: Install packages
         run: |
           brew update
-          brew install ninja
-          python3 -m pip install --pre meson
+          brew install ninja python
+          python3 -m pip install git+https://github.com/mesonbuild/meson
 
       - name: Sanity Checks
         run: |
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -63,10 +63,16 @@ jobs:
       - name: Fetch tags and unshallow
         run: git fetch --unshallow --tags
 
+      - run: brew update
+      # github actions overwrites brew's python. Force it to reassert itself, by running in a separate step.
+      - name: unbreak python in github actions
+        run: |
+          find /usr/local/bin -lname '*/Library/Frameworks/Python.framework/*' -delete
+          sudo rm -rf /Library/Frameworks/Python.framework/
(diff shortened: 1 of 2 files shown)


.github/workflows/build_all.yml
```diff
@@ -63,10 +63,16 @@ jobs:
       - name: Fetch tags and unshallow
         run: git fetch --unshallow --tags
 
+      - run: brew update
```
neheb, line 66: A lot of work :)

What's wrong with the stock Python in the runner?

eli-schwartz: I do not trust Github at all, there was a cascading laundry list of problems with it and I *do* trust brew.

But mostly, the issue is that github broke brew itself, and any formula that attempts to install python, either directly or as a dependency, will error out -- and that includes any wrapdb package that specifies `brew_packages` if that brew package pulls in python anywhere in the dependency tree.

So we must unbreak brew in order to guarantee that arbitrary `brew install XXX` will "just work" -- and that means nuking the github actions one from high orbit until no traces remain.

neheb: The CI also supports vcpkg. Wonder which is better...

neheb: I don't know how true that indirect python comment is. For example, Orcus specifies boost, which definitely pulls python. CI didn't fail though.

eli-schwartz: Which CI is that? I didn't think we recently changed orcus.

neheb: I ran manual build all runs on here after my changes.

eli-schwartz: I went and looked up boost in homebrew, it doesn't depend on python at all. boost_python3 does, but that's a separate formula. (It will also install `python@3.11`, which isn't the same as the `python` alias, but does conflict once again with Github Actions.)

neheb: Will look when CI is done.

neheb: Hmm seems correct.

PR #835, opened 2022-12-18
lame: Remove LIBRARY entry in lame.def

nirbheek:
LIBRARY entry is using the MinGW name:

    LIBRARY  libmp3lame.DLL

This makes the import library useless on MSVC, because it uses the wrong DLL name.


nirbheek:
I had to add `fix_def.py` to `PER_PROJECT_PERMITTED_FILES`.

neheb:
Ummm, why not just fix the library name to be that?

nirbheek:
> Ummm, why not just fix the library name to be that?

Library name is compiler-specific, and can also be specified in the target itself. Better to let the linker generate it automatically, since it will always know.

neheb:
not sure what you mean.

library can be called 'libmp3lame' with name_prefix set to '' and soversion set to '' for windows.

PR #836, opened 2022-12-18
expat: Override the dependency so [provide] works

nirbheek:


eli-schwartz (changes requested):
Commit message is incorrect, `[provide]` already works.

This wrap file is using the `dep = varname` form, not the `dependency_names = dep` form, which means that override_dependency is not needed (or used).

This does, however, break compatibility with debian oldstable and Ubuntu LTS, which I'd rather not do without a compelling need.

What this PR actually does is change a backwards-compatible syntax for one that isn't, and doesn't migrate the .wrap file at the same time anyway.

eli-schwartz:
Particularly, override_dependency is interesting because you do NOT need a provide section at all. Or even a wrap file. If a subproject is activated, such as via implicit matching on the `subproject name == dependency name`, meson will try configuring the subproject and seeing if it overrides the dependency. This is often handy in upstream projects, but a moot point for the WrapDB, which always provides a wrap file with a provide section.

nirbheek:
Ah, I missed the `expat` line in the wrap. I thought wrapdb would auto-populate the provided dependencies list if the build file overrode it. My bad!

PR #838, opened 2022-12-19
entt: prefer `src` directory over `single_include`

crispyricepc:
Using `#include <entt/entt.hpp>` is still possible with this change, but now people who prefer the forward declarations in `fwd.hpp` can also use it


neheb:
need to bump releases.json

neheb:
eh no you have to make a separate entry.

eli-schwartz:
The changes should be squashed together, BTW.

crispyricepc:
Okay sorry, first PR over here. Think I fixed it?

neheb:
reverse the order in releases.json

crispyricepc:
Ah ff-
Thanks for your patience

eli-schwartz (approved):
LGTM once CI passes

PR #840, opened 2022-12-21
box2d: don't build unit-tests

neheb:
Fixes CI under Ubuntu 22.04


subprojects/packagefiles/box2d/meson.build
```diff
@@ -13,4 +13,7 @@ box2d_dep = declare_dependency(link_with: box2d_lib,
     include_directories: incdir,
     version: meson.project_version())
 
-subdir('unit-test')
+# Upstream has an old versionm of doctest.h which is imcompatible with glibc >=2.34
```
eli-schwartz, line 16: Sounds like we should be using *our* doctest subproject instead.

eli-schwartz (changes requested):

subprojects/packagefiles/box2d/meson.build
neheb: I....don't know how possible that is. There's a doctest.h in that directory.

neheb: Speficially, the error is:

```
../subprojects/box2d-2.4.1/unit-test/doctest.h:4021:47: error: size of array ‘altStackMem’ is not an integral constant-expression
 4021 |         static char             altStackMem[4 * SIGSTKSZ];
```

neheb: Huh. It's possible with implicit_include_directories: false . Cool.

eli-schwartz (approved):

PR #845, opened 2022-12-21
pcre2: update to 10.42

neheb:


subprojects/packagefiles/pcre2/meson.build
```diff
@@ -233,15 +230,15 @@ if (
     'RunTest',
     runtest,
     should_fail: host_machine.system() == 'windows' and c_compiler.get_argument_syntax() != 'msvc',
-    env: ['srcdir=@0@'.format(meson.current_source_dir())],
+    env: 'srcdir=@0@'.format(meson.current_source_dir()),
```
eli-schwartz, line 233: Not obvious to me why we're removing these. It's semantically correct, env is a list of strings.

neheb: That goes for other stuff too. Removing [] is shorter and cleaner.

PR #848, opened 2022-12-23
speexdsp: Compile arm test program using SIMD flags and libmath

kcgen:
Feedback from the prior commit indicated it didn't get us all way on the Pi, and the test program needs NEON compiler flags as well as libmath included.


neheb:
the muon fmt config I use is:

```
max_line_len=120
indent_by='  '
space_array=false
wide_colon=false
```

eli-schwartz:
I'd be happy to use four spaces instead of overriding it to two, personally. :P

eli-schwartz (dismissed):
If you're going to format the files, please do it in a separate commit.

kcgen:
> If you're going to format the files, please do it in a separate commit.

Updated; all logical changes (to the SIMD blocks)  are now in the first commit, which involved some nested level changes, so large chunks are still showing at changed.

The second commit is now purely formatting / zero logic (and described as such). 

I wasn't sure which formatting style I should use; @eli-schwartz, you mentioned 4-space indentation and @neheb mentioned 2; happy to use either. Let me know.

neheb:
The latter is the predominant one, mostly because of me. There's no real standard AFAIK.

Unlike python, formatting doesn't seem to be mandated by meson.

kcgen:
Thanks @neheb - will do!

kcgen:
> The latter is the predominant one, mostly because of me. There's no real standard AFAIK.
> 
> Unlike python, formatting doesn't seem to be mandated by meson.

Done and re-tested:

Issue #830, opened 2022-12-17
OpenSSL cross compile with macOS build and arm host


jhgorse:
Greetings,

I was playing with meson because it seemed fast. I compiled OpenSSL in 7.5 seconds as the wrapdb subproject compared to 5:11 seconds for regular build. This inspired me to try to cross compile to 32 bit ARM linux host/target using macOS M1 as the build box.

At first I tried linaro/arm embedded toolchains, but neither support macOS to arm-linux, just macOS to baremetal arm-none. Okay, so what about clang?

cross-file.txt
```
[binaries]
c = 'clang'
cpp = 'clang++'
ar = 'ar'
strip = 'strip'
as = 'as'

[build_machine]
system = 'darwin'
cpu_family = 'aarch64'
cpu = 'aarch64'

[host_machine]
system = 'linux'
cpu = 'arm'
cpu_family = 'arm'
endian = 'little'
```

build.meson
```
project('tutorial', 'c', default_options : ['c_std=c11', 'cpp_std=c++11'])
ssl = subproject('openssl')
executable('demo', 'main.c')
```

I had to clear the `openssl_libraries` `atomic` entry in `/subprojects/openssl-3.0.7/generated-config/archs/linux-armv4/asm/meson.build`. Couldn't figure out how to override it at the top level meson file or the cross-file.
```
openssl_libraries = [
  'dl',
  # 'atomic',
]
```

We get to the point where it can't handle the assembly.
```
test_build % meson compile -C builddir/
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /opt/homebrew/bin/ninja -C /private/tmp/test_build/builddir
ninja: Entering directory `/private/tmp/test_build/builddir'
[898/970] Compiling C object subprojects/openssl-3.0.7/libcrypto.so.p/generated-config_archs_linux-armv4_asm_crypto_bn_armv4-gf2m.S.o
FAILED: subprojects/openssl-3.0.7/libcrypto.so.p/generated-config_archs_linux-armv4_asm_crypto_bn_armv4-gf2m.S.o
[...]

PR #850, opened 2022-12-24
Update generator.sh to be more portable

jhgorse:
xargs has different behavior across platforms, causing some paths to be omitted from the generation. This fixes those omissions.

This is based on @eli-schwartz's proposal here: https://github.com/mesonbuild/wrapdb/issues/830#issuecomment-1356671325

### DIFF
--- a/subprojects/packagefiles/openssl/generator.sh
+++ b/subprojects/packagefiles/openssl/generator.sh
@@ -31,11 +31,12 @@ rm -rf config/archs
 LANG=C make -C config
 
 # Copy generated files back into correct place
-find config/archs -name 'meson.build' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
-find config/archs -name '*.asm' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
-find config/archs -name '*.c' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
-find config/archs -name '*.h' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
-find config/archs -iname '*.s' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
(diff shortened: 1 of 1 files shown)


neheb:
Needs releases.json bumped.

subprojects/packagefiles/openssl/generator.sh
```diff
@@ -31,11 +31,16 @@ rm -rf config/archs
...
-find config/archs -name '*.asm' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
-find config/archs -name '*.c' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
-find config/archs -name '*.h' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
-find config/archs -iname '*.s' | xargs -I % sh -c 'mkdir -p ../../../generated-$(dirname %); cp % ../../../generated-%'
+cmd () {
+  mkdir -p ../../../generated-$(dirname "$1")
+  cp "$1" ../../../generated-"$1"
+}
+export -f cmd
+find config/archs -name 'meson.build' -exec sh -c "cmd {}" \;
```
eli-schwartz, line 39: I mentioned in the other ticket how this could be done without `export -f`, but if you're going to use `export -f` then you also need to explicitly run `-exec bash -c` because "sh" might be a shell that doesn't implement this, such as dash or busybox.

jhgorse: Fixed. Removed export -f for more portable sh friendly approach.

jhgorse:
@neheb 

> Needs releases.json bumped.

Done.

eli-schwartz:
Erm, but.

We don't need to update releases.json as this doesn't need to produce a new WrapDB release that runs generator.sh, creates the generated-config/ directory, and produces a patch overlay that refuses to run generator.sh "because generated-config/ already exists". :)

jhgorse:
Removed useless single quotes.

jhgorse:
Removed releases.json bump. We don't need to re-run generator.sh to get the wrap to run and this should produce the same output as the original linux-based run.

eli-schwartz:
Seems reasonable. Let's do it. :D

tp-m:
Would be nice if commit messages and PR titles could be prefixed with the wrap/project in question next time :slightly_smiling_face:

PR #854, opened 2022-12-25
zlib: use xz tarball

neheb:
Smaller. Also add fallback URL.

### DIFF
--- a/subprojects/zlib.wrap
+++ b/subprojects/zlib.wrap
@@ -1,6 +1,7 @@
 [wrap-file]
 directory = zlib-1.2.13
 source_url = http://zlib.net/fossils/zlib-1.2.13.tar.gz
+source_fallback_url = https://github.com/madler/zlib/releases/download/v1.2.13/zlib-1.2.13.tar.gz
 source_filename = zlib-1.2.13.tar.gz
 source_hash = b3a24de97a8fdbc835b9833169501030b8977031bcb54b3b3ac13740f846ab30
 patch_directory = zlib


subprojects/zlib.wrap
```diff
@@ -1,8 +1,9 @@
 [wrap-file]
 directory = zlib-1.2.13
-source_url = http://zlib.net/fossils/zlib-1.2.13.tar.gz
-source_filename = zlib-1.2.13.tar.gz
-source_hash = b3a24de97a8fdbc835b9833169501030b8977031bcb54b3b3ac13740f846ab30
+source_url = http://zlib.net/fossils/zlib-1.2.13.tar.xz
```
eli-schwartz, line 3: This does not even exist, why are we changing to it?

eli-schwartz (dismissed):

subprojects/zlib.wrap
neheb: Ah it was changed. Makes sense.

eli-schwartz: No, this has always been how zlib operates. The current version is available as gzip from a top-level download url, and old versions are available also as gzip in the fossils/ directory.

The current version, and only the current version, may have other download formats and perhaps PGP signatures as well. It used to be that the current version was additionally available in .bz2 and .zip, although currently it is xz instead of bz2.

eli-schwartz:
Also:
> Smaller.

It is smaller by 200kb, I'm not convinced this is worth using non-permanent urls.

neheb:
OTOH one reason to update releases.json would be to force an update of the wrap file. That is, meson wrap install zlib before this has no fallback and after this it does.

eli-schwartz:
I'm not strongly opinionated one way or the other, but people who want this fallback url can add it without a wrapdb release of it. They'd have to know or care, though. Bumping it would cause people who already have the wrap to see an update which might not be really necessary -- although it does not really cause harm.

Anecdotally/personally, I haven't encountered this "bad tarballs" problem.

neheb:
CI did apparently: https://github.com/mesonbuild/wrapdb/actions/runs/3774064318/jobs/6415944704

PR #857, opened 2022-12-26
Theora fixes: proper library versioning, fix build on x86 with assembly enabled

tp-m:
Various fixes for the Meson build discovered as part of GStreamer's binary packaging in https://gitlab.freedesktop.org/gstreamer/cerbero/-/merge_requests/1065


tp-m:
cc @amyspark

subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -170,6 +170,8 @@ libtheoradec = library(
   vs_module_defs: '../win32/xmingw32/libtheoradec-all.def',
   c_args: compile_args,
   link_args: theoradec_version_flag,
+  soversion: theoradec_soversion,
+  version: theoradec_libversion,
```
eli-schwartz, line 174: I don't think we need both the soversion and the version, do we? The soversion defaults to the first dotted component of the version, which is what most projects do (including this one).

subprojects/packagefiles/theora/lib/meson.build
```diff
@@ -224,6 +225,8 @@ libtheora = library(
   c_args: compile_args,
   link_args: theora_version_flag,
   vs_module_defs: 'theora.def', # Missing in the official tarballs
+  soversion: theora_soversion,
```
nirbheek, line 228: `soversion:` is redundant here (and above). If `version:` is specified, the first component is set as `soversion:`.

nirbheek: Also need `darwin_versions:` set to the following:
```
libtheora.0.dylib (compatibility version 4.0.0, current version 4.10.0)
libtheoradec.1.dylib (compatibility version 3.0.0, current version 3.4.0)
libtheoraenc.1.dylib (compatibility version 3.0.0, current version 3.2.0)
```

(copying from the gstreamer issue so it's updated in both places)

subprojects/packagefiles/theora/lib/meson.build
tp-m: Fixed, thanks.

subprojects/packagefiles/theora/lib/meson.build
tp-m: Updated accordingly (I think).

nirbheek (approved):

neheb:
No releases.json bump

PR #858, opened 2022-12-26
libjpeg-turbo: update to 2.1.4

bgilbert:
I've checked the upstream diff and verified that there are no relevant build system changes.


eli-schwartz:
At some point it would be really good to add the upstream tests...

PR #859, opened 2022-12-26
libtiff: fix build when using libjpeg-turbo from a subproject

bgilbert:
`meson setup build -Dwraps=libjpeg-turbo,libtiff` fails:

    ../subprojects/tiff-4.5.0/test/raw_decode.c:73:10: fatal error: jpeglib.h: No such file or directory

Remember to include the libjpeg dependency when building `raw_decode`.


neheb:
...wow how did I miss this?

PR #861, opened 2022-12-27
libpng: enable SSE support on x86 CPUs

bgilbert:
Unconditionally enable the SSE source files and `#define` on x86/x86_64 processors, matching the behavior of `configure.ac`.  `pngpriv.h` checks compiler-defined preprocessor macros to determine what, if any, optimizations to actually enable.


neheb:
This will definitely fail on MinGW32. It does not default to passing SSE2 by default. Maybe I'd replace that host_machine check with cc.check_header('xmmintrin.h')

neheb:
I assume support is not compiled in.

neheb:
Related: https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/lame/meson.build#L94

eli-schwartz:
> I assume support is not compiled in.

It's not clear to me how this was intended to answer my request for a fact check and citations for the upstream build system.

bgilbert:
With my copy of MinGW (`mingw32-gcc-12.2.1-4.fc37.x86_64`), `__SSE2__` is defined if `-msse2` is specified to the compiler, and isn't defined otherwise.  `pngpriv.h` will then enable or disable the optimizations, respectively.  To me this seems like the correct behavior: we shouldn't use SSE2 instructions if the program is being compiled to run on older machines.  What's the failure mode you're concerned about?

eli-schwartz:
https://github.com/glennrp/libpng/blob/61bfdb0cb02a6f3a62c929dbc9e832894c0a8df2/configure.ac#L301-L345

So, this has an option to enable/disable "hardware optimization". It defaults to empty, and if manually passed no/off, it disables hardware optimizations, and by default it enables them by adding the implementation files and defining PNG_INTEL_SSE_OPT to 1 (for i?86|x86_64, and different defines/implementation files for other CPUs).

The implementation files are completely ifdef'ed out, and only if pngpriv.h detects the compiler defines as available, do the implementation files actually provide functions.

This seems straightforward and correct to me.

jpakkane:
As a data point:

> SSE2 (Streaming SIMD Extensions 2) is one of the Intel [SIMD](https://en.wikipedia.org/wiki/SIMD) (Single Instruction, Multiple Data) [processor supplementary instruction](https://en.wikipedia.org/wiki/Processor_supplementary_instruction) sets first introduced by [Intel](https://en.wikipedia.org/wiki/Intel) with the initial version of the [Pentium 4](https://en.wikipedia.org/wiki/Pentium_4) in 2000.

So every Intel CPU  made in the last _22 years_ supports SSE2 and further that is only on 32 bit mode, it has always available in 64 bit mode. Do we _really_ need to support processors from the 90s or could we just always enable SSE2 and wait for bug reports and only fix it then (which is _very_ unlikely to ever happen)?

bgilbert:
Not every embedded x86 clone will support them.  (That isn't as relevant for MinGW, but may be relevant for e.g. Linux.)  And it seems surprising to override the compiler options configured by the user?

neheb:
@jpakkane in OpenWrt, non SSE capable processors are supported. Not that I wish to imply this wrap will be used there.

eli-schwartz:
SSE2 capable or not, seems to be missing the point.

Isn't the purpose of this PR to enable the functionality of "add the implementation files, and let the compiler figure out whether the compiler thinks it's supposed to generate SSE2 code for its own intrinsics as well as user code based on the current compiler flags"?

If the compiler defines for SSE2 aren't defined then this all becomes a no-op. Simply don't pass `-msse2` on systems where your CPU doesn't support it (or pass a -no option).

neheb:
I tested this on mingw32 and it didn't fail. I was curious.

```
$ strings libpng16-16.dll | grep png_init_filter
!png_init_filter_functions
_png_init_filter_functions
```

64-bit:
```
$ strings libpng16-16.dll | grep png_init_filter
png_init_filter_functions_sse2
7png_init_filter_functions
"png_init_filter_functions_sse2
png_init_filter_functions
png_init_filter_functions_sse2
```

It's as I suspected. It's nice that it doesn't fail to compile like with LAME.

bgilbert:
Any remaining blockers to getting this in?

jpakkane:
Not really, it can be improved later if need be.

PR #862, opened 2022-12-27
pcre2: small fixes

bgilbert:
I hadn't noticed that #845 was pending and was just about to PR a similar update.  A few remaining fixes from my branch:

- Update the GitHub organization name in the source URL
- Add a test case added in pcre2 10.42
- Skip grep test if grep is disabled


neheb:
I also noticed after doing https://github.com/mesonbuild/wrapdb/pull/863 :). I'll merge this first though.

eli-schwartz:
- I don't like that other PR at all
- if we were going to merge multiple updated to one wrap, I'd prefer they all be grouped together into a single release....

neheb:
Needs a rebase

bgilbert:
Rebased.

PR #863, opened 2022-12-27
pcre2: configure properly

neheb:
Instead of using a generic config.

Need to test MinGW to make sure it configures properly.


eli-schwartz:
I'm not a fan of having such a gigantic generated file checked into git here. It's not obvious what is supposed to be wrong with the old method.

neheb:
Old config is a generic one that only enables certain features and headers. For example, only pcre2_8 is enabled instead of also 16 and 32.

Generating the config as done here also allows using gnu_symbol_visibility: 'hidden'.

eli-schwartz:
I think you may be missing my point:

- this doesn't *explain* what the difference is
- I more than anything else do not believe it's reasonable to check the file into git and then fail to update it on new pcre2 releases.

Solve the first problem by describing the goal in the commit message.
Solve the second problem by producing the template file with:

```meson
configure_file(
    input: 'src/config.h.in',
    output: 'config.h.meson',
    command: [find_program('config-h-to-mesondefine.py'), '@INPUT@', '@OUTPUT@'],
)
```

neheb:
failing to update happened with liblzma, so I'm aware of the issue. See https://github.com/mesonbuild/wrapdb/commit/62a7f6658f786fb5ab52aa2f9f45c22d83d2e35b

An alternative to checking in a config.h.meson file is using the cmake config in the tarball. Unfortunately, it's very incomplete.

eli-schwartz:
> failing to update happened with liblzma, so I'm aware of the issue. See [62a7f66](https://github.com/mesonbuild/wrapdb/commit/62a7f6658f786fb5ab52aa2f9f45c22d83d2e35b)

And I would like to move away from that by not doing it anymore.

So I think we need to fix liblzma too, to either use configure_file in command mode or else openssl's generator.sh support (generator.py?) to produce it in the final wrap archive.

PR #864, opened 2022-12-27
proxy-libintl: add

bgilbert:
The upstream meson.build declares a version of `1` and has done so for several releases, since the Meson config was first introduced.  I've used the tag version here.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1807,6 +1807,14 @@
       "1.3.0-1"
     ]
   },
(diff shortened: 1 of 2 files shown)


neheb:
https://github.com/mesonbuild/wrapdb/pull/456 :)

bgilbert:
Eek, and it's word-for-word identical.  Closing.

PR #865, opened 2022-12-27
libxml2: fix build with zlib from a subproject; improve iconv detection

bgilbert:
`meson setup build -Dwraps=zlib,libxml2` fails:

    subprojects/libxml2-2.10.3/meson.build:160:8: ERROR: Dependencies must be external dependencies

We can't use `cc.has_header()` because zlib hasn't been built yet.  If we found zlib, assume we have the header as well.

Also allow disabling iconv support, since I have a use case that needs that, and fix iconv detection with Meson < 0.60.


bgilbert:
Added a commit to allow disabling iconv support, since I have a use case that needs that.

eli-schwartz:
> We can't use `cc.has_header()` because zlib hasn't been built yet. If we found zlib, assume we have the header as well.

This makes sense, since we're using a proper dependency() search, and that will guarantee one of 3 things:
- there's a pkg-config file for it
- Meson's builtin `clib_compiler.find_library('z')` search found zlib and *also* found `clib_compiler.has_header('zlib.h')` already!
- it's a subproject dependency fallback and assumed not-incomplete

Under no conditions do we need to be worried that zlib is only available as, say, a Debian/Fedora runtime package but not the `-devel` component, so checking is pointless.

subprojects/packagefiles/libxml2/meson.build
```diff
@@ -78,9 +78,9 @@ if math_lib.found()
 endif
 
 if meson.version().version_compare('>= 0.60')
-  iconv_lib = dependency('iconv', required: false)
+  iconv_lib = dependency('iconv', required: get_option('iconv'))
 else
-  iconv_lib = dependency('', required: false)
+  iconv_lib = dependency('', required: get_option('iconv'))
```
eli-schwartz, line 83: This doesn't work. It's a dummy never-found dependency, so building with `-Diconv=enabled` will make this error out.

That being said, this code has never actually worked before, checking for iconv support on Meson below 0.60 has always required checking cc.has_function with and without the result of cc.find_library('iconv'), and before the 0.60 support was added, this just set:

```meson
withcdata.set10('WITH_ICONV', iconv_lib.found())
```

so iconv was only configured *iff* you had an external iconv.

eli-schwartz (changes requested):

subprojects/packagefiles/libxml2/meson.build
bgilbert: Okay, I think I have the logic correct now.

subprojects/packagefiles/libxml2/meson.build
```diff
@@ -78,9 +78,23 @@ if math_lib.found()
 endif
 
 if meson.version().version_compare('>= 0.60')
-  iconv_lib = dependency('iconv', required: false)
+  iconv_lib = dependency('iconv', required: get_option('iconv'))
+elif get_option('iconv').disabled()
+  iconv_lib = disabler()
 else
-  iconv_lib = dependency('', required: false)
+  iconv_lib = dependency('iconv', required: false)
```
eli-schwartz, line 85: This is probably a waste of time, since iconv won't ever exist as a dependency without the builtin check. Upstream iconv certainly doesn't ship a pkg-config file.

eli-schwartz (commented):
I'd probably check the <0.60 stage like this:

```meson
code = '''
#include <iconv.h>
int main() {
    iconv_open("","");
}
'''
iconv_lib = dependency('', required: false)
if get_option('iconv').disabled()
    iconv_found = false
else
    iconv_found = cc.links(code, name: 'iconv_open')
    if not iconv_found
        iconv_lib = cc.find_library('iconv')
        iconv_found = cc.links(code, name: 'iconv_open in libiconv' dependencies: iconv_lib)
    endif
endif
```

bgilbert:
Okay, took another shot at it.  Thanks for the reviews.

neheb:
That is a very interesting issue. My strategy of using a dictionary here looks like is pointless.

subprojects/packagefiles/libxml2/meson.build
```diff
@@ -78,9 +78,29 @@ if math_lib.found()
 endif
 
 if meson.version().version_compare('>= 0.60')
-  iconv_lib = dependency('iconv', required: false)
+  iconv_lib = dependency('iconv', required: get_option('iconv'))
 else
-  iconv_lib = dependency('', required: false)
+  if get_option('iconv').disabled()
+    iconv_lib = dependency('', required: false)
+  else
```
neheb, line 85: I didn't bother doing this because of all the work involved. In any case, I doubt many will be using such an old meson version.

neheb:
I was going to say to convert that dictionary to a simple list, but thread_dep and dl_dep can't really be subprojects. Well, maybe dl_dep for Windows.

neheb:
```
subprojects/libxml2-2.10.3/meson.build:160:8: ERROR: Dependencies must be external dependencies
```

My dl_dep theory is correct.

PR #866, opened 2022-12-27
libffi: fix MinGW build

bgilbert:
Fix the `FFI_EXEC_STATIC_TRAMP` test to match the logic in `configure.ac`. Defining that macro on MinGW causes many build errors in `tramp.c`.


neheb:
ping @xclaesse

xclaesse:
This was already fixed in the upstream PR, and is part of the update there: https://github.com/mesonbuild/wrapdb/pull/873.

PR #867, opened 2022-12-28
dlfcn-win32: add

neheb:


neheb:
```
subprojects/libarchive-3.6.2/meson.build:396:8: ERROR: Dependencies must be external dependencies
```

Of course...

neheb:
this dl.h is suspicious.

jpakkane:
This adds a dependency called `dl` but there is currently no way to say that it only works on Windows. Which is to say it might get downloaded and built on unixy platforms. Which leads to build failures. Dunno if we should guard against it somehow or whether we ignore this as too rare a case.

neheb:
OTOH most usages of dependency('dl') here have required: false or are a feature option.

PR #868, opened 2022-12-28
pcre2: add build options for 8/16/32 bit libs and disable 16/32 variant by default

tp-m:


subprojects/packagefiles/pcre2/meson.build
```diff
@@ -95,72 +95,106 @@ endforeach
...
 
+if not get_option('pcre2-8').disabled()
+  config_h_defs += ['-DSUPPORT_PCRE2_8']
+endif
+if get_option('pcre2-16').enabled()
+  config_h_defs += ['-DSUPPORT_PCRE2_16']
+endif
+if get_option('pcre2-32').enabled()
+  config_h_defs += ['-DSUPPORT_PCRE2_32']
+endif
```
xclaesse, line 109: IMHO this is a bit abusing a feature option, you only use 2 states not 3, that should be a boolean. "auto" has no meaning here.

tp-m: I think it's "using", not "abusing". As I see it these are features, so it makes sense to make these feature options. of course I admit they serve no purpose here at this point in time (but they might in future, e.g. one might want to build 16-bit width support by default on Windows but not elsewhere).

This is an ideological discussion that I feel we've rehashed multiple times elsewhere already :)

eli-schwartz: Instead of playing language games with what the type name of the option is, we should look at what the option *does*.

The option is an "option" that has the following values:
- enabled
- auto
- disabled

You are attempting to use it as something that doesn't distinguish between "auto" and "disabled", which means that anyone who attempts to pass `=auto` discovers that it's a valid value, but doesn't do what they want.

(It doesn't matter what they want, someone is going to complain they wanted the opposite behavior.)

Instead, I propose that since you really want the term "enabled", and don't like booleans, we compromise.

I'll approve this PR if you change the option to a combo option with the allowed string choices: `['enabled', 'disabled']`, and in meson.build you can check it with:

```meson
if get_option('pcre2-32') == 'enabled'
    config_h_defs += ['-DSUPPORT_PCRE2_32']
endif
```

There we have it, you get your enablement names, users aren't allowed to specify semantically nonsense values.

tp-m: You've got to be joking.

eli-schwartz: It was a mostly-sarcastic rationale intended to highlight why booleans are the right option type, yeah.

(I believe in booleans as the one true approach, and rate feature options where not all the options do something, *below* hand-rolled combo options.)

PR #869, opened 2022-12-28
SDL2: check target machine instead of host machine

Jan200101:
We inherited this from the initial taisei meson.build, but I don't think it was correct.


eli-schwartz:
Can you explain how this works? Remember that target machine means a cross compiler that produces output on the host machine for the target machine

jpakkane:
Yes, the only time you'd need it is when creating a cross compiler.

Jan200101:
Yep, this was a misreading of the documentation that made the purpose a bit unclear.

PR #870, opened 2022-12-28
pcre2: add jit support

Jan200101:
- adds general and sealloc jit support
- uses the cmake config header instead of the generic one


neheb:
Needs a rebase.

subprojects/packagefiles/pcre2/meson.build
```diff
@@ -6,25 +6,76 @@ project(
...
+platform_is_cygwin = (host_machine.system() == 'cygwin')
+
+platform_is_unixlike = (
+    platform_is_bsd or
+    platform_is_emscripten or
+    platform_is_haiku or
+    platform_is_linux or
+    platform_is_darwin or
+    platform_is_cygwin or
+false)
```
neheb, line 41: typo?

subprojects/packagefiles/pcre2/meson.build
```diff
@@ -6,25 +6,76 @@ project(
...
+
+platform_is_linux_desktop = (host_machine.system() == 'linux')
+platform_is_android = (host_machine.system() == 'android')
+platform_is_linux = platform_is_android or platform_is_linux_desktop
+
+platform_is_darwin = (host_machine.system() == 'darwin')
+
+platform_is_cygwin = (host_machine.system() == 'cygwin')
+
+platform_is_unixlike = (
```
neheb, line 34: you could simplify with something like:

```
platform_is_unixlike = host_machine.system() in [x, y, z]
```

subprojects/packagefiles/pcre2/meson.build
Jan200101: is this in regard to cygwin?
This is a straight translation from the upstream `CMakeLists.txt` that uses `UNIX` which is true on cygwin.

neheb: The last false

Jan200101: This was originally taken from SDL https://github.com/mesonbuild/wrapdb/blob/4f0620ca571968a58113f6e26126274de99eca91/subprojects/packagefiles/sdl2/meson.build#L71-L76
but there seems to be no good reason for it to be there.

Appears to be something we inherited from the taisei SDL2 port

neheb:
LGTM

PR #872, opened 2022-12-29
pcre2: fix library versioning

tp-m:


jpakkane:
Missing an entry in the release json file.

tp-m:
> Missing an entry in the release json file.

Fixed, thanks.

PR #873, opened 2022-12-30
libffi: Update to latest upstream PR version

xclaesse:
Upstream PR received a few contributions and fixes: https://github.com/libffi/libffi/pull/746


subprojects/packagefiles/libffi/meson.build
```diff
@@ -75,11 +74,12 @@ endforeach
 # Default values
 c_sources = ['ffi.c']
 asm_sources = ['sysv.S']
+masm_args = []
 targetdir = ''
 have_long_double = ''
 have_long_double_variant = false
 
-if host_cpu_family == 'aarch64' and is_windows
+if host_cpu_family == 'aarch64' and host_system in ['cygwin', 'windows']
```
nirbheek, line 82: armasm and cygwin? that sounds wrong.

subprojects/packagefiles/libffi/meson.build
```diff
@@ -95,7 +95,7 @@ elif host_cpu_family.startswith('alpha')
 elif host_cpu_family.startswith('arc')
   target = 'ARC'
   asm_sources = ['arcompact.S']
-elif host_cpu_family == 'arm' and is_windows
+elif host_cpu_family == 'arm' and host_system in ['cygwin', 'windows']
```
nirbheek, line 98: Same here, does cygwin even run on arm?

xclaesse: That's a translation of https://github.com/libffi/libffi/blob/master/configure.host#L9, I have no idea if it makes sense or not...

nirbheek: ... I guess it doesn't hurt since it's an impossible situation but would be good to get the maintainer to confirm

PR #874, opened 2023-01-02
SDL2: add option to use system iconv

Jan200101:
also removes extra `or false`'s.

The logic looks sound to me, but I would appreciate if others could look over.


subprojects/packagefiles/sdl2/meson.build
```diff
@@ -367,11 +368,21 @@ endif
...
     ibus_dep,
 ]
 
+if not opt_system_iconv.disabled()
+    iconv_dep = cc.find_library('iconv', required : opt_system_iconv)
+else
+    iconv_dep = fake_dep
+endif
+
+if cc.has_function('iconv_open', dependencies: iconv_dep)
```
Jan200101, line 381: to me this looks sound in theory.
But when testing it said it tested for `iconv_open` while linking with `-liconv`.
I have no libiconv on my system, was this only a visual aspect or did it link regardless of if it found anything?

eli-schwartz: Meson has a built-in `dependency('iconv')` that handles this correctly. But using it requires a fairly recent Meson.

You can emulate it for older versions of Meson by checking for iconv_open without any dependencies, and then if that fails, checking for it with `cc.find_library('iconv')`. Additional handling is needed to manage feature options correctly, and of course some systems do really weird things with has_function returning a dummy built-in that doesn't actually link, so you don't want to use has_function after all -- instead you want `cc.links()` on a crafted code snippet that actually includes iconv.h and runs a no-frills `int main() { iconv_open(...); }`.

It's all kinds of super fun, that's why I added the built-in dependency handling. :P

neheb:
dependency('iconv') should be used. Guard it with

```
```

to prevent warnings.

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -367,11 +368,24 @@ endif
...
     ibus_dep,
 ]
 
+if not opt_system_iconv.disabled()
+    iconv_dep = cc.find_library('iconv', required : opt_system_iconv)
+else
+    iconv_dep = fake_dep
+endif
+
+if cc.links('#include <iconv.h>\nint main() { iconv_open("", ""); }', name : 'iconv_open test')
```
eli-schwartz, line 381: This does need to add iconv_dep as a dependency, because if iconv is standalone the links() check will fail to link without it.

Jan200101: Yep, forgot to do that.

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -367,11 +368,28 @@ endif
...
-    cc.find_library('iconv', required : false),
     dbus_dep,
     ibus_dep,
 ]
 
+if not opt_system_iconv.disabled()
+    if meson.version().version_compare('>=0.60')
+        iconv_dep = dependency('iconv', required : opt_system_iconv)
+    else
+        iconv_dep = cc.find_library('iconv', required : opt_system_iconv)
```
eli-schwartz, line 379: If iconv is required, but it's available as a built-in libc thing, this will error out because find_library() fails.

subprojects/packagefiles/sdl2/meson.build
```diff
@@ -367,11 +368,28 @@ endif
...
+    endif
+else
+    iconv_dep = fake_dep
+endif
+
+if cc.links('#include <iconv.h>\nint main() { iconv_open("", ""); }', name : 'iconv_open test', dependencies : iconv_dep)
+    cdata.set10('HAVE_ICONV', true)
+    cdata.set10('HAVE_SYSTEM_ICONV', true)
```
eli-schwartz, line 387: If iconv is disabled this check still runs, and may find it as a built-in and set these defines to enable its use.

It should only run inside the `if not opt_system_iconv.disabled()` block.

eli-schwartz (changes requested):

subprojects/packagefiles/sdl2/meson.build
Jan200101: This option is not to disable iconv but to disable relying on an iconv library.

When the option is disabled it will check if libc provides it and otherwise disable iconv support.

Since we have features in meson the idea is
`auto` = lib if available otherwise libc
`enabled` = lib only
`disabled` = libc only

`dependency('iconv')` is useful, but will overall require more logic to handle the above.
Does the dependency solution do anything we cannot recreate using `find_library` and `links` ?

eli-schwartz: That seems like a strange option. Why would someone want to make that decision?

"automatically enable iconv functionality if it's a builtin, otherwise allow the user to choose"?

Jan200101: This copies the upstream behavior for enabled/disabled but adds extra functionality for auto which checks for both in the order of library and then libc.

Jan200101: Also some more context:
this is wanted for Lite-XL, where Windows builds currently link against libiconv because of SDL, even if thats not what is wanted.
The point is not to add an iconv toggle.

Jan200101: would it be better if I simply replaced it with a boolean and made it behave like upstream?

Jan200101: bump

Jan200101: one last bump before I consider this PR stale

eli-schwartz:
I don't understand why bad UX counts as "intentional, let's do it".

subprojects/packagefiles/sdl2/meson.build
eli-schwartz: >  this is wanted for Lite-XL, where Windows builds currently link against libiconv because of SDL, even if thats not what is wanted.

So... lite-xl just wants to turn iconv off unconditionally? Am I understanding correctly?

Jan200101: > So... lite-xl just wants to turn iconv off unconditionally? Am I understanding correctly?

No
Lite-XL wants iconv IF its not an extra dependency.
Otherwise this would require iconv.dll on Windows and libiconv.so on BSD.

eli-schwartz: So unconditionally enable it on Linux and unconditionally disable it everywhere else?

Jan200101: not necessarily.
iconv being included in libc is not a given and is just something glibc does and musl decided to copy for compatibility.
Other libcs that may be used on Linux may not implement it.

The idea is to seperate between toolchain iconv and library iconv.

eli-schwartz: We don't do that for libm or libdl, even though both are more likely to require a shared library than iconv is -- e.g. it's a shared library on glibc.

What if you have libiconv.a installed, which is suitable for not having to require libiconv.so or iconv.dll?

>  Other libcs that may be used on Linux may not implement it.

Which libcs for Linux are we talking about here? We're not exactly spoiled for choice -- a libc is a complex endeavor that, unless being statically linked -- in which case none of this discussion even matters, since you *have* to statically link iconv too -- requires extensive platform support, you need to compile your entire distro for it, or run multiple ***system libcs***.

Your choices are exactly glibc and musl.

Even if you wanted to statically link against a private libc, you don't really have a lot of choices when it comes to a libc. Every remaining option specifically targets embedded devices, bionic is not actually used outside of Android, dietlibc doesn't even support shared libraries, if you say "klibc" I'm just going to laugh, and...

... all of them other than klibc *still* support iconv as a builtin, not as an external library. For uclibc you might need to enable iconv support at configure time.

eli-schwartz: Soooo.

I reiterate: when building redistributables, unconditionally enable it on Linux, unconditionally disable it everywhere else.

Jan200101: #882

PR #878, opened 2023-01-16
minizip-ng: update to 3.0.8

neheb:


neheb:
```
glib| WARNING: CMake Toolchain: Failed to determine CMake compilers state
```

no idea. ping @eli-schwartz

PR #882, opened 2023-01-26
SDL2: implement system_iconv exactly like upstream it does

Jan200101:
Does it *exactly* like upstream does it.

This should leave no room for argumentation.


Jan200101:
making CI happy

eli-schwartz:
> Does it *exactly* like upstream does it.
> 
> This should leave no room for argumentation.

I can and will argue, both because:
- "the way upstream does it" is due to the fact that CMake *can't* do feature options properly so it doesn't seem completely odd to implement better granularity, though this is subjective
- your current implementation is broken and errors out on the common case, even though iconv exists and is exactly what you want it to be

Jan200101:
>     * "the way upstream does it" is due to the fact that CMake _can't_ do feature options properly so it doesn't seem completely odd to implement better granularity, though this is subjective

So?
I see wrapdb as a collection of meson setups for third party projects that intend to do the same as upstream does with some things changed to fit the proper meson way.

The upstream flag is for using libiconv or toolchain iconv, otherwise SDL2 will fall back onto its own implementation.

My suggested way with features was being able to choose between libiconv, toolchain iconv or whatever was found on the system.

Clearly we disagree on such, so using the upstream way of doing it, no matter how flawed, is preferred since it is the canonical way of how it should be done.

I encourage you to open a Pull Request to do it the way you want it, I am unwilling to continue arguing about this.

eli-schwartz:
> My suggested way with features was being able to choose between libiconv, toolchain iconv or whatever was found on the system.

Maybe I really don't understand what the option does, but equally, I think you've done a poor job explaining what the option does.

It really didn't look to me like you offered any such choice?

Honestly, it sounds like what you really want is `dependency('iconv', method: 'builtin')`?

If you actually wanted to make it a choice which one to use, then instead of using a feature to turn iconv on, and having different levels of "require it, probe for it, disable it" be used to determine where to look for it? I would have assumed you'd have one (feature) option to control whether iconv functionality is requested, and another combo option determining what types to allow.

Lacking an understanding of what all this is trying to accomplish, I can't really say what the best approach to take here is -- aside for being pretty sure that if the goal is providing redistributables, the build environment can pretty well guarantee what's going to be found, and the answer is *very* much tied to a) the operating system, of which all Linux ones have a built-in one, b) whether non-glibc systems have explicitly installed the enhanced, more featureful libiconv extracted from glibc and provided as a standalone library.

xclaesse:
Meson already has custom code to lookup iconv dependency, I think `dependency('iconv', required: get_option('iconv'))` is all we need here. Everything else can be removed.

Jan200101:
> Meson already has custom code to lookup iconv dependency, I think `dependency('iconv', required: get_option('iconv'))` is all we need here. Everything else can be removed.

yes but
1. it would bump the required meson version from `0.48.0` to `0.60.0`, which is quite a drastic bump in my opinion
2. the order in which `dependecy('iconv')` checks for iconv is reversed with how upstream does it.

1 is generally not a problem (only distro with meson < 0.60 I now of is debian buster (oldstable)) but 2 might be because GNU libiconv supports more encoding than e.g. musl.

xclaesse:
> 2 might be because GNU libiconv supports more encoding than e.g. musl.

Maybe we should reverse the order in Meson then.

eli-schwartz:
We will not change the order in Meson.

Detecting the builtin is the most likely case, because many libcs provide builtins (not even exclusively Linux).

If both exist, the builtin will always fail to link successfully, which means we fall back to the external "system" class.

If the builtin exists, and the external "system" one only has the runtime component installed, but not the development component, then checking for the system first will "succeed", and result in overlinking to an external shlib/DLL, because what it actually detected was... the fact that the builtin works.

Jan200101:
> If both exist, the builtin will always fail to link successfully, which means we fall back to the external "system" class.

How come?

Jan200101:
> Maybe we should reverse the order in Meson then.

At the end of the day there is no correct order, the developer should be given a choice to prefer one over the other.

eli-schwartz:
> How come?

Because you should not have multiple libraries in the same process space with the same symbols, it's a really bad idea. And one of the libraries is "built-in", after all. ;) So you can't *not* load libc at runtime.

In order to safely use GNU libiconv in combination with a libc that has a built-in iconv, you *must* ... call it something other than iconv. Which is exactly what GNU libiconv does. It installs its own header that incompatibly clashes with the libc one, and uses the magic of header defines to make your code, which contains "iconv_open", be preprocessed into something else.

If you use that header, you must use `-liconv` or the software fails to link.

If you do not use that header, then it doesn't matter how much you use `-liconv`, you ***will not get the enhanced function***. You will be stuck with the built-in musl function that doesn't support as many encodings.


And -- again if you do not use that header -- despite this, you will frequently still have to distribute libiconv.so in your application bundle, because you linked to it[^1], even though it's not used and you gain no benefit from it.

[^1]: If your linker supports the `--as-needed` flag, Meson will use it by default. You can configure this behavior with `-Db_asneeded=false`. If used, this flag tells the compiler that since `-liconv` wasn't actually used, it should skip adding a requirement tag `DT_NEEDED` to the executable metadata.

neheb:
I'll just point out the way OpenWrt does it: https://github.com/openwrt/openwrt/blob/master/include/nls.mk

The choice is global, not local. Works with regular meson dependency just fine.

xclaesse:
> In order to safely use GNU libiconv in combination with a libc that has a built-in iconv, you must ... call it something other than iconv. Which is exactly what GNU libiconv does.

Ah right, I remember we had the discussion on the past. So at the end of the day, if I remember correctly, what determine which implementation we use is which header is picked with `#include <iconv.h>`, right? So order does not matter, only one will work?

eli-schwartz:
Yes. This is why we look for the built-in first, because we need to tell whether `-liconv` is needed.

Jan200101:
> Because you should not have multiple libraries in the same process space with the same symbols, it's a really bad idea.

Being a bad idea has never actually stopped anyone from doing it.

> And one of the libraries is "built-in", after all. ;) So you can't _not_ load libc at runtime.

libc is, arguably, an optional dependency.

> In order to safely use GNU libiconv in combination with a libc that has a built-in iconv, you _must_ ... call it something other than iconv.

Yes but also no.
There is no problem when statically linking against a library iconv, but dynamic linking *may* be a problem.
I couldn't find anything that defines in what order dynamic symbols are resolved.

> Which is exactly what GNU libiconv does. It installs its own header that incompatibly clashes with the libc one, and uses the magic of header defines to make your code, which contains "iconv_open", be preprocessed into something else.

and that preprocessing is guarded behind a check to see if `LIBICONV_PLUG` is not defined.
Nothing would prevent someone from e.g. having a libiconv wrap that could act as a statically linked libc alternative.

Further more, the libiconv developers are aware that there may be a clash between software expecting libc iconv and libiconv, so they (according to the comments) try to be binary compatible.

> If you use that header, you must use `-liconv` or the software fails to link.

Another incompatibility is that glibc provides its own `iconv.h`

> If you do not use that header, then it doesn't matter how much you use `-liconv`, you _**will not get the enhanced function**_. You will be stuck with the built-in musl function that doesn't support as many encodings.

...

> And -- again if you do not use that header -- despite this, you will frequently still have to distribute libiconv.so in your application bundle, because you linked to it[1](#user-content-fn-1-913700293a7c00aad1f527b9934d9591), even though it's not used and you gain no benefit from it.
> 
> ...

...

> So, how do you choose which iconv to use? Very simple: choose which directories to add to the header include paths. But there's a big problem with that -- it's usually already on the default path (it will be in a linux package manager, unless that is NixOS which is an entirely different kettle of fish). And if it isn't on the default path, adding it to a meson-based build via CFLAGS will force the use of the external one. You cannot actually choose whether or not to use it, because your choice was obsolete by that time.

I'm not quite sure about this.
Checking for the library header is pointless since glibc also ships `iconv.h`, so it can only tell if you there is *some* iconv implementation.

And what if I prefer statically linking to libiconv over the libc implementation?

Looking at the actual dependency code in meson I also found that it only checks for an iconv library and never verifies if its actually usable.
We are in the lucky situation that unlike libm, libdl and pthreads, libc is not required to ship a libiconv.
But in theory what is preventing a libc implementation from doing that to fit in with the other libraries libc provides and possibly fix software that explicitly depends on a libiconv?

eli-schwartz:
> libc is, arguably, an optional dependency.

No? It really isn't?

If you're statically linking libc, you're not then dynamically loading libiconv.

If you're building in freestanding mode for embedded, then you'll not find iconv anyway at `dependency()` time. if you're statically linking both libc and libiconv, then maybe kind of sort of... except in both these cases, you're not loading multiple libraries into the same process space! And furthermore, this is just "And what if I prefer" from down below.

> I couldn't find anything that defines in what order dynamic symbols are resolved.

For good reason, it's not exactly well specified. IIRC in practice this is "first one wins" so libc probably always wins.

> Another incompatibility is that glibc provides its own `iconv.h`

This is not an incompatibility, this is exactly what I'm talking about in the first place.

> Checking for the library header is pointless since glibc also ships `iconv.h`, so it can only tell if you there is some iconv implementation.

Then it is a good thing that we don't just check for the library header.

> And what if I prefer statically linking to libiconv over the libc implementation?

Then this works fine using the preprocessed name.

> Looking at the actual dependency code in meson I also found that it only checks for an iconv library and never verifies if its actually usable.

Please check again, I fixed this bug a long time ago.

> We are in the lucky situation that unlike libm, libdl and pthreads, libc is not required to ship a libiconv.
> But in theory what is preventing a libc implementation from doing that to fit in with the other libraries libc provides and possibly fix software that explicitly depends on a libiconv?

I presume it would do the same thing that libdl does, and provide an empty libdl.a with no contents, to satisfy the link editor. Pretty sure there were plans to drop libm.so the same way.

Note that there is no requirement to ship either one -- posix merely states that the compiler when executed in link mode must accept `-lm`, the compiler is free to accept this via a case statement and not have an on-disk file. This part of why Meson checks those special names with `-l` in addition to searching for the file. In practice, everyone uses empty static archives.

Compiling with and without a do-nothing library makes no difference, but skipping the do-nothing library means shorter command lines.

> Further more, the libiconv developers are aware that there may be a clash between software expecting libc iconv and libiconv, so they (according to the comments) try to be binary compatible.

And the only goal there is to save you from UB if you link both into the same process and it's not specified which one gets guaranteed to be used. And I guess for LD_PRELOAD.

This isn't a valid technology stack. This is emergency safeguards to save people doing UB from themselves.

jpakkane:
> No? It really isn't?

On Windows it is feasible and some people actually do it.

eli-schwartz:
Ultimately this all seems to be distracting from the point.

Meson has had actual thought go into the default order of lookup. I don't anticipate this changing.

If for some reason you absolutely must have a specific iconv version, you can do that! There's a couple good reasons to do it. For example you might know that FreeBSD has an inferior icing, and mandate that the lookup method for that OS is "system" instead of "auto".

This doesn't detract from the fact that by default you want to do the commonly correct thing, and allow users to do so. I've already said above that it sounds like you just want to sometimes pass a method to this dependency. e.g. using the method "builtin" you can guarantee that Meson will only detect an iconv dependency if you can use it without linking to additional libraries. Or `static: true` should resolve either a builtin dependency, or "a system library but only if it can be statically linked".

eli-schwartz:
> On Windows it is feasible and some people actually do it.

So that's effectively just the freestanding case, surely? Not really a good reason to say that we should lookup iconv via an external library before looking it up in a built-in?

Jan200101:
> > libc is, arguably, an optional dependency.
> 
> No? It really isn't?

Yes really it is

> If you're statically linking libc, you're not then dynamically loading libiconv.
> 
> If you're building in freestanding mode for embedded, then you'll not find iconv anyway at `dependency()` time. if you're statically linking both libc and libiconv, then maybe kind of sort of... except in both these cases, you're not loading multiple libraries into the same process space! And furthermore, this is just "And what if I prefer" from down below.

You can build C programs without libc just fine.
Not even freestanding, just without the stdlib.

> For good reason, it's not exactly well specified. IIRC in practice this is "first one wins" so libc probably always wins.

Personal testing on Fedora 37 x86_64 I have found that the library wins.
But this may depend on how things were compiled.

> > And what if I prefer statically linking to libiconv over the libc implementation?
> 
> Then this works fine using the preprocessed name.
> 
> > Looking at the actual dependency code in meson I also found that it only checks for an iconv library and never verifies if its actually usable.
> 
> Please check again, I fixed this bug a long time ago.

<https://github.com/mesonbuild/meson/blob/01275fb09ec083aa8328ba100ed665bb9034c914/mesonbuild/dependencies/misc.py#L485-L494>
```py
class IconvSystemDependency(SystemDependency):
    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
        super().__init__(name, env, kwargs)
        self.feature_since = ('0.60.0', "consider checking for `iconv_open` with and without find_library('iconv')")

        h = self.clib_compiler.has_header('iconv.h', '', env)
        self.link_args = self.clib_compiler.find_library('iconv', env, [], self.libtype)

        if h[0] and self.link_args:
            self.is_found = True
```

is this not the code?


----

No matter `dependency('iconv')` is not worth bumping the meson version that much so talking about changing the behavior is pointless.

eli-schwartz:
Sorry, I misread that, we fixed the code that might break -- the built-in -- to do a full link test.

For the system dependency, is there a case where the library and header might exist but isn't usable?

Jan200101:
> Sorry, I misread that, we fixed the code that might break -- the built-in -- to do a full link test.
> 
> For the system dependency, is there a case where the library and header might exist but isn't usable? I could not think of a case that could naturally occur, only cases that could be rigged specifically to prove that it's technically possible. I figured it probably wasn't worth the build performance hit of checking that, for the same reason we don't do compile checks after successfully finding a pkg-config dependency.

Arguably it should be fine but I prefer sanity checking the environment so that a function that is suppose to return an iconv dependency actually returns iconv

Jan200101:
Welp
This isn't going anywhere and I like cutting my losses short:

Going forward I would not like to contribute to or be associated with the SDL2 wrap anymore.

PR #885, opened 2023-01-31
google-brotli: add dllimport to deps

neheb:
Fixes: warning C4273: '': inconsistent dll linkage


neheb:
hmm now I get:

```
warning: 'BrotliDefaultFreeFunc' redeclared without dllimport attribute after being referenced with dll linkage
```

with MinGW GCC

PR #886, opened 2023-01-31
github: use ninja from pip

neheb:
More up to date

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -26,8 +26,7 @@ jobs:
       - name: Install packages
         run: |
           sudo apt-get update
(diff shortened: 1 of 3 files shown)


eli-schwartz:
That sounds wrong, especially for homebrew. And why is this changing things other than ninja, anyway?

Note: I am not fond of the pip version of ninja specifically, because it's not actually upstream ninja. It's a fork with patches that make ninja behave a bit differently (not a lot differently). I don't anticipate it causing different CI results, but why  go out of the way here?

...

Is there any specific problem this PR was motivated by?

neheb:
Not sure what you mean. The apt command only installs ninja. The rest is preinstalled.

PR #889, opened 2023-02-01
Draft: libsrtp2: update to v2.5.0

tp-m:
Untested so far, fwiw.


tp-m:
Passed GStreamer CI builds at least (Windows + Linux).

PR #892, opened 2023-02-06
flatbuffers: update to 23.1.21

jmillan:


jmillan:
> why remove patch_directory = flatbuffers ?

Oh, that was unintentioal.

jmillan:
Fixed.

eli-schwartz:
In the future, please use `git commit --amend` and force push -- do NOT delete something in one commit and add it back in another commit, that results in very messy commit logs... :)

(We should not be merging it like that, IMO.)

jmillan:
> In the future, please use git commit --amend and force push

Ok, I will :+1:

PR #894, opened 2023-02-10
zstd: upstream update to 1.5.4

eli-schwartz:
merge upstream changes to meson.build; as usual, drop:
- subdir() paths rewritten away from build/meson/
- re-adding the project version script


eli-schwartz:
I am *finally* able to update this with clean CI, even works on Windows. :D This zstd release has a bunch of fixes I asked for.

PR #899, opened 2023-02-17
Add Asio

### LINKED ISSUE #896: Request: Asio
https://github.com/chriskohlhoff/asio

Header-only by default, and no required dependencies. Will probably implement Meson here myself.

According to https://think-async.com/Asio/asio-1.24.0/doc/asio/using.html, a shared lib can also be build and there are optional dependencies. Not sure if I will add it, have to take a look at it first.

stephanlachnit:
Superseeds #523 (mostly copied mesonization).
Closes #896.

### DIFF
--- a/subprojects/packagefiles/asio/meson.build
+++ b/subprojects/packagefiles/asio/meson.build
@@ -1,0 +1,22 @@
+project('asio',
+    'cpp',
+    version : '1.24.0',
+    license : 'BSL-1.0',
+)
+
+asio_deps = []
+
+# For windows builds, ASIO depends on ws2_32
+if target_machine.system() == 'windows'
+    compiler = meson.get_compiler('cpp')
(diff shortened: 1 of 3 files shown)


subprojects/packagefiles/asio/meson.build
```diff
@@ -0,0 +1,22 @@
...
+    version : '1.24.0',
+    license : 'BSL-1.0',
+)
+
+asio_deps = []
+
+# For windows builds, ASIO depends on ws2_32
+if target_machine.system() == 'windows'
+    compiler = meson.get_compiler('cpp')
+    ws2_32_dep = compiler.find_library('ws2_32', required: true)
```
jpakkane, line 12: You don't need `required: true`. It is the default.

stephanlachnit: Ah classic copy-pasta, thx!

PR #904, opened 2023-03-01
Add c-flags

DieTime:
Adds c-flags 1.5.4

### DIFF
--- a/subprojects/c-flags.wrap
+++ b/subprojects/c-flags.wrap
@@ -1,0 +1,8 @@
+[wrap-file]
+directory = c-flags-1.5.4
+source_url = https://github.com/DieTime/c-flags/archive/1.5.4.tar.gz
+source_filename = c-flags-1.5.4.tar.gz
+source_hash = 890345a0a378bc1a2dcff8c39b0e7745947153bd2c2e1fd0e6b60944f78b46ae
+
+[provide]
+c-flags = libcflags_dep
--- a/ci_config.json
+++ b/ci_config.json
@@ -28,6 +28,11 @@
     "rtaudio",
     "tinyxml2"
   ],
+  "c-flags": {
+    "build_options": [
+      "c-flags:tests=true"
+    ]
+  },
   "cexception": {
     "build_options": [
       "cexception:werror=false"
--- a/releases.json
+++ b/releases.json
@@ -92,6 +92,14 @@
       "2.3.1-2"
     ]
   },
+  "c-flags": {
+    "dependency_names": [
+        "c-flags"
+    ],
+    "versions": [
+        "1.5.4-1"
+    ]
+  },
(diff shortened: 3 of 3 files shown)


subprojects/c-flags.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = c-flags-1.5.2
+source_url = https://github.com/DieTime/c-flags/archive/1.5.2.tar.gz
+source_filename = c-flags-1.5.2.tar.gz
+source_hash = f4f4aed5156dca60af2434cecfa45ccd706eef37bf8b731459ebd41ca1daf150
+
+[provide]
+cflags = cflags_dep
+cflags_static = cflags_static_dep
```
eli-schwartz, line 9: This is interesting to see, as usually projects tend to rely on meson's `default_library={static|shared|both}` to generate whichever one is suitable. It's also something that a subproject fallback can set so that one links to the shared library if it's globally installed, but the static library if a subproject is used.

DieTime: Thanks, I changed the build script, now the library is built using the `library` command, so that the user can specify the type of library using `default_library`.

DieTime: Also fixed compatibility for Windows. The last PR fails on Windows workflow.

subprojects/c-flags.wrap
```diff
@@ -0,0 +1,8 @@
+[wrap-file]
+directory = c-flags-1.5.3
+source_url = https://github.com/DieTime/c-flags/archive/1.5.3.tar.gz
+source_filename = c-flags-1.5.3.tar.gz
+source_hash = 81dd6d961c46f0039c58a381d2cb2aba786d553f02271f39d8b34722a2c4689d
+
+[provide]
+libcflags = libcflags_dep
```
eli-schwartz, line 8: Hmm... the dependency which you provide as a wrap interface is "libcflags" but the pkg-config file installed is "c-flags.pc", so this doesn't provide the same interface. The wrap dependency name should really correspond to the name that `dependency()` would detect via pkg-config.

DieTime: Do I understand correctly that the dependency name `c_flags` will be correct in case of `c-flags.pc`?

eli-schwartz: No, but `c-flags` will be correct in the case of `c-flags.pc`. :)

DieTime: Thanks, I fixed the mistake.

DieTime: Also corrected `releases.json`

DieTime: In version 1.5.4 corrected the export of symbols.
Pull Request: DieTime/c-flags#2.

PR #905, opened 2023-03-02
Add libzmq 4.3.5

stephanlachnit:
This is a first draft of adding the core ZeroMQ library libzmq to WrapDB. Should work on Linux.

Things to fix:
- [x] Fix Windows support
- [x] Check MacOS / BSD support (BSD not actually tested, but should be fine)
- [x] Things mentioned as "broken in [...] upstream" (https://github.com/zeromq/libzmq/pull/4521, https://github.com/zeromq/libzmq/pull/4522, https://github.com/zeromq/libzmq/pull/4523)
- [ ] ~At least one test / example~ (postponed, already used extensively in a project from us)
- [ ] ~Find out how to depend on VMCI~ (not important enough)
- [ ] ~PGM version as build option~ (not important enough)
- [x] Better summary
- [x] CI config for optional features (draft, tls, pgm, norm)
- [ ] ~Run cppzmq with libzmq wrap~ (postponed, do this in a separate pull request)

/cc @bluca let me know if you are interested in this upstream (maybe to replace autoconf)

### DIFF
--- a/subprojects/libzmq.wrap
+++ b/subprojects/libzmq.wrap
@@ -1,0 +1,10 @@
+[wrap-file]
+directory = libzmq-4.3.5
+source_url = https://github.com/zeromq/libzmq/archive/refs/tags/v4.3.5.tar.gz
+source_filename = libzmq-4.3.5.tar.gz
+source_hash = 6c972d1e6a91a0ecd79c3236f04cf0126f2f4dfbbad407d72b4606a7ba93f9c6
+patch_directory = libzmq
+diff_files = libzmq/fix-mingw-afunixh.diff
+
+[provide]
+libzmq = libzmq_dep
--- a/subprojects/packagefiles/libzmq/meson.build
+++ b/subprojects/packagefiles/libzmq/meson.build
@@ -1,0 +1,568 @@
+project('libzmq', ['cpp', 'c'],
+  version: '4.3.5',
+  license: 'LGPL-3.0-or-later',
+)
+
+cc = meson.get_compiler('c')
+host_os = host_machine.system()
+
+thread_dep = dependency('threads')
+
+libzmq_inc_private = include_directories('src', 'external/wepoll')
+libzmq_inc_public = include_directories('include')
+
+libzmq_deps_private = [thread_dep]
+libzmq_deps_public = []
+
+libzmq_flags_private = ['-DZMQ_CUSTOM_PLATFORM_HPP', '-D_REENTRANT', '-D_THREAD_SAFE']
+libzmq_flags_public = []
+
+libzmq_src = files(
+  'src/precompiled.cpp',
+  'src/address.cpp',
+  'src/channel.cpp',
+  'src/client.cpp',
+  'src/clock.cpp',
+  'src/ctx.cpp',
+  'src/curve_mechanism_base.cpp',
+  'src/curve_client.cpp',
+  'src/curve_server.cpp',
+  'src/dealer.cpp',
+  'src/devpoll.cpp',
+  'src/dgram.cpp',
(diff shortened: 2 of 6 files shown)


stephanlachnit:
Note: feel free to already review this

subprojects/packagefiles/libzmq/meson.build
```diff
@@ -0,0 +1,480 @@
+project('libzmq', ['cpp', 'c'],
+    version: '4.3.4',
+    license: 'LGPL-3.0-or-later',
+    default_options: ['cpp_std=gnu++11'],
```
neheb, line 4: c++11 or 14

subprojects/packagefiles/libzmq/meson.build
```diff
@@ -0,0 +1,480 @@
...
+)
+
+cc = meson.get_compiler('c')
+thread_dep = dependency('threads')
+
+libzmq_inc_private = include_directories('.', 'src')
```
neheb, line 10: . is unnecessary

stephanlachnit:
In principle this can (almost) be merged for linux. Windows and other UNIXes are broken but not really a priority for me. Is there a policy for this?

neheb:
policy is make it fail for Windows and mark to not build on windows and msys2 in ci_config.json.

stephanlachnit:
MinGW requires patching due to https://github.com/zeromq/libzmq/pull/4678 (waiting is not really an option giving how infrequent libzmq has releases). I bumped the minimum meson version to `0.63.0` to allow for this.

stephanlachnit:
Overall this is ready to review  🎉 

The main point open for discussion is the bump to meson `0.63.0` and the addition of patch files. I could avoid this by throwing an error if mingw is detected and disabling the mingw CI. Please let me know what you prefer this over the version bump.

I see three minor TODOs for the future:
- Use libzmq wrap in cppzmq wrap and enable CI on Windows (very easy)
- Test static libraries, e.g. there is a `ZMQ_STATIC` setting (probably not too hard)
- Run the tests (a bit more work)

Once this is merged I will open a follow-up for cppzmq.

subprojects/libzmq.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = libzmq-4.3.5
+source_url = https://github.com/zeromq/libzmq/archive/refs/tags/v4.3.5.tar.gz
+source_filename = libzmq-4.3.5.tar.gz
+source_hash = 6c972d1e6a91a0ecd79c3236f04cf0126f2f4dfbbad407d72b4606a7ba93f9c6
+patch_directory = libzmq
+diff_files = libzmq/fix-mingw-afunixh.diff
```
jpakkane, line 7: WrapDB has a policy that things in it may not have functional changes, only build system updates. Weird that there is not a test for this.

stephanlachnit: There is, but I added an exception.

jpakkane: In any case it is not permitted. We can't really have any functionality delta because we don't have the manpower to keep maintaining them and if anything goes wrong, then it's on us.

stephanlachnit: Alright - then I will revert those changes and wait until a new version is released

stephanlachnit:
Is there a possibility to declare a dependency with different arguments if linking shared/static? I.e. if I `default_library=both_libraries`, I can simply set `cpp_static_args` and `cpp_shared_arsgs`. However I also need to export a static arg to dependencies (namely `ZMQ_STATIC`). Alternative would be to error out if `default_library=both_libraries`.

PR #908, opened 2023-03-03
Add fallback URL for upstream tarball

xclaesse:
Some upstream servers are not reliable and randomly fail on build-all
CI. We can automatically upload the tarball to github release artifacts
as fallback.

enet, libwebsockets: Reupload to add fallback URL


tools/create_release.py
```diff
@@ -151,6 +152,14 @@ def upload(self, path: Path, mimetype: str):
         response = requests.post(self.upload_url, headers=headers, params=params, data=path.read_bytes())
         response.raise_for_status()
 
+    def create_source_fallback(self):
+        response = requests.get(self.wrap_section['source_url'])
```
eli-schwartz, line 156: Shouldn't we use meson subprojects download for this, which performs its own safety checks?

xclaesse: Meson is not installed in release CI job and the url/checksum should already have been checked by the CI before merging. I think that should be good enough, no?

eli-schwartz: Hmm, fair point.

tools/create_release.py
```diff
@@ -151,6 +152,14 @@ def upload(self, path: Path, mimetype: str):
...
+        response.raise_for_status()
+        filename = Path(self.wrap_section['source_filename'])
+        filename.write_bytes(response.content)
+        self.upload(filename, 'application/zip')
```
eli-schwartz, line 160: What is the content type even used for? Seems like a pointless aspect of GitHub 's API. They should not and probably do not care whether it's a zip or a tarball. So the fact that we get this wrong probably does not matter.

xclaesse: I could use `mimetypes.guess_type()` but I don't think it matters.

PR #912, opened 2023-03-04
libarchive: fix `progs` option

benoit-pierre:
- actually build the programs specified
- add `bsdcpio` to the list of supported programs


neheb:
@eli-schwartz is dependency_names the recommended way?

subprojects/packagefiles/libarchive/tar/meson.build
```diff
@@ -0,0 +1,22 @@
+src = [
```
neheb, line 1: use files().

subprojects/packagefiles/libarchive/tar/meson.build
```diff
@@ -0,0 +1,22 @@
...
+
+if host_machine.system() == 'windows'
+  src += 'bsdtar_windows.c'
+endif
+
+bsdtar = executable(
+  'bsdtar',
+  dependencies: [libarchive_fe_dep, rx_dep],
+  install: true,
+  sources: src,
```
neheb, line 19: sources: is optional.

subprojects/packagefiles/libarchive/tar/meson.build
benoit-pierre: FWIW, the `meson rewrite` does not use files in those cases.

neheb: files() helps with muon's formatter.

eli-schwartz: muon's formatter is capable of discerning that order is not significant for arguments to the files() function, but order *might* be significant in arbitrary arrays.

This allows muon to alphanumerically sort arguments. Assuming you care. I don't care one way or another, myself.

subprojects/packagefiles/libarchive/cat/meson.build
```diff
@@ -0,0 +1,11 @@
+bsdcat = executable(
+  'bsdcat',
+  files(
+    'bsdcat.c',
+    'cmdline.c',
+  ),
```
eli-schwartz, line 6: Since we're discussing moving over to files() I think it's worth noting that *inlining* the files() function as an argument to executable() is totally pointless. It's also less readable.

subprojects/packagefiles/libarchive/cpio/meson.build
```diff
@@ -0,0 +1,17 @@
...
+  'cpio.c',
+  ]
+
+if host_machine.system() == 'windows'
+  src += 'cpio_windows.c'
+endif
+
+bsdcpio = executable(
+  'bsdcpio',
+  files(src),
```
eli-schwartz, line 12: If you're going to use files() on the basis that `muon fmt` can discern the order of literal arguments and lexicographically sort them, then that's one thing. Passing an array as a variable reference to files() is pointless though because it doesn't help `muon fmt` *and* it's unneeded / potentially less readable.

subprojects/packagefiles/libarchive/cat/meson.build
eli-schwartz: (Now granted, `muon fmt` doesn't discern that order isn't significant here. But it *could*.)

subprojects/packagefiles/libarchive/tar/meson.build
eli-schwartz: I should say that it's *arguably* not significant. There's no technical reason why people cannot use a list of `srcs = files(...)`, and then use `srcs[0]` for one specific executable and other indexes for other targets.

It is just that it's a reasonably safe assumption that doing so is quite unusual, so while technically sorting them is semantically changing the build file and thus unsafe, it's likely safe to do so in the common case. Whereas arrays are inherently not particularly likely to have files in them in the first place.

subprojects/packagefiles/libarchive/cat/meson.build
benoit-pierre: > Since we're discussing moving over to files() I think it's worth noting that _inlining_ the files() function as an argument to executable() is totally pointless. It's also less readable.

I actually find the separation between the name of the executable and the sources more readable, but would just use an array:
```python
bsdcat = executable(
  'bsdcat',
  [
    'bsdcat.c',
    'cmdline.c',
  ],
  dependencies: libarchive_fe_dep,
  install: true,
)
```

subprojects/packagefiles/libarchive/cpio/meson.build
benoit-pierre: Right, that one was dumb…

benoit-pierre:
Since I have other `libarchive` related changes, would rather I update this PR (multiple commits of course), or do you prefer separate ones?

eli-schwartz:
I would update this one and fix all the issues at once, the main difference really is that you'd just bump the version once rather than once per fix. No point in uploading a new wrap release if you know you're just going to upload a new one in a few minutes. :D

subprojects/packagefiles/libarchive/cat/meson.build
eli-schwartz: Sure, ultimately the important thing is a) self-consistency, followed by b) what the people primarily contributing to that wrap find easy to work with.

With that in mind, whichever one you choose should be consistent between here and `libarchive_fe`. :)

benoit-pierre: > Sure, ultimately the important thing is a) self-consistency, followed by b) what the people primarily contributing to that wrap find easy to work with.

You forgot: potential compatibility with dumb tools (`meson rewrite`…) :P.

> With that in mind, whichever one you choose should be consistent between here and `libarchive_fe`. :)

Dammit!

benoit-pierre:
Updated with other changes:
 - fix `progs` option: actually build the programs specified
      and add `bsdcpio` to the list of supported programs.
 - fix build when some dependencies are internal (e.g. provided
      by a subproject): zlib, zstd
 - simplify iconv checks

Note:
- I only fixed some of the dependency checks (for those features I need)
- I bumped the meson version required to `0.60.0` (for `iconv`)

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -2,7 +2,7 @@ project(
   'libarchive',
   'c',
   version: '3.6.2',
-  meson_version: '>=0.49.0',
```
neheb, line 5: Unwritten rule is to keep compatibility with meson 0.49, which is included with Debian 9. I don't see a compelling reason to increase to 0.60.

benoit-pierre: This really should be mentioned in the "How to contribute new wraps" section…

neheb: One of the reasons it's unwritten is because wraps aren't really tested on Debian 9. Last time I did that, I found that meson 0.49 crashes with ternary operators (?: instead of if else). Another reason is that some wraps really need a newer version.

Anyway, 0.60 is really new. I don't think most OSes include it.

benoit-pierre: That does not make any sense. Either it's a rule, and it should be enforced by running the CI with the minimum supported version, or it's just a lie.

Anyway, even without the iconv change, it does not work with `0.49` (`disabler().partial_dependency(…)` does not work), would need at least `0.54` anyway (for the former issue and `meson.override_dependency`).

I don't think it's too much to ask for a semi-recent version of meson.

eli-schwartz: It's not even a rule, really.

But it is a good rule of *thumb*, that wraps should work on the oldest version of Meson that it isn't too painful to support. I'm not a fan of people marking wraps as requiring very recent Meson "just because that is all I tested it with", and sometimes it's very easy to support old versions with low effort.

The right balance is probably "review comment: you could probably avoid bumping the meson_version so high by just doing XXX. I would do so, but it's up to you in the end".

eli-schwartz: 0.49.0 is a very robust checkpoint because it unlocks use by people running Debian oldstable using apt instead of pip to install meson.

0.56.0 is another good checkpoint because Debian stable has it.

eli-schwartz:
> multiple commits of course

Btw what happened to that? :)

subprojects/packagefiles/libarchive/meson.build
benoit-pierre: Unclear guidelines waste both our time. It may not be painful to support older versions, but it does take more time. At the very least all wraps should be updated with `meson_version: '>=0.49.0',` so you get a warning when using a feature only provided by a more recent version.

benoit-pierre:
> > multiple commits of course
> 
> Btw what happened to that? :)

I interpreted `fix all the issues at once` as: one commit…

eli-schwartz:
The partial_dependency() uses are probably not necessary, because if a `dependency()` succeeds then the canonical header should never not exist.

There's 3 types of dependency detection:
- pkg-config file exists, packaged in the same distro package as the headers
- meson built-in detection for e.g. zlib, will check for headers already, manually, as a prerequisite to reporting it as found
- subproject fallback which is assumed to set up the overridden dependency with includes support

I genuinely don't know why software does one check for a dependency and another check for HAVE_FOO_H, it's rather pointless as they can only be used together.

The only time you need to check for both is when you check for a library with cc.find_library() and need to detect the case where e.g. on Debian the runtime libraries are installed but not the devel component. Because raw library checks are awkward.

eli-schwartz:
> I interpreted `fix all the issues at once` as: one commit…

Sorry, I intended that as a response to "one fix per PR". Multiple commits are good for review and bisectability, one PR is good for avoiding administrative overhead in release management.

benoit-pierre:
`0.49.0` support is out: 
```
libarchive/meson.build:148:0: ERROR:  #mesondefine does not contain exactly two tokens: #cmakedefine ZLIB_WINAPI 1
```

benoit-pierre:
`0.54.0` seems to bork out for similar reasons…

benoit-pierre:
OK, `0.54.1` minimally required, and partial support for building with internal dependencies (only `zlib`/`zstd`, although I don't even need it anymore thanks to `--auto-features=disabled`).

benoit-pierre:
And I forgot to remove the call to `meson.override_find_program`…

benoit-pierre:
> I genuinely don't know why software does one check for a dependency and another check for HAVE_FOO_H, it's rather pointless as they can only be used together.

Well… Just look at the failed macOS CI run: the `libpcreposix` dependency was found, version `8.44`, but `pcreposix.h` is apparently missing (it is part of the official source distribution for `8.44`, I checked).

benoit-pierre:
OK, I changed the `regex` feature code to (partially) mirror what is done in `CMakeLists.txt`:
- check for glibc support first
- `librex` availability second
- `libpcreposix` last

eli-schwartz:
> Well… Just look at the failed macOS CI run: the `libpcreposix` dependency was found, version `8.44`, but `pcreposix.h` is apparently missing (it is part of the official source distribution for `8.44`, I checked).

Amazing! Based on some debugging: https://github.com/mesonbuild/wrapdb/actions/runs/4360296271/jobs/7623052633

macOS appears to provide a /usr/lib version of pcre, with a pkg-config file that specifies the library is in /usr/lib but a general prefix of /usr/local, and doesn't ship any headers in either directory anyway.

Totally broken, the pkg-config is a lie and there's no way to get anything workable there like at all, I guess? IMHO it's safe to say that you only need `cc.has_header('pcreposix.h', ...)` if host_machine.system() == 'darwin'. I would not otherwise bother doing this sort of double check. macOS sometimes has incredibly broken Apple-provided system components, but that's the only place I can think of where this ever happens.

eli-schwartz:
No idea how I hit "close with comment" halfway through responding, sorry...

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -299,16 +299,42 @@ else
...
+if not rx_dep.found()
+  if cc.has_header('regex.h') and cc.has_function('regcomp')
+    rx_dep = declare_dependency()
+    cdata.set('HAVE_REGEX_H', 1)
+    rx_support = 'glibc'
   endif
 endif
+if not rx_dep.found()
+  regex_dep = cc.find_library('regex', disabler: true, required: false)
+  if cc.has_header_symbol('regex.h', 'regcomp', dependency: regex_dep.partial_dependency(compile_args: true, includes: true))
```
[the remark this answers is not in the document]
benoit-pierre: Right, I just mirrored the glibc check (with the added dependency).

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -299,16 +299,42 @@ else
...
-  if cc.has_function('regcomp', dependencies: regex_dep)
-    deps += regex_dep
+# If `regex` is support is enabled, check for
+# - glibc support first
+# - `libregex` second
+# - `libpcreposix` last
+rx_support = false
+rx_dep = get_option('regex').disabled() ? declare_dependency() : disabler()
+if not rx_dep.found()
+  if cc.has_header('regex.h') and cc.has_function('regcomp')
```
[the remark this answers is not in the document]
benoit-pierre: Done.

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -299,16 +299,42 @@ else
...
-elif get_option('regex').auto()
-  regex_dep = cc.find_library('regex', required: false)
-  if cc.has_function('regcomp', dependencies: regex_dep)
-    deps += regex_dep
+# If `regex` is support is enabled, check for
+# - glibc support first
+# - `libregex` second
+# - `libpcreposix` last
+rx_support = false
+rx_dep = get_option('regex').disabled() ? declare_dependency() : disabler()
```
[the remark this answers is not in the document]
benoit-pierre: Done.

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -299,16 +299,42 @@ else
...
+    cdata.set('HAVE_PCREPOSIX_H', 1)
+    rx_support = 'libpcreposix'
+  endif
+endif
+if not rx_dep.found()
+  if get_option('regex').enabled()
+    error('Feature regex cannot be enabled')
+  endif
+  rx_dep = declare_dependency()
+endif
```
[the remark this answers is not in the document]
benoit-pierre: But I think the error message is more confusing for the user.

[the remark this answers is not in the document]
eli-schwartz: I can hear that logic.

Do you think that the `not_found_message: 'some message'` kwarg to dependency() helps here? It is printed as `message()` immediately after the dependency if it isn't found -- and then if it's required, the failure is re-raised. It would looks like this:

```
Run-time dependency foobar found: NO (tried pkgconfig and cmake)
Message: this is a not found message

meson.build:3:0: ERROR: Dependency "foobar" not found, tried pkgconfig and cmake
```

Or

```
Dependency foobar skipped: feature foo disabled
Message: this is a not found message
Build targets in project: 1

Found ninja-1.11.0.git at /usr/local/bin/ninja
```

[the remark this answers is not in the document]
benoit-pierre: I think it's still less readable (not the last message, not an error), plus in this case the possibility of missing `pcreposix.h` still need to be accounted for. Although the macOS CI build is currently happy with the glibc provided support, but I have no idea if older / other macOS version can be assumed to also provide it.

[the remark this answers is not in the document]
eli-schwartz: All right then.

PR #915, opened 2023-03-05
expat: fix dllexport

neheb:
Overriding XMLIMPORT like this prevents the code from applying dllexport


neheb:
the code has no dllexport. Amazing.

PR #920, opened 2023-03-07
imguizmo: add exports for Windows

neheb:


neheb:
hmm didn't work. oh well.

PR #877, opened 2023-01-16
fluidsynth: update to 2.3.2

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>

PR #925, opened 2023-03-10
glib: update to 2.76.1

bgilbert:


bgilbert:
This seems to be hitting the same failure as https://github.com/mesonbuild/wrapdb/pull/877:

```
glib| WARNING: CMake Toolchain: Failed to determine CMake compilers state
```

eli-schwartz:
cmake -- what joy. :roll_eyes: 

I tracked down the issue, I think? No idea what the fix is.

neheb:
@eli-schwartz is cmake needed on CI to find anything?

PR #930, opened 2023-03-12
simdjson: install lib and header

stephanlachnit:
Forgotten in #897


stephanlachnit:
Not sure why Windows fails...

jpakkane:
[Almost certainly because of this](https://mesonbuild.com/FAQ.html#my-project-works-fine-on-linux-and-mingw-but-fails-to-link-with-msvc-due-to-a-missing-lib-file-fatal-error-lnk1181-why).

neheb:
Yep. Windows does not export symbols by default

stephanlachnit:
Ah of course, what else^^ Does MSVC support `gnu_symbol_visibility`?

neheb:
https://github.com/simdjson/simdjson/blob/cd13b02bf09f68c50571e04e575d1947d5721ed8/include/simdjson/common_defs.h

Need to define SIMDJSON_VISUAL_STUDIO if argument syntax is msvc for dep and library and SIMDJSON_BUILDING_WINDOWS_DYNAMIC_LIBRARY for the library. SIMDJSON_USING_WINDOWS_DYNAMIC_LIBRARY for the dep

stephanlachnit:
> Need to define SIMDJSON_VISUAL_STUDIO if argument syntax is msvc for dep and library and SIMDJSON_BUILDING_WINDOWS_DYNAMIC_LIBRARY for the library. SIMDJSON_USING_WINDOWS_DYNAMIC_LIBRARY for the dep

Thanks!

eli-schwartz:
Awesome!

By the way, since upstream does seem to support setting the DLL API macro automatically on Unix to the gnu symbol visibility attribute, maybe we should also define `gnu_symbol_visibility: true`? It doesn't prevent failing installs, but it does protect against symbol leakage and provide optimization opportunities to the compiler.

neheb:
Please also rebase to current master.

PR #934, opened 2023-03-18
github CI: test clang-cl

neheb:
Various wraps here handle clang-cl. Test for it.


neheb:
```
 meson.build:1:0: ERROR: Compiler clang-cl can not compile programs.
```

hmmmm

PR #939, opened 2023-03-24
zlib: add meson option for tests.

dabrain34:


eli-schwartz:
This test doesn't include any additional dependencies etc. -- any particular reason you'd rather disable it entirely?

(Note that it's possible to configure tests but not run them, simply pass the name of the superproject as a suite name to `meson test`.)

neheb:
why not just bool?

eli-schwartz:
> I dont need to run this test in my project

And as I said, it's possible to filter the tests so that you only run the ones from your project, regardless of how many subprojects you have or what options they support to control it. Works for more than just zlib, even. :)

eli-schwartz:
> why not just bool?

Indeed, if no functionality of a tristate is being used then it seems odd to use a tristate.

dabrain34:
I followed the gstreamer policy as zlib can be often a subproject (for gstreamer for example) it can yield the parent project.

xclaesse:
There is a subtle side effect of using feature option: `-Dauto_features=disabled` will also disable tests which is probably a good thing because that's used to make minimal builds.

nirbheek:
Feature options are mechanically "just tri-state" but semantically they also make sense for features, as the name suggests, anything that is an optional feature that can be disabled or enabled as per the user's requirements.

PR #949, opened 2023-03-30
CI: test xcode and vs backends

neheb:


eli-schwartz:
Maybe these should be additional jobs instead. Backend-specific quirks can often rear their heads if you don't verify, but it's not necessarily specific to not-ninja.

neheb:
Goal currently is to run build all and close this PR. https://github.com/neheb/wrapdb/actions/runs/4568275989 partially failed.

neheb:
xcode is totally broken. vs seems to work:

```
13 failed: fluidsynth, glew, glib, harfbuzz, json-glib, libffi, sdl2, sdl2_image, sdl2_mixer, sdl2_net, sdl2_ttf, theora, zstd
```

theora and zstd seem like they need to be fixed.

Issue #950, opened 2023-03-30
ICU 72.1 wrap doesn't build, not configurable


srl295:
Hi! I'm not new to ICU. but I'm pretty new to meson.

Trying to add ICU as a subproject… meson 0.63.0 on mac.

```
icu4c = subproject('icu', default_options: [ 'default_library=static', 'cpp_std=c++17'])

icu_uc = icu4c.get_variable('icuuc_dep')


…
executable(…
   dependencies: [icu_uc],
)
```

Errors below.  Now, is there a way I can set switches on the ICU subproject? 

I'd set something like the following

```
# subprojects/icu/meson.build
add_project_arguments('-DUCONFIG_NO_BREAK_ITERATION=1', '-DUCONFIG_NO_LEGACY_CONVERSION=1', '-DUCONFIG_NO_FORMATTING=1', language: 'cpp')
add_project_arguments('-DU_ATTRIBUTE_DEPRECATED=', language: 'cpp')
```

… but these should be switchable by the caller.

-----

```
[1/417] Compiling C++ object subprojects/icu/source/common/libicuuc.72.dylib.p/ucnvhz.cpp.o
FAILED: subprojects/icu/source/common/libicuuc.72.dylib.p/ucnvhz.cpp.o 
clang++ -arch x86_64 -Isubprojects/icu/source/common/libicuuc.72.dylib.p -Isubprojects/icu/source/common -I../../../subprojects/icu/source/common -fcolor-diagnostics -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -Werror -std=c++17 -O3 -DU_COMMON_IMPLEMENTATION -MD -MQ subprojects/icu/source/common/libicuuc.72.dylib.p/ucnvhz.cpp.o -MF subprojects/icu/source/common/libicuuc.72.dylib.p/ucnvhz.cpp.o.d -o subprojects/icu/source/common/libicuuc.72.dylib.p/ucnvhz.cpp.o -c ../../../subprojects/icu/source/common/ucnvhz.cpp
../../../subprojects/icu/source/common/ucnvhz.cpp:558:9: error: 'ucnv_safeClone_72' is deprecated [-Werror,-Wdeprecated-declarations]
        ucnv_safeClone(((UConverterDataHZ*)cnv->extraInfo)->gbConverter, &localClone->subCnv, &size, status);
        ^
[...]

PR #951, opened 2023-03-30
icu: quell deprecation warnings during build

srl295:
Ref: #950


neheb:
need to add version bump in releases.json

srl295:
@neheb  let me see if i can make any other improvements then

srl295:
@neheb hm, that requires a new tag. won't this need to be merged first and then tagged? anyway i added a commit with a new version tag

neheb:
it gets tagged after merge.

subprojects/packagefiles/icu/source/i18n/meson.build
```diff
@@ -1,255 +1,5 @@
...
-  #  'decimalformatpattern.cpp',
-  #  'ucurr.cpp',
-  #  'digitlst.cpp',
-  #  'uspoof_wsconf.cpp',
-  #  'decfmtst.cpp',
-  #  'locdspnm.cpp',
-  #  'identifier_info.cpp',
-  #  'filteredbrk.cpp',
-)
+fs = import('fs')
```
neheb, line 1: the fs variable leaks from the main meson.build file. no need to specify it again.

srl295:
@eli-schwartz  attributed you on the sources.txt commit^

srl295:
this quells almost all of the warnings. There are some about sprintf (which was changed to snprintf or similar in ICU 73)

eli-schwartz:
Nice!

Were there any other changes you wanted to make right now or do you think this is ready to merge now and come back to any other improvements another day?

srl295:
@eli-schwartz I need to do the actual integration which won't be able to depend on this yet due to meson version 😆 

I'll probably need to come back to the switches and stuff later, so maybe let's let this merge and i'll do that another time?

Also I did send out a query to ICU on upstreaming everything.  if the build files were upstreamed, we've already seen good PRs from this project to upstream patches, so I'd assume we could still get help keeping it up to date upstream? (I.e. if the icu/meson.build files were upstreamed, (and i'm not promising that) would they be likely to be maintained there)?

eli-schwartz:
Sure, I am generally happy to be pinged on any issues that a project has with handling tasks in meson. In fact, there is a decent chance I'd spontaneously show up at projects to offer advice, and sometimes to diagnose something as a meson bug and fix it...

I can officially offer to be a code owner and be looped in for reviews too, if that helps.

PR #952, opened 2023-04-03
liblzma: allow threads to be disabled

Gottox:
This is useful for systems that change their behavior based on the presence of threads, notably emscripten.


jpakkane:
You need to add a new entry to `releases.json`, specifically `5.2.11-2`.

Gottox:
fixed.

eli-schwartz:
create_release failed with: "Failed to resolve action download info. Error: Internal Server Error"

Was wondering why no automatically triggered release...

neheb:
@Gottox you think an emscripten CI should be added?

Gottox:
Well, why not? 😄

PR #953, opened 2023-04-04
github CI: test clang-cl

neheb:
Various wraps here handle clang-cl. Test for it.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -54,6 +54,29 @@ jobs:
         run: |
           python tools/sanity_checks.py
 
+  VisualStudio-clang-cl:
+    if: github.event_name != 'schedule' || github.repository == 'mesonbuild/wrapdb'
+    runs-on: windows-latest
+    env:
+      CC: clang-cl
+      CXX: clang-cl
+      TEST_BUILD_ALL: 1
+      TEST_FATAL_WARNINGS: ${{ github.event.inputs.fatal_warnings }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Fetch tags and unshallow
+        run: git fetch --unshallow --tags
+
+      - name: Install packages
+        run: |
+          python -m pip install git+https://github.com/mesonbuild/meson
+
+      - uses: ilammy/msvc-dev-cmd@v1
+      - name: Sanity Checks
+        run: |
+          python tools/sanity_checks.py
+
   MacOS:
     if: github.event_name != 'schedule' || github.repository == 'mesonbuild/wrapdb'
     runs-on: macos-latest
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -27,7 +27,6 @@ jobs:
 
   VisualStudio:
     runs-on: windows-latest
-
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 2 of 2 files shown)


neheb:
hmmm from 4 to 8...

neheb:
```
2023-04-04T07:04:13.2875808Z 8 skipped: arduinocore-avr, google-woff2, libgpiod, ludocode-mpack, m4, mocklibc, termbox, tinyply
2023-04-04T07:04:13.2876156Z 4 failed: fluidsynth, harfbuzz, icu, range-v3
2023-04-04T07:04:13.2876345Z 2 errored: slirp, wayland
```

not too bad.

subprojects/packagefiles/fluidsynth/meson.build
```diff
@@ -158,6 +158,9 @@ endif
 
 if openmp_dep.found()
   conf.set('HAVE_OPENMP', 1)
+  if cc.get_id() == 'clang-cl'
```
neheb, line 161: @eli-schwartz this looks like it should be fixed in meson proper.

neheb:
```
2023-04-06T03:59:05.1866943Z 26 failed: centurion, cjson, flac, fluidsynth, glib, harfbuzz, imgui-sfml, json-c, json-glib, libffi, libmicrohttpd, libssh2, libuv, libwebsockets, lzo2, mpdecimal, range-v3, rdkafka, sdl2, sdl2_image, sdl2_mixer, sdl2_net, sdl2_ttf, sfml, spng, theora
```

A lot more actually...

neheb:
Let's just merge so packages can be fixed as added.

neheb:
ping @jpakkane

.github/workflows/build_all.yml
```diff
@@ -54,6 +54,29 @@ jobs:
         run: |
           python tools/sanity_checks.py
 
+  VisualStudio-clang-cl:
+    if: github.event_name != 'schedule' || github.repository == 'mesonbuild/wrapdb'
+    runs-on: windows-latest
+    env:
+      CC: clang-cl
+      CXX: clang-cl
```
eli-schwartz, line 62: Probably we can do this DRY with a matrix, since the only difference compared to the VisualStudio job is CC/CXX, right?

.github/workflows/sanity_checks.yml
```diff
@@ -40,6 +56,9 @@ jobs:
 
       - uses: ilammy/msvc-dev-cmd@v1
       - name: Sanity Checks
+        env:
+          CC: clang-cl
+          CXX: clang-cl
```
eli-schwartz, line 61: Same here?

eli-schwartz (commented):
Non-blocking review, in general testing with clang-cl seems reasonable.

.github/workflows/build_all.yml
neheb: that's the only difference, yes.

Not sure what you mean with DRY.

.github/workflows/sanity_checks.yml
neheb: IIRC using a matrix messes up the Jobs view as the jobs get grouped together. See here for an example: https://github.com/mesonbuild/wrapdb/actions/runs/4603869421

.github/workflows/build_all.yml
eli-schwartz: DRY -- "Don't Repeat Yourself"

.github/workflows/sanity_checks.yml
eli-schwartz: Oh, I didn't think of that. I know the side bar view shows them all in a column, and that's the part I usually look at.

neheb: If that view is acceptable I'll revert back.

eli-schwartz: Honestly I don't hugely care. Which one do you prefer?

neheb: current version. I usually click the middle of the screen to get to a specific run. one less click than with a matrix.

eli-schwartz: Cool, then let's keep it.

PR #954, opened 2023-04-04
icu: fix compilation with clang-cl

neheb:
Also disabled a bunch of warnings on MSVC.

ping @srl295


srl295 (commented):
I saw this after it was merged, and LGTM.

PR #972, opened 2023-04-07
pugixml: fix compilation with additional options

lanedis:
If you try to use a project what uses pugixml with additional options it will fail to compile.
Because 'args' - have to be the list rather then string.

Steps to reproduce:
test_pugi.cpp:
```c++
int main(int argc, char **argv) { return 0; }
```
meson.build:
```meson
project('test_pugi', 'cpp')
pugixml_proj = subproject('pugixml', default_options : ['no_xpath=true'])
exe = executable('test_pugi', 'test_pugi.cpp', dependencies : dependency('pugixml'))
```
Add pugixml:
`meson wrap install pugixml`

Possible error on compilation:
```
FAILED: subprojects/pugixml-1.13/libpugixml.so.1.13.p/src_pugixml.cpp.o 
ccache c++ -Isubprojects/pugixml-1.13/libpugixml.so.1.13.p -Isubprojects/pugixml-1.13 -I../subprojects/pugixml-1.13 -fvisibility=hidden -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c++14 -O0 -g -fPIC '-DPUGIXML_API=__attribute__((visibility("default")))-DPUGIXML_NO_XPATH' -MD -MQ subprojects/pugixml-1.13/libpugixml.so.1.13.p/src_pugixml.cpp.o -MF subprojects/pugixml-1.13/libpugixml.so.1.13.p/src_pugixml.cpp.o.d -o subprojects/pugixml-1.13/libpugixml.so.1.13.p/src_pugixml.cpp.o -c ../subprojects/pugixml-1.13/src/pugixml.cpp
```


jpakkane:

PR #973, opened 2023-04-07
Add bshoshany-thread-pool

v1993:
Add https://github.com/bshoshany/thread-pool - a header-only C++ thread pool implementation. I've [PRed](https://github.com/bshoshany/thread-pool/pull/106) `meson.build` upstream, so hopefully next release will include it, allowing us to only keep wrap around. Includes built-in tests.

As for wrap name, it was taken from conan and vcpkg versions of this library.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -95,6 +95,14 @@
       "2.3.1-2"
     ]
   },
+  "bshoshany-thread-pool": {
+    "dependency_names": [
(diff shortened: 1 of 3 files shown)


v1993:
Any idea why MacOS builder is failing? It seems to be completely unrelated to this PR.

PR #974, opened 2023-04-08
Add glbinding

v1993:
Add https://github.com/cginternals/glbinding - a C++ OpenGL binding/loader with enhanced type safety. Includes glbinding-aux library, which is not built by default.

Upstream relies on CMake-generated headers that are far too complex to reproduce nicely with meson, so I've bundled them and tweaked where needed.

I'm not quite sure about tests - it's an OpenGL library, so I doubt CI will manage to run them.

### DIFF
--- a/subprojects/packagefiles/glbinding/meson.build
+++ b/subprojects/packagefiles/glbinding/meson.build
@@ -1,0 +1,176 @@
+project('glbinding', 'cpp',
+  version : '3.3.0',
+  license : 'MIT',
+  meson_version: '>=0.62.0',
+  default_options: ['cpp_std=c++11', 'warning_level=0'],
+)
+
+cpp = meson.get_compiler('cpp')
+
+if get_option('default_library') == 'both' and cc.get_argument_syntax() == 'msvc'
+  error('default_library=both not supported when compiling with MSVC')
+endif
+
+system_name_arg = '-DSYSTEM_@0@'.format(host_machine.system().to_upper())
+
+glbinding_args = [system_name_arg]
+glbinding_aux_args = [system_name_arg]
+glbinding_public_args = []
+glbinding_aux_public_args = []
+
+if host_machine.system() != 'windows'
+  dl_dep = dependency('dl')
+else
+  dl_dep = declare_dependency()
+endif
+
+if get_option('default_library') == 'shared' or get_option('default_library') == 'both'
+  glbinding_args += ['-Dglbinding_EXPORTS']
+  glbinding_aux_args += ['-Dglbinding_aux_EXPORTS']
+else
+  glbinding_args += '-DGLBINDING_STATIC_DEFINE'
+  glbinding_public_args += '-DGLBINDING_STATIC_DEFINE'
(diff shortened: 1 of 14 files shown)


v1993:
Checks keep failing due to unauthorized files, even though I've added them to script as exceptions. I guess CI runs script from master branch instead of PR?

Also, MSVC emits a truly ridiculous amount of warning C4251, but it doesn't seem to be out fault.

neheb:
The entries in sanity_checks.py need to have the subdir.

subprojects/packagefiles/glbinding/meson.build
```diff
@@ -0,0 +1,185 @@
...
+
+cpp = meson.get_compiler('cpp')
+
+if get_option('default_library') == 'both' and cc.get_id() == 'msvc'
+  error('default_library=both not supported when compiling with MSVC')
+endif
+
+is_shared = get_option('default_library') == 'shared'
+
+if cpp.has_argument('-fvisibility=hidden')
```
neheb, line 14: just do

```
visibility_hidden_args = cpp.get_supported_arguments('-fvisibility=hidden')
```

same thing.

subprojects/packagefiles/glbinding/meson.build
```diff
@@ -0,0 +1,185 @@
...
+glbinding_includes = ['source/glbinding/include', 'bundled_headers']
+
+use_system_khrplatform = cpp.has_header('KHR/khrplatform.h')
+summary('Use system KHR/khrplatform.h', use_system_khrplatform)
+
+if not use_system_khrplatform
+  glbinding_includes += 'source/3rdparty/KHR/include'
+endif
+
+glbinding_sources = [
```
neheb, line 52: use files() instead of []

subprojects/packagefiles/glbinding/meson.build
neheb: wait this is wrong. Just use gnu_symbol_visibility: 'hidden

v1993:
Fixed issues pointed out by review.

v1993:
TODO for self (working on a different wrap rn): use `protocol: gtest` for test.

subprojects/packagefiles/glbinding/pch/pch.h
```diff
@@ -0,0 +1 @@
+#include <glbinding/Binding.h>
```
v1993, line 1: About this file...

Upstream has [its own precompiled header](https://github.com/cginternals/glbinding/blob/master/source/glbinding/source/Binding_pch.h) - which is why I've decided to make use of this feature in the first place. However, that header is included by other files (which meson docs warns not to do, although it doesn't seem to break in practice) and has a `#pragma once` guard, which produces a compiler warning. 

As for build time - there's a clear benefit, reducing it from ~47 seconds to ~36 seconds on my machine.

v1993:
```
cc1plus.exe: out of memory allocating 69631 bytes
```
So, what now?

neheb:
I’d squash some of these commits. That’s a hilarious error. Wonder if it can be worked around.

neheb:
That is indeed a monstrosity. Cool that it works now.

PR #975, opened 2023-04-09
sdl2: some backports from the Taisei wrap

Akaricchi:
The find-dylib-name change is needed on some Linux systems that have libraries installed in places the compiler is not aware of. For example, on my system (openSUSE Tumbleweed), libjack.so is provided via a pipewire compatibility layer in `/usr/lib64/pipewire-0.3/jack/`, which doesn't appear in `cc -print-search-dirs`. For anything more complicated, it might be a good idea to allow specifying these library names statically in a machine file.

The WGI joystick toggle is added because SDL's WGI code is very brittle and prone to breakage. It seems to rely on some very specific name mangling of an API that was never meant to be called directly from C. It doesn't compile on at least some versions of the [llvm-mingw](https://github.com/mstorsjo/llvm-mingw) toolchain. Losing WGI is not a big deal; the plethora of other joystick drivers available on windows is more than enough.


neheb:
```
100755 → 100644 
```

hmmm. IIRC find_program will not succeed if mode has no x.

Akaricchi:
Ugh, I definitely haven't changed the mode intentionally. I only used `meson subprojects packagefiles` to bounce the files back and forth between the subproject directory and packagefiles; maybe there's a bug there?

In any case, I'll fix the mode, but the accidental change didn't seem to have caused any problems on my end.

neheb:
need a releases.json bump as well.

Akaricchi:
Ah sorry, not used to how things are done here. Should I increment release in every commit, or is this going to get squashed and only needs one increment for the whole PR?

neheb:
just single for whole PR.

Akaricchi:
Not sure if i'm doing this right.

neheb:
```
Run ./tools/sanity_checks.py
s..
Ran 3 tests in 0.194s

OK (skipped=1)
```


hmmmmmmm

Akaricchi:
No idea what to do with this now:

```
[556/556] Linking target subprojects/SDL2-2.26.5/test/testoffscreen
ninja: no work to do.
No tests defined.
ninja: Entering directory `/data/git/wrapdb/_build'
ninja: no work to do.
Installing subprojects/SDL2-2.26.5/libsdl2main.a to /data/git/wrapdb/_build/pkg/usr/local/lib64
Installing subprojects/SDL2-2.26.5/libsdl2.so to /data/git/wrapdb/_build/pkg/usr/local/lib64
.
FAIL: test_releases (__main__.TestReleases) (step='patch_directory', name='sdl2')
  File "/data/git/wrapdb/./tools/sanity_checks.py", line 203, in test_releases
    self.check_files(name, patch_path)
  File "/data/git/wrapdb/./tools/sanity_checks.py", line 403, in check_files
    self.fail('Not permitted files found: ' + not_permitted_str)
AssertionError: Not permitted files found: subprojects/packagefiles/sdl2/.meson-subproject-wrap-hash.txt

Ran 3 tests in 19.420s

FAILED (failures=1, skipped=1)
```

Akaricchi:
Actually just removing that file works, but I don't know how it even got there.

Issue #961, opened 2023-04-05
sdl2_image wrap needs to expose configuration options


v1993:
(Opening this partially to keep myself reminded about this, I'm hoping to get to this issue somewhat soon.)

Currently, sdl2_image wrap is pretty bare-bones and could use at least some improvements:

* It builds all provided source files, but only uses features from a few of them. Others end up completely `#ifdef`ed away, including the example xxx file - which is meant as a template for adding new formats, not for building.
* There's no way to require TIFF support nicely - it simply uses `required: false` for its library finder.
* There's no way to request support for formats that aren't PNG/JPG/TIFF but still supported by SDL2 image library.
* There's no way to build without libpng and libjpg, which is especially frustrating since SDL2 image itself allows using stb_image loader shipped as part of its source, thus functioning with core SDL2 library as the only dependency.

The proper solution would be to add feature options per supported loader (and remove file `IMG_xxx.c` from source list entirely), passing each feature option to corresponding `dependency` call, then add defines/files to library accordingly.

PR #976, opened 2023-04-09
sdl2_image: update to 2.6.3

neheb:


v1993:
Partially fixes #961 as is, thanks for this!

I suggest adding option `use_stbimage`. When specified, it defines `USE_STBIMAGE` macro, adds `IMG_stb.c` and drops dependencies on libpng and libjpeg for png/jpeg loaders, so those are enabled unless explicitly disabled.

`IMG_xxx.c` is a *template files to help with adding new loaders*. There's no reason to add it as source and upstream CMake indeed does not do that.

Loaders are disabled unless explicitly enabled with a define (`IMG_stb.c` is an exception due to its nature as a helper rather than loader - and even then it's `ifedef`ed beyond `USE_STBIMAGE`). This means that there's no point in adding source files unless you also enable respective `LOAD_*` macros. As a result, all unconditionally built files, sans BMP (for which `-DLOAD_BMP` is defined), do exactly nothing. IMO it would make sense to expose those as feature options as well, with `auto` value behavior mirroring defaults of upstream CMake builds.

v1993:
Also: upstream CMake defaults avif, jxl, tiff and webp support to off. IMO it would make sense to default their values to `disabled` in this wrap as well.

Notably, upstream builds also *default* to using dependency-free STB loader for PNG/JPEG, so by default library does not depend on anything external (except SDL2, obviously).

v1993:
Thinking about it a little more - it probably will be the best if this gets merged in current state, then I'll build upon improvements from this PR and make a separate one.

neheb:
fair enough.

PR #980, opened 2023-04-10
sdl2: provide sdl2main and sdl2_test dependencies, don't link to sdl2main by default

v1993:
This better reflects how sdl2main should be used and actually works when a library (e.g. sdl2_image) links to sdl2 in a project that builds with `with_main=true`.

Mark `with_main` deprecated and default it to false - this is a sensible default now that linking to sdl2main the right way is simple.

sdl2_test is used for testing by some other projects, e.g. sdl2_image - which motivated me to do this change in the first place.

Additionally, bump required meson version to 0.60 and replace deprecated functions with alternatives.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -2079,9 +2079,12 @@
   },
   "sdl2": {
     "dependency_names": [
-      "sdl2"
+      "sdl2",
+      "sdl2main",
+      "sdl2_test"
     ],
     "versions": [
+      "2.26.5-4",
       "2.26.5-3",
       "2.26.5-2",
       "2.26.5-1",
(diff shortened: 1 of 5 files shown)


releases.json
```diff
@@ -2082,6 +2082,7 @@
       "sdl2"
```
neheb, line 2082: needs updating here.

v1993:
> this breaks sdl2_mixer btw.

It's probably `playmus` executable needing to link to `sdl2main`. To quote sdl2_image:

```meson
if sdl2_dep.type_name() == 'internal'
  sdl2main_dep = dependency('sdl2main')
else
  sdl2main_dep = cc.find_library('SDL2main')
endif
```

and then include it as execuatable's dependency.

P.S.: I guess that's another SDL library to work on now that sdl2_image is merged.

PR #982, opened 2023-04-10
sdl2_image: major overhaul

### LINKED ISSUE #961: sdl2_image wrap needs to expose configuration options

v1993:
Closes #961. Depends on #980 ~~- currently includes it, I'll rebase on master and convert from draft once it's merged~~.

* Add support for STB, Image I/O and WIC backends - default build now only needs SDL2
* Add support and options for all formats
* Add support for *disabling* saving PNG/JPEG
* Add summary for enabled features
* Add test suite

The only missing feature compared to official build system is support for loading libraries plugin-style instead of linking to them either statically or dynamically. However, I wouldn't expect this to be an issue in most uses of wrap.


subprojects/packagefiles/sdl2_image/meson.build
```diff
@@ -2,12 +2,87 @@ project(
...
+  meson_version: '>=0.59.0',
 )
 
-feature_args = ['-DLOAD_BMP']
+feature_args = []
 
-options = {
+if host_machine.system() == 'windows' and get_option('default_library') != 'static'
+    feature_args += '-DDLL_EXPORT'
+endif
```
v1993, line 12: From the looks of it - dynamic linking on windows probably never worked in the first place. This part likely needs to be ported to sdl2_mixer and sdl2_ttf wraps.

v1993:
Rebased on master, ready for review.

neheb:
```
warning C4828: The file contains a character starting at offset 0x5d9 that is illegal in the current source character set (codepage 65001).
```

funny.

subprojects/packagefiles/sdl2_image/meson.build
```diff
@@ -2,12 +2,88 @@ project(
...
+)
+
+sdl2_dep = dependency('sdl2')
+
+deps = [sdl2_dep]
+
+# Backends provide support for certain formats, allowing not to depend on external library or use potentially better optimized algorithm
+backend_provided_loaders = []
+
+use_stbimage_backend = get_option('use_stbimage').allowed()
```
neheb: why not just use a bool here?

subprojects/packagefiles/sdl2_image/meson.build
```diff
@@ -2,12 +2,88 @@ project(
...
+use_imageio_backend = get_option('use_imageio').require(host_machine.system() == 'darwin', error_message: 'Image I/O backend is exclusive to MacOS').allowed()
+if use_imageio_backend
+  add_languages('objc', native: false)
+  sources += files('IMG_ImageIO.m')
+  deps += dependency('appleframeworks', modules : ['ApplicationServices'])
+
+  backend_provided_loaders += ['png', 'jpg', 'tif', 'bmp', 'gif', 'tga']
+endif
+
+if not use_imageio_backend and host_machine.system() == 'darwin'
```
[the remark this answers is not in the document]
v1993: I've considered this before and think the intent is more clear the current way due to explicit `and` instead of `if` nesting.

subprojects/packagefiles/sdl2_image/meson.build
v1993: 1. IMO having all backends as features is better than having them as mix of booleans and features.
2. `auto` gained different behavior in later commit.

PR #984, opened 2023-04-10
CI: Add MSYS 

neheb:


neheb:
This PR adds entries to https://github.com/mesonbuild/wrapdb/actions/caches which is...interesting.

neheb:
```
../subprojects/SDL2-2.26.5/src/test/SDL_test_harness.c:354:1: internal compiler error: in assign_stack_local_1, at function.cc:429
```


wow...

neheb:
hmm can't figure this out.

neheb:
good enough.

@eli-schwartz this is ready. should I add build_all support?

eli-schwartz:
I think build_all support makes sense.

Quick question: this looks like it adds 5 variants of msys2, correct? You'd probably know better than me, but are all those variants necessary in order to robustly test functionality on msys2?

Or to put it another way, what are the odds that something which succeeds on one is going to fail on the others? In particular, do we need to test both clang and gcc on both 32/64 bit? Maybe we could check e.g. mingw32 and clang64 and ucrt64 to save on two jobs?

neheb:
yes.

the backends are named after compilers/libc but it's more than that.

```
MinGW = GCC/MSVCRT/LIBSTDC++
UCRT = GCC/UCRT/LIBSTDC++
CLANG = CLANG/UCRT/LIBC++
```

If I had to get rid of some, it would be UCRT64 and CLANG32. The former is somewhat pointless and the latter somewhat experimental. Note the compiler error I posted above.

neheb:
Removed both MINGW platforms.

MSVCRT is for versions prior to Windows 10, which are all EOL. Microsoft recommends using UCRT now.

neheb:
@eli-schwartz set to ready

jpakkane:
Fine by me at least.

eli-schwartz:
Fine by me as long as you think all the matrix jobs are helpful to catch bugs in meson.build definitions.

The only reason I commented about it is because we do sometimes run into bottlenecks on CI parallelism which slows down the feedback cycle. If adding more jobs helps us uncover issues we want to know about, then that's great. But if it turns out something was not needed and we didn't really care, then paying the cost is a bit annoying in terms of global wait time.

...

tl;dr everything looks fine, go ahead and merge it.

xclaesse:
@neheb In https://github.com/mesonbuild/wrapdb/pull/1109 I'm facing issues with CLANG32 (works on CLANG64). How can I reproduce a similar env on my Windows machine to reproduce?

neheb:
clang32 is available with MSYS2.

xclaesse:
I noticed I had a very old installation. Reinstalled from scratch and I get clang64 and ucrt64 just fine, but not clang32. At least I don't see the terminal launcher.

xclaesse:
oh, ok, I see /msys64/clang32.exe, clicking on it opens the terminal. But it seems it's not creating a shortcut for it by default.

neheb:
It is for me. But then again I install through scoop.

neheb:
```
2023-08-25T22:45:24.7808164Z 19 failed: backward-cpp, cairo, cppzmq, harfbuzz, icu, libccp4c, libffmpegthumbnailer, liblastfm, libobsd, libuv, mpdecimal, nng, openh264, openssl, orcus, taglib, theora, unit-system, vorbis
2023-08-25T22:45:24.7808904Z 2 errored: libffi, wayland
```

a lot are not easy to fix unfortunately.

PR #985, opened 2023-04-11
Add physfs

v1993:
Add [physfs](https://github.com/icculus/physfs) - a portable filesystem/archive access wrapper often used by games. Does not depend on anything except threads (and a bit of platform-specific stuff which is not really notable).

I've decided to disable most archivers by default, since they are pretty unlikely to be used by most applications these days, only leaving zip and 7z. However, users are free to re-enable those archivers by explicitly enabling options.

I've included a very simple test to verify that it links and runs, since `test.c` from upstream is interactive and not very portable.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1899,6 +1899,14 @@
       "10.23-1"
     ]
   },
+  "physfs": {
+    "dependency_names": [
+      "physfs"
+    ],
+    "versions": [
+      "3.2.0-1"
+    ]
+  },
   "pkgconf": {
     "dependency_names": [
       "libpkgconf"
--- a/subprojects/packagefiles/physfs/meson.build
+++ b/subprojects/packagefiles/physfs/meson.build
@@ -1,0 +1,80 @@
+project(
+  'physfs',
+  'c',
+  version: '3.2.0',
+  meson_version: '>=0.54.0',
+)
+
+feature_args = []
+extra_library_args = []
+
+if get_option('default_library') == 'static'
+    extra_library_args += '-DPHYSFS_STATIC'
+endif
+
+# Files that don't correspond to configurable archivers
+sources = files(
+  'src/physfs.c',
+  'src/physfs_byteorder.c',
+  'src/physfs_unicode.c',
+  'src/physfs_platform_posix.c',
+  'src/physfs_platform_unix.c',
+  'src/physfs_platform_windows.c',
+  'src/physfs_platform_os2.c',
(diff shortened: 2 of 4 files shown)


tools/sanity_checks.py
```diff
@@ -77,6 +77,9 @@
     'openal-soft': [
         'hexify.py'
     ],
+    'physfs': [
+        'simple_test.c'
+    ],
```
eli-schwartz, line 82: In general this is considered a bad idea, and the reason this override table exists is really for the sake of buildsystem helper scripts and not for source code.

Please try submitting this upstream, IMHO it would make sense for them to be willing to accept the contribution of an automated sanity-check test program.

eli-schwartz (changes requested):

tools/sanity_checks.py
v1993: I've removed this. I'm not sure if it'll make it upstream any time soon, given that 3.2 branch is a bugfix-only and physfs 4 is still in the planning phase.

eli-schwartz: IDK, I'd call this a bugfix.

It's a workflow fix for the issue where you tried to build the project and something mysterious went wrong but wasn't detected during the build.

But also as far as I can tell there's no 4.0 branch at all so I don't see why upstream would have an issue making this change at the bleeding edge of whatever development there *is*.

Maybe try it just to find out if it would be accepted? :)

eli-schwartz: The active development in question is exactly one commit from November that is also cherry-picked to the 3.2 maintenance branch.

I'm familiar with repositories where actual, meaningful differences between branches exist. Even there, there's always a branch where changes *can* be applied... but in this case, it seems like it would actually be more work to have commits be exclusive to a particular branch.

eli-schwartz: (Meson itself is an example of where one branch exists for active development while another branch exists to backport fixes.)

v1993: Sure, like I said, I'll try to PR tests upstream when I have time. Until that (and a new release containing tests) happens, would it be possible to get this merged as-is, without tests?

neheb:
Please rebase

eli-schwartz:
There are commits in this PR that cancel each other out. Maybe squash them into one commit?

v1993:
> There are commits in this PR that cancel each other out. Maybe squash them into one commit?

Done.

PR #990, opened 2023-04-12
sdl2_mixer: link playmus to sdl2main

v1993:
#980 broke this.


neheb:
Probably should update this package mixed with fixes.

PR #991, opened 2023-04-12
libxml2: update to 2.10.4

bgilbert:


bgilbert:
```console
libxml2| WARNING: Project targets '>=0.54.0' but uses feature introduced in '0.54.1': cmakedefine without exactly two tokens.
```

Hmm.  `config.h.cmake.in` hasn't changed at all since the last release.  I could just bump the minimum version to 0.54.1, but I don't understand where this error is coming from.  All of the `#cmakedefine`s in `config.h.cmake.in` have two tokens.  I tried explicitly setting all the substitution variables in configuration data, in case an undefined variable was the cause, but it didn't help, and the log doesn't specify the bad define either.

bgilbert:
Okay, bumped the minimum to 0.54.1.

neheb:
I'd look at the clang-cl output and fix as many warnings as you can. Grep the tree for _CRT_SECURE_NO_WARNINGS

bgilbert:
Those warnings would have been there before, right?  My goal at the moment is to get the [CVEs](https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.10.4) fixed, rather than general cleanup.

PR #995, opened 2023-04-21
physfs: enable GNU C extenstions

v1993:
Whoops, missed this the first time around. Prevents warnings for undeclared readlink and lstat on POSIX platforms.


eli-schwartz:
It should also work in the per-subproject default_options kwarg to project()

v1993:
@eli-schwartz Is that the preferred place to set it?

neheb:
Yes. Another way is to define _GNU_SOURCE. Probably.

neheb:
Thinking about it some more, _GNU_SOURCE is better. meson gained subproject standard setting in 0.61 or 0.62. Using physfs as a subproject with earlier meson will not work properly.

@eli-schwartz worth it to add a warning for this?

v1993:
@neheb Fine, I've switched to defining `_XOPEN_SOURCE=500` instead, since that's what [lstat](https://man7.org/linux/man-pages/man2/lstat.2.html) and [readlink](https://man7.org/linux/man-pages/man2/readlink.2.html) are guarded by.

v1993:
After trying to fix missing alloca on MacOS, which seemingly requires `__GNUC__` over there, it seems to me that the best option is just to stick to original idea with `c_std=gnu89` (maybe `gnu99`). Simple, portable, allows meson and compiler to handle features.

v1993:
Switched back to default_options with `c_std=gnu99`. Bumped minimal meson version to 0.62 to ensure that it is applied.

neheb:
Should do both honestly. meson 0.62 is quite recent.

eli-schwartz:
You don't need to bump the minimum version for this... default_options works quite fine for a project on its own even on very old versions. For subproject usage specifically it either needs a new-ish version of meson OR setting the global c_std, but doing the latter is quite feasible.

It's completely silly to establish a cliff for wraps where everything arbitrarily declares meson >=0.62 despite it not actually being a requirement.

v1993:
Unbumped meson version. Anything else to do?

neheb:
@eli-schwartz with some wraps here, I needed to add _GNU_SOURCE as lighttpd uses c99 for the project version.

v1993:
Like I've explained - defining `_GNU_SOURCE` solves nothing here, as relevant functions are guarded by

neheb:
Sounds broken on Apple's part.

PR #998, opened 2023-04-23
CI: don't test clang32

neheb:
For some reason it keeps failing on generating symbols. The other backends seem to be fine.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -86,7 +86,7 @@ jobs:
       TEST_FATAL_WARNINGS: ${{ github.event.inputs.fatal_warnings }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


eli-schwartz:
I don't think it has anything to do with clang32, this is a known random flake that happens in the meson CI too. That uses mingw32 and mingw64.

https://github.com/mesonbuild/meson/blob/01420bf8fc1ee6c1466e2fa4c9805267192dce26/.github/workflows/msys2.yml#L54-L56

eli-schwartz:
There's a dirty hack to work around it, but restarting a failing job works as well.

neheb:
The real problem is build_all

eli-schwartz:
Nonetheless, setting winjitdebug so that msys2 detects a debugging environment and refuses to crash, seems safer.

neheb:
I'll set to draft for now.

PR #1001, opened 2023-04-23
lzo2: use find_program's found()

neheb:
Using build machine like this seems wrong. find_program can check if .sh can be executed.


v1993:
How about doing `find_program('sh', required: false)` and passing script as an argument to it?

neheb:
not sure what you mean.

this script looks simple enough to reimplement in meson.

subprojects/packagefiles/lzo/lzotest/meson.build
```diff
@@ -2,8 +2,9 @@ t = executable('lzotest', 'lzotest.c', dependencies: lzo2_dep)
 
 test('basic lzotest', t, args: ['-mlzo', '-n2', '-q', files('../COPYING')], suite: 'basic')
 test('exhaustive lzotest', t, args: ['-mavail', '-n10', '-q', files('../COPYING')])
-if build_machine.system() != 'windows'
-    test('lzotest directory check', find_program('../util/check.sh'),
+check = find_program('../util/check.sh', required: false)
+if check.found()
+    test('lzotest directory check', check,
       args: [topdir],
```
v1993, line 8:
```suggestion
shell = find_program('sh', required: false)
if shell.found()
    test('lzotest directory check', shell,
      args: [topdir / 'util/check.sh', topdir],
```

This is what I meant.

eli-schwartz:
> Using build machine like this seems wrong. find_program can check if .sh can be executed.

See commit bab7f28f1da21f9101db5f20aa01e92d09982382 -- this is checking build_machine because the script as implemented by upstream requires not only sh, but also the directory location /tmp for writing a tempfile.

build_machine is thus an appropriate check.

> this script looks simple enough to reimplement in meson.

This too is a possibility, but if the script is going to be used directly then I'm not sure there's much to change here...

subprojects/packagefiles/lzo/lzotest/meson.build
eli-schwartz: It's not clear to me why that is especially advantageous, since on unix the sh shell always exists, and on Windows it probably doesn't but Meson will anyways manually parse the shebang, extract /bin/sh, and search for /bin/sh as the "real" program.

v1993: Interesting, why is meson trying to run this test on windows for now then?

eli-schwartz: Because you can install the "sh" program on Windows, and the search algorithm for /bin/sh -- a POSIX path -- on Windows is to remove the POSIX path and search for "sh".

neheb:
Hmm I thought MSYS2 also has /tmp

subprojects/packagefiles/lzo/lzotest/meson.build
v1993: But actually trying to run that file fails with code 127, which is AFAIK exactly because `/bin/sh` is not found by OS. I guess implementing change I've suggested would fix this, but it also raises a question if path stripping should be done in this case - meson reports program as found yet it can't be executed in practice, which is undesirable.

neheb:
Don't feel like fixing this.

PR #1009, opened 2023-05-07
Add cimg

v1993:
Add [cimg](https://cimg.eu/) - an image processing library for C++. While library itself is header-only, it has a rather wide range of configuration options, including support for other (non-header-only) libraries. As such, it makes for a rather sizable amount of meson options.

It does not include tests, so I've opted to enable building examples it ships with in CI.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -41,6 +41,9 @@
   "cimg": {
     "build_options": [
       "cimg:examples=true"
+    ],
+    "brew_packages": [
+      "libomp"
     ]
   },
   "cli11": {
--- a/ci_config.json
+++ b/ci_config.json
@@ -38,6 +38,11 @@
       "cexception:werror=false"
     ]
   },
+  "cimg": {
+    "build_options": [
+      "cimg:examples=true"
+    ]
+  },
   "cli11": {
     "build_options": [
       "cpp_std=c++14"
--- a/releases.json
+++ b/releases.json
@@ -192,6 +192,14 @@
       "6.2.2-1"
     ]
   },
+  "cimg": {
+    "dependency_names": [
+      "cimg"
+    ],
+    "versions": [
+      "3.2.3-1"
+    ]
+  },
   "cjson": {
     "dependency_names": [
       "libcjson",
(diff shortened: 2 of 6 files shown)


v1993:
Guess I just should stop defining `cimg_use_openmp` altogether... Code from library, indented and commented for clarity:

```cpp
#if !defined(cimg_use_openmp)
  #ifdef _OPENMP
    #define cimg_use_openmp 1
  #else
    #define cimg_use_openmp 0
  #endif
#else
  // WTF???
  #undef cimg_use_openmp
  #define cimg_use_openmp 1
#endif

#if cimg_use_openmp!=0
  #include <omp.h>
  #define cimg_pragma_openmp(p) cimg_pragma(omp p)
#else
  #define cimg_pragma_openmp(p)
#endif
```

v1993:
Looks like apple's clang defines `_OPENMP` without providing actual OpenMP support? Just great. Guess I'll have to try to install libomp from brew and pray that works...

In the meantime, on Windows clang-cl decided to warn about deprecated functions despite file defining constants that should prevent that from happening. Good job, two different yet equally painful variations of clang...

v1993:
Alright, now both builds are failing because of OpenMP issues - MacOS still can't find OpenMP even when installed from homebrew and clang-cl fails at linking step due to unresolved references to functions from supporting OpenMP library (meson bug?). Kinda clueless what to do next.

jpakkane:
If no-one can come up with a solution, then you can tag it so CI won't try to build it. It's better than not having it available at all.

neheb:
to fix clang-cl you need something like this: https://github.com/mesonbuild/wrapdb/commit/528f003861aac7cfb0124f1766eb264abdc4305f

upstream meson bug: https://github.com/mesonbuild/meson/pull/11648/commits/2d60aa6b393177cb6e8fa3504c2568217dde2f64

neheb:
brew.sh 's libomp is Keg-only. meaning you need to add non default paths to use it.

v1993:
I'm no longer interested in this library (decided to go with a different solution), so closing this PR. It should be a fairly solid foundation in case someone decides to get this finished and merged.

PR #1010, opened 2023-05-08
libpng: Fix windows x86_64 build

talisein:
It seems the current github runner returns a __ARM_NEON define that is a single space.

When the current build is run with the following:
```
debug('raw: "', c_compiler.get_define('__ARM_NEON'), '"')
debug('underscorify: ', c_compiler.get_define('__ARM_NEON').underscorify())
```

In the meson-log I see (output trimmed)
```
Fetching value of define "__ARM_NEON" :   (cached)
Debug: raw: "   "
Debug: underscorify:  _
```


neheb:
need a releases.json bump.

PR #1011, opened 2023-05-08
liburing: update to 2.3

pobrn:
Changes included:
 - update list of tests
 - update list of examples
 - update list of man pages
 - actually use `no_glibc_statx` in compat.h
 - define `CONFIG_HAVE_NVME_URING` in config-host.h when applicable
 - remove `CONFIG_HAVE_MEMFD_CREATE` as it is no longer used


eli-schwartz:
You should mark the `build_on` mapping in ci_config.json to avoid building liburing on msys2 as well.

neheb:
No

PR #1014, opened 2023-05-10
liburing: drop "uring" dependency name

pobrn:
The upstream project ships a pkg-confing file with the "liburing" name, so use that.


eli-schwartz:
GitHub is in severely degraded mode today.

PR #1016, opened 2023-05-10
catch2: Update to 3.3.2

mikecrowe:


jpakkane:

mikecrowe:
> You need to add a new entry to `releases.json`.

Whoops. Thanks. Hopefully that's done now.

The:
`catch2| WARNING: Project targets '>=0.50.0' but uses feature introduced in '0.54.1': cmakedefine without exactly two tokens.`
sanity check failure confuses me since there were no `cmakedefine` changes between Catch2 v3.2.0 and v3.3.2.

neheb:
that's an upstream meson change to now warn about a bug fixed in 0.54.1

v1993:
Warning fixed upstream, next release should be good.

PR #1023, opened 2023-05-17
Add littlecms.

jpakkane:


eli-schwartz:
Duplicate of https://github.com/mesonbuild/wrapdb/blob/master/subprojects/lcms2.wrap so please just upgrade that in place.

Note: CI fails because:
```
meson.build:1:0: ERROR: Multiple wrap files provide 'lcms2' dependency: littlecms.wrap and lcms2.wrap
```

jpakkane:
Hmm, didn't see that because I searched for "littlecms" rather than "lcms2". :( Well, can't be helped now I suppose.

eli-schwartz (approved):

PR #1025, opened 2023-05-23
fixing indent for c-flags

noah1510:
This fixes the indentation for c-flags.
Before its version and dependency_names had an indent of 4 now they have one of 2 like the rest.

The only reason for this is to allow automatic updates of the releases file since every formatter requires consistent indentation (or at least they will produce them when writing).
In other words this pr is to prevent changes accidental changes to the releases file in the future.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -107,10 +107,10 @@
   },
   "c-flags": {
     "dependency_names": [
-        "c-flags"
+      "c-flags"
     ],
     "versions": [
-        "1.5.4-1"
+      "1.5.4-1"
     ]
   },
   "catch": {


eli-schwartz:
At a minimum, please squash both these changes into a single commit. One commit adds a bug, the next commit reverts it.

neheb:
clang-format can handle this.

noah1510:
Clang-format can produce the same formatting as this pr does.
However atm there is no .clang-format file in this repo and the format is no checked prior to each pr.
This just makes the format of the file consistent, to prevent json tools from producing unwanted deltas in the future.

neheb:
I’d add those in a separate PR. No idea how to run it on git commit.

PR #1026, opened 2023-05-24
Install fmt lib.

jpakkane:


xclaesse:
You probably also want to install headers then.

jpakkane:
Updated.

eli-schwartz:
It should also install a pkg-config file (and generate a fmt-uninstalled.pc for the resulting devenv).

subprojects/packagefiles/fmt/meson.build
```diff
@@ -29,3 +35,9 @@ fmt_header_only_dep = declare_dependency(
   include_directories: 'include',
   compile_args: '-DFMT_HEADER_ONLY',
 )
+
+pkg_mod = import('pkgconfig')
+pkg_mod.generate(fmt_lib,
+  name: 'fmt',
+  filebase: 'fmt',
```
eli-schwartz, line 42: Shouldn't the name and filebase be inferred in this case?

PR #1030, opened 2023-05-25
abseil-cpp: Sync dependency names list with distro pkgconfig names

nirbheek:
Only include the dependency names that actually link to a library, because the rest are auto-generated header-only deps that we can skip.

There are some 'aggregate' pkgconfig names as well that link to libs indirectly, but those are harder to maintain so let's skip for now.


nirbheek:
CC: @neheb

PR #1034, opened 2023-05-28
Add oatpp-sqlite

Totto16:
Add oatpp-sqlite
The repo can be found [here](https://github.com/oatpp/oatpp-sqlite)

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1890,6 +1890,14 @@
       "1.3.0-1"
     ]
   },
+  "oatpp-sqlite": {
+    "dependency_names": [
+      "oatpp-sqlite"
+    ],
+    "versions": [
+      "1.3.0-1"
+    ]
+  },
   "oatpp-swagger": {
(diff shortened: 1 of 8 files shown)


Totto16:
I found an error in the base module 'oatpp' (only on windows, when you use it as dependency) so this can only be merged after #1039 is merged

Totto16:
After some investigating, the same errors from 'oatpp' (See #1031) occurs on windows, they are remainigs from that module, but fixed already in upstream, but no new release has been made yet. So disabling the tests in windows, as long as the version is to be known to have bugs is the best solution, as also discussed in the PR above discussed.

Totto16:
This is now ready and squashed

Totto16:
@neheb LOL is this supposed to happen:
image

neheb:
nope. I'm guessing some race condition.

Totto16:
Interesting :joy:

PR #1038, opened 2023-05-28
Add oatpp-libressl

Totto16:
Add oatpp-libressl
The repo can be found [here](https://github.com/oatpp/oatpp-libressl)

To pass the tests libressl has to be installed, but since there's no package under e.g debian, either a wrap must be made for libressl, or some other solution has to be found.

Therefore this is a draft atm.


neheb:
conflicts

PR #1035, opened 2023-05-28
Add oatpp-swagger

Totto16:
Add oatpp-swagger
The repo can be found [here](https://github.com/oatpp/oatpp-swagger)

This depends on #1039 and can only be merged after that!

PR #1039, opened 2023-05-29
fix two options on 'oatpp' module:

Totto16:
this fixes 2 options, that the CI didn't catch, since it only affects projects, that have this as dependency and only windows msvc builds, and I didn't build those projects locally on windows to see this issue, i saw it in some oatpp-modules in the CI

It bumps the version of the module 'oatpp' to '1.3.0-2'

(some changes were already made by accident in the merge of 'oatpp-swagger', where I also edited the 'oatpp' module. :sweat:


jpakkane:
[You should never merge your branches to update them, but use rebase instead](https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date).

Totto16:
The reason for the merge is, that the commits in #1035 are wrong, they also modify the module 'oatpp' which they shouldn't. I couldn't modify these in time, after I found the  issue. So a rebase could corrupt it. but I'll do that :+1:  I knew of the rule, :sweat:  but decided, that since it's affected by something, that YOU SHOULD NEVER DO, modifying another module, without bumping it's number :sweat:

Totto16:
it is rebases and squashed now

PR #1041, opened 2023-05-29
add oatpp-zlib module

Totto16:
Add oatpp-zlib
The repo can be found [here](https://github.com/oatpp/oatpp-zlib)

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -2120,6 +2120,14 @@
       "1.3.0-1"
     ]
   },
+  "oatpp-zlib": {
+    "dependency_names": [
+      "oatpp-zlib"
+    ],
+    "versions": [
+      "1.3.0-1"
+    ]
+  },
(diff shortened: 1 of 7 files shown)


neheb:
rebase maybe?

Totto16:
Like I thought and looked up on the official Documentation, 32 Bit is not supported, it happens to work with all other `oatpp-*` modules I ported so far, but for this one it does not, it is complicated to setup a test environment for me, so I think it would be okay, to not support 32 bit here...

But how can that be done? Any suggestions?
Just fail in the meson.build if the e.g. ```build_machine.cpu_family() == 'x86'`` ?`

neheb:
So clang32 is the only 32 bit CI. Might fail on Linux 32-bit too.

Totto16:
❤️

PR #1042, opened 2023-05-30
abseil-cpp: Provide headers for projects to install

nirbheek:
Some projects that use abseil-cpp, such as webrtc-audio-processing, use abseil types in public headers. They need to install the headers, so provide them as variables that can be accessed via the subproject object.


jpakkane:
Why don't we make the subproject install them directly? Is it because they need to be installed in some special subproject-only include directory?

nirbheek:
> Why don't we make the subproject install them directly? Is it because they need to be installed in some special subproject-only include directory?

The subproject is all static libraries that are linked to directly by whatever project consumes it. The subproject does not install anything, so it is up to the project to decide whether it wants to install the headers.

jpakkane:
> The subproject does not install anything

The question here being, should it? That is what all other projects do (whether static or shared linked), but unfortunately Abseil is a Google project and they have a tendency to not follow distribution best practices.

nirbheek:
There's no point installing the libraries in this project, it's meant to be used as vendored sources. Honestly, the only reason why webrtc-audio-processing uses its types in its API headers is because that code is from libwebrtc, which is *also* a google project that doesn't care about modularization.

PR #1043, opened 2023-05-30
add 'stduuid' wrap

Totto16:
This adds the header only library 'stduuid' from [here](https://github.com/mariusbancila/stduuid)

### DIFF
--- a/subprojects/packagefiles/stduuid/test/meson.build
+++ b/subprojects/packagefiles/stduuid/test/meson.build
@@ -1,0 +1,32 @@
+test_src_files = files('main.cpp', 'test_generators.cpp', 'test_uuid.cpp')
+
+test_inc_dirs = inc_dirs
+test_inc_dirs += include_directories('../catch')
+
+test_flags = flags
+test_override_options = override_options
+
+
+if host_machine.system() == 'windows'
+    test_flags += [
+        windows_options_prefix + 'EHc',
+        windows_options_prefix + 'Zc:hiddenFriend',
+        windows_options_prefix + 'D_SCL_SECURE_NO_WARNINGS',
+    ]
+else
+    test_flags += ['-fexceptions']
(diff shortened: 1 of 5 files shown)


Totto16:
I just wanted to ask, if there's a better way, to determine, if the compiler is in c++xx > mode or is this the best you can get:
```c++
is_in_cpp17_mode = cpp.compiles(
        '''
#if defined(__cplusplus) && (__cplusplus >= 201703L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L))
#else
#error in cpp 17 mode
#endif
'''
 ```

**Comment by Totto16** [comment]
So it's the correct way, it's just also checking for the MSVC alternative, which might be the case in some MSVC compilers, it would be neat to have something like this in meson built-in, but since it isn't, my solution is correct, it throws an error (#error directive) when it's not c++17 or above, so `cpp.compiles()` returns false, so everything is correct, it also works (locally and in the ci) :+1:

**Comment by Totto16** [comment]
I saw a small logic error, but other than that, it is now correct

File: subprojects/packagefiles/stduuid/test/meson.build [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+if host_machine.system() == 'windows'
+    test_flags += [
+        windows_options_prefix + 'EHc',
+        windows_options_prefix + 'Zc:hiddenFriend',
+        windows_options_prefix + 'D_SCL_SECURE_NO_WARNINGS',
+    ]
+elif host_machine.system() == 'darwin'
+    test_flags += ['-fexceptions', '-g', '-Wall']
+else
+    test_flags += ['-fexceptions', '-g', '-Wall']
```
jpakkane, line 18: You must _never_ add `-g` or `-Wall`. That's what debug and warning levels are for.

PR #366, opened 2022-04-02
Add Cairo.

jpakkane:

PR #1045, opened 2023-05-31
Add cairo, fontconfig, fribidi and pixman

xclaesse:
This also add support for disabling cairo tests (see https://github.com/mesonbuild/wrapdb/pull/366#issuecomment-1235580822) and increase test timeout for pixman.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -93,6 +93,9 @@
       "windows": false
     }
   },
+  "fontconfig": {
+    "fatal_warnings": false
+  },
   "fuse": {
     "_comment": "- relies on Linux and BSD specific APIs",
     "build_on": {
--- a/ci_config.json
+++ b/ci_config.json
@@ -429,6 +429,12 @@
       "--timeout-multiplier=2"
     ]
   },
+  "pixman": {
+    "_comment": "Some tests takes more than 2min to complete.",
+    "test_options": [
+      "--timeout-multiplier=3"
+    ]
+  },
   "pkgconf": {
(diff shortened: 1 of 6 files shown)


eli-schwartz:
Oh... From the other PR:

> Running tests requires specific hacks: https://gitlab.freedesktop.org/cairo/cairo/-/blob/master/.gitlab-ci.yml#L202. Even `-Dcair:tests=disabled` is not enough, which is an upstream issue.

That's sad (assuming cair is a typo for cairo and the option was correct when running).

xclaesse:
Fontconfig is now blocked by https://gitlab.freedesktop.org/fontconfig/fontconfig/-/merge_requests/275, unless we have a way to skip clang-cl CI here?

xclaesse:
> That's sad (assuming cair is a typo for cairo and the option was correct when running).

I gave it another try since that comment was pretty old, but it still fails: https://github.com/mesonbuild/wrapdb/actions/runs/5134739149/jobs/9239166651?pr=1045. That's not a functional test that get disabled with `-Dtests=disabled`. Certainly an upstream bug that would need to be investigated. But as said in the old PR, we are not upstream's CI and Meson build definition is upstream.

xclaesse:
Fixed the root cause there: https://github.com/mesonbuild/wrapdb/pull/1047

xclaesse:
@eli-schwartz failed CI is clang-cl (fixed upstream, should not be blocker for us) and msys symbolextractor crash (random). I think this PR should be merged regardless of those failures. Ok?

xclaesse:
@eli-schwartz Rebased, LGTM?

PR #1047, opened 2023-05-31
expat: Depend on Meson >=0.54.1

xclaesse:
It was already silently required, but Meson >= 1.0 started to warn about it: https://github.com/mesonbuild/meson/pull/11511.

```
expat| WARNING: Project targets '>= 0.49.0' but uses feature introduced
in '0.54.1': cmakedefine without exactly two tokens.
```


eli-schwartz:
Kind of disappointed whenever I see cmakedefine format being used at all, if I have to be honest...

eli-schwartz (approved):

PR #1048, opened 2023-06-02
Update unit-system to 0.7.1

noah1510:


eli-schwartz:
@noah1510,

This wrap no longer builds due to checksum errors. Looking at the repo it appears that you've renamed it from unit-system/ to unit-system-meson/ and therefore the generated tarball's root directory has changed, invalidating the checksum and also requiring to change the wrap-file's `directory = ` entry.

noah1510:
How should I update this then?
Should I create a PR with the updated url, filename and dir?
Does this also update the old versions or are they also going to be broken?

eli-schwartz:
Every old version is broken.

What's the goal of adding "-meson" to the project name anyway?

noah1510:
It is more than just adding this the repo was moved to an organization.
-meson is added since this is the generated source code for the meson build system.
There is also an arduino and python output so I streamlined the naming to make it more clear which output each repo is.

eli-schwartz:
It's a fairly common practice for people to have multiple repos centered around a project and name some/all the repos with a suffix indicating the programming language it is relevant for. So naming one of them "unit-system-python" is familiar to me.

I assume the Arduino repo has changed the actual code in ways specific to the Arduino environment, so that makes sense too.

Why is"unit-system Desktop C++14" named after the build system though? I'd expect it to be either "unit-system" or "unit-system-cpp", with a mild preference for the former.

If you did use the former then the repo name would stay the same. And just redirecting the user/org name has no effect on tagged archives, nor the "git clone" downloaded directory name.

noah1510:
I also considered it naming "unit-system-cxx" or rather "unit-system-cpp".
Internally the desktop C++ and Arduino share a lot of implementation details so their common tooling dir and command is cpp.

I don't want to use simply "unit-system" for one specific target due to the fact that the name is too ambiguous.
So no matter what the archive name will have to be changed.

At the moment all outputs are named after the build system because that simplifies or rather speeds up the testing of the build system files.
All the code is produced by a code generator either way so creating an output for autotools or cmake would keep the outputs cleaner and make it easier to spot issues.
Changing this will require a lot of changes to the tooling so I would have to do it later this week.

noah1510:
I finally had the time to look at this again.
The tooling changes were less than expected so the new repo name is unit-system-cppDesktop: https://github.com/unit-system-exports/unit-system-cppDesktop

PR #1050, opened 2023-06-05
Add wrap for ffmpeg (libraries only)

grep0:
Basic wrap for ffmpeg libraries (binaries not supported yet)

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -2935,6 +2935,14 @@
       "0.7.3-1"
     ]
   },
+  "quickjs": {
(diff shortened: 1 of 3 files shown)


eli-schwartz:
If you just want to use the external project module then you don't need the WrapDB as it's pretty easy to locally set up.

On the other hand, use of the WrapDB is supposed to provide a portable wrap that works well OOTB, which invoking a posix shellscript on Windows is not.

PR #1051, opened 2023-06-10
liburing: update to 2.4

pobrn:
Changes:
 - sync list of examples
 - sync list of tests
 - sync list of man pages
 - compile `version.c` as well
 - create `io_uring_version.h`
 - make examples depend on `thread_dep`
 - only use `-Wno-sign-compare` for tests
 - define `UAPI_LINUX_IO_URING_H_SKIP_LINUX_TIME_TYPES_H` in `compat.h`
 - add `sys/fanotify.h` check


benoit-pierre:
It looks the source tarballs contains symlinks, maybe that's the issue?

benoit-pierre:
Or rather a case insensitive filesystem issue?

benoit-pierre:
Yes, but with with the same name (ignoring case) as the target.

benoit-pierre:
It's a link to itself, an error on dereferencing seems logical.

neheb:
right. the issue is the tarball. Wonder if there's a way to preprocess it.

neheb:
ChatGPT gave me this.

It's probably not even the right function.

```
@@ -31,7 +31,13 @@ def read_archive_files(path: Path, base_path: Path):
             archive_files = set(base_path / i for i in archive.namelist())
     else:
         with tarfile.open(archive_path) as archive:
-            archive_files = set(base_path / i.name for i in archive)
+            archive_files = set()
+            for member in archive:
+                if member.islnk():
+                    symlink_path = base_path / member.name
+                    extract_symlink(member, symlink_path)
+                else:
+                    archive_files.add(base_path / member.name)
     return archive_files
 
 if __name__ == '__main__':
```

eli-schwartz:
There appears to be an ffi library upstream, now.

eli-schwartz:
Maybe? What about my ffi comment.

neheb:
What happens if you apply https://github.com/neheb/wrapdb/commit/3878b0fb507e5f2706c917f3e328cdc73007edce ?

neheb:
yep, worked: https://github.com/neheb/wrapdb/actions/runs/6511968588

No solution for macOS.

PR #1053, opened 2023-06-14
fluidsynth: update to 2.3.3

kcgen:


kcgen:
I'm not sure what to do here; is this a wait




(Side comment.. FluidSynth's dependence on glib has generated a higher-than-normal rate of CI problems whenever that dependency tree needs to be compiled from source instead of provided via system lib. It's been a source of years of CI pain for DOSBox Staging, and often breaks under vcpkg, too.  :face_exhaling: )

neheb:
@kcgen try rebasing

neheb:
Why close?

Issue #1052, opened 2023-06-14
protobuf.wrap: how to get protoc.exe?


dvrogozh:
Reiterating question asked by @nanthony21 at https://github.com/mesonbuild/protobuf/issues/16

protobuf.wrap does expose few library dependencies, but leaves `protoc.exe` unexposed as it seems. Is there currently a way to get access to the built `protoc.exe`? Or .wrap file needs to be fixed with `meson.override_find_program()`? Program is needed in some cases when you have to generate some protobuf content on your side.

neheb:
IIRC I avoided adding code like that as the generated include paths get messed up. No real idea how to fix.

dvrogozh:
Does meson support the case when wrap file exposes both some dependencies and programs, i.e.:
```
<...>
[provide]
dependency_names = protobuf
program_names = protoc
```
And with meson.build getting patched:
```
<...>
meson.override_dependency('protobuf', protobuf_dep)
meson.override_find_program('protoc', protoc)
```

In this case I am getting:
```
subprojects/protobuf-<version>/meson.build:14:9: ERROR: Recursive include of subprojects: protobuf => protobuf
```

Mind that I am seeing this on my own protobuf.wrap which uses some pre-built stuff...

neheb:
No idea. I had it working but abandoned it after protoc was not working correctly as a subproject. I don't know of a way to get it working without code changes.

dvrogozh:
> Does meson support the case when wrap file exposes both some dependencies and programs

Yes it does. That was my mistake in a meson.build for file-wrap. I used `find_program` to look for the program and that gave recursive program. I.e. don't use `find_program()` in the file-wrap for binaries. I ended up using `files(protoc.exe)` instead and that worked.

dvrogozh:
> Or .wrap file needs to be fixed with `meson.override_find_program()`?

I have tried to add this to protobuf wrap in wrapdb and this seems working working for me. I will submit PR. Let's discuss there possible issues if any.

dvrogozh:
> I will submit PR. Let's discuss there possible issues if any.

Submitted #1054. It will need approval for ci since I did not contribute to this project before.

PR #1054, opened 2023-06-14
protobuf: provide protoc program

dvrogozh:
Fixes: #1052

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -2049,7 +2049,11 @@
       "protobuf",
       "protoc"
     ],
+    "program_names": [
+      "protoc"
+    ],
     "versions": [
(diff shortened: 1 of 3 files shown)


neheb:
Need to bump releases.json

dvrogozh:
> Need to bump releases.json

Added "3.21.12-3" to versions. I hope I guessed right what needs to be done from other commits to releases.json.

neheb:
Ah yes program_names is also needed. See the rest of the file for a hint.

dvrogozh:
> Ah yes program_names is also needed. See the rest of the file for a hint.

Added.

PR #1061, opened 2023-06-24
libtiff: update to 4.5.1

bgilbert:
Upstream added a Windows resource file and refactored endian detection.


subprojects/packagefiles/libtiff/libtiff/meson.build
```diff
@@ -57,6 +57,13 @@ sources = files(
   'tif_@0@.c'.format(host_machine.system() == 'windows' ? 'win32' : 'unix'),
 )
 
+if host_machine.system() == 'windows' and meson.get_compiler('c').get_argument_syntax() == 'msvc'
```
neheb, line 60: Not compatible with MSYS2?

bgilbert: Personally I'd make the check less restrictive, but to be safe, I'm mirroring the one in the upstream CMakeLists.

bgilbert:
Okay, this is ready for review.

subprojects/packagefiles/libtiff/libtiff/meson.build
neheb: Fishy. I'd try just testing for windows.

bgilbert: Okay, done.

PR #1062, opened 2023-06-29
libpng: update to 1.6.40

bgilbert:
Drop manual `__aarch64__` definition on MSVC ARM64, since the underlying issue was fixed in libpng 1.6.38.

Disable Windows CRT deprecation warnings in both libpng and pngtest, as is done in the upstream CMakeLists.


subprojects/packagefiles/libpng/meson.build
```diff
@@ -18,8 +18,11 @@ include = include_directories('.')
 
 c_args = []
 
-if host_machine.system() == 'windows' and get_option('default_library') != 'static'
-  c_args += '-DPNG_BUILD_DLL'
+if host_machine.system() == 'windows'
+  c_args += ['-D_CRT_NONSTDC_NO_DEPRECATE', '-D_CRT_SECURE_NO_DEPRECATE']
```
neheb, line 22: still an issue. See clang-cl output.

btw: these are specific to msvc.

bgilbert: The config wasn't setting the defines for `pngtest`, only for `libpng`.  Fixed now.  The build still has warnings for zlib but not for anything in libpng.

`_CRT_NONSTDC_NO_DEPRECATE` is also respected by the mingw-w64 headers.

PR #1063, opened 2023-07-03
Add CRoaring

N-Wouda:
Adds a Meson build definition for [RoaringBitmap/CRoaring](https://github.com/RoaringBitmap/CRoaring), a widely used efficient bitmap impementation. 

(I have not done this before, so I don't know if this is correct, but it works in my project! 😄)

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -369,7 +369,7 @@
       "croaring"
     ],
     "versions": [
-      "1.3.0"
+      "1.3.0-1"
     ]
   },
   "cxxopts": {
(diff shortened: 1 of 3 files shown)


N-Wouda:
Is the OS X build broken because of this PR, or something unrelated?

neheb:
Unrelated.

eli-schwartz:
I recommend in the future using `git commit --amend && git push --force` to fix issues with your previous commit... there is one logical change here but spread across 4 commits, which is highly confusing.

(It is normally policy to split commits by logical change.)

PR #1066, opened 2023-07-05
freetype: update to 2.13.1

neheb:
Added fallback URL


xclaesse:
No need to add fallback URL, wrapdb automatically upload the tarball to github, see https://github.com/mesonbuild/wrapdb/releases/tag/opus_1.4-1 for example.

neheb:
Fails on Build All otherwise. This mirror is officiall.

PR #1068, opened 2023-07-07
libattr: add

neheb:


subprojects/libattr.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = attr-2.5.1
+source_url = http://download.savannah.nongnu.org/releases/attr/attr-2.5.1.tar.xz
```
bgilbert, line 3: https?

(Yes we have the `source_hash`, but still.)

subprojects/packagefiles/libattr/meson.build
```diff
@@ -0,0 +1,47 @@
...
+  'libattr/attr_copy_action.c',
+  'libattr/attr_copy_check.c',
+  'libattr/attr_copy_fd.c',
+  'libattr/attr_copy_file.c',
+  'libattr/libattr.c',
+  'libattr/syscalls.c',
+  include_directories: incdirs,
+  install: true,
+  link_args: link_args,
+)
```
bgilbert, line 76: Upstream does have a libtool version.

PR #1076, opened 2023-07-11
libjpeg-turbo: update to 3.0.0

bgilbert:
`so_minor` wasn't reset to 0 upstream.


subprojects/packagefiles/libjpeg-turbo/meson.build
```diff
@@ -130,11 +127,10 @@ incdir = include_directories('.')
 subdir('simd')
 cdata.set('WITH_SIMD', have_simd)
 if not have_simd
-  simd = static_library('simd', 'jsimd_none.c')
+  simd = []
 endif
```
jpakkane, line 131: This splits the simd handling needlessly in two different locations. The if block should be deleted and instead in the `simd` subdir you should do this:

```
if simd_opt.disabled()
  simd = []
  subdir_done()
endif
```

bgilbert: Updated!

PR #1027, opened 2023-05-24
Update magic_enum to 0.9.0

tmayoff:

PR #1078, opened 2023-07-11
Bumped magic_enum version to 0.9.2

tmayoff:
The last [PR](https://github.com/mesonbuild/wrapdb/pull/1027) got a bit messed up so moved to another uptodate branch


Totto16:
What's the status of this? The fails seem related to catch, and not the dependency itself?

neheb:
sure. tests can be disabled AFAIK.

tmayoff:
Can the tests be disabled just for the wrapdb CI, or do I need to add a patch to remove the tests?

Issue #1082, opened 2023-07-14
protobuf: std::system_error thrown running protoc on Linux


dvrogozh:
Consider the following example running on Linux:
```
$ cat meson.build
project('proto-bug', 'c')

protoc = find_program('protoc')

custom_target('proto',
  input : 'some.proto',
  output : ['som.pb.h', 'some.pb.cc'],
  command : [protoc, '--cpp_out=@OUTDIR@', '-I=@CURRENT_SOURCE_DIR@',
    # We can't use @INPUT@ which would be more correct from meson perspective
    # for the reason that protoc has a requirement that -I or --proto_path
    # must be exact prefix of .proto file names otherwise protoc aborts with
    # an error. Unfortunately meson did not guarantee that for the @CURRENT_SOURCE_DIR@
    # and @INPUT@ pair.
    '@CURRENT_SOURCE_DIR@/some.proto'],
  build_always : true)

$ cat some.proto
syntax = "proto3";
package some;
message SomeMsg { uint64 _some = 1; }

$ meson wrap install protobuf
Installed protobuf version 3.21.12 revision 3

$ meson setup --wrap-mode=forcefallback _build
<...>
$ meson compile -C _build
<...>
[1/1] Generating proto with a custom command
FAILED: som.pb.h some.pb.cc
/home/dvrogozh/bugs/proto-bug/_build/subprojects/protobuf-3.21.12/src/protoc --cpp_out=. -I=../ ..//some.proto
terminate called after throwing an instance of 'std::system_error'
  what():  Unknown error -1
ninja: build stopped: subcommand failed.
```

I was using meson 1.1.1. I don't think that's Linux OS specific, but formally I was running Cent OS 7 with toolsets as follows:
```
scl enable devtoolset-9 rh-python38 -- meson <...>
```

dvrogozh:
This issue is the same as https://github.com/protocolbuffers/protobuf/issues/4958. It's caused by missing dependency from 'threads' in rules for 'protoc'. I raised PR #1083 to address this. Once applied I don't see the issue anymore and example posted in issue description builds fine.

PR #1083, opened 2023-07-14
protobuf: add dependency from threads for protoc

dvrogozh:
Fixes: #1082


neheb:
Great. I have an incomplete update to 3.22: https://github.com/neheb/wrapdb/tree/3 . Would be great if you could take over.

PR #1087, opened 2023-07-23
libarchive: update to 3.7.0

neheb:


benoit-pierre:
This update break in some cross-compilation cases, there's already a new [v3.7.1 release](https://github.com/libarchive/libarchive/releases/tag/v3.7.1) with a fix (https://github.com/libarchive/libarchive/commit/bd074c25).

PR #1096, opened 2023-08-07
lua: Only build as DLL on Windows

PixelyIon:
`-DLUA_BUILD_AS_DLL` was being applied to all non-POSIX platforms rather than only Windows, this would lead to `__declspec(dllexport)`/`__declspec(dllimport)` being used which is Windows-specific causing compilation to fail on non-Windows platforms.


jpakkane:
You should have added a new entry to `releases.json`. Because it was missing there won't be a new release.

PixelyIon:
Ah, given this PR has already been merged do I create a new one to bump it?

xclaesse:
Oops, missed that indeed, yes please make another PR.

eli-schwartz:
It also means that CI did not run.

PR #1099, opened 2023-08-09
sanity_checks: Verify releases.json has been updated

xclaesse:


eli-schwartz:
Seems reasonable. Wanna temporarily push a dummy update to subprojects/ as a followup commit to check that it fails in practice?

xclaesse:
It works: https://github.com/mesonbuild/wrapdb/actions/runs/5812175409/job/15756985660?pr=1099.

eli-schwartz:
Nice!

PR #1110, opened 2023-08-17
update: tabulate-1.5.0, indicators-2.3.0

rtgiskard:


tools/sanity_checks.py
```diff
@@ -305,6 +305,8 @@ def check_source_url(self, name: str, wrap_section: configparser.SectionProxy, v
         elif name == 'netstring-c':
             # There is no specific version for netstring-c
             return True
+        elif name in ('indicators', 'tabulate'):
+            version = '.'.join(version.split('.')[:2])
```
jpakkane, line 309: Could use use version numbers with two digits instead? It would be better not to have custom code for edge cases like this.

rtgiskard: Thanks for the note, updated :)

PR #1114, opened 2023-08-19
Fix formatting + remove deceptive description

meator:
Some descriptions of options contained superfluous whitespace. This commit removes it.

Descriptions for `loguru_debug_logging` and `loguru_debug_checks` say that these options default to the opposite of `NDEBUG`. This is true in general but meson fully overrides these options so that they are enabled by default independently of `NDEBUG` (or `b_ndebug`).

This PR also shortens description of `loguru_with_streams` because the code sample doesn't look good without newlines (which aren't displayed in `meson configure`).

Should I bump the revision of the wrap?


neheb:
Also prefix the commit title.

PR #1117, opened 2023-08-20
giflib: add version 5.2.1

benoit-pierre:
Note: there's an unnecessary include of `unistd.h` in `gif_hash.h` that breaks the MSVC build, so I had to include a patched one… There hasn't been any activity on the upstream repository since the last release (4 years ago), so I don't think a fix is forthcoming.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -114,6 +114,15 @@
       "windows": false
     }
   },
+  "giflib": {
+    "_comment": "Unnecessary <unistd.h> include in gif_hash.h breaks the build",
+    "build_on": {
+      "windows": false
+    },
+    "msys_packages": [
+      "diffutils"
+    ]
+  },
   "glbinding": {
     "debian_packages": [
       "libxi-dev",
--- a/releases.json
+++ b/releases.json
@@ -708,6 +708,14 @@
       "1.14.1-1"
     ]
   },
+  "giflib": {
+    "dependency_names": [
+      "giflib"
+    ],
+    "versions": [
+      "5.2.1-1"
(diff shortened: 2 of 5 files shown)


jpakkane:
Unfortunately we have a strict policy that wrapd does not permit code changes, only build definitions.

benoit-pierre:
> Unfortunately we have a strict policy that wrapd does not permit code changes, only build definitions.

OK, so ship a version that does not build with MSVC.

I thought adding the following section to `ci_config.json` would to that:
```json
    "_comment": "Unnecessary <unistd.h> include in gif_hash.h breaks the build",
    "build_on": {
      "windows": false
    },
```
But that does not seem to be the case:
```
subprojects\giflib-5.2.1\meson.build:38:5: ERROR: C header 'unistd.h' not found

A full log can be found at D:\a\wrapdb\wrapdb\_build\meson-logs\meson-log.txt
s.
ERROR: test_releases (__main__.TestReleases) (step='check_new_release', name='giflib')
  File "D:\a\wrapdb\wrapdb\tools\sanity_checks.py", line 258, in test_releases
  File "D:\a\wrapdb\wrapdb\tools\sanity_checks.py", line 401, in check_new_release
    raise Exception(f'Wrap {name} failed to configure due to bugs in the wrap, rather than due to being unsupported')
Exception: Wrap giflib failed to configure due to bugs in the wrap, rather than due to being unsupported
```

benoit-pierre:
OK, I've removed the header, the setup phase will now fail when building with MSVC.

PR #1118, opened 2023-08-20
leptonica: tweak giflib support

benoit-pierre:
Check for `dependency('giflib')` first, before falling back to `compiler.find_library('gif', has_headers: 'gif_lib.h')`, so using our `giflib.wrap` is supported.

Depends on #1117.


benoit-pierre:
I've also doubled the tests' timeout to avoid failures on the CI.

PR #1123, opened 2023-08-24
ci: fix transient MSYS2 failures

benoit-pierre:
Make sure Python is updated to >=3.11 (fix https://github.com/msys2/MINGW-packages/issues/17415).

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -88,6 +88,8 @@ jobs:
             pkgconf:p
             python-certifi:p
             python-pip:p
+          # Make sure Python is updated to >=3.11 (fix https://github.com/msys2/MINGW-packages/issues/17415).
+          update: true
 
       - name: Install packages
         shell: msys2 {0}


eli-schwartz:
Do you offhand know how much longer the CI will take when msys2 has to apply updates?

(I assume we only need it until a new base install is deployed, no idea what the timeframe for that is.)

benoit-pierre:
No noticeable difference apparently:
- [before](https://github.com/mesonbuild/wrapdb/actions/runs/5968264501/job/16191730146#logs): 2m 22s (for the `msys2/setup-msys2@v2` phase)
- [after](https://github.com/mesonbuild/wrapdb/actions/runs/5968563614/job/16192638546?pr=1123): 2m 26s

eli-schwartz:
Oh ha, that makes sense... it's only even updating pacman and msys2-runtime. :D The rest are going to have to be downloaded either way.

neheb:
@benoit-pierre this should be applied to Build All as well.

PR #1124, opened 2023-08-25
libunibreak: add version 5.1

benoit-pierre:
Upstream: https://github.com/adah1972/libunibreak.


eli-schwartz:
The tests fail miserably, and I'm totally unsurprised since it downloads random content from the internet during the tests and apparently that has now changed what the tests expect. This feels... wrong.

benoit-pierre:
Same as upstream… We can change the links to `https://www.unicode.org/Public/15.0.0/ucd/auxiliary/LineBreakTest.txt` and so on, that should fix the tests.

PR #1131, opened 2023-08-28
add sqlitecpp 3.3.1-1

ninjaoflight:
adds [SQLiteCpp](https://github.com/SRombauts/SQLiteCpp) version 3.3.1 to the list of wraps

If there is anything else required just let me know
> a previous PR for this wrap was in #816

### DIFF
--- a/subprojects/sqlitecpp.wrap
+++ b/subprojects/sqlitecpp.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = SQLiteCpp-3.3.1
+source_url = https://github.com/SRombauts/SQLiteCpp/archive/refs/tags/3.3.1.zip
+source_filename = sqlitecpp-3.3.1.zip
+source_hash = e1bdc1ff47f201bc2b77703de1001ea5e525151fa57c22ed43f06335b4784603
+patch_directory = sqlitecpp
+
+[provide]
+sqlitecpp = sqlitecpp_dep
--- a/ci_config.json
+++ b/ci_config.json
@@ -578,6 +578,18 @@
       "windows": false
     }
   },
+  "sqlitecpp": {
+    "_comment": "required to not use the default sqlite3 in MacOS to avoid compilation errors/tests failures, also add pkgconfig path to find sqlite3(keg-only)",
+    "build_options": [
+      "sqlitecpp:SQLITECPP_BUILD_TESTS=true",
+      "sqlitecpp:SQLITECPP_BUILD_EXAMPLES=true",
+      "pkg_config_path=/usr/local/opt/sqlite/lib/pkgconfig"
+    ],
+    "brew_packages": [
+      "sqlite"
+    ]
+
+  },
   "tinyfsm": {
     "build_options": [
       "tinyfsm:build_examples=true"
--- a/releases.json
+++ b/releases.json
@@ -2600,6 +2600,14 @@
       "3.34.1-1"
     ]
   },
+  "sqlitecpp": {
+    "dependency_names": [
+      "sqlitecpp"
+    ],
+    "versions": [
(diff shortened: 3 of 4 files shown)


ninjaoflight:
seems that the CI is not passing due to some errors regarding gtest

neheb:
```
error: ‘std::enable_if_t’ has not been declared
  302 |   template <typename T, std::enable_if_t<std::is_convertible<T, int64_t>::value,
```

Needs C++14.

neheb:
Add sqlitecpp:std_cpp=c++14 I guess.

jpakkane:
You need to set the C++ std version to something newer. Some platforms seem to default to c++11 or even older.

neheb:
Yeah all projects using gtest 1.14 need to be using C++14: https://github.com/google/googletest/releases/tag/v1.14.0

ninjaoflight:
yes, I noticed that also, the project uses by default c++11 and gtest 1.14.0 so I will send a PR to the main project in the meantime

jpakkane:
Their readme specifically mentions depending only on c++11, so that might be a hard sell.

neheb:
Readme seems outdated. I'd just mention that C++11 has been deprecated by at least MSVC.

ninjaoflight:
> Their readme specifically mentions depending only on c++11, so that might be a hard sell.

I guess some users can just use an older GTest if required, specially the ones using older compilers in ubuntu or debian

> Readme seems outdated. I'd just mention that C++11 has been deprecated by at least MSVC.

Yes, in fact for windows C++14 is already the default.

I made a PR on the main repo: SRombauts/SQLiteCpp#448

neheb:
for now, you can add the fixed meson.build file to packagefiles such that it gets overwritten.

ninjaoflight:
forgot to add the patch directory, the build is passing for all except macos, as is using included sqlite

ninjaoflight:
the CI is failing on MacOS due to the included SQLite3, is there a way to force the fallback? 
on the SQLitecpp repo '--force-fallback-for=sqlite3' was used

neheb:
For macOS you can probably install SQLite with brew and use it. ci_config has examples with ffmpeg for example.

subprojects/sqlitecpp.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = SQLiteCpp-3.3.1
+source_url = https://github.com/SRombauts/SQLiteCpp/archive/refs/tags/3.3.1.zip
+source_filename = sqlitecpp-3.3.1.zip
+source_hash = e1bdc1ff47f201bc2b77703de1001ea5e525151fa57c22ed43f06335b4784603
+patch_directory = sqlitecpp
+
+
```
neheb, line 8: extra line

ninjaoflight: fixed

ninjaoflight:
for macos still failing even after brew install

```
sqlite is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have sqlite first in your PATH, run:
  echo 'export PATH="/usr/local/opt/sqlite/bin:$PATH"' >> /Users/runner/.bash_profile

For compilers to find sqlite you may need to set:
  export LDFLAGS="-L/usr/local/opt/sqlite/lib"
  export CPPFLAGS="-I/usr/local/opt/sqlite/include"

For pkg-config to find sqlite you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/sqlite/lib/pkgconfig"
```

neheb:
yes I told you how to fix it. You need to add pkgconfig paths. this is done with libffmpegthumbnailer for example.

PR #1133, opened 2023-08-30
wayland-protocols: update to 1.32

dcbaker:


neheb:
Didn’t this get macOS support?

dcbaker:
The protocols themselves might build okay on Windows or MacOS, but the tests require wayland-scanner, which only works on DRM based kernels (Linux, *BSD, Solaris) by design.

neheb:
I remember a wayland project gaining macOS(kthread) support. Maybe libwayland or something.

dcbaker:
It might in the newer wayland 1.22 release, but we need new wayland protocols to build that.

dcbaker:
Apparently yes, 1.22 does work on macos.

neheb:
So I guess upstream needs to fix protocols as well.

dcbaker:
I think this is a check and egg problem, but lets see what happens. I've added an update for wayland to 1.22 (https://github.com/mesonbuild/wrapdb/pull/1134), and then rebased this on that. Let's see what happens here.

dcbaker:
Nope, more reading. There's a port of wayland to MacOS, but it doesnt look like it was ever upstreamed.

dcbaker:
So I'm not sure what to do here. It seems a little odd to ask upstream to add "unsupported platform" errors just for our CI, but?

PR #1137, opened 2023-08-31
harfbuzz: update to 8.1.1

benoit-pierre:


benoit-pierre:
Is there a mechanism for the sanity check tool to handle conditional dependency names? `harfbuzz-cairo` is only provided if cairo support was compiled in.

neheb:
Probably not. If it’s a feature option, you could set it to enabled in ci_config

neheb:
I see 404 errors. Maybe add a source_fallback

benoit-pierre:
> I see 404 errors. Maybe add a source_fallback

For a subproject (freetype2).

benoit-pierre:
> > Is there a mechanism for the sanity check tool to handle conditional dependency names?
> 
> I see you found the answer in your fixup commit. Please squash however.

Of course, I'll squash once everything is ready.

neheb:
Freetype was recently updated and this was overlooked.

benoit-pierre:
The wrap in this repo has a fallback, so does the WrapDB entry, but the subproject wrap in harfbuzz does not.

benoit-pierre:
And the fallback URL should be removed from this repo wrap, correct?

amyspark:
@neheb Hi! Sorry, I did not think about x86's need for underscore prefixing. @benoit-pierre will need to modify 

https://github.com/mesonbuild/wrapdb/blob/06700e5d6b1fd17f6195e477a1bd4230469d03c9/subprojects/packagefiles/icu/source/data/meson.build#L33-L39

as follows:

```meson
elif cpp.get_argument_syntax() == 'gcc'
  icudata_asm_output = '@0@.S'.format(U_ICUDATA_NAME)
    if host_machine.cpu() == 'x86'
      icudata_command += ['-a', 'gcc-cygwin']
    else
      icudata_command += ['-a', 'gcc-mingw64']
    endif
  else
    icudata_command += ['-a', 'gcc']
  endif
```

The `mingw64` generator was not intended to be used with x86 outputs. There is no functional difference between the two generators except that `cygwin` prepends the needed underscore.

Apart from that, I expect CI will continue to fail. In my local testing, Meson is leaking Harfbuzz's `cpp_rtti=false` default clause into the ICU build, and the latter **needs** RTTI to work.

neheb:
@huh that’s interesting. One CI I did not add is x86 MSVC. Wonder if it would also fail there. Although it’s probably soft deprecated at this point. clang32 is different as it’s relatively new.

neheb:
icu was fixed for clang32. maybe rebase.

neheb:
clang-cl failure is unrelated.

benoit-pierre:
> clang-cl failure is unrelated.

So mark it as ready and merge with the build broken with clang-cl?

neheb:
Yes

PR #1139, opened 2023-08-31
[TEST] Libffi update

benoit-pierre:


benoit-pierre:
`closures / unwindtest_ffi_call` fails multiple times, even after increasing the timeout. I mark it as xfailed, and of course, it passes! ;(

neheb:
mhmm. you're going to have fun with this. IIRC there are upstream bugs that need fixing.

benoit-pierre:
> `closures / unwindtest_ffi_call` fails multiple times, even after increasing the timeout. I mark it as xfailed, and of course, it passes! ;(

Nope, I'm just an idiot…

benoit-pierre:
Either I screwed up again, or `should_fail` has no impact on a test timeout.

benoit-pierre:
Yes, it's your PR, with some tweaks to fix the CI.

xclaesse:
> `closures / unwindtest_ffi_call` fails multiple times, even after increasing the timeout. I mark it as xfailed, and of course, it passes! ;(

I don't think it's a timeout, on other platforms that return immediately. Probably a deadlock. I think failing CI are upstream bugs, clang-cl is not tested and mingw32 all tests fail with autotools in upstream CI but they ignore test failure.

benoit-pierre:
> > `closures / unwindtest_ffi_call` fails multiple times, even after increasing the timeout. I mark it as xfailed, and of course, it passes! ;(
> 
> I don't think it's a timeout, on other platforms that return immediately. Probably a deadlock. I think failing CI are upstream bugs, clang-cl is not tested and mingw32 all tests fail with autotools in upstream CI but they ignore test failure.

Yep, let's just skip that particular test.

PR #1141, opened 2023-09-01
libusb: honor 'examples' and 'tests' options

Spadi0:
The Meson build options `examples` and `tests` did previously exist, but were not honored.


neheb:
I'd prefer using subdir_done() instead.

Spadi0:
@neheb I've changed the code to use `subdir_done()` instead. :+1:

Spadi0:
:thinking: So the CI fails, but the problem is not in a part of the code I've touched. I can bump the minimum Meson version as another commit in this PR, which should fix this problem - do you want me to do that?

neheb:
I'd rather fix the problem directly than increase the version. Although it probably doesn't build with meson that old anyway...

Spadi0:
Ok, I authored a new commit which fixes the build without bumping the minimum required Meson version. It seems to build fine on my machine - hopefully the CI concurs... :crossed_fingers:

PR #1144, opened 2023-09-02
libevdev: add 1.13.1

benoit-pierre:
Upstream: https://gitlab.freedesktop.org/libevdev/libevdev.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -411,6 +411,17 @@
       "windows": false
     }
   },
+  "libevdev": {
+    "alpine_packages": [
+      "bash"
+    ],
+    "build_options": [
+      "libevdev:documentation=disabled"
+    ],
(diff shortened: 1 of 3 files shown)


benoit-pierre:
So how do you handle Linux only packages that support meson, but don't show an `unsuported …` error on other OSs?

neheb:
Fix upstream I guess

benoit-pierre:
Seeing as `tools/sanity_checks.py` would be happy with a failed required dependency check, then I don't see why a failed `cc.find_library(…)` or `cc.has_header(…)` check is not similarly accepted.

neheb:
So one way to fix this is to copy the meson.build file to packagefiles and fix it up such that it passes CI.

benoit-pierre:
Can this be merged as-is? On Linux is supported by upstream, I see no point in messing around with the original meson.build.

jpakkane:
Failing tests are not permitted, you need to edit the wrapdb metadata files so that the build is skipped on Windows.

PR #1148, opened 2023-09-09
sanity checks: don't install recommended APT packages

benoit-pierre:
Rational: speedup installs by selecting only what's strictly necessary.


jpakkane:
Thanks.

PR #1149, opened 2023-09-09
curl: add 8.0.1

### LINKED ISSUE #561: Request: libcurl
Please consider adding libcurl (in a way that can be built for windows from linux (cross-platform)).
Thanks

benoit-pierre:
Note:
- not the latest version (8.0.1 is from 2023, latest is 8.2.1 released July 2023), because that's what I started with.
- advanced / experimental options are not supported: Unicode (Windows), HTTP/3, Hyper HTTP backend, WebSocket…
- not all SSL backends are supported: only OpenSSL, Secure Channel (Windows), and Secure Transport (Windows)
- not all LDAP providers are supported: only OpenLDAP `>= 2.4.48` (2019) and Windows.

Close #561.

### DIFF
--- a/subprojects/packagefiles/curl/meson.build
+++ b/subprojects/packagefiles/curl/meson.build
@@ -1,0 +1,1875 @@
+# Init. {{{
+
+project(
+  'curl',
+  'c',
+  version: '8.0.1',
+  # Version gated use of `test(…, verbose: true)` needs `0.62.0`.
+  meson_version: meson.version().version_compare('>=0.62.0') ? '>=0.62.0' : '>=0.60.0',
+)
+
+fs = import('fs')
+pkg = import('pkgconfig')
+cc = meson.get_compiler('c')
+cdata = configuration_data()
+
+sys_deps = []
+lib_deps = []
+
+# Constants. {{{
+
+CURL_PUBLIC_INC = include_directories('include')
+CURL_INTERNAL_INC = include_directories('lib')
+CURL_VERSION = cc.get_define(
+  'LIBCURL_VERSION',
+  include_directories: CURL_PUBLIC_INC,
+  prefix: '#include"curl/curlver.h"',
+).strip('"')
+CURL_VERSION_NUM = cc.get_define(
+  'LIBCURL_VERSION_NUM',
+  include_directories: CURL_PUBLIC_INC,
+  prefix: '#include"curl/curlver.h"',
+)
+
+if CURL_VERSION != meson.project_version()
+  error(
+    'versions mismatch, meson has @0@, while `culver.h` reports @1@!'.format(
+      meson.project_version(),
+      CURL_VERSION,
+    ),
+  )
+endif
+
+WINDOWS_ONLY_ERROR = 'only supported on Windows'
+MACOS_ONLY_ERROR   = 'only supported on macOS'
+
+MISSING_LIBS_ERROR = 'required libraries missing'
+
+SSL_DISABLED_ERROR  = 'feature ssl is disabled'
+HTTP_DISABLED_ERROR = 'feature http is disabled'
+
+DISABLED_OPT = get_option('_disabled')
+ENABLED_OPT  = get_option('_enabled')
+
+# }}}
+
+# Basic feature options. {{{
+# (not dependent on checks)
+
+foreach _opt : [
+  'cookies',
+  'crypto-auth',
+  'dict',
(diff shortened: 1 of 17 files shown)


xclaesse:
> 8.0.1 is from 2013, latest is 8.2.1 released July 2013

You mean 2023 or it had no release in 10 years?

benoit-pierre:
> > 8.0.1 is from 2013, latest is 8.2.1 released July 2013
> 
> You mean 2023 or it had no release in 10 years?

2023, of course…

neheb:
so a bunch of packages are not set to be built on MSVC because of missing curl. Would be interesting to undo that and see if they build.

benoit-pierre:
> so a bunch of packages are not set to be built on MSVC because of missing curl. Would be interesting to undo that and see if they build.

Those packages would need an explicit `subproject/curl.wrap` file, no?

neheb:
well, this PR provides it.

benoit-pierre:
Oh, right, the superproject provides it.

benoit-pierre:
So let's say I want to test if `cpr` no builds, and patch-out `"build_on": { "windows": false }` in `ci_config.json`, do I need to bump the revision in `releases.json` too?

neheb:
In order to get it tested in this PR, yes. A commit like that can be removed prior to merging.

subprojects/packagefiles/curl/meson.build
```diff
@@ -0,0 +1,1866 @@
+# Init. {{{
+
+project(
+  'curl',
+  'c',
+  version: '8.0.1',
+  # Version gated use of `test(…, verbose: true)` needs `0.62.0`.
+  meson_version: meson.version().version_compare('>=0.62.0') ? '>=0.62.0' : '>=0.60.0',
```
neheb, line 8: that is...fascinating. I remember doing something like this with an older meson version and got an error. Can't remember which it was. I'd just set to 0.62.0 personally.

subprojects/packagefiles/curl/meson.build
```diff
@@ -0,0 +1,1866 @@
...
+# }}}
+
+# Basic checks. {{{
+
+check_args = []
+check_headers = []
+
+_headerlist = {
+  'arpa/inet.h': [
+    'getpeername',
```
neheb, line 131: that is so hard to read. Maybe put the array on the same line.

subprojects/packagefiles/curl/meson.build
neheb: actually, why is verbose: true even needed?

subprojects/packagefiles/curl/meson.build
benoit-pierre: It's easier as soon as they are multiple functions, and `muon fmt` will automatically switch to multiple-lines in that case anyway. I try to limit my fighting with its choices.

subprojects/packagefiles/curl/meson.build
benoit-pierre: The whole curl testsuite is run by `tests/runtests.pl`, so from meson's point of view, there's only one test. And said test can take a while, so verbose output is a much nicer default.

neheb:
I'd remove liblastfm from the test commit. It's known broken on Windows. I need to fix it at some point.

benoit-pierre:
Looks like the libupnp build is broken with msvc too.

neheb:
ah right. libupnp needs some special threading library that I never bothered with.

neheb:
first casualty:

```
2023-09-17T04:36:03.7999038Z D:\a\wrapdb\wrapdb\subprojects\libmicrohttpd-0.9.77\src\include\microhttpd.h(126): error C2371: 'ssize_t': redefinition; different basic types
2023-09-17T04:36:03.7999720Z D:\a\wrapdb\wrapdb\subprojects\libnpupnp-5.0.1\src\inc\config.h(43): note: see declaration of 'ssize_t'
```

The problem here is that libmicrohttpd defines it as ptrdiff_t whereas this defines it as __int64. Should be the former I think.

neheb:
```
meson -Dwraps=libmicrohttpd,curl a

subprojects/curl-8.0.1/lib/meson.build:217:6: ERROR: Tried to override dependency 'libcurl' which has already been resolved or overridden at /home/mangix/devstuff/wrapdb/subprojects/libmicrohttpd-0.9.77/meson.build:181:
```

PR #1150, opened 2023-09-10
tweak cURL dependent packages

benoit-pierre:
Follow up to #1149.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -313,7 +313,7 @@
     "debian_packages": [
       "libcurl4-gnutls-dev"
     ],
-    "_comment": "libcurl is not available with MSVC CI",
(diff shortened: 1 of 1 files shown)


ci_config.json
```diff
@@ -292,7 +292,7 @@
     "build_options": [
       "liblastfm:tests=enabled"
     ],
-    "_comment": "libcurl is not available with MSVC CI",
+    "_comment": "broken",
```
neheb, line 295: msys too

benoit-pierre: Good catch, amended.

neheb:
you changed libnpupnp, not libupnp. Not accurate for the former. Will change in an upcoming PR.

PR #1153, opened 2023-09-11
Add libsass 3.6.4

lazka:
This is a direct copy of https://github.com/lazka/libsass/tree/meson which I've created for gtk back in the days.

There is an open PR for adding meson support upstream in https://github.com/sass/libsass/pull/3073 but the project is considered deprecated and hasn't seen a commit in over a year, so there is not much hope for progress there.

There is a newer 3.6.5 release out, but I wanted to copy things as is from my fork first and look at updating it later.

### DIFF
--- a/subprojects/packagefiles/libsass/meson.build
+++ b/subprojects/packagefiles/libsass/meson.build
@@ -12,11 +12,9 @@ add_project_arguments(
   language: ['cpp'])
 
 cpp = meson.get_compiler('cpp')
-if cpp.get_id() != 'msvc'
-  add_project_arguments(
-    cpp.get_supported_arguments(['-Wno-non-virtual-dtor']),
-    language: ['cpp'])
-endif
+add_project_arguments(
+  cpp.get_supported_arguments('-Wno-non-virtual-dtor'),
+  language: ['cpp'])
 
 inc = include_directories('include')
 winres_path = files(join_paths('res', 'resource.rc'))
--- a/subprojects/packagefiles/libsass/meson.build
+++ b/subprojects/packagefiles/libsass/meson.build
@@ -11,11 +11,6 @@ add_project_arguments(
   '-DLIBSASS_VERSION="@0@"'.format(meson.project_version()),
   language: ['cpp'])
 
-cpp = meson.get_compiler('cpp')
-add_project_arguments(
-  cpp.get_supported_arguments('-Wno-non-virtual-dtor'),
-  language: ['cpp'])
-
 inc = include_directories('include')
 winres_path = files(join_paths('res', 'resource.rc'))
 
--- a/subprojects/packagefiles/libsass/meson.build
+++ b/subprojects/packagefiles/libsass/meson.build
@@ -1,0 +1,25 @@
+project('libsass', 'c', 'cpp',
+        version: '3.6.4',
+        meson_version : '>= 0.48.0',
+        default_options: [
+            'c_std=c99',
+            'cpp_std=c++11',
+            'buildtype=debugoptimized',
+        ])
+
+add_project_arguments(
+  '-DLIBSASS_VERSION="@0@"'.format(meson.project_version()),
+  language: ['cpp'])
+
+cpp = meson.get_compiler('cpp')
+if cpp.get_id() != 'msvc'
+  add_project_arguments(
+    cpp.get_supported_arguments(['-Wno-non-virtual-dtor']),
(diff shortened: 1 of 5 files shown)


tristan957 (approved):

subprojects/packagefiles/libsass/src/meson.build
```diff
@@ -0,0 +1,107 @@
...
+]
+
+sass_sources = cpp_sources + c_sources
+
+if host_machine.system() == 'windows'
+  windows = import('windows')
+  sass_sources += [windows.compile_resources(winres_path)]
+endif
+
+dl_dep = cpp.find_library('dl', required : false)
```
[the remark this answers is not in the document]
lazka: I added a version check

subprojects/packagefiles/libsass/meson.build
```diff
@@ -0,0 +1,23 @@
...
+        ])
+
+add_project_arguments(
+  '-DLIBSASS_VERSION="@0@"'.format(meson.project_version()),
+  language: ['cpp'])
+
+cpp = meson.get_compiler('cpp')
+add_project_arguments(
+  cpp.get_supported_arguments('-Wno-non-virtual-dtor'),
+  language: ['cpp'])
```
eli-schwartz, line 17: Meson 1.0.0 removed the `-Wnon-virtual-dtor` that this switches off, so you can probably just drop it entirely. See https://github.com/mesonbuild/meson/pull/10339

subprojects/packagefiles/libsass/src/meson.build
```diff
@@ -0,0 +1,116 @@
...
+if host_machine.system() == 'windows'
+  windows = import('windows')
+  sass_sources += [windows.compile_resources(winres_path)]
+endif
+
+if meson.version().version_compare('>= 0.62')
+  dl_dep = dependency('dl', required : false)
+else
+  dl_dep = cc.find_library('dl', required : false)
+endif
```
eli-schwartz, line 83: It looks like libsass unconditionally requires the dl interface on non-Windows, and the point of the `dependency('dl')` addition to Meson was specifically so that people who need it can *require* it, without having to search for the "dl" found library and assume that if it isn't found, it is present inside your libc.

subprojects/packagefiles/libsass/src/meson.build
```diff
@@ -0,0 +1,116 @@
...
+
+lib_args = []
+if get_option('default_library') != 'static'
+  lib_args += ['-DADD_EXPORTS']
+endif
```
eli-schwartz, line 88: It's probably fine to add this define even on non-Windows as it's simpler from meson.build to do so -- although it is unused on non-Windows, and simply makes the command line longer...

subprojects/packagefiles/libsass/src/meson.build
```diff
@@ -0,0 +1,116 @@
...
+conf_data.set('includedir', join_paths('${prefix}', get_option('includedir')))
+conf_data.set('VERSION', meson.project_version())
+
+pkg_install_dir = join_paths(get_option('libdir'), 'pkgconfig')
+
+configure_file(
+  input: 'support/libsass.pc.in',
+  output: 'libsass.pc',
+  configuration: conf_data,
+  install_dir: pkg_install_dir)
```
eli-schwartz, line 116: It's a very straightforward .pc file, I would probably just use meson's builtin generator and get libsass-uninstalled.pc support for free.

subprojects/packagefiles/libsass/src/meson.build
```diff
@@ -76,14 +76,18 @@ if host_machine.system() == 'windows'
...
-  dl_dep = dependency('dl', required : false)
+if host_machine.system() != 'windows'
+  if meson.version().version_compare('>= 0.62')
+    dl_dep = dependency('dl')
+  else
+    dl_dep = cc.find_library('dl', required : false)
+  endif
 else
-  dl_dep = cc.find_library('dl', required : false)
+  dl_dep = declare_dependency()
```
tristan957, line 86: The more common idiom is dependency('')

lazka: Do you mean `dependency('', required: false)` ?

neheb: Yes. The difference is the return value of found()

lazka:
Thanks!

PR #1155, opened 2023-09-11
asio: fetch archive from github.com

dvrogozh:
Fetching asio's tar.gz archive from github.com instead of sourceforge.net allows to improve 2 aspects:

* Sometimes sourceforge.net is under maintenance and building projects which depend on asio fail. Per my observatons this happenned at least twice over past 2 months. Since many other wrapdb projects already fetch from github, plus asio git home is on github, it's reasonable to switch to minimize fetches from different resources.

* Archive from sourceforge.net is dist tarball which weights more than tagged sources from github (4.3MB vs. 2.8MB). So fetching from github we somewhat optimize both footprint and download time.

CC: @stephanlachnit


subprojects/asio.wrap
```diff
@@ -1,8 +1,8 @@
 [wrap-file]
-directory = asio-1.24.0
-source_url = https://sourceforge.net/projects/asio/files/asio/1.24.0%20%28Stable%29/asio-1.24.0.tar.gz/download
-source_filename = asio-1.24.0.tar.gz
-source_hash = 2f23ef6eada06ecc1472af5df6365ed4f15452ccd07dc0a6851fa20d571dba94 
+directory = asio-asio-1-24-0
+source_url = https://github.com/chriskohlhoff/asio/archive/refs/tags/asio-1-24-0.tar.gz
```
dvrogozh, line 3: @neheb, @eli-schwartz : what would be a recommended solution to handle this wrapdb sanity check error? As you can see asio upstream project uses somewhat specific versioning scheme diverging between tag names (1-24-0) and release versions (1.24.0).
```
AssertionError: False is not true : Version 1.24.0 not found in https://github.com/chriskohlhoff/asio/archive/refs/tags/asio-1-24-0.tar.gz
```

eli-schwartz:
> Sometimes sourceforge.net is under maintenance and building projects which depend on asio fail. Per my observatons this happenned at least twice over past 2 months.

Recent versions of the wrapdb tooling upload a fallback_url copy of the tarball to the wrapdb repository for reliability while still using the canonical upstream tarball for auditability. The original asio submission predates this, but updating the wrap to the latest version would automatically provide this fallback.

> Since many other wrapdb projects already fetch from github, plus asio git home is on github, it's reasonable to switch to minimize fetches from different resources.

Asio's git hosting may be on GitHub, but its official website frontpage says the download location is exclusively sourceforge. The fact that GitHub autogenerates a `git-archive` output whether the project wishes to or not, doesn't mean it has to be used.

It's also got badly formatted git tags :( and the asio sources are in an additional subdirectory of the git repo which is odd.

IMHO it's best to limit fetching to GitHub to cases where GitHub is the canonical location. Of course, many projects don't upload their own release assets and rely on GitHub autogenerated tarballs for that.

subprojects/asio.wrap
dvrogozh: Answering for myself.... There are seems to be couple other projects with similar problem, so, that's handled as exception right in the sanity check script:
https://github.com/mesonbuild/wrapdb/blob/b160efaee3d15614ab3cb0ff26086e14f62c1c5c/tools/sanity_checks.py#L305-L309

I'll add similar exception for asio.

dvrogozh:
> IMHO it's best to limit fetching from GitHub to cases where GitHub is the canonical location.

Ok, noted.

> Recent versions of the wrapdb tooling upload a fallback_url copy of the tarball to the wrapdb repository<...>
> The original asio submission predates this, but updating the wrap to the latest version would automatically provide this fallback.

@eli-schwartz : This sounds like a very good improvement which should address the issue I noticed and tried to address with this PR... Can you explain more what should be done? You mean I need update meson to the latest version? or something needs to be added to the asio.wrap file? Sorry, I did not quite follow...

> It's also got badly formatted git tags :(

That's due to how asio maintainers name tags - they use `asio-1-24-0` and github adds `asio` prefix resulting in the noted archive name.

> and the asio sources are in an additional subdirectory of the git repo which is odd.

That's how asio project is structured. See https://github.com/chriskohlhoff/asio - they have `asio` folder on the top level. Github just honestly created archive with sources available in the repo.

eli-schwartz:
Right, what I'm saying is it feels a bit like the github repo is an inferior source altogether. :)

eli-schwartz:
> This sounds like a very good improvement which should address the issue I noticed and tried to address with this PR... Can you explain more what should be done? You mean I need update meson to the latest version? or something needs to be added to the asio.wrap file? Sorry, I did not quite follow...

The current wrapdb version of asio is asio 1.24.0, but upstream has released asio 1.28.0 and if you contribute an updated asio 1.28.0 to the wrapdb, it will automatically take advantage of the new tooling.

It may be as trivial as `sed 's/4/8/g'` on the wrap file and meson.build file, but I haven't checked.

dvrogozh:
> if you contribute an updated asio 1.28.0 to the wrapdb, it will automatically take advantage of the new tooling.

I see. So, that's internal improvement on wrapdb side. I will need to check whether I can switch to asio 1.28.0 on my side right away. Hopefully yes. I will try it out and post an update.

xclaesse:
Note that simply adding `1.24.0-2` in releases.json is enough to generate a new identical release which will have the fallback URL.

eli-schwartz:
I'd much rather avoid micro optimization churn for outdated releases though.

dvrogozh:
> Note that simply adding 1.24.0-2 in releases.json is enough

That's still good to know since other components might not have new release we can switch to...

dvrogozh:
I have submitted PR #1162 to update asio to latest 1.28.1. It works on my side. I am closing current PR,

eli-schwartz:
Thanks for the version update. :)

On the topic of generating identical releases, I *think* it should be sufficient for a wrapdb maintainer to manually upload an additional release asset for the fallback tarball, and replace the .wrap file with one that has the additional fallback entry. People newly installing the wrap file would see the change, but you would not get updated by meson wrap update. This is probably okay since it's very easy to manually update if you are affected by sourceforge flakiness.

dvrogozh:
@eli-schwartz : I have a question. I updated asio.wrap on my side and I indeed see what you were talking about - now there is source_falback_url pointing to wrapdb re-hosted tarball. Questions:

1. In which order urls will be tried? I assume first will be `source_url` (sourceforge.net), then `source_fallback_url` (wrapdb). Right?
2. On which errors on fetching `source_url` meson will attempt `source_fallback_url`?

I am asking above because actual issue I saw was sha256 checksum mismatch. Thus, if `source_url` will be attempted first, will checksum mismatch be qualified to try `source_fallback_url` or that will be considered fatal error?

```
[wrap-file]
directory = asio-1.28.1
source_url = https://sourceforge.net/projects/asio/files/asio/1.28.1%20%28Stable%29/asio-1.28.1.tar.gz/download
source_filename = asio-1.28.1.tar.gz
source_hash = eef0eddce419827876f4929c6d7523e010528031e5289ec57914e89fb9ab9d8d
patch_filename = asio_1.28.1-1_patch.zip
patch_url = https://wrapdb.mesonbuild.com/v2/asio_1.28.1-1/get_patch
patch_hash = 9a8f47f1e9e60c4250142b7fe5b1ada3a2e3c73dd057d4c6d4198147fc5d25a9
source_fallback_url = https://github.com/mesonbuild/wrapdb/releases/download/asio_1.28.1-1/asio-1.28.1.tar.gz
wrapdb_version = 1.28.1-1

[provide]
asio = asio_dep
```

eli-schwartz:
That should gracefully handle fallback, yes. We catch all wrap-specific exception types (including "checksum mismatch") and retry the entire download stage with the fallback url.

PR #1156, opened 2023-09-11
gdk-pixbuf: Add version 2.42.9

hrxi:


benoit-pierre:
Use `"fatal_warnings": false`.

neheb:
```
WARNING: Gettext not found, all translation (po) targets will be ignored.
```

sounds like an upstream bug.

neheb:
https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/156

subprojects/packagefiles/gdk-pixbuf/disable-jpeg-test-205.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/tests/pixbuf-jpeg.c b/tests/pixbuf-jpeg.c
+index be2c6b4fe..5a21668a1 100644
+--- a/tests/pixbuf-jpeg.c
++++ b/tests/pixbuf-jpeg.c
+@@ -216,7 +216,7 @@ main (int argc, char **argv)
+   g_test_add_func ("/pixbuf/jpeg/comment", test_comment);
+   g_test_add_func ("/pixbuf/jpeg/at_size", test_at_size);
+   g_test_add_func ("/pixbuf/jpeg/issue70", test_jpeg_markers);
+-  g_test_add_func ("/pixbuf/jpeg/issue205", test_jpeg_fbfbfbfb);
++  //g_test_add_func ("/pixbuf/jpeg/issue205", test_jpeg_fbfbfbfb);
```
xclaesse, line 10: This is patching source code in WrapDB, that's against our policy. I also don't think .patch files are going to be included into wrapdb releases, so that won't work.

IMHO you could just skip tests. When the Meson port is upstream, we don't need to be their CI.

hrxi: It sounds like the policy is suboptimal in this case, where it leads to not running the tests at all vs simply excluding one known-to-be-bad test.

I can also include the whole test file from the meson.build if that's better, it wouldn't be patching source code.

.github/workflows/sanity_checks.yml
```diff
@@ -33,6 +33,11 @@ jobs:
       - name: Fetch tags and unshallow
         run: git fetch --unshallow --tags
 
+      # https://github.com/actions/runner-images/issues/5459#issuecomment-1532856844
+      - name: Fix patch binary in search path
+        # Needed to use GNU patch instead of Strawberry Perl patch.
+        run: echo "C:\Program Files\Git\usr\bin" >> $env:GITHUB_PATH
```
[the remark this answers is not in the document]
hrxi: Sounds good.

subprojects/packagefiles/gdk-pixbuf/disable-jpeg-test-205.patch
xclaesse: In this case running unit tests is just a waste of CI time, we just distribute upstream package as-is. Running tests makes sense when we maintain the Meson port in WrapDB because we have to make sure our build actually works. IMHO the bug is reported upstream, now it's not our problem anymore.

hrxi: > IMHO the bug is reported upstream, now it's not our problem anymore.

There are still three test failures on macOS. We can certainly pretend they're not there, but I think looking at them individually and assessing whether they represent a problem is more healthy. If we ignore the tests, then this step will not be done for upgrades of the library.

As an aside: Why does wrapdb default to running the tests if it's mostly a waste of CI time?

hrxi:
Works now with CI workaround for Strawberry Perl patch.

Disabled the tests as requested.

PR #1162, opened 2023-09-13
asio: update to 1.28.1

dvrogozh:
This update also has a goal to use benefits of wrapdb fallback mechanism in case original source location (sourceforge.net) is not available. See discussion in #1155.

CC: @stephanlachnit


eli-schwartz (approved):
Thanks!

PR #1163, opened 2023-09-13
jbig2dec: add 0.20

benoit-pierre:
Upstream: https://github.com/ArtifexSoftware/jbig2dec.

### DIFF
--- a/subprojects/packagefiles/jbig2dec/src/jbig2dec.def
+++ b/subprojects/packagefiles/jbig2dec/src/jbig2dec.def
@@ -1,0 +1,13 @@
+EXPORTS
+; Public API.
+jbig2_complete_page
+jbig2_ctx_free
+jbig2_ctx_new_imp
+jbig2_data_in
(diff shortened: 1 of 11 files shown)


subprojects/packagefiles/jbig2dec/src/jbig2dec.def
```diff
@@ -0,0 +1,13 @@
...
+; Public API.
+jbig2_complete_page
+jbig2_ctx_free
+jbig2_ctx_new_imp
+jbig2_data_in
+jbig2_global_ctx_free
+jbig2_make_global_ctx
+jbig2_page_out
+jbig2_release_page
+; Private, but used by the jbig2dec executable.
```
neheb, line 11: that's somewhat interesting. I'm guessing this doesn't use visibility attributes either.

benoit-pierre: Nope, the shared library exports everything…

PR #1164, opened 2023-09-13
libtiff: update to 4.6.0; set library version

bgilbert:
Upstream canonically keeps the library version in `configure.ac` in `current:revision:age` format, and converts it to a version number in the cmake scripts.  Mirror the cmake logic.

Also stop naming the library binary `libtiff4`; upstream doesn't do it.


subprojects/packagefiles/libtiff/libtiff/meson.build
```diff
@@ -65,11 +65,12 @@ if host_machine.system() == 'windows'
 endif
 
 tiff4_lib = library(
-  'tiff4',
```
neheb, line 68: Does this impact the generated pkgconfig file?

bgilbert: The filename is the same, and the only change to the contents is the `-l` argument:

```patch
@@ -6,6 +6,6 @@
 Description: Tag Image File Format (TIFF) library.
 Version: 4.6.0
 Requires.private: libjpeg, liblzma, zlib
-Libs: -L${libdir} -ltiff4
+Libs: -L${libdir} -ltiff
 Libs.private: -lm
 Cflags: -I${includedir}
```

PR #1171, opened 2023-09-15
libarchive: Define `_GNU_SOURCE` when checking for types

PixelyIon:
In glibc, certain types such as `id_t` are gated behind checks for `__USE_XOPEN` and similar defines that are set set when `_GNU_SOURCE` is defines which is done by libarchive's cmake `config.h.in`. Without this, the types visible to meson during the checks and the compiler during compilation can differ resulting in these types can be defined twice which can cause compilation errors.

Here's an excerpt from glibc's `sys/types.h` with `id_t` being conditionally defined:
```c
#if (defined __USE_XOPEN || defined __USE_XOPEN2K8) \
    && !defined __id_t_defined
typedef __id_t id_t;
# define __id_t_defined
#endif
```
and `features.h`:
```c
#ifdef _GNU_SOURCE
# undef  _ISOC95_SOURCE
# define _ISOC95_SOURCE	1
# undef  _ISOC99_SOURCE
# define _ISOC99_SOURCE	1
# undef  _ISOC11_SOURCE
# define _ISOC11_SOURCE	1
# undef  _ISOC2X_SOURCE
# define _ISOC2X_SOURCE	1
# undef  _POSIX_SOURCE
# define _POSIX_SOURCE	1
# undef  _POSIX_C_SOURCE
# define _POSIX_C_SOURCE	200809L
# undef  _XOPEN_SOURCE
# define _XOPEN_SOURCE	700
# undef  _XOPEN_SOURCE_EXTENDED
# define _XOPEN_SOURCE_EXTENDED	1
# undef	 _LARGEFILE64_SOURCE
# define _LARGEFILE64_SOURCE	1
# undef  _DEFAULT_SOURCE
# define _DEFAULT_SOURCE	1
# undef  _ATFILE_SOURCE
# define _ATFILE_SOURCE	1
# undef  _DYNAMIC_STACK_SIZE_SOURCE
# define _DYNAMIC_STACK_SIZE_SOURCE 1
#endif
```


PixelyIon:
Ouch it seems like a list of prefixes was only added in meson 1.0.0, I hope it's an acceptable workaround to use `prefix: '#define _GNU_SOURCE 1\n#include <sys/types.h>'` instead.

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -169,7 +169,7 @@ types = {
 }
 
 foreach t, v : types
-  if not cc.has_type(t, prefix: '#include <sys/types.h>')
+  if not cc.has_type(t, prefix: '#define _GNU_SOURCE 1\n#include <sys/types.h>')
```
benoit-pierre, line 172: You could use `args: '-D_GNU_SOURCE'` too.

dcbaker (approved):

PR #1173, opened 2023-09-15
libuv: update to 1.46.0

neheb:


neheb:
that is...quite the issue.

So the tests call icacls on Windows, which comes with Windows but not MSYS2. Meaning the paths have to leak into MSYS2. Unfortunate.

neheb:
leaking paths bother me. so I'll wait a bit before merging.

PR #1178, opened 2023-09-16
orcus: try to fix MSYS2 compilation

neheb:


neheb:
need to update instead

PR #1181, opened 2023-09-17
curl: fix compilation with libmicrohttpd

neheb:
Fixes a redefinition error. Most projects that define ssize_t for Windows use ptrdiff_t, not __int64.

While at it since curl is here now, get rid of it in ci_config.


neheb:
figured.

PR #1182, opened 2023-09-17
libnpupnp: update to 5.0.2

neheb:


neheb:
Upstream has this fixed: https://framagit.org/medoc92/npupnp/-/commit/5781a107ab3b31806980e815976639fdff35ac35

ping @medoc92

neheb:
let's just merge. MSVC fails but that's been the case.

PR #1185, opened 2023-09-18
Add sassc 3.6.2

lazka:
This is a direct copy of https://github.com/lazka/sassc/tree/meson which I've created for gtk back in the days.

This project is in a similar situation as #1153. There exists an upstream PR for adding meson (https://github.com/sass/sassc/pull/269), but the project is unmaintained, so unlikely to see progress there.

### DIFF
--- a/subprojects/packagefiles/sassc/meson.build
+++ b/subprojects/packagefiles/sassc/meson.build
@@ -31,9 +31,9 @@ if host_machine.system() == 'windows'
   win_res = windows.compile_resources(
     'res/libsass.rc',
     depend_files: ['res/libsass.ico'])
-  sassc_sources += [win_res]
-  incs += [include_directories(join_paths('win', 'posix'))]
-  sassc_sources += [join_paths('win', 'posix', 'getopt.c')]
+  sassc_sources += win_res
+  incs += include_directories('win/posix')
+  sassc_sources += 'win/posix/getopt.c'
 endif
 
 sassc = executable(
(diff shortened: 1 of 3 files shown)


lazka:
The next step after this would be to try to get rid of all references of my forks in GNOME and replace them with wrapdb: https://github.com/search?q=org%3AGNOME+https%3A%2F%2Fgithub.com%2Flazka%2Fsassc&type=code

subprojects/packagefiles/sassc/meson.build
```diff
@@ -0,0 +1,47 @@
...
+sassc_sources = ['sassc.c']
+
+incs = []
+if host_machine.system() == 'windows'
+  windows = import('windows')
+  win_res = windows.compile_resources(
+    'res/libsass.rc',
+    depend_files: ['res/libsass.ico'])
+  sassc_sources += [win_res]
+  incs += [include_directories(join_paths('win', 'posix'))]
```
jpakkane, line 35: You don't need to wrap `include_directories` with brackets. You also don't need to use `join_paths` at all. So this should be:

   incs += include_directories('win/posix')

And the same below for `getopt.c`.

lazka:
Thanks!

lazka:
> The next step after this would be to try to get rid of all references of my forks in GNOME and replace them with wrapdb: https://github.com/search?q=org%3AGNOME+https%3A%2F%2Fgithub.com%2Flazka%2Fsassc&type=code

Turns out both projects in GNOME using this depend on the wraps being used in flatpak as separate modules and not within meson, and I couldn't figure out how to use wrapdb without meson in flatpaks, so this project is currently on ice.

PR #1194, opened 2023-09-21
libffmpegthumbnailer: partially fix MSYS2 compilation

neheb:


neheb:
Bad part about this is that it's dead upstream.

PR #1195, opened 2023-09-21
catch: remove

neheb:
This package was and is a mistake. This is in reality catch2. Last version of catch was 1.12.0.


neheb:
downgrading doesn't work either. guess no way to move forward.

eli-schwartz:
What problem is this trying to solve? This wouldn't be the only deprecated wrap name, IIRC.

neheb:
what other?

Problem this means to solve is that libffmpegthumbnailer uses catch as a dependency for the tests. meson --wrap-mode=forcefallback will cause usage of this wrap and fail to compile.

jpakkane:
FTR the dependency name of Catch 2 is`catch2`, at least according to the pc file it installs:

    /usr/share/pkgconfig/catch2.pc

If someone is using `catch` as a dep name they are probably looking for Catch 1. The "correct" solution would be to update the project to use Catch 2 instead but that's probably not feasible.

A big fundamental issue here is that we have chosen to make wrapdb append only and keep the old releases around. Which is basically the thing you want but sometimes things are not great.

Just as a wild guess, if you were to add a provide entry for `catch2` in the wrap file of `catch`, would the lookup then fail?

neheb:
catch can't be used using dependency(). so no error here.

PR #1197, opened 2023-09-23
{libopenjp2,libxml2,sqlite3}: don't link with pthreads on Windows

bgilbert:
libopenjp2 and sqlite3 were configured to use the Win32 threading primitives on Windows, but were being linked with `-pthread` anyway.  Stop doing that.  For libopenjp2, we were defining both `MUTEX_win32` and `MUTEX_pthread`; stop doing that too.

libxml2 can use the Win32 primitives, but was taking a dependency on winpthreads instead.  Switch to the Win32 primitives.


subprojects/packagefiles/sqlite3/meson.build
```diff
@@ -17,15 +17,17 @@ libm = cc.find_library('m', required: false)
 sources = files('shell.c', 'sqlite3.c')
 if host_machine.system() == 'windows'
   sqlite_args = get_option('default_library') != 'static' ? '-DSQLITE_API=__declspec(dllexport)' : []
+  libthreads = declare_dependency()
```
neheb, line 20: dependency('', required: false) might be better.

neheb:
If I had to guess, more packages have this problem. Good enough here.

PR #1198, opened 2023-09-24
gumbo-parser: update to 0.12.1

benoit-pierre:
Note:
- switch to new maintained fork (also used by [Arch Linux ](https://archlinux.org/packages/extra/x86_64/gumbo-parser/))
- currently broken on MSVC because `visualc/include/strings.h` was removed

@txgk: can `visualc/include/strings.h` be added back? I would have opened an issue on codeberg, but registrations are disabled… (I also can contribute the port to meson if you're interested, BYTW)


txgk:
> @txgk: can visualc/include/strings.h be added back?

Hello, yes of course. I deleted it earlier because I thought that no one needed it, but since it has value still, let's return it. After 0.12.0 there were several good fixes, so I immediately released it, so just change the version to 0.12.1 and `visualc/include/strings.h` file will be back in place.

> I would have opened an issue on codeberg, but registrations are disabled…

Very strange. I guess they are testing something...

> I also can contribute the port to meson if you're interested, BYTW

I think it'd be better to stay with autotools for now to keep the gap between new and old repositories smaller (that way we'll save maintainers a good chunk of time I believe).

benoit-pierre:
> > @txgk: can visualc/include/strings.h be added back?
> 
> Hello, yes of course. I deleted it earlier because I thought that no one needed it, but since it has value still, let's return it. After 0.12.0 there were several good fixes, so I immediately released it, so just change the version to 0.12.1 and `visualc/include/strings.h` file will be back in place.

Thanks.

> > I would have opened an issue on codeberg, but registrations are disabled…
> 
> Very strange. I guess they are testing something...

Too much spam, apparently… ¯\\_(ツ)_/¯

> > I also can contribute the port to meson if you're interested, BYTW
> 
> I think it'd be better to stay with autotools for now to keep the gap between new and old repositories smaller (that way we'll save maintainers a good chunk of time I believe).

Sure, I understand.

PR #1203, opened 2023-09-27
github: add dependabot

neheb:
Bot to update CI.

This will update at least actions/checkout on merge.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -18,7 +18,7 @@ jobs:
       TEST_BUILD_ALL: 1
       TEST_FATAL_WARNINGS: ${{ github.event.inputs.fatal_warnings }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
@@ -39,7 +39,7 @@ jobs:
       TEST_BUILD_ALL: 1
       TEST_FATAL_WARNINGS: ${{ github.event.inputs.fatal_warnings }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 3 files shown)


eli-schwartz:
There's nothing to do here, it accomplishes zero things of value. I also generally object to dependabot because it's absolutely horrible around batching PRs.

If we were to use anything it should be renovate-bot, which is well behaved in large part because it is also open source. But I don't really think we should do anything.

neheb:
```
git grep uses: | cut -d ':' -f 3 | sort
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 actions/checkout@v3
 ilammy/msvc-dev-cmd@v1
 ilammy/msvc-dev-cmd@v1
 ilammy/msvc-dev-cmd@v1
 ilammy/msvc-dev-cmd@v1
 msys2/setup-msys2@v2
 msys2/setup-msys2@v2
```
are subject to update.

eli-schwartz:
My understanding is that the only thing that major version updates for actions/checkout actually changes is the internal nodejs version, which is the biggest fattest nothingburger ever. The previous major release is still fully supported with the same new features because there's rarely ever breaking changes, GitHub just considers the internal nodejs version to be a breaking change because GitHub Enterprise customers would need to update their deployment to new Actions runners.

neheb:
Meh. Guess I'll manually do it.

eli-schwartz:
My question isn't about whether there are things "subject to update", but rather "are there ever updates of substance".

What are we updating manually other than a nodejs internal version?

neheb:
@eli-schwartz reworked PR. Please review.

PR #1206, opened 2023-09-29
gtk4: add 4.13.1

hrxi:


releases.json
```diff
@@ -931,6 +931,14 @@
       "1.7.0-2"
     ]
   },
+  "gtk4": {
+    "dependency_names": [
+      "gtk4"
+    ],
+    "versions": [
+      "4.13.1-1"
```
xclaesse, line 939: AFAIK, odd numbers are unstable releases, I think we should instead add latest of 4.12.x.

hrxi: Seems to be correct according to [their homepage](https://gtk.org/):
> Latest stable 4.12.3
> Latest unstable 4.13.1
> Latest old stable 3.24.38

v1993:
Do you plan on working on this further? Linux builds should have `libxkbcommon-dev` package added in `ci_config.json`. I'm less sure about what to do about missing gstreamer on windows, though.

hrxi:
> Do you plan on working on this further?

Yes.

Thanks for your hints, will look into them when I get back to it.

PR #1208, opened 2023-09-29
libadwaita: add 1.4.0

hrxi:


hrxi:
```
subprojects/libadwaita-1.4.0/src/meson.build:272:10: ERROR: Dependency 'gio-2.0' is required but not found.
```

Can I force meson to disregard the system dependency and build it from the wrap instead?

xclaesse:
Wondering why it has glib on this CI... The problem is pretty interesting here. libadwaita does `gnome.compile_resources()` before checking for glib/gio dependency. That means the first time Meson lookup search for gio is inside the GNOME module where we don't specify a version. Later when libadwaita lookup for gio with a specific verion it's too late for Meson to use a fallback.

There are 2 ways of fixing this:
- Add `--wrap-mode=forcefallback` in ci_config.json.
- Patch libadwaita to do `gio_dep = dependency('gio-2.0', version: glib_min_version)` before `gnome.compile_resources()` in https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/meson.build.

xclaesse:
Maybe wrapdb should have forcefallback by default?

hrxi:
libadwaita 1.4.0 doesn't compile with MSVC: https://gitlab.gnome.org/GNOME/libadwaita/-/issues/739. Is that a reason to package an earlier version?

neheb:
> Maybe wrapdb should have forcefallback by default?

that'll break a bunch.

PR #1209, opened 2023-09-29
Bump fmt to 10.1.1

fennewald:
Updates fmt to it's most recent version, 10.1.1.

Only changes in underlying library are related to the naming of some macros, which are updated in the meson.build file.

I have been using this wrap in prod for the last month, and have not yet had any issues. Confirmed it works with `spdlog`.


fennewald:
Oops. That's embarrassing

jpakkane:
Please squash this to a single commit. The additional commits do not provide any value.

PR #1216, opened 2023-10-07
giflib: fix build with busybox

neheb:
busybox head supports -c but not --bytes


neheb:
ping @benoit-pierre

benoit-pierre (approved):

neheb:
ugh. two unwanted changes slipped in. I'll fix at some point.

PR #1221, opened 2023-10-09
nghttp2: add 1.56.0

hrxi:


hrxi:
Known build problem on MSVC: https://github.com/nghttp2/nghttp2/issues/1633. What should I do? Ignore the build failures?

Ignoring them in `ci_config.json` doesn't work, it expects the meson setup call to fail.

neheb:
Define it with add_project_arguments?

hrxi:
> Define it with add_project_arguments?

I think that wouldn't propagate to dependents, but they require it for the header. I guess I need to add it to `declare_dependency`, too.

subprojects/packagefiles/nghttp2/lib/meson.build
```diff
@@ -0,0 +1,63 @@
+subdir('includes')
+configure_file(
```
neheb, line 2: IMO use the regular pkgconfig module instead of patching the pc file.

neheb:
yes.

Windows has a SSIZE_T in BaseTsd.h FWIW. Maybe that can be adapted for upstream.

neheb:
@hrxi new version available. Also, this doesn't work with newer clang where implicit declarations become errors.

Issue #1017, opened 2023-05-11
"[provides]" section cannot be omitted in wrap files


pobrn:
The [wrapdb documentation](https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section) states that (since meson ???)

> A wrap file named `foo.wrap` implicitly provides the dependency name foo even when the `[provide]` section is missing.

But this is not accounted for in `sanity_checks.py` as far as I can see.

eli-schwartz:
Not all wraps necessarily perfectly correspond to that mapping. Either way, an implicit provide is a bit fragile since it means we are *guessing* -- and there's no downside to being explicit, is there?

PR #1223, opened 2023-10-10
glib-networking: add 2.78.0

hrxi:


hrxi:
This fails with
```
FAIL: test_releases (__main__.TestReleases) (step='provide', name='glib-networking')
  File "D:\a\wrapdb\wrapdb\tools\sanity_checks.py", line 224, in test_releases
    self.assertIn('provide', config.sections())
AssertionError: 'provide' not found in ['wrap-file']
```

However, it is kind of by design that this wrap doesn't provide any libraries to link to. glib-networking creates/installs gio plugins that allow gio to make TLS connections. You don't directly depend on that.

xclaesse:
I would add a new flag in ci_config.json that tells this wrap is not providing anything.

benoit-pierre:
There's a dirty workaround: you can add an empty `dependency_names =` or `program_names =` line to the provide section. Example:
```yml
[provide]
program_names =
```
:P

benoit-pierre:
Why even add yet another key to `ci_config.json`? `sanity_check.py` should allow a `release.json` entry with no `dependency_names` / `program_names`.

hrxi:
> Why even add yet another key to `ci_config.json`? `sanity_check.py` should allow a `release.json` entry with no `dependency_names` / `program_names`.

I have no opinion on this. If you come to a conclusion of what to do, I'll do it.

benoit-pierre:
[This step](https://github.com/mesonbuild/wrapdb/blob/master/tools/sanity_checks.py#L236) checks that the list of provides in `release.json` and the wrap are the same, so IMHO [that step](https://github.com/mesonbuild/wrapdb/blob/master/tools/sanity_checks.py#L222) can just be removed.

xclaesse:
> Why even add yet another key to `ci_config.json`? `sanity_check.py` should allow a `release.json` entry with no `dependency_names` / `program_names`.

Originally that was because it's a common mistake to forget listing dependencies a wrap provide. A new contributor and uncareful reviewer could omit dependency names in both [provide] section and releases.json. But maybe we should just assume we'll review carefully...

hrxi:
CC #1017

PR #1227, opened 2023-10-11
imgui: add docking branch as a separate dependency

tmayoff:
I have a lot of projects that utilize the docking branch of imgui, figured it'd be useful for others as well to have it available via wrapdb


tmayoff:
Anyone have an idea on how to fix the windows/macos builds

benoit-pierre:
It looks like it's missing a system dependency: dxgi.

tmayoff:
It's using the same patch files as the normal imgui, and they shouldn't differ all that drastically. I couldn't see any extra configs for imgui anywhere else other than some debian packages for the CI. This should work if the normal imgui does

benoit-pierre:
What I meant is `meson.build` need to be adapted to add the `dxgi` dependency on MSCV.

yretenai:
This branch is now merged into main, and released as imgui v1.90.

tmayoff:
I'm not sure if it's a good idea to be modifying the main imgui patch meson.build, but I'd like to reduce duplicate code. 

I'm not familiar enough with macos builds, so if someone knows what's happening or can debug it to help that'd be great.

benoit-pierre:
> I'm not sure if it's a good idea to be modifying the main imgui patch meson.build, but I'd like to reduce duplicate code.

No way around it if the 2 variants need different dependencies.

> I'm not familiar enough with macos builds, so if someone knows what's happening or can debug it to help that'd be great.

It's missing a framework, this one I think:
```meson
dependency('appleframeworks', modules : 'QuartzCore')
```

PR #1228, opened 2023-10-11
cmocka: fix mingw cross compile

neheb:
This wrap is fairly broken. The same configuration is used for both native and non native libraries, which only accidentally worked.


neheb:
haha this error is hillarious. On MinGW, clock_gettime needs threads as a dependency. It only accidentally works on clang as it's a builtin function.

PR #1234, opened 2023-10-11
gRPC: Add wrap and build files for gRPC

### LINKED ISSUE #1093: Request: gRPC Support
There are a number of google libraries supported already, as well as protobuf, so I think gRPC would be a great/natural addition.

klokik:
Closes #1093

Provides `grpc` and `grpc++` dependencies.

Some of the upstream libraries are not included yet as they need more work: I have to figure out yet how to gracefully specify protobuf pathes to protoc compiler when generating sources, and implement that in my meson.build generator script; as maintaining it manually, is more work than I can sign on to.

For the time being I have applied a workaround for missing protobuf header (in the old version currently provided with wrapdb) required for C# protoc plugin to be build.

Fixes for wrapdb version of abseil-cpp are in the works

Verified to work on GNU/Linux and macOS.
Will look towards windows support later.


neheb:
conflicts

klokik:
Fixed conflict (rebased)

Enabled Windows MSVC and clang-cl, Alpine.

Still have to verify that users will not have linking issues while using as static libraries, so don't merge yet

neheb:
https://github.com/mesonbuild/wrapdb/pull/1251 fixes the openssl issue. Unfortunately it seems to be NACKed,

klokik:
- Rebased
- Fixed `protoc` dependency search for the pkg-config case
- added a few inherited dependencies to grpc ones, for static clinking case
- target_machine -> host machine

Can this get merged?

Known issues:

- cross-compilation may be broken (unless you run it thru qemu-user): protoc plugins are compiled for `host_machine`, and in case of cross-compiling may fail to run on build machine to generate sources.
- testing targets not included
- lesser used `grpcpp_channelz`, `grpc_reflection` libraries are not included. These rely on grpc plugins generating sources.

I will address these follow-up merge requests, as I iron-out known corner-cases.

Issue #1145, opened 2023-09-05
zlib needs LARGEFILE support


neheb:
```
wasm-ld-13: warning: function signature mismatch: lseek
>>> defined as (i32, i32, i32) -> i32 in subprojects/zlib-1.3/libz.a(gzlib.c.o)
>>> defined as (i32, i64, i32) -> i64 in /usr/share/emscripten/cache/sysroot/lib/wasm32-emscripten/libc-debug.a(lseek.o)
```

found with emscripten.

PR #1239, opened 2023-10-13
zlib: fix emscripten size mismatch

neheb:
meson passes _FILE_OFFSET_BITS=64 by default but emscripten also requires _LARGEFILE64_SOURCE to match.


neheb:
Fixes: https://github.com/mesonbuild/wrapdb/issues/1145

This matches what CMakeLists.txt does.

PR #1240, opened 2023-10-13
libopenjp2: avoid pulling-in unnecessary dependencies

benoit-pierre:
Example: avoid downloading & configuring a cURL subproject when using `-Dbuild_jpip_server=false --wrap-mode=forcefallback`.

I'm tempted to change the `build_jpip_server` option to a feature for simpler code, which might break some configs when updating the wrap, thoughts?


neheb:
Go for it.

benoit-pierre:
OK, never mind, I don't have the patience to go through the mess that this wrap is for now… (`curl_dep` and `fcgi_dep` are used elsewhere for no good reason in another part of that half-backed, clearly never tested port).

neheb:
probably needs to be rewritten to modern standards honestly.

benoit-pierre:
Yep.

PR #1241, opened 2023-10-13
libarchive: fix wrong functions checks

neheb:
I removed these upstream in order to use the macros in the headers.


benoit-pierre:
Meaning?

benoit-pierre:
`HAVE__MKGMTIME` is still used, for example.

benoit-pierre:
Wait, never mind, I'm stupid and/or too tired…

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -107,6 +107,12 @@ if cc.has_header_symbol('errno.h', 'EILSEQ')
   cdata.set('HAVE_EILSEQ', 1)
 endif
 
+foreach w : ['_mkgmtime', 'ctime_s', 'gmtime_s', 'localtime_s']
+  if cc.has_header_symbol('time.h', w)
+    cdata.set('HAVE_@0@'.format(w.to_upper()), 1)
+  endif
+endforeach
+
```
benoit-pierre, line 115: Maybe move the check for `gmtime_r` here too?

neheb: No need. I moved the code upstream to try the _s versions first and then the _r. _r is implemented in terms of _s as a macro IIRC. _r on POSIX systems is available as a function.

neheb:
Reference: https://github.com/libarchive/libarchive/commits?author=neheb

neheb:
Added missing __MINGW_USE_VC2005_COMPAT. We don't test MinGW32 but might as well match upstream.

PR #1248, opened 2023-10-15
giflib: fix tests under busybox

neheb:
BusyBox does not support these diff options.

ping @benoit-pierre


benoit-pierre:
What's the use case for supporting BusyBox?

neheb:
Alpine Linux

benoit-pierre:
And `apk add diffutils` is not a valid solution?

The `--strip-trailing-cr` was for Windows, I think (which build after removing an unnecessary `unistd.h` include, if I remember correctly).

It looks like the diff version of BusyBox stupidly output an unified diff by default anyway, and support `-u` but not the long option variant.

benoit-pierre:
Not that I'm opposed to supporting Alpine, but I think it makes sense to expect a minimum standard for the development tools, and I don't think bare Alpine cuts it.

neheb:
hmm? CI passes.

benoit-pierre:
Yes, because:
```yaml
  "giflib": {
    "_comment": "Unnecessary <unistd.h> include in gif_hash.h breaks the build",
    "build_on": {
      "windows": false
    },
```

benoit-pierre:
Anyway, at least change it to `diff -u`.

neheb:
busybox does not support -u.

benoit-pierre:
I tested in docker, and it works fine (it's not listed in the help, but that's not the only option missing).

neheb:
busybox diff -u . works in neither fedora nor alpine.

benoit-pierre:
```bash
 ▹ podman run --rm -ti docker.io/library/alpine:latest
/ # diff -u <(echo foo) <(echo bar)
--- /dev/fd/64
+++ /dev/fd/65
@@ -1 +1 @@
-foo
+bar
/ # apk list
WARNING: opening from cache https://dl-cdn.alpinelinux.org/alpine/v3.18/main: No such file or directory
WARNING: opening from cache https://dl-cdn.alpinelinux.org/alpine/v3.18/community: No such file or directory
alpine-baselayout-3.4.3-r1 x86_64 {alpine-baselayout} (GPL-2.0-only) [installed]
alpine-baselayout-data-3.4.3-r1 x86_64 {alpine-baselayout} (GPL-2.0-only) [installed]
alpine-keys-2.4-r1 x86_64 {alpine-keys} (MIT) [installed]
apk-tools-2.14.0-r2 x86_64 {apk-tools} (GPL-2.0-only) [installed]
busybox-1.36.1-r2 x86_64 {busybox} (GPL-2.0-only) [installed]
busybox-binsh-1.36.1-r2 x86_64 {busybox} (GPL-2.0-only) [installed]
ca-certificates-bundle-20230506-r0 x86_64 {ca-certificates} (MPL-2.0 AND MIT) [installed]
libc-utils-0.7.2-r5 x86_64 {libc-dev} (BSD-2-Clause AND BSD-3-Clause) [installed]
libcrypto3-3.1.3-r0 x86_64 {openssl} (Apache-2.0) [installed]
libssl3-3.1.3-r0 x86_64 {openssl} (Apache-2.0) [installed]
musl-1.2.4-r1 x86_64 {musl} (MIT) [installed]
musl-utils-1.2.4-r1 x86_64 {musl} (MIT AND BSD-2-Clause AND GPL-2.0-or-later) [installed]
scanelf-1.3.7-r1 x86_64 {pax-utils} (GPL-2.0-only) [installed]
ssl_client-1.36.1-r2 x86_64 {busybox} (GPL-2.0-only) [installed]
zlib-1.2.13-r1 x86_64 {zlib} (Zlib) [installed]
```

neheb:
oh interesting.

```
busybox diff -u /dev/null /dev/null
```

works.

benoit-pierre (approved):

neheb:
@benoit-pierre you mentioned MSVC. Where does diff come from in that case?

benoit-pierre:
CR/LFs in outputs.

benoit-pierre:
No sure if that's what you're asking. The diff executable itself?

neheb:
yes.

neheb:
ah it probably comes with git for windows.

neheb:
@benoit-pierre funny. I hit this when cross compiling MinGW. Not that it matters really. Code should be fixed.

PR #1250, opened 2023-10-15
freeglut: add missing dependencies

neheb:
ping @kusma


kusma:
Looks reasonable. 

I suspect you might need this for the CI, though:

```diff
diff --git a/ci_config.json b/ci_config.json
index ba7a533..14c31b5 100644
--- a/ci_config.json
+++ b/ci_config.json
@@ -105,6 +105,7 @@
     ],
     "debian_packages": [
       "libgl-dev",
+      "libopengl-dev",
       "libglu1-mesa-dev",
       "libxi-dev"
     ]
```

Not 100% sure why we have both `libgl-dev` and `libopengl-dev`, but I suspect the reason is somehow related to GLVND...

PR #1251, opened 2023-10-15
openssl: add bash dependency

neheb:
generator.sh relies on bash. Gives a better error message when not found.


neheb:
@nazar-pc should this be rewritten in python?

eli-schwartz:
It "only runs on Linux", so I see very little point in rewriting the generator.sh script in python for portability.

But if you're going to check for dependencies I'd check for perl and GNU Make too...

subprojects/packagefiles/openssl/generator.py
```diff
@@ -0,0 +1,69 @@
...
+os.environ['LANG'] = 'C'
+subprocess.run(["make", "-C", "config"])
+
+# Copy generated files back into the correct place
+find_cmd = 'mkdir -p ../../../generated-$(dirname "$1"); cp "$1" ../../../generated-"$1"'
+subprocess.run(["find", "config/archs", "-name", "meson.build", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
+subprocess.run(["find", "config/archs", "-name", "*.asm", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
+subprocess.run(["find", "config/archs", "-name", "*.c", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
+subprocess.run(["find", "config/archs", "-name", "*.h", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
+subprocess.run(["find", "config/archs", "-iname", "*.s", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
```
eli-schwartz, line 53: I still do not understand the purpose of this PR.

But this right here.

This is not rewriting the generator from bash into python. This is just... no.

eli-schwartz (dismissed):

subprojects/packagefiles/openssl/generator.py
neheb: A proper rewrite would be much more complex to deal with globbing and whatnot.

eli-schwartz: Ok?

So what exactly is wrong with the bash version.

neheb: Alpine CI is lightweight and does not come with bash by default. It's probably not a good idea to mandate bash.

eli-schwartz: Install a normal userland on the alpine CI then? idk what the problem is.

bash is more lightweight than python, that's for sure.

neheb: but python is needed for meson anyway.

neheb: @eli-schwartz updated version and ported more of these calls to python.

eli-schwartz: There's still a bunch of find -execs in there...

eli-schwartz:
I'd feel a lot more comfortable if @nazar-pc okay'ed this, as the person most likely to be updating it in the future.

subprojects/packagefiles/openssl/generator.py
neheb: I came up with

```
+def copy_files(extension):
+    for root, _, files in os.walk("config/archs"):
+        for file in files:
+            if file.lower().endswith(extension.lower()):
+                source_path = os.path.join(root, file)
+                relative_path = os.path.relpath(source_path, search_dir)
+                destination_dir = os.path.join("../../../generated", os.path.dirname(relative_path))
+                destination_path = os.path.join(destination_dir, file)
+
+                os.makedirs(destination_dir, exist_ok=True)
+                shutil.copy(source_path, destination_path)
+
 # Change to the directory of the Bash script
 script_dir = os.path.dirname(os.path.abspath(__file__))
 os.chdir(script_dir)
@@ -47,12 +59,11 @@ os.environ['LANG'] = 'C'
 subprocess.run(["make", "-C", "config"])

 # Copy generated files back into the correct place
-find_cmd = 'mkdir -p ../../../generated-$(dirname "$1"); cp "$1" ../../../generated-"$1"'
-subprocess.run(["find", "config/archs", "-name", "meson.build", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
-subprocess.run(["find", "config/archs", "-name", "*.asm", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
-subprocess.run(["find", "config/archs", "-name", "*.c", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
-subprocess.run(["find", "config/archs", "-name", "*.h", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
-subprocess.run(["find", "config/archs", "-iname", "*.s", "-exec", "sh", "-c", find_cmd, "_ignored", "{}", ";"])
+copy_files("meson.build")
+copy_files(".asm")
+copy_files(".c")
+copy_files(".h")
+copy_files(".s")
```

I don't think it's equivalent.

nazar-pc:
If we want to make it usable on Windows, this makes some sense, otherwise not really. This is exactly that kind of stuff Bash is well suited for: running a series of simple CLI commands one after another.

I don't really know Python, so maintaining Bash would probably be easier and it is arguably nicer to read than a bunch of `subprocess.run`. And except those enjoying doing OpenSSL upgrades, there is little value in this since CI will run this stuff once on Linux and publish all the necessary files as a wrap, so shouldn't really be necessary for most people to run this manually, let alone on Windows.

While I appreciate the effort, I'm not convinced this is better or necessary.

neheb:
Goal is Alpine compatibility. The Python code currently is not much different than the bash script.

nazar-pc:
> Goal is Alpine compatibility.

`apk add bash`?

eli-schwartz:
Yeah but the original PR simply installed bash on Alpine (and the PR title still says that's what it does). So it's not clear why the goal suddenly became to rewrite it in python (especially if that causes it to be less easy for people working on updating the openssl wrap to do so).

"Install the bash package" is pretty compatible with Alpine, no?

Tachi107:
> Goal is Alpine compatibility

What about removing the script's bashisms so that it runs on Alpine's sh? Doesn't look that hard.

eli-schwartz:
> Doesn't look that hard.

Do tell.

neheb:
IIRC, the python script ran faster when I tested it. May have been a fluke.

benoit-pierre:
Or just install bash?

eli-schwartz:
Still waiting to hear what the POSIX sh version of the BASH_SOURCE bashism is.

Tachi107:
> Still waiting to hear what the POSIX sh version of the BASH_SOURCE bashism is.

Not exactly as useful as `BASH_SOURCE`, but I've used `$0` in the past and in worked for me

subprojects/packagefiles/openssl/generator.py
```diff
@@ -0,0 +1,74 @@
...
+# Clone the upstream OpenSSL repository
+shutil.rmtree("openssl")
+subprocess.run(["git", "clone", "--depth", "1", "--branch", "openssl-" + openssl_version, "https://github.com/openssl/openssl.git"])
+
+# Remove the config/archs directory
+shutil.rmtree("config/archs")
+
+# Run the make command with LANG set to C
+os.environ['LANG'] = 'C'
+subprocess.run(["make", "-C", "config"])
```
klokik, line 47:
```suggestion
subprocess.run(["make", "-C", "config"], env={"LANG": "C"})
```

subprojects/packagefiles/openssl/generator.py
```diff
@@ -0,0 +1,74 @@
...
+    shutil.rmtree(d)
+
+# 32-bit s390x supported in Meson
+shutil.rmtree("../../../generated-config/archs/linux32-s390x")
+
+# Remove build info files
+for d in glob.glob("../../../generated-config/archs/*/asm_avx2"):
+    shutil.rmtree(d)
+for d in glob.glob("../../../generated-config/archs/*/*/crypto/buildinf.h"):
+    os.remove(d)
```
klokik, line 68: maybe go over a list of globs and remove everything within in a nested loop

subprojects/packagefiles/openssl/generator.py
```diff
@@ -0,0 +1,74 @@
...
+
+if openssl_version is None:
+    with open("../../openssl.wrap", encoding="utf-8") as wrap_file:
+        for line in wrap_file:
+            if "directory = " in line:
+                openssl_version = line.strip().split()[-1]
+                break
+
+# Clone the Node.js repository
+subprocess.run(["git", "clone", "--depth", "1", "-b", node_version, "https://github.com/nodejs/node.git"])
```
klokik, line 24:
```suggestion
git = find_program('git')
subprocess.run([git, "clone", "--depth", "1", "-b", node_version, "https://github.com/nodejs/node.git"])
```

subprojects/packagefiles/openssl/generator.py
neheb: not sure I follow.

subprojects/packagefiles/openssl/generator.py
neheb: wait a minute, this is horrible advice.

python has no find_program.

subprojects/packagefiles/openssl/generator.py
neheb: python has find_program?

benoit-pierre: No, but there's [`shutil.which`](https://docs.python.org/3/library/shutil.html?highlight=which#shutil.which), as suggested in [`subprocess.Popen`'s documentation](https://docs.python.org/3/library/subprocess.html?highlight=subprocess%20run#subprocess.Popen).

eli-schwartz: The subprocess.Popen documentation is a bit weird except in the sense that you may want to invoke APIs which do not search PATH, i.e. direct use of `os.exec*()` functions that do not have a "p" in them. OR if by "maximum reliability" one means "reliable on non-POSIX platforms like Windows, the most common platform around". Which... fair enough lol.

I do find it somewhat baffling however that reading the subprocess.py source code it appears Python will refuse to use the optimized path of posix_spawn if given a command without a dirname (a fact totally unrelated to whether it is an absolute path).

eli-schwartz: Anyway the only point of `find_program('make')` would be doing it ***in meson.build***, if one wanted to use meson.build to check which tools are available rather than having generator.sh fail on its own.

PR #1256, opened 2023-10-16
libusb: fix printf for mingw

neheb:


neheb:
bah. Upstream needs fixing.

PR #1257, opened 2023-10-16
CI: add alpine

neheb:
Alpine uses musl, which makes an interesting opportunity to test packages for compilation.

Also, ninja is not used but samurai instead.

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -24,6 +24,25 @@ jobs:
         run: |
           ./tools/sanity_checks.py
 
+  Alpine:
+    runs-on: ubuntu-latest
+    container:
+      image: alpine
+    steps:
+      - name: Install packages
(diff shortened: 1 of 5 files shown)


neheb:
https://repology.org/project/samurai/versions

Debian is missing oddly enough.

.github/workflows/sanity_checks.yml
```diff
@@ -24,6 +24,25 @@ jobs:
...
           ./tools/sanity_checks.py
 
+  Alpine:
+    runs-on: ubuntu-latest
+    container:
+      image: alpine
+    steps:
+      - name: Install packages
+        run: |
+          apk add cmake g++ gcc git linux-headers pkgconf py3-pip samurai sudo
```
neheb, line 34: sudo here is needed because of sanity_checks.py. Maybe replace with su -c?

PR #1260, opened 2023-10-18
libpsl: add 0.21.2

hrxi:


hrxi:
Windows compatibility of the psl tool (not required for the library) was fixed in https://github.com/rockdaboot/libpsl/pull/193, but there has been no release since then.

Does the no-patching-upstream-projects rule preclude packaging a working version of libpsl for Windows, short of a new release?

Or am I allowed to simply not compile the tool?

neheb:
Patched meson.build files are allowed.

neheb:
Btw localtime_r is available with MinGW. Probably better to add localtime_s support upstream though.

hrxi:
> Probably better to add localtime_s support upstream though.

They already did that upstream, but they haven't added a new release.

hrxi:
Works everywhere except clang-cl, I opened an issue with upstream.

PR #1263, opened 2023-10-20
Protobuf: make default include path available to protoc when used from submodule

### LINKED ISSUE #1259: Feature request: provide an import path to be used with `protoc` for compiling *.proto files
Protobuf ships with a bunch of standard *.proto files (any, descriptor, empty, timestamp) that are expected to be found, if imported within another proto file, during `protoc` executable invocation.

Protoc accepts a list of `--proto_path pathXXX` arguments to specify import search locations. Unfortunately if using protobuf as a subproject, one can not expect to find these at standard locations (like /usr/share/protobuf/google/....). Thus cannot use protoc from subproject to generate sources at meson project build time.

I didn't found any example within existing wrapdb files on how it may be specified elsewhere, outside of `protobuf` packagefiles itself, which is sell aware of its sources locations.

Suggestion: populate `protobuf` dependency variables with a path to this location.

klokik:
Closes #1259

Protoc populates default `proto_path`s by searching for directories named `include` in its surroundings

This adds a dependency for `protoc` on creating a symlink to make such directory available.

Because FS module in meson does not provide a mechanism to create symlinks, a python script is used for this job.


subprojects/packagefiles/protobuf/src/symlink.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+
+from pathlib import Path
```
neheb, line 3: I'd run with pylint

klokik: Done
```
$ pylint symlink.py 

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

PR #1266, opened 2023-10-22
gumbo-parser: fix compilation with newer clang

neheb:
Newer clang errors on implicit functions.

```
subprojects/gumbo-parser/examples/get_title.p/get_title.c.o -c ../../home/runner/work/wrapdb/wrapdb/subprojects/gumbo-parser/examples/get_title.c
2023-10-22T02:21:12.8570068Z ../../home/runner/work/wrapdb/wrapdb/subprojects/gumbo-parser/examples/get_title.c:28:12: error: call to undeclared function 'fileno'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
2023-10-22T02:21:12.8572088Z    28 |   int fd = fileno(fp);
2023-10-22T02:21:12.8572738Z       |            ^
```

ping @benoit-pierre


subprojects/packagefiles/gumbo-parser/examples/meson.build
```diff
@@ -19,5 +19,6 @@ foreach _src : [
     _src.split('.')[0],
     _src,
     dependencies: gumbo_dep,
+    c_args: '-D_DEFAULT_SOURCE',
```
benoit-pierre, line 22: Wouldn't `_POSIX_C_SOURCE` be better?

neheb: seems to work. Fixed.

PR #1268, opened 2023-10-24
Add DirectXMath wrap

seungha-yang:
Adding DirectXMath (header only library) wrap.
See also https://github.com/microsoft/DirectXMath

### DIFF
--- a/subprojects/packagefiles/directxmath/meson.build
+++ b/subprojects/packagefiles/directxmath/meson.build
@@ -1,0 +1,29 @@
+project('directxmath', 'cpp', version: '3.1.8')
+
+incdirs = include_directories('Inc')
+
+directxmath_headers = files(
+  'Inc/DirectXCollision.h',
+  'Inc/DirectXCollision.inl',
+  'Inc/DirectXColors.h',
+  'Inc/DirectXMath.h',
+  'Inc/DirectXMathConvert.inl',
+  'Inc/DirectXMathMatrix.inl',
+  'Inc/DirectXMathMisc.inl',
+  'Inc/DirectXMathVector.inl',
+  'Inc/DirectXPackedVector.h',
+  'Inc/DirectXPackedVector.inl',
+)
+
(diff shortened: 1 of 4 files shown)


subprojects/packagefiles/directxmath/meson.build
```diff
@@ -0,0 +1,7 @@
+project('directxmath', 'cpp', version: '3.1.8')
+
+incdirs = include_directories('Inc')
+
+directxmath_dep = declare_dependency(
```
seungha-yang, line 5: probably I should install headers with pkg-config file

subprojects/packagefiles/directxmath/meson.build
```diff
@@ -0,0 +1,7 @@
+project('directxmath', 'cpp', version: '3.1.8')
```
nirbheek, line 1: Where is this version `3.1.8` coming from? The release tag says `3.18`: https://github.com/microsoft/DirectXMath/releases/tag/dec2022

seungha-yang: uh, it seems to be inconsistent with version specified in cmake file then https://github.com/microsoft/DirectXMath/blob/dec2022/CMakeLists.txt#L6

subprojects/packagefiles/directxmath/meson.build
seungha-yang: done

subprojects/packagefiles/directxmath/meson.build
seungha-yang: I guess it might be mistake in their cmake file but I'd follow the cmake project version

seungha-yang: Filed an issue https://github.com/microsoft/DirectXMath/issues/169

nirbheek (approved):
LGTM, but I'll wait 24h to see if someone else has comments before merging since this edits `tools/sanity_checks.py`.

PR #1269, opened 2023-10-25
utf8proc: update to 2.9.0

benoit-pierre:
Additionally, fix the install rules, add some tests, and fix GCC/Clang symbols visibility.


benoit-pierre:
I have conditioned the tests on curl being present, but since not all tests require downloading data, maybe it would make more sense to only enable those conditionally?

subprojects/packagefiles/utf8proc/meson.build
```diff
@@ -31,3 +34,58 @@ pkg.generate(
 )
 
 install_headers('utf8proc.h')
+
+curl = find_program('curl', required: get_option('tests'))
+
+if get_option('tests').require(curl.found()).allowed()
```
neheb, line 41: this looks like a more complicated version  of

if curl.found()

subprojects/packagefiles/utf8proc/meson.build
```diff
@@ -31,3 +34,58 @@ pkg.generate(
...
+      _test,
+      executable(
+        _test,
+        'test/tests.c',
+        'test/@0@.c'.format(_test),
+        dependencies: utf8proc_dep,
+      ),
+      args: _args,
+      depends: _depends,
+      workdir: meson.current_build_dir(),
```
neheb, line 88: is this not default?

benoit-pierre: From my experience, you can't rely on the working directory if not specified: it will vary depending on how the project is compiled (e.g. as subproject or not).

subprojects/packagefiles/utf8proc/meson.build
benoit-pierre: The goal is to be able to use `--auto-features=disabled`, as always.

PR #1273, opened 2023-10-27
sqlpp11: Initial WrapDB files for sqlpp11

j-carl:


subprojects/packagefiles/sqlpp11/meson_options.txt
```diff
@@ -0,0 +1,5 @@
+option('sqlite3_connector', type : 'boolean', value : false, description : 'Build SQLite3 connector')
```
neheb, line 1: why not use feature options?

j-carl: I want to give the user the power to select which database connector he wants to use and not automatically enable them. It could lead to that all database connectors are enabled.

neheb: Sounds like this should be a combo option then.

j-carl: No, not really. The user should still be able to select more then one database connector. A use-case would be transferring data from SQLite3 to PostgreSQL.

neheb:
vcpkg has the proper stuff. Unfortunately, it needs proper CI setup.

Alternative would be to add libmariadb and libpostgresql to wrapdb. Or just use feature options.

PR #1275, opened 2023-11-02
[Bug report only] Unable to extract source code from zip file, when `patch_directory` is defined.

antonysigma:
**Observation:**

After running `meson setup -Dwraps=freetype1 build/`, the configuration step failed. Only contents of the `meson.build` files are found in the path `subproject/freetype-1.4/` directory. I expect the source code to be present, followed by the patched files.

**Environment:**

Meson version 1.2.3
OS: 22.6.0 Darwin Kernel Version 22.6.0
Ninja 1.11.1.git.kitware.jobserver-1


benoit-pierre:
That's because the source zip is missing a common root directory, add `lead_directory_missing = true` to the wrap.

antonysigma:
> That's because the source zip is missing a common root directory, add `lead_directory_missing = true` to the wrap.

@benoit-pierre Oh silly me. Let me retract this PR, as well as the bug report.

subprojects/freetype1.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = freetype-1.4
+source_url = https://sourceforge.net/projects/gnuwin32/files/freetype/1.4/libttf-1.4-src.zip/download#
+source_filename = libttf-1.4-src.zip
+source_hash = 533341269bdded2619622222e79af8af066b98b9183dbafda663907daabef309
+
+patch_directory = freetype1
+
+[provide]
+freetype1 = freetype_dep
```
benoit-pierre, line 10: The value must match the name of the variable used in `meson.build` (currently `ttf_dep`).

antonysigma: Fixed and `git push`ed. 

How about the license field in `project()` ? It is a custom license with a list of active Patents.

neheb: Needs a rebase against master.

PR #1276, opened 2023-11-05
libsignal-protocol-c: add 2.3.3

hrxi:


hrxi:
The project doesn't use `__declspec(dllexport)` for the library exports. Upstream is archived. Short of patching the project ourselves, this is not buildable in MSVC. This is not allowed for wrapdb AFAIUI, so it should be good to go like it currently is.

benoit-pierre:
You could use a def file; e.g. for gumbo-parser: https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/gumbo-parser/src/gumbo.def.

hrxi:
Permalink: https://github.com/mesonbuild/wrapdb/blob/e3f96a5c8f42458dc69fcb66b4f3dd842f578168/subprojects/packagefiles/gumbo-parser/src/gumbo.def

That works. It does not count as modifying the project? The upstream CMake build doesn't support MSVC, due to the missing attributes.

benoit-pierre:
> Permalink: https://github.com/mesonbuild/wrapdb/blob/e3f96a5c8f42458dc69fcb66b4f3dd842f578168/subprojects/packagefiles/gumbo-parser/src/gumbo.def
> 
> That works. It does not count as modifying the project?

It's already used more than a dozen of times in `subprojects/packagefiles`.

neheb:
Yeah it's fine to do. Recommended actually.

hrxi:
I think the rest of the build problems are inherent to libsignal-protocol-c.

Can be merged now.

PR #1277, opened 2023-11-05
qrencode: add 4.1.1

hrxi:


neheb:
weird that clang-cl builds but not cl...

hrxi:
It's a problem of missing `__declspec(dllexport)`, I believe.

neheb:
Yeah need to add a def file.

hrxi:
Also opened an upstream PR fixing MSVC support properly: https://github.com/fukuchi/libqrencode/pull/214.

PR #1279, opened 2023-11-06
nghttp2: update to 1.58.0

hrxi:


subprojects/packagefiles/nghttp2/meson.build
```diff
@@ -20,9 +22,53 @@ configuration.set('PROJECT_VERSION_MAJOR', meson.project_version().split('.')[0]
...
-configuration.set('includedir', get_option('includedir'))
-configuration.set('libdir', get_option('libdir'))
-configuration.set('prefix', get_option('prefix'))
+
+if cc.has_function_attribute('noreturn')
+    configuration.set('HINT_NORETURN', '__attribute__((noreturn))')
+else
+    configuration.set('HINT_NORETURN', '')
+endif
+if host_machine.system() == 'windows'
```
neheb, line 31: only on 32-bit. 64-bit is __int64 IIRC.

hrxi: Windows doesn't have `ssize_t`. `nghttp2` wants `int`:

https://github.com/nghttp2/nghttp2/blob/d07a3d2a24688fe125cdc5beda72361394dab9e1/CMakeLists.txt#L291-L292

https://github.com/nghttp2/nghttp2/blob/d07a3d2a24688fe125cdc5beda72361394dab9e1/cmakeconfig.h.in#L4-L5

hrxi:
I'd like to patch the project so that it builds on MSYS2-CLANG*. Upstream PR is here: https://github.com/nghttp2/nghttp2/pull/1997.

Since patching the .c file is not allowed AFAIK, this PR can be merged with failing checks instead.

PR #1280, opened 2023-11-06
curl: add CURL_STATICLIB to curl_dep when needed

Tachi107:
The public `curl/curl.h` header requires `CURL_STATICLIB` to be defined by user code to properly define `CURL_EXTERN` to the correct value. Failing to do so causes linking failures when linking against a static libcurl on Windows, since `CURL_EXTERN` gets wrongly defined to `__declspec(dllimport)` instead of the empty string. You can see this in [curl/curl.h#L129](https://github.com/curl/curl/blob/60359ad5049c42f333ac179a58cda87d6e4143a0/include/curl/curl.h#L129).

This patch adds `-DCURL_STATICLIB` to the set of compiler args of both curl_dep and the generated pkg-config file when performing a static build.

Cc: @benoit-pierre


neheb:
something tells me we should add static CI for Windows.

benoit-pierre (approved):

benoit-pierre:
> something tells me we should add static CI for Windows.

Make it a cross-mingw build, kill 2 birds with one stone.

neheb:
funny, I tried that once. The main issue I ran into was properly declaring exe_wrapper. I don't know if it can be declared with environment variables. I also wonder about pkg-config. Ubuntu doesn't have that for MinGW IIRC.

benoit-pierre:
AUR as one: https://aur.archlinux.org/cgit/aur.git/tree/mingw-pkgconfig.sh?h=mingw-w64-pkg-config.

neheb:
hrm biggest problem with static run turns out to be libpng.

PR #1282, opened 2023-11-08
nowide: fix mistake in test

neheb:
Fixes:

NameError: name 'stdout' is not defined

ran through pylint too.


neheb:
ugh I think the error is because Alpine is being ran as a container.

PR #1283, opened 2023-11-10
openssl: Make it build with MSVC

### LINKED ISSUE #181: openssl wrap doesn't build on windows (vs via ninja)
Hi,

```
PS C:\Users\anfreund\src\pg-meson\build-ninja-x64> ninja .\subprojects\openssl-1.1.1l\openssl.exe
[1/1] Linking target subprojects/openssl-1.1.1l/openssl.exe
FAILED: subprojects/openssl-1.1.1l/openssl.exe subprojects/openssl-1.1.1l/openssl.pdb
"link" @subprojects/openssl-1.1.1l/openssl.exe.rsp
LINK : fatal error LNK1181: cannot open input file 'subprojects\openssl-1.1.1l\crypto.lib'
```
As far as I can tell the reason for the failure to build is that the buildsystem contained in the wrap doesn't take care of exporting symbols. Due to that link.exe doesn't produce an import library, which then causes this issue.

CC: @nazar-pc

Regards,

Andres

hrxi:
MSVC needs a way of knowing which symbols are supposed to be exported, otherwise it does not create a DLL. This means the wrap was previously unusable as a shared library.

The OpenSSL build process does that by passing a `.def` file to the linker. This reimplements OpenSSL's `util/mkdef.pl` script to make the meson build do the same as the normal build process.

Fixes #181.


neheb:
You could implement something similar to https://github.com/mesonbuild/wrapdb/pull/1251 so that Alpine builds.

nazar-pc (commented):
This looks potentially useful, but can you fill PR description, please?

Try to explain what was the issue that you have encountered, what were the ways to solve it, which approach did you take and why, etc.?

I'm not using Windows, I don't deal with MSVC unless I have to and it is never fun, I am rarely dealing with build systems or even Meson. So it would be really helpful for code dumps to be accompanied with nice and detailed description of changes, it seriously helps with code reviews.

eli-schwartz:
The `*.def` file is basically the Microsoft equivalent of `-Wl,--version-script=/path/to/libfoo.map`, i.e. https://sourceware.org/binutils/docs/ld/VERSION.html

They are actually used by any compiler toolchain on Windows, not just MSVC but also when you compile on Cygwin or MSYS2 using a mingw GCC.

So in theory this is a very reasonable change to make, although I haven't looked at all into how the generator script works or whether it emits the same symbol exports mapping that openssl upstream uses on Windows-based targets.

hrxi:
> This looks potentially useful, but can you fill PR description, please?
> 
> Try to explain what was the issue that you have encountered, what were the ways to solve it, which approach did you take and why, etc.?
> 
> I'm not using Windows, I don't deal with MSVC unless I have to and it is never fun, I am rarely dealing with build systems or even Meson. So it would be really helpful for code dumps to be accompanied with nice and detailed description of changes, it seriously helps with code reviews.

Added some description to the commit/PR, does that work?

hrxi:
> They are actually used by any compiler toolchain on Windows, not just MSVC but also when you compile on Cygwin or MSYS2 using a mingw GCC. So I would generally agree that this doesn't actually seem like a "make it build with MSVC", just a "make it build _better_ on Windows".

AFAIK this is only really necessary for MSVC. AFAIK, gcc can just treat any symbol of global visibility as exported. MSVC refuses to do so and needs a `__declspec(dllexport)` annotation or a .def file. Of course, GCC or clang *can* consume a .def file or `__declspec(dllexport)` annotations on Windows.

> (Note as well that you only need a defs file for producing shared `*.dll`s, just like you only need a unix ld version script when producing `*.so`s.)

You're probably aware, but because of the ambiguous formulation: You don't need ld version scripts to produce a shared object. Of course, you never need such a script when linking statically.

hrxi:
> You could implement something similar to #1251 so that Alpine builds.

Alpine seems pretty orthogonal to this PR. Should I open another PR trying to fix Alpine?

eli-schwartz:
> AFAIK this is only really necessary for MSVC. AFAIK, gcc can just treat any symbol of global visibility as exported. MSVC refuses to do so and needs

Correct. As I said, it's used, not mandatory. GCC defaults to the "same" thing as it does on Unix, and defaults undetermined symbol types to "is public". IIRC this is only when the alternative is "the DLL has zero symbols, that's obviously wrong, so this must not be annotating the difference and we shall simply give you all the symbols and let you deal with it".

> You're probably aware, but because of the ambiguous formulation: You don't need ld version scripts to produce a shared object. Of course, you never need such a script when linking statically.

Read: in the event your project has an ld version script or a .def file, you only need to use it to obtain the intended functionality, when producing a shared object. You can refrain from using it at any time, but the consequence is that you don't obtain the intended functionality.

(And the intended functionality isn't just exposing symbols that would otherwise be hidden, it's also hiding symbols that would otherwise be exposed, or giving a symbol a versioned name so it can coexist with other versions of the same shared library. Hiding symbols that would otherwise be exposed both avoids accidentally breaking ABI compat, and also makes libraries smaller and faster.)

eli-schwartz:
> Alpine seems pretty orthogonal to this PR. Should I open another PR trying to fix Alpine?

It's relevant in the sense that this PR has failing CI. The problem, despite attempts to argue to the contrary, is that Alpine needs to have GNU bash installed but does not. GNU bash is perfectly reasonable to require, and the openssl wrap does require it, so let's just install it lmao. No need for wild goose chases down the road of "port a bash script to a python script that just repeatedly runs `subprocess.run(shell_command)`".

hrxi:
> Read: in the event your project has an ld version script or a .def file, you only need to use it to obtain the intended functionality, when producing a shared object. You can refrain from using it at any time, but the consequence is that you don't obtain the intended functionality.

Yes, precisely. :)

hrxi:
> It's relevant in the sense that this PR has failing CI.

I do not want to fix the Alpine CI currently. If that makes the PR unmergeable, feel free to close it.

eli-schwartz:
It doesn't make the PR unmergeable.

@nazar-pc as the PR has been updated, what do you think?

nazar-pc (approved):
I will not pretend to understand how this works (thanks for PR description!), but some comments in the code would be helpful. Otherwise if it works and doesn't break existing usage (I only used static builds before) I think it is better to have this than not.

hrxi:
Fixed the unused variable. Should be good to merge now.

PR #1285, opened 2023-11-14
openssl: Fix alpine CI

hrxi:


neheb:
```
Can't locate Text/Template.pm in @INC (you may need to install the Text::Template module) (@INC contains: /__w/wrapdb/wrapdb/subprojects/openssl-3.0.12/node/deps/openssl/config/../openssl/util/perl/OpenSSL ../openssl /usr/local/lib/perl5/site_perl /usr/local/share/perl5/site_perl /usr/lib/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib/perl5/core_perl /usr/share/perl5/core_perl) at ./generate_headers.pl line 6.
```

interestingly enough, https://github.com/mesonbuild/wrapdb/pull/1251 does not suffer from this. It's NACKed though.

eli-schwartz:
> interestingly enough, #1251 does not suffer from this. It's NACKed though.

Considering the code that failed is:
```sh
make -C config
```
 and the PR you refer to also runs:
```python
subprocess.run(["make", "-C", "config"])
```

I strongly suspect there must be a deeper reason than "does not suffer from this".

hrxi:
I think I'd prefer not to be the focus of some sort of disagreement I don't care about.

PR #1292, opened 2023-11-18
fmt: fix build with Cygwin

neheb:
Cygwin requires gnu compiler flag.


neheb:
That is,...hmmm.

So passing _GNU_SOURCE or _POSIX_C_SOURCE doesn't work. The compiler flag must be gnu. No idea how to fix.

PR #1293, opened 2023-11-19
Add c-ares unit tests, export symbols on MSVC required by unit tests

klokik:
This will help to address linking issues that may be encountered by users in different configurations


neheb:
Unit tests require private symbols?

klokik:
Unfortunatelly yes. Or a separate static library that will be used for the test-suite executable.

neheb:
I think the latter is more widely done here.

klokik:
https://github.com/c-ares/c-ares/blob/4982f76a2f7a0d629c32370427a689584443e3ea/CMakeLists.txt#L49

klokik:
Maybe, but primarily l want this unit test here to have an easy-to-reach sample user of the library, to make sure that it does not have unexpected linking issues somewhere. Like the previous version was causing undefined symbols because of missing `compile_args` provided by `cares_dep` in the case of the static build.

benoit-pierre:
Do the tools require private symbols too? If not, that can be used to test linking with the final library, while linking the tests with a static version.

klokik:
Yes: at least on windows `tools` use an internal wrapper implementing `strcasecmp()`.

klokik:
How bad is it exporting these internal symbols only if `tests` flags is set?

benoit-pierre:
It's kinda icky to get a different library whether you enabled tests or not.

klokik:
All these functions are properly prefixed so unlikely to cause collisions. But I dread to think of using `--export-all-symbols` on windows+gcc

klokik:
Rebased, that was a pain due to a sneaky formatting change. Thanks @neheb :upside_down_face: 

No hacky symbols exporting this time: building static lib for tools and unit tests if required.

neheb:
:)

Should have waited before merging. Oh well.

PR #1295, opened 2023-11-20
msgpackc-cxx: rename dependency name to msgpack-cxx and add new options

stephanlachnit:
Since version 6.0.0 `msgpackc-cxx` is now found via CMake with `msgpack-cxx` (see https://github.com/msgpack/msgpack-c/pull/1054).

This PR does the following things:
- Rename the dependency to `msgpack-cxx` while leaving the wrap name untouched
- Make the `boost` option functional with the `MSGPACK_NO_BOOST` compile definition
- Add an option for the `std::variant` adaptor (since version 6.1.0)
- Fix indent size to follow editorconfig specifications
- Update list of tests


stephanlachnit:
Regarding the failing sanity check: how should the changed dependency name be handled? I purposely removed `msgpackc-cxx` from the provides in the wrap file, as this is a dependency name used only for version <6.0.0. Should it be removed from `releases.json`?

subprojects/packagefiles/msgpackc-cxx/meson.build
```diff
@@ -1,27 +1,39 @@
...
+  version: '6.1.0',
+  meson_version: '>=0.47.0',
 )
 
 args = []
+deps = []
+
+boost_dep = dependency('boost', required: get_option('boost'))
+if boost_dep.found()
+  deps += [boost_dep]
```
neheb, line 12: why not just pass this to dependencies directly?

neheb:
> Regarding the failing sanity check: how should the changed dependency name be handled? I purposely removed `msgpackc-cxx` from the provides in the wrap file, as this is a dependency name used only for version <6.0.0. Should it be removed from `releases.json`?

yeah.

subprojects/packagefiles/msgpackc-cxx/meson.build
stephanlachnit: Ah, I would then only check do
```meson
if not boost_dep.found()
  args += ['-DMSGPACK_NO_BOOST']
endif
```
right?

neheb: yeah. passing not found dependencies to dependencies: is fine. they won't be used/linked.

subprojects/packagefiles/msgpackc-cxx/meson.build
```diff
@@ -1,27 +1,36 @@
...
+
+boost_dep = dependency('boost', required: get_option('boost'))
+if not boost_dep.found()
+  args += ['-DMSGPACK_NO_BOOST']
+endif
+
 if get_option('x3_parse')
-    args += ['-DMSGPACK_USE_X3_PARSE']
+  if not boost_dep.found()
+    error('Boost is required for X3 parse')
```
neheb, line 16: now this is interesting. why is there a boot option? Sounds like you could replace get_option('boost') with get_option('x3_parse')

subprojects/packagefiles/msgpackc-cxx/meson.build
```diff
@@ -1,27 +1,36 @@
...
+  compile_args: args,
+  include_directories: include_directories('include'),
+  dependencies: boost_dep,
 )
 
-if meson.version().version_compare('>=0.54.0')
-    meson.override_dependency('msgpackc-cxx', msgpackc_cxx_dep)
-endif
-
 if get_option('tests')
```
neheb, line 31: My preference is to subdir('test') unconditionally and inside if dependencies are missing, pass subdir_done()

subprojects/packagefiles/msgpackc-cxx/meson.build
stephanlachnit: No, because there are Boost components enabled by default independent of X3.

subprojects/packagefiles/msgpackc-cxx/meson.build
stephanlachnit: Sounds reasonable

subprojects/packagefiles/msgpackc-cxx/meson.build
neheb: So boost is mandatory?

subprojects/packagefiles/msgpackc-cxx/meson.build
stephanlachnit: Wait no sorry, I don't get the `subdir_done()` part. The reason why I check for this is that this cannot work if `-DMSGPACK_NO_BOOST` is one of the compile arguments. The only scenario here is that someone uses `boost=disabled` and `tests=true`, which will not work.

subprojects/packagefiles/msgpackc-cxx/meson.build
stephanlachnit: For most users, yes. These components can be disabled (which is experimental), but that would be a very specific use case (e.g. when someone wants to avoid Boost at all costs).

subprojects/packagefiles/msgpackc-cxx/meson.build
neheb: Simplest way is to do:

```
if get_option('tests') and not boost_dep.found()
  subdir_done() or error()
endif
```

in tests/meson.build. Although it might make more sense to do

```
tboost_dep = dependency('boost', modules: ['unit_test_framework', 'system'], required: get_option('tests'))
```

and match subdir_done() on that.

stephanlachnit: > Simplest way is to do:
> 
> ```
> if get_option('tests') and not boost_dep.found()
>   subdir_done() or error()
> endif
> ```
> 
> in tests/meson.build.

Just so that I get it right, that would be
```meson
# only run tests if required
if not get_option('tests')
  subdir_done()
endif

# ensure boost was not disabled
if not boost_dep.found()
  error('Boost is required for tests')
endif
```

> Although it might make more sense to do
> 
> ```
> tboost_dep = dependency('boost', modules: ['unit_test_framework', 'system'], required: get_option('tests'))
> ```
> 
> and match subdir_done() on that.

No, because that is not enough. We need to check that `boost` has not been disabled (due to the compile definition), so this needs to happen before.

neheb: Sure. check for boost_dep first and then that.

stephanlachnit: This only really makes sense if `tests` is a feature, not a bool. So I changed that, it now detects automatically if the tests can be build or not. Also allows for the CI to run on Windows.

neheb (approved):

PR #1296, opened 2023-11-21
pv: add version 1.8.5

benoit-pierre:
Upstream: https://codeberg.org/a-j-wood/pv/.

### DIFF
--- a/subprojects/pv.wrap
+++ b/subprojects/pv.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = pv-1.8.5
+source_url = https://codeberg.org/a-j-wood/pv/releases/download/v1.8.5/pv-1.8.5.tar.gz
+source_hash = d22948d06be06a5be37336318de540a2215be10ab0163f8cd23a20149647b780
+source_filename = pv-1.8.5.tar.gz
+patch_directory = pv
+
+[provide]
+program_names =
--- a/ci_config.json
+++ b/ci_config.json
@@ -648,6 +648,21 @@
       "pkgconf:tests=false"
     ]
   },
+  "pv": {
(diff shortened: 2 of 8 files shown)


eli-schwartz:
Since pv is a CLI program whose job is to display terminal progress, I am curious what the general utility of it is as part of the wrapdb.

benoit-pierre:
Still useful for assembling stuff (e.g. as part of an embedded distribution or program).

benoit-pierre:
Similarly, I also use the libarchive wrap to get a tar executable for the target platform, not for the library itself.

subprojects/pv.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = pv-1.8.5
+source_url = https://codeberg.org/a-j-wood/pv/releases/download/v1.8.5/pv-1.8.5.tar.gz
+source_hash = d22948d06be06a5be37336318de540a2215be10ab0163f8cd23a20149647b780
+source_filename = pv-1.8.5.tar.gz
+patch_directory = pv
+
+[provide]
+program_names =
```
eli-schwartz, line 9: baffled how this works.

neheb: I assume the context is muon failing on this?

eli-schwartz: The context is rather that I don't understand the intention of defining this in a non-functional manner.

The fact that muon fails on this is interesting, but not as interesting as the question of "what was the goal here".

For example, the goal clearly wasn't to make meson work with `find_program('pv')`.

bgilbert: https://github.com/mesonbuild/wrapdb/pull/2344

PR #1299, opened 2023-11-21
libssh2: update to 1.11.0 

neheb:


neheb:
```
 ../subprojects/libssh2-libssh2-1.11.0/src/libssh2.rc(5) : fatal error RC1015: cannot open include file 'libssh2.h'.
```

@eli-schwartz 

I have include_directories explicit. No idea why it's not picking it up

neheb:
hacked with

```
libssh2_sources += win.compile_resources('src/libssh2.rc', include_directories: meson.current_source_dir() / 'include')
```

doesn't look right.

PR #1312, opened 2023-11-27
abseil-cpp: fix linking issue

jmillan:
`absl_string_lib` needs to be linked with `absl_crc_lib`.

'absl/strings/internal/cord_rep_crc.cc' file makes use of `crc_internal::CrcCordState` which is defined in
'absl/crc/internal/crc_cord_state.cc'


jmillan:
For reference, we are hitting this issue [here](https://github.com/versatica/mediasoup/actions/runs/6982919260/job/19003064065?pr=1201)

PR #1315, opened 2023-11-28
libuv: update to v1.47.0

jmillan:


jmillan:
Wow, it seems that Alpine GitHub image is running as `root`. This must have changed in the recent days, otherwise previous libuv releases would have failed for the same reason.

jmillan:
> This must have changed in the recent days, otherwise previous libuv releases would have failed for the same reason.

Oh, so Alpine was added to the CI Runners `after` the current libuv version was added, so indeed this is the first time the libuv tests are running in Alpine.

The problem is that libuv tests do not allow to be run by `root`. cc @neheb

ibc:
> The problem is that libuv tests do not allow to be run by `root`. cc @neheb

Unless `UV_RUN_AS_ROOT` environment variable is set:

https://github.com/libuv/libuv/blob/v1.x/test/run-tests.c#L69
```c
int main(int argc, char **argv) {
#ifndef _WIN32
  if (0 == geteuid() && NULL == getenv("UV_RUN_AS_ROOT")) {
    fprintf(stderr, "The libuv test suite cannot be run as root.\n");
    return EXIT_FAILURE;
  }
#endif
```

jmillan:
>  If there is no real problem then adding an env in the CI would work:

There is a reason for that check as per the commit message https://github.com/libuv/libuv/commit/af45b6ba2f2c506ae18dd24481e24db17b8269e7

```txt
The test suite is known to have issues when run as root. This
commit prevents the test suite from running when the euid is 0.
```

jmillan:
The [libuv issue template](https://raw.githubusercontent.com/libuv/libuv/v1.x/.github/ISSUE_TEMPLATE.md) clearly states not to run tests with root.

jmillan:
Issue created in GitHub runner-images: https://github.com/actions/runner-images/issues/8915

jmillan:
> In order to not run as root, the workflow file must be changed. PR with such a change welcome.

Which change? Alpine runs as a Docker image, meaning everything run inside will have an effective user id 0. 

Do you mean removing Alpine from the workflow? Or somehow creating a 'test' user that runs 'cargo test' to avoid tests being run with effective user id 0?

neheb:
AFAIK a user needs to be created and switched to it using su -c or something

.github/workflows/sanity_checks.yml
```diff
@@ -39,6 +39,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Create non privileged user
+        run: |
+          adduser wrapdb --disabled-password
+          su wrapdb
```
ibc, line 45: AFAIK this step runs in its own process so once it's completed it will exit and next step ("Sanity Checks") will run in a fresh process. So it won't work (indeed it doesn't).

I'd try doing this in "Sanity Checks" step below:

```yaml
su - wrapdb -c "./tools/sanity_checks.py"
```

However it won't work because it won't find `./tools/sanity_checks.py`. It should work if we add absolute path (not sure which one it is).

jmillan:
The final fix:

* Create an unprivileged user.
* Change ownership of './' to the new user.
* Run `sanity_checks.py` with the new user.

neheb:
That last commit should be applied to build_all as well

PR #1321, opened 2023-12-03
mpdecimal: don't run tests if random is missing

neheb:
random() is needed for tests.


skrah:
> Given this project seems to not accept patches or have a git, it will remain broken.

Magically, `mpdecimal 4.0.0` supports `MinGW` out of the box:

https://www.bytereef.org/mpdecimal/changelog.html

PR #1324, opened 2023-12-06
glib: update to 2.78.2

bgilbert:


bgilbert:
## Alpine and VS-clang-cl

```
proxy-libintl| subprojects/proxy-libintl-0.4/meson.build:14: WARNING: Project targets '>= 0.46.0' but uses feature introduced in '0.48.0': darwin_versions arg in library.

subprojects/proxy-libintl-0.4/meson.build:14:11: ERROR: Fatal warnings enabled, aborting
```

Fixed in https://github.com/frida/proxy-libintl/pull/6.

## VS

```
glib| DEPRECATION: None of the values ['gnu99'] are supported by the c compiler.
glib| However, the deprecated gnu99 std currently falls back to c99.
glib| This will be an error in the future.
glib| If the project supports both GNU and MSVC compilers, a value such as
glib| "c_std=gnu11,c11" specifies that GNU is prefered but it can safely fallback to plain c11.

subprojects\glib-2.78.2\meson.build:1:0: ERROR: Fatal warnings enabled, aborting
```

Projects [can't fix this warning](https://github.com/mesonbuild/meson/blob/7368795d13081d4928a9ba04d48498ca2442624b/docs/markdown/Release-notes-for-1.3.0.md#c_std-and-cpp_std-options-now-accepts-a-list-of-values) without requiring Meson 1.3.0+, which they may not be willing to do, so I imagine it'll show up frequently in this repo.  (Maybe not specifically for glib, which has a pretty aggressive [Meson update policy](https://gitlab.gnome.org/GNOME/glib/-/blob/main/docs/meson-version.md?ref_type=heads).)

PR #1329, opened 2023-12-08
sdl2: disable unity builds

takase1121:
Unity builds doesn't seem to work with SDL2, it throws a bunch of build errors.
Please advise if this isn't the correct way to disable it, and I should disable it when including SDL2 as a dependency, etc.

### DIFF
--- a/subprojects/packagefiles/sdl2/meson.build
+++ b/subprojects/packagefiles/sdl2/meson.build
@@ -1207,7 +1207,7 @@ sdl2_test = static_library('sdl2_test',
     c_args : core_args,
     link_args : core_ldflags,
     dependencies : [sdl2_dep, test_deps],
-    override_options : ['c_std=none', 'unity=off'],
+    override_options : ['c_std=none'],
     build_by_default : false,
 )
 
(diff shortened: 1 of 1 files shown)


jpakkane:
You should use `override_options` on the target itself.

takase1121:
Sorry, I didn't fully understand that - does that mean I should disable unity builds on the target that _uses_ SDL2? I'm still pretty novice when it comes to meson.

jpakkane:
You should add that override on the target that can't be built as a unity build, in this case to the main SDL2 library target. [Relevant doc link](https://mesonbuild.com/Reference-manual_functions.html#arguments10).

takase1121:
If I understood correctly, this means I should put `override_options` on only sdl2_dep (which contains the sdl2 library target).

jpakkane:
No, to the target that defines the SDL 2 library. What it means is "this target that is currently being defined can not be built in unity mode so do not use it for this target even if it has been enabled globally".

eli-schwartz:
Any particular reason this was closed?

Note that this should really bump the revision in releases.json...

PR #1330, opened 2023-12-09
add wrapfile for tree-sitter

Gottox:


ci_config.json
```diff
@@ -851,5 +851,7 @@
       "wayland:tests=false",
       "wayland:dtd_validation=false"
     ]
+  },
+  "tree_sitter": {
```
jpakkane, line 855: This is not needed if there contents is empty.

Gottox: fixed

PR #1333, opened 2023-12-11
gRPC: Add `channelz`, reflection

klokik:
This introduces targets that depend on `.proto` files being compiled.
To avoid patching sources, I use a wrapper script around `protoc` to create expected include paths within `builddir`


klokik:
Oh well, it seems that one needs to bump version of something (protobuf?) to get this building

PR #1335, opened 2023-12-14
Add whisper.cpp

tmayoff:


subprojects/packagefiles/whisper_cpp/meson.build
```diff
@@ -0,0 +1,66 @@
...
+
+cc = meson.get_compiler('c')
+
+math = cc.find_library('m', required: false)
+threads = dependency('threads')
+
+if build_machine.system() == 'linux'
+    add_project_arguments('-D_GNU_SOURCE', language: 'c')
+
+    if not get_option('no_avx')
```
klokik, line 11: Don't do double negation, e.g. `if get_option('avx').enabled() ..,`

klokik: And I would prefer x86 extensions opt-in, rather than opt-out. Chances of it being compiled elsewhere are high.
At least check for `host_machine.cpu_family() == 'x86_64'`

subprojects/packagefiles/whisper_cpp/meson.build
```diff
@@ -0,0 +1,66 @@
+project('whisper.cpp', ['cpp', 'c'], version: '1.5.2')
+
+cc = meson.get_compiler('c')
+
+math = cc.find_library('m', required: false)
+threads = dependency('threads')
+
+if build_machine.system() == 'linux'
```
klokik, line 8: 1. `host_machine`, not `build_machine`
2. checking for `cc.get_argument_syntax() == 'gcc'` would be better IMO.

subprojects/packagefiles/whisper_cpp/meson_options.txt
```diff
@@ -0,0 +1,10 @@
+option('no_avx', type: 'boolean', value: false, description: 'Disable AVX')
+option('no_avx2', type: 'boolean', value: false, description: 'Disable AVX2')
+option('no_fma', type: 'boolean', value: false, description: 'Disable FMA')
+option('no_f16C', type: 'boolean', value: false, description: 'Disable F16c')
```
klokik, line 4:
```suggestion
option('no_f16C', type: 'feature', value: 'disabled', description: 'Disable F16c')
```

subprojects/packagefiles/whisper_cpp/meson_options.txt
```diff
@@ -0,0 +1,10 @@
...
+option(
+    'build_examples',
```
klokik, line 6:
```suggestion
    'examples',
```
just examples would be enough

subprojects/packagefiles/whisper_cpp/meson.build
```diff
@@ -0,0 +1,66 @@
...
+
+whisper_lib = library(
+    'whisper',
+    'whisper.cpp',
+    dependencies: [ggml_dep],
+)
+whisper_cpp_dep = declare_dependency(
+    include_directories: '.',
+    dependencies: [ggml_dep],
+    link_with: whisper_lib,
```
klokik, line 61: consider `install`ing it, adding `version` and `soversion`

Installing headers and pkg-config `*.pc` file would be a nice, as a stretch goal.

klokik (changes requested):

subprojects/packagefiles/whisper_cpp/meson.build
tmayoff: I'll switch these around, this was just a one for one port of the original cmake build files, I didn't want to mess with things too much

PR #1336, opened 2023-12-14
Expose spdlog option to use std::format instead of internal fmt lib

hdoc:
Now that `std::format` is available in recent compiler versions (Clang 17, GCC 13), it's possible to use the compiler-provided `std::format` instead of fmtlib.

spdlog already provides this functionality upstream and [exposes it in CMake](https://github.com/gabime/spdlog/blob/v1.x/CMakeLists.txt#L90), but it's not an option with Meson.

This change adds support for that functionality to our wrap.


neheb:
Might want to add a check if std::format can be used.

hdoc:
@neheb added a check for the `format` header, let me know if that's what you had in mind.

subprojects/packagefiles/spdlog/meson.build
```diff
@@ -24,7 +24,9 @@ if get_option('external_fmt').enabled()
 endif
 
 if get_option('std_format').enabled()
-  spdlog_compile_args += '-DSPDLOG_USE_STD_FORMAT'
+  if meson.get_compiler('cpp').has_header('format', required: true)
```
neheb, line 27: required can take a feature option directly.

neheb:
The check is not enough actually.

If compiling in a C++ mode lower than 17, it will still succeed. I haven't tested if check_header works. That would be the cleanest solution.

In exiv2, I currently use

```
if not cpp.has_header_symbol('format', '__cpp_lib_format')
  deps += dependency('fmt')
endif
```

subprojects/packagefiles/spdlog/meson.build
hdoc: I'm aware that required can take `get_option('std_format'))` directly but this whole branch is gated by the  `if get_option('std_format').enabled()` anyway so the header check is always required.
If the option isn't specified in the first if statement, the `has_header()` check will never execute.
Setting required to `true` is a logical optimization.

This is what I think you're proposing, please clarify if my assumption is wrong:
```meson
if get_option('std_format').enabled()
  if meson.get_compiler('cpp').has_header('format', required: get_option('std_format'))
    spdlog_compile_args += '-DSPDLOG_USE_STD_FORMAT'
  endif
endif
```

hdoc:
Updated the check to use the [new C++20 library feature test macros](https://en.cppreference.com/w/cpp/utility/feature_test) like you proposed, and verified it worked locally with GCC 13 on macOS. 

One question: do we need to do any feature gating for the C++ standard version here? If a user is using C++17 and using the `compiled_lib` option I think this will fail. Perhaps we should hide the `std_format` option behind a `is_in_cpp20_mode` check [like here](https://github.com/mesonbuild/wrapdb/blob/fa17f829e49d0f2b6d0defd6d3392fab59055605/subprojects/packagefiles/stduuid/meson.build#L63-L70). Please let me know what you think is best for the end-user.

neheb:
I think it's best to modify the build file as such:

```
if meson.get_compiler('cpp').has_header_symbol('format', '__cpp_lib_format', required: get_option('std_format'))
  spdlog_compile_args += '-DSPDLOG_USE_STD_FORMAT'
else
  fmt_dep = dependency('fmt', version: '>=8.1', required: get_option('external_fmt').enabled())
  if fmt_dep.found()
    spdlog_dependencies += fmt_dep
    spdlog_compile_args += '-DSPDLOG_FMT_EXTERNAL'
  endif
endif
```

hdoc:
Looks reasonable and passed tests locally. I didn't realize `get_option()` could be used in this way. Thank you

subprojects/packagefiles/spdlog/meson.build
```diff
@@ -15,20 +15,16 @@ thread_dep = dependency('threads')
 spdlog_dependencies = [thread_dep]
 spdlog_compile_args = []
 
-if get_option('external_fmt').enabled()
-  fmt_dep = dependency('fmt', version: '>=8.1', required: true)
+if meson.get_compiler('cpp').has_header_symbol('format', '__cpp_lib_format', required: get_option('std_format'))
+  spdlog_compile_args += '-DSPDLOG_USE_STD_FORMAT'
+else
+  fmt_dep = dependency('fmt', version: '>=8.1', required: get_option('external_fmt').enabled())
```
neheb, line 21: you don't need enabled() here.

hdoc: Fixed

PR #1337, opened 2023-12-15
Update glm.wrap to point to the right url

germandiagogomez:
The link seems to be outdated.


germandiagogomez:
My bad. This is some kind of ssl problem. Closing.

PR #1339, opened 2023-12-17
curl: pass CURL_STATICLIB for static libraries

Barakat:
cURL is not compiled with `CURL_STATICLIB` when building a static library. cURL symbols end up being exported in the executable.


neheb:
Need a bump in releases.json. ping @benoit-pierre

benoit-pierre:
It would be better to add `public_args` to the library `c_args`, instead of using `add_project_arguments` (which impacts all other compiled library/executables).

Barakat:
I added `curl_lib_c_args` for this to keep these two, `public_args`, separate.

benoit-pierre:
> I added `curl_lib_c_args` for this to keep these two, `public_args`, separate.

Why? It makes perfect sense to use `public_args`.

Barakat:
`public_args` appears to be intended for the consumers of the library, and not for the library itself. I added `curl_lib_c_args` to make it explicit that the flags are for different audience.

benoit-pierre:
`public_args` is for the library compilation flags that are public, exported to consumers too. So no, I don't agree with the duplication.

Barakat:
@benoit-pierre I have passed it with `public_args`.

benoit-pierre (approved):

PR #1341, opened 2023-12-17
protobuf: update to v25.2

klokik:
Wanted by #1333

This now depends on `abseil-cpp` :cry:


klokik:
blocked by this https://github.com/protocolbuffers/protobuf/issues/14602

neheb:
MinGW is failing too.

neheb:
```
 D:\a\wrapdb\wrapdb\_build/../subprojects/abseil-cpp-20230802.0/absl/synchronization/internal/per_thread_sem.cc:93:(.text+0x385): undefined reference to `absl::lts_20230802::synchronization_internal::Win32Waiter::Wait(absl::lts_20230802::synchronization_internal::KernelTimeout)'
```

abseil issue I guess.

klokik:
Righ, my favorite chains of broken stuff

klokik:
@neheb  how do I add `diff_files` without triggering a sanity-check warning for a `minimum meson_version '>=0.58.0' but uses features which were added in newer versions` ? Or what would you suggest?
Or it would be acceptable to bump master project requirement to meson 0.63.0 ?

jpakkane:
> Would it be acceptable to bump master project requirement to meson>=0.63.0 ?

If it fixes your issue, sure.

neheb:
since you're fixing exports, I wonder if gnu_symbol_visibility is applicable.

klokik:
> I wonder if gnu_symbol_visibility is applicable.

For `protobuf(-lite)` yes, but not for `libprotoc`.

I had to default to static library, because of weird linking issues with MINGW. The rest of the platforms are fine, but I have no idea why `gcc` starts to export `absl` symbols there.

klokik:
Checks green, but cross-compilation will not work. Fixing `protoc-native` target ...

klokik:
Fixed cross-compilation: should be ready to go!

jpakkane:
WrapDB does not permit code changes. We use releases as-is with only build system changes. Otherwise we have to carry the maintenance burden of keeping them up to date which is something we don't even have resources to do.

klokik:
@jpakkane What would you suggest me to do? Should I drop MINGW from this release?

Me too isn't happy patching the source, but then, I'm afraid of falling into the rabbit-hole of fixing everything down the line. Everything is always broken somewhere and only may get fixed in the next release, that will inevitably bring other 🐞s

klokik:
And MSVC version will be only available as a static library compiled with `/permissive`

klokik:
Ok, that worked: downstream `abseil-cpp` is already patched, what a surprise 😏

klokik:
Think it's ready to go?

PR #1345, opened 2023-12-22
add libmpdclient

MaxKellermann:


eli-schwartz:
You should also update releases.json with the name, version number, and list of provides.

MaxKellermann:
Oh, right. Added. 
Do I need to add something to ci_config.json? I tried to run the test, but couldn't figure out how to test only libmpdclient, and testing everything seems to take a looong time...

eli-schwartz:
Can you sort releases.json as well?

As for ci_config.json this is used only in cases where you need to pass extra meson options to a specific wrap, or install packages with apt/brew.

eli-schwartz:
> As for ci_config.json this is used only in cases where you need to pass extra meson options to a specific wrap, or install packages with apt/brew.

... Which yeah, needs to be done at least to pass test=true.

MaxKellermann:
Sorted& added some foo to ci_config.json, mostly guessed. I don't know if this works on Windows, but the CI will tell me, I guess.

eli-schwartz:
I think you also still need `'build_options': ['test=true']`?

MaxKellermann:
Oh yes, it's false by default - sorry for try'n'erroring my way through....

eli-schwartz:
> Sorted& added some foo to ci_config.json, mostly guessed. I don't know if this works on Windows, but the CI will tell me, I guess.

Yup, doesn't work on Windows. The msys2 errors appear from a quick glance to be source level issues, the Visual Studio missing .lib file indicates that there is no vs_module_defs file nor a dllexport annotation. MSVC defaults to the equivalent of GNU symbol visibility (-fvisibility) "hidden", only explicitly exported symbols enter the shared DLL interface. It then, very cleverly, reports success when you link a useless library with zero symbols, because MSVC knows that error messages are important...

MaxKellermann:
The VS problem is one I've been fighting with yesterday - it fails in the libmpdclient CI the same way, but I have no idea what to do about it: https://github.com/MusicPlayerDaemon/libmpdclient/actions/runs/7300420548/job/19895017607 - if you have a clue, please tell me..

The MSYS problems are obviously a portability bug in libmpdclient. I'll need to fix that in the next release. Can I disable the tests only on Windows?

MaxKellermann:
> a dllexport annotation

Ah this one. Windows is weird. I wonder why this was never a problem with mingw.

MaxKellermann:
As a workaround until the dllexport thing is fixed, would it be possible to default to building static libraries? I guess if somebody uses a wrap, they want static libraries anyway - I was wondering why that isn't Meson's default for subprojects...

eli-schwartz:
GCC/binutils on Windows acts like GCC/binutils on Linux and exports all symbols by default. To play nicely with the Windows ecosystem it does this by first checking if there are *any* dllexports, and if there aren't any, it assumes that you developed the library Linux style and want all symbols exported, so it does so.

If you use any dllexports, it assumes that you took care to ensure that it builds a DLL correctly with MSVC, and have annotated all your exported symbols -- so it knows it is safe to hide all the rest just like MSVC does.

tl;dr msys2 actually wants users to have usable software, hence this works out of the box.

eli-schwartz:
> As a workaround until the dllexport thing is fixed, would it be possible to default to building static libraries?

Sure, no problem. Please leave a json comment (`'_comment': 'json sucks'`) saying why you did it.

> I guess if somebody uses a wrap, they want static libraries anyway - I was wondering why that isn't Meson's default for subprojects...

Not necessarily, e.g. using DLLs means that you can swap it out at runtime which may be important in some contexts. Especially for stuff like sideloading an updated libssl that has gotten security patches applied.

MaxKellermann:
Static linking doesn't solve all the problems, because as soon as I enable that, MSVC will complain about the lack of socketpair() in the unit test. So this shortcut won't work. I need to fix the unit test first, but I don't have enough time left this year for that. I'll get back to this PR next month. Thanks so far, @eli-schwartz.

MaxKellermann:
Okay, I just figured that I don't need to make the unit test portable .... because "check" isn't portable anyway. It doesn't even compile with mingw because it uses `sigaction()`.

Therefore, my suggestion is to disable MSVC/MSYS for now.

The next libmpdclient release will have a solution for the dllexport thing. Then we may enable the Wrapdb CI test for libmpdclient on Windows, but for that, I need a way to disable the unit tests on Windows, but enable them on all others. It looks like it can't be done currently with the wrapdb ci script.

eli-schwartz:
You may want the meson.build to automatically refuse to use unittests on MinGW but allow it on MSVC (if I'm understanding correctly) by checking for build_machine.system() in combination with cc.get_id()

MaxKellermann:
Yes, but it needs to refuse that on all Windows targets (mingw, msvc, msys) because "check" doesn't work on any of them. All those years I've been using this test library, I never knew or cared.

MaxKellermann:
Strange, the check website (https://libcheck.github.io/check/) claims it does work on Windows, but the "check" wrap throws many compiler errors when I try that. Maybe just the "check" wrap is broken?

MaxKellermann:
Apparently, "check" is just buggy. There's a configure-time check for `sigaction()`, but in one line, it calls `sigaction()` without checking `HAVE_SIGACTION`. Sigh, I stepped into a fractal problem again.

neheb:
seems fairly dead upstream.

PR #1348, opened 2023-12-25
pango: add additional provides based on exported dependencies

eli-schwartz:


eli-schwartz:
Compilation fails under clang-cl -> not an issue from this PR.

PR #1352, opened 2023-12-30
lcms2: update to 2.16

neheb:


neheb:
ping @mm2

mm2:
Thanks for reporting. You triggered that issue, was due to https://github.com/mesonbuild/meson/pull/7556
I fixed it in my sources.  https://github.com/mm2/Little-CMS/commit/f2a7a20abe58d3f5bfa2db8b7b6db809b0920788
Then found it is better to just remove the option.

PR #1358, opened 2024-01-06
pixman: update to 0.43.2

bgilbert:
Marking draft for now, in hope of [clarification](https://gitlab.freedesktop.org/pixman/pixman/-/issues/87) on whether this is a stable release.


bgilbert:
No official clarification from upstream, either in the [issue](https://gitlab.freedesktop.org/pixman/pixman/-/issues/87) or a [mailing list thread](https://lists.cairographics.org/archives/cairo/2024-January/029519.html).  However, releases with odd minor numbers [automatically enable](https://gitlab.freedesktop.org/pixman/pixman/-/blob/8e4be8c2dbd605d54c79484d9f5d8befd42bd6f3/pixman/pixman-private.h#L1060-1073) some [debug code](https://gitlab.freedesktop.org/pixman/pixman/-/blob/8e4be8c2dbd605d54c79484d9f5d8befd42bd6f3/pixman/pixman-region.c#L90-95) which claims to have performance consequences, so it seems safest to treat this as a development release and not package it.

I'll leave this PR open as documentation unless wrapdb maintainers would prefer otherwise.

bgilbert:
Bumped to 0.43.2.  Per [this PR](https://gitlab.freedesktop.org/pixman/pixman/-/merge_requests/91), odd minors are no longer development releases, and the corresponding debug code has been removed.

This should be ready for review.

jpakkane:
```
pixman| Project name: pixman
pixman| Project version: 0.43.2
pixman| DEPRECATION: None of the values ['gnu99'] are supported by the c compiler.
pixman| However, the deprecated gnu99 std currently falls back to c99.
pixman| This will be an error in the future.
pixman| If the project supports both GNU and MSVC compilers, a value such as
pixman| "c_std=gnu11,c11" specifies that GNU is prefered but it can safely fallback to plain c11.

subprojects\pixman-0.43.2\meson.build:21:0: ERROR: Fatal warnings enabled, aborting
```

bgilbert:
wrapdb CI is going to see that failure a lot, I think.  `c_std=gnu11,c11` requires Meson &ge; 1.3.0, and many projects may want to continue supporting older Meson.

neheb:
wonder if it's possible for CI to fail for wraps that have custom meson.build files only.

PR #1371, opened 2024-01-22
glib: update to 2.78.4

bgilbert:


bgilbert:
Test failures are the expected ones.

neheb:
mmm I don't like that alpine failure.

Add gettext-tiny-dev to alpine_packages

bgilbert:
Ah yeah, good call.  Updated.

PR #1072, opened 2023-07-08
taglib: update to 1.13.1

neheb:

PR #1372, opened 2024-01-23
sanity_checks: don't compare revision against a version component

bgilbert:
`1.3-5` was sorting greater than `1.3.1-1` because the normalized `[1, 3, 5]` is greater than `[1, 3, 1, 1]`.

Ensure a revision is always compared with another revision and not with a version component.  To simplify the implementation, split off the revision into a separate field of the `Version` object.  This also allows us to verify that the revision exists and is numeric.

### DIFF
--- a/tools/utils.py
+++ b/tools/utils.py
@@ -23,8 +23,15 @@ class Version:
     def __init__(self, s: str) -> None:
         self._s = s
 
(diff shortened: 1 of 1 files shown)


neheb:
Oh hey. This looks like it fixes https://github.com/mesonbuild/wrapdb/pull/1072

PR #1387, opened 2024-02-02
cwalk: new package

antoniovazquezblanco:


subprojects/packagefiles/cwalk/meson.build
```diff
@@ -0,0 +1,15 @@
+project('cwalk', 'c',
+  license: 'MIT',
+  meson_version: '>= 0.57.0'
```
neheb, line 3: pretty sure you can reduce to 0.49.0

antoniovazquezblanco: ..\subprojects\cwalk-1.2.7\meson.build:15: WARNING: Project targets '>= 0.49.0' but uses feature introduced in '0.50.0': include_directories kwarg of type string. Use include_directories('include') instead

antoniovazquezblanco:
A release was pushed with fixed meson version.

https://github.com/likle/cwalk/issues/42

antoniovazquezblanco:
It seems the default option for `default_libraries` is `shared`. See https://mesonbuild.com/Builtin-options.html

For some reason, Visual Studio check script is checking for the presence of the static one... Something wrong on my end?

Thanks!

neheb:
```
ERROR: File 'subprojects/cwalk-1.2.8/cwalk.lib' could not be found
```

is caused by no functions being exported. This can be fixed with a custom def file.

antoniovazquezblanco:
This is being fixed upstream https://github.com/likle/cwalk/pull/46

antoniovazquezblanco:
Now all tests pass! :D

subprojects/packagefiles/cwalk/meson.build
```diff
@@ -0,0 +1,29 @@
+project('cwalk', 'c',
```
jpakkane, line 1: Upstream already has a `meson.build` file. Could this not use that one instead?

antoniovazquezblanco: It needs a fix that has just merged. We will be able to use upstream meson file in the next version.

subprojects/packagefiles/cwalk/meson.build
```diff
@@ -0,0 +1,29 @@
+project('cwalk', 'c',
+  license: 'MIT',
+  version: '1.2.9',
+  meson_version: '>= 0.57.0'
+)
+
+cwalk_inc = include_directories('include')
+
+cwalk_c_args = []
+if get_option('default_library') == 'shared'
```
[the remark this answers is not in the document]
antoniovazquezblanco: Done! Thanks

antoniovazquezblanco:
Upstream fix at https://github.com/likle/cwalk/pull/47

antoniovazquezblanco:
From upstream, @likle asks:

> I am not verify familiar with meson, doesn't "both" mean it builds shared as well as static?
> Wouldn't that set it to CWK_SHARED for the static build done using the both option?

antoniovazquezblanco:
Ping @neheb :)

antoniovazquezblanco:
Thanks @neheb!

Do you have an answer for https://github.com/mesonbuild/wrapdb/pull/1387#issuecomment-2098203749 ?

How does meson deal with dual builds?

Thanks again!!

neheb:
both builds both shared and static. When using in a dependency, the shared is used. In wraps around here, typically != static is checked to enable a SHARED macro.

benoit-pierre:
IMHO, wraps that need special shared/static arguments should use `<lang>_shared_args` / `<lang>_static_args` when available and otherwise error on meson < 1.3.0 when building with `--default-library=both`.

neheb:
Right. It’s usually not an issue with GCC based compilers but not MSVC. The best solution for MSVC currently is to avoid these and use def files.

PR #1390, opened 2024-02-03
libxmlpp: update to 5.2.0

neheb:


neheb:
I have no idea why Alpine fails like this.

PR #1391, opened 2024-02-03
protobuf: fix deps problem, allow disabling install

v-for-vandal:
* Now dependency from abseil is correctly carried to 'upstream' project
  when building protobuf as subproject
* Allow disabling installation. This patch adds new option
  'should_install' and uses it as value for 'install' argument
  in all appropriate meson invocations. The intended use is to disable
  installing protobuf when it is bundled as subproject.


ninjaoflight:
seems that you need to update the 'releases.json' file to include a new version, something like '25.2-2'

v-for-vandal:
@eli-schwartz Right, sorry, different conventions. I fixed the commit name and [hopefully] better explained why one may need 'should_install' option.
Some projects prefer 'one features - one pr', so if you want, I can split it into two separate PRs - bugfix for dependencies and new option.
I have a questions: should I write a test for this new option ? Can I event write it ?

eli-schwartz:
> Some projects prefer 'one features - one pr', so if you want, I can split it into two separate PRs - bugfix for dependencies and new option.

I would say it's better to have two commits in one PR than to have two PRs with one commit each.

A single PR will, on merging, activate the release tooling to publish a new edition of the wrap. While it's fine and even encouraged to have atomic commits, there's no good reason to publish a new wrap revision for each commit if you know it will be immediately superseded by a new revision.

v-for-vandal:
@eli-schwartz All tests are green. Is there something I should do, or is it up to you now?

eli-schwartz:
> @eli-schwartz All tests are green. Is there something I should do, or is it up to you now?

Nah. Sorry, was just busy :D fortunately another maintainer was available to handle the final step.

PR #1393, opened 2024-02-05
libpng: update to 1.6.42

bgilbert:
Reuse the soversion from 1.6.41 because upstream does.


benoit-pierre:
This install a `libpng16.so.16.42.0` library, but on my Arch Linux install, the shared library version is still `16.41.0` (same when compiling `1.6.42` from source with the provided CMakefiles).

bgilbert:
Ah yeah, good catch.  I've special-cased a workaround so the soversion isn't accidentally left unchanged in the next version bump.

PR #1394, opened 2024-02-06
x-plane-sdk: new wrap

judemille:
Packaging up the [X-Plane plugin SDK](https://developer.x-plane.com/sdk/plugin-sdk-downloads/) for easy use with Meson.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -869,5 +869,10 @@
       "wayland:tests=false",
       "wayland:dtd_validation=false"
     ]
+  },
+  "x-plane-sdk": {
+    "build_options": [
+      "x-plane-sdk:cpp=enabled"
+    ]
   }
 }
--- a/subprojects/packagefiles/x-plane-sdk/meson.build
+++ b/subprojects/packagefiles/x-plane-sdk/meson.build
@@ -1,0 +1,128 @@
+project(
+    'x-plane-sdk',
+    'c',
+    version: '4.0.1',
+    license: 'MIT',
+    default_options: 'warning_level=0',
+    meson_version: '>= 0.60.0',
+)
+
+xplm_incdir = include_directories('SDK/CHeaders/XPLM')
+widgets_incdir = include_directories('SDK/CHeaders/Widgets')
+wrappers_incdir = include_directories('SDK/CHeaders/Wrappers')
+
+cc = meson.get_compiler('c')
+
+ca = []
+xplm_level = get_option('xplm-level')
+
+if xplm_level >= 200
+    ca += '-DXPLM200'
+endif
+
+if xplm_level >= 210
+    ca += '-DXPLM210'
+endif
+
+if xplm_level >= 300
+    ca += '-DXPLM300'
+endif
+
+if xplm_level >= 301
+    ca += '-DXPLM301'
+endif
+
+if xplm_level >= 303
+    ca += '-DXPLM303'
+endif
+
+if xplm_level >= 400
+    ca += '-DXPLM400'
+endif
+
+if target_machine.system() == 'windows'
+    ca += ['-DAPL=0', '-DIBM=1', '-DLIN=0']
+    xplm_dep = declare_dependency(
+        compile_args: ca,
+        include_directories: [xplm_incdir],
+        dependencies: [
+            cc.find_library(
+                'XPLM_64',
+                dirs: [meson.current_source_dir() / 'SDK' / 'Libraries' / 'Win'],
+            ),
+        ],
+    )
+    xpwidgets_dep = declare_dependency(
+        compile_args: ca,
+        include_directories: [widgets_incdir],
+        dependencies: [
+            xplm_dep,
+            cc.find_library(
+                'XPWidgets_64',
+                dirs: [meson.current_source_dir() / 'SDK' / 'Libraries' / 'Win'],
+            ),
+        ],
+    )
+elif target_machine.system() == 'darwin'
+    ca += ['-DAPL=1', '-DIBM=0', '-DLIN=0']
+    xplm_dep = declare_dependency(
+        compile_args: ca,
+        link_args: [
+            '-F' + meson.current_source_dir() / 'SDK' / 'Libraries' / 'Mac',
+            '-framework',
+            'XPLM',
+        ],
+        include_directories: [xplm_incdir],
+    )
+    xpwidgets_dep = declare_dependency(
+        compile_args: ca,
+        link_args: [
+            '-F' + meson.current_source_dir() / 'SDK' / 'Libraries' / 'Mac',
+            '-framework',
+            'XPWidgets',
+        ],
+        include_directories: [widgets_incdir],
+        dependencies: [xplm_dep],
+    )
+else
+    ca += ['-DAPL=0', '-DIBM=0', '-DLIN=1']
+    xplm_dep = declare_dependency(
+        compile_args: ca,
+        include_directories: [xplm_incdir],
+    )
+    xpwidgets_dep = declare_dependency(
+        compile_args: ca,
+        include_directories: [widgets_incdir],
+        dependencies: [xplm_dep],
+    )
+endif
+
+if add_languages('cpp', native: false, required: get_option('cpp'))
+    xpcpp_dir = 'SDK/CHeaders/Wrappers'
+    # There is no reason to make this a dynamic library.
+    # Static is the only sensible choice for this particular library.
+    xpcpp = static_library(
+        'xpcpp',
+        cpp_args: ca,
+        include_directories: [wrappers_incdir],
+        pic: true,
+        dependencies: [xplm_dep, xpwidgets_dep],
+        sources: [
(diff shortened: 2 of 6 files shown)


judemille:
Sanity checks pass.

ci_config.json
```diff
@@ -869,5 +869,8 @@
       "wayland:tests=false",
       "wayland:dtd_validation=false"
     ]
+  },
+  "x-plane-sdk": {
+    "_comment": "- nothing special."
```
eli-schwartz, line 874: If there's no config being passed to the CI then I'd expect to not have an entry at all.

subprojects/packagefiles/x-plane-sdk/meson.build
```diff
@@ -0,0 +1,121 @@
+project(
+    'x-plane-sdk', 'c', 'cpp',
```
eli-schwartz, line 2: From a quick glance I didn't see a C compiler being used for anything (other than library lookups that could be done with a C++ compiler just as easily). Do we need both?

tools/sanity_checks.py
```diff
@@ -300,6 +300,10 @@ def check_source_url(self, name: str, wrap_section: configparser.SectionProxy, v
         elif name == 'directxmath':
             # DirectXMath source url contains only tag name without version
             return True
+        elif name == 'x-plane-sdk':
+            segs = version.split('.')
+            assert(len(segs) == 3)
```
eli-schwartz, line 305: Any reason to use a bare assert here?

Note that a bare assert should NOT use parentheses, as it is not a function. Spell this as:

```python
assert len(segs) == 3
```

The extra parentheses are interpreted as
```python
to_check = (len(segs) == 3)

assert to_check
```

eli-schwartz: Ok after taking another look I notice this is copied from sqlite which does the same broken thing a few lines up... sigh...

eli-schwartz (changes requested):

ci_config.json
judemille: Ah okay, wasn't sure of the procedure. I'll chop that out.

subprojects/packagefiles/x-plane-sdk/meson.build
judemille: The library, except for the C++ wrappers, is exposed as C. Am I mistaken to mark it as having C, in that case?

tools/sanity_checks.py
judemille: I'm innocent, I promise!

subprojects/packagefiles/x-plane-sdk/meson.build
eli-schwartz: Those parts appear to be precompiled SDK binaries?

You don't need a C compiler to link to C libraries, only to compile them. In theory, someone that has a C++ compiler installed but doesn't have a C compiler (what *does* that mean, exactly??? who knows) could link to precompiled libraries in their C++ only software.

judemille: I've updated the build script -- a couple fixes, as well as making building the C++ wrappers optional, since not all may use them (instead using [PPL](https://github.com/PhilippMuenzel/PPL) if they still use C++).

This way, only people who really want the included wrappers get them, and it only asks for a C++ compiler if using them. Granted, the only platforms X-Plane supports all have C++ compilers, so I would be concerned if one managed to end up without one, but I suppose for minimal installs on Linux?

judemille:
oops, UI didn't update the CI status until after i clicked the re-request review button. fantastic UI design lmao.
lemme fix that

eli-schwartz:
> oops, UI didn't update the CI status until after i clicked the re-request review button. fantastic UI design lmao. lemme fix that

The brave new world of JavaScript... everything is dynamic up until you actually want it to be. :D

Started the CI

subprojects/packagefiles/x-plane-sdk/meson.build
```diff
@@ -0,0 +1,129 @@
...
+    xpwidgets_dep = declare_dependency(
+        compile_args: ca,
+        include_directories: [widgets_incdir],
+        dependencies: [xplm_dep],
+    )
+endif
+
+if get_option('cpp').enabled()
+    add_languages('cpp')
+    xpcpp_dir = 'SDK/CHeaders/Wrappers'
```
eli-schwartz, line 102: The point of feature options is to support auto, otherwise we'd just use boolean true/false.

You can pass the feature option in as the value of "required" to various functions such as add_languages, and check the return value to see if it's found. Then use that as the "if" condition.

judemille:
I forgor to enable the C++ wrappers when testing that.

subprojects/packagefiles/x-plane-sdk/meson.build
```diff
@@ -0,0 +1,129 @@
...
+    )
+
+    xpcpp_dep = declare_dependency(
+        compile_args: ca,
+        dependencies: [xplm_dep, xpwidgets_dep],
+        include_directories: [wrappers_incdir],
+        link_with: [xpcpp],
+    )
+else
+    xpcpp_dep = disabler()
```
eli-schwartz, line 128: This means that anything which depends on this will silently be disabled as well, which feels like the wrong way to interact with a different subproject (e.g. a parent project). I would suggest instead evaluating this as `dependency('', required: false)` which will show up as a "NOT FOUND dependency.

eli-schwartz (changes requested):

subprojects/packagefiles/x-plane-sdk/meson.build
judemille: Right, that makes sense.

subprojects/packagefiles/x-plane-sdk/meson.build
judemille: Ahhh, I knew I was missing something.

judemille:
@eli-schwartz Is this good to go now?

subprojects/packagefiles/x-plane-sdk/meson.build
```diff
@@ -0,0 +1,128 @@
+project(
+    'x-plane-sdk',
+    'c',
+    version: '4.0.1',
+    license: 'MIT',
+    default_options: 'warning_level=0',
```
eli-schwartz, line 6: Added in 0.50.0: https://github.com/mesonbuild/meson/commit/a9e63568fe4a1ef420cdb8bacb85fbf6d621dc0d

subprojects/packagefiles/x-plane-sdk/meson_options.txt
```diff
@@ -0,0 +1,4 @@
+option('xplm-level', type: 'integer', min: 0, max: 400, value: 0,
+      description: 'The XPLM compatibility level to use. See the documentation.')
+option('cpp', type: 'feature', value: 'disabled',
```
eli-schwartz, line 3: This started working in 0.60.0: https://github.com/mesonbuild/meson/commit/8a0d12ec295780977d0e2cb7e2f64f1e3a3e0067

Before then it said "Option name cpp is reserved."

It is not clear to me what the commit tried to do or how it tried to do it. There is no FeatureNew for this, nor release notes documentation, and the functionality is not tested, so it would appear to be an accidental side effect...

@xclaesse thoughts?

eli-schwartz (changes requested):

subprojects/packagefiles/x-plane-sdk/meson.build
judemille: One day I'll get the version number right.

subprojects/packagefiles/x-plane-sdk/meson_options.txt
judemille: Huh. I will admit, it didn't occur to me to test with a version so old, but it makes sense. I wonder if any distros even have a version pre-0.60.0 packaged still?

eli-schwartz: Debian oldstale (not a typo) has 0.56, Ubuntu 20.04 LTS (Long Term Suckage) has 0.53.

Some people care, some do not. This is not exactly core infrastructure here so I doubt anyone will blink twice if, unlike zlib, this package doesn't build with prehistoric mesons. Even if the distros exist, it doesn't follow that x-plane-sdk users use those distros.

judemille: From experience: some people do some very cursed stuff in building X-Plane plugins. libacfutils, AFAIK, just does not build if your compilers and stuff are too new. and it has tarballs and stuff in tree. wild stuff.

but yeah i'll just make it `0.60.0` and call it a day

judemille: the kinds of people crazy enough to build X-Plane plugins with super old distros are probably not the same kinds of people to use meson

judemille:
ok am i dumb or something?

PR #1398, opened 2024-02-07
cpputest: fixed wrap

SteelPh0enix:
Current wrap of CppUTest is fundamentally broken in multiple ways:

* Overwrites main dependency (by default), making it impossible to link it and use the wrap
* It's ignoring the existence of platform-specific implementations, forcing the user to provide them manually
* Typo in options

This PR fixes all those issues and makes this wrap usable.


SteelPh0enix:
Any suggestions how could i keep the extensions off by default while keeping the CI green?

jpakkane:
You can specify build options in `ci_config.json`.

SteelPh0enix:
I've added CI settings, fixed some stuff and performed sanity checks locally, it should be green now.

SteelPh0enix:
Huh, so that's how this switch works...

All right, i'll make a compiler check in there for MSVC and GCC/Clang, should be good enough for the CI and "most people".

SteelPh0enix:
Done, hopefully that's the last issue

subprojects/packagefiles/cpputest/src/Platforms/meson.build
```diff
@@ -3,7 +3,7 @@ cpputest_platform = get_option('cpputest_platform')
 if cpputest_platform != 'none'
   if cpputest_platform == 'autodetect'
     compiler_id = meson.get_compiler('cpp').get_id()
-    if compiler_id == 'clang' or compiler_id == 'gcc'
+    if compiler_id == 'gcc' or compiler_id == 'clang' or compiler_id == 'clang-cl'
```
neheb, line 6: should it be treated as GCC or VisualCpp?

SteelPh0enix:
Hm, maybe `clang-cl` should be bound to `VisualCpp` instead...
Yet the CI passes, hopefully it stays that way. Leaving that comment for the future me if this ever becomes a problem. It's 2AM, CI is green, I'm going to sleep.

subprojects/packagefiles/cpputest/src/Platforms/meson.build
SteelPh0enix: Yeah, it passes here so I'm not sure it's just a funky compatibility layer, or intended behaviour.
I'll throw it to VisualCpp and let's see what happens

SteelPh0enix: After quick read, yeah, this should definitely be treated as VisualCpp

PR #1399, opened 2024-02-08
spdlog: fix compilation with cpp_std >= gnu++20

polesapart:
When compiling a meson project with cpp_std=c++20 (actually tried with gnu++20), the spdlog subproject wrap fails to build, because on subproject root's  meson.build the following check will succeed:

`if meson.get_compiler('cpp').has_header_symbol('format', '__cpp_lib_format', required: get_option('std_format'))`

Leading to fmt_dep to be undefined:

```
if meson.get_compiler('cpp').has_header_symbol('format', '__cpp_lib_format', required: get_option('std_format'))
  spdlog_compile_args += '-DSPDLOG_USE_STD_FORMAT'
else
 fmt_dep = dependency('fmt', version: '>=8.1', required: get_option('external_fmt'))
endif
```

The patch adds a falsy fmt_dep as documented in meson.

This allow the src/meson.build to succeed:

```
if not fmt_dep.found()
  src += files('bundled_fmtlib_format.cpp')
endif
```

instead of aborting because of an undefined fmt_dep.

With that change, all tests passed.


dcbaker (approved):

neheb (approved):

subprojects/packagefiles/spdlog/meson.build
```diff
@@ -16,6 +16,7 @@ spdlog_compile_args = []
 
 if meson.get_compiler('cpp').has_header_symbol('format', '__cpp_lib_format', required: get_option('std_format'))
   spdlog_compile_args += '-DSPDLOG_USE_STD_FORMAT'
+  fmt_dep = depemndency('', required: false)
```
Totto16, line 19: Sorry to interrupt, but isn't `depemndency` spelled wrong here?
Shouldn't it be `dependency`

polesapart: @Totto16 You're absolutely right. Not sure how this typo went it before I submit it.

PR #1406, opened 2024-02-14
backward-cpp: fix typo in feature flags

NewbiZ:
Remove the typo adding a trailing `S` to backward-cpp feature flags. See https://github.com/bombela/backward-cpp/blob/master/backward.hpp#L139 for the proper names.


NewbiZ:
> The change here looks correct to me, there's a few missing bits for the wrapdb infrastructure though

Could you clarify what needs to be done? I am not familiar with how the wrapdb repository works.

dcbaker:
I think in this case the only thing you need to do is edit releases.json and add a new release of backward cpp. In this case that would be "1.6-3" I think

PR #1408, opened 2024-02-16
spdlog: remove default c++ version

### LINKED ISSUE #1404: spdlog c++ standard should be upgraded
### Intent
I want to use the `spdlog` wrap with the standard `std::format` instead of the bundled `fmt::format`. This is possible since C++20 onward. `spdlog` offers this possibility with the use of `-DSPDLOG_USE_STD_FORMAT`.

### Description of the issue
The current `spdlog` wrap uses `default_options: ['cpp_std=c++14']`.
This is incompatible with the `std_format` feature, because it specifically checks for `__cpp_lib_format`, which is only defined if the C++ standard is set to 20 onward. This effectively makes the `std_format` feature unusable.

### References
- The `spdlog` wrap forces the c++ standard to c++ 14: https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/spdlog/meson.build#L6
- The `spdlog` wrap exposes the `std_format` feature and checks that `__cpp_lib_format` is defined: https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/spdlog/meson.build#L17
- The `__cpp_lib_format` feature test macro is only defined on C++20 onward: https://en.cppreference.com/w/cpp/feature_test#cpp_lib_format

NewbiZ:
Remove default `c++14` version since some features require `c++20` onward.

Fixes #1404


dcbaker:
This doesn't seem right. setting no version means the compiler picks it's default, which assumes that any version is fine. That isn't true, it sounds like the version should be set to `c++20`, since it requires c++20 features?

neheb:
No.

spdlog as a subproject inherits the c++ standard from the main project. It can work back to C++11.

NewbiZ:
> This doesn't seem right. setting no version means the compiler picks it's default, which assumes that any version is fine. That isn't true, it sounds like the version should be set to `c++20`, since it requires c++20 features?

I don't think there's an universal answer here. The library itself is mainly c++11, can leverage c++14 constexpr, and has defines that requires c++20, and some that don't.

PR #1410, opened 2024-02-16
elfutils: create wrap

NewbiZ:
`elfutils` (https://sourceware.org/elfutils/) is a set of libraries and tools to inspect ELF, DWARF, etc.

> **libelf**
> elf32, elf64 and gelf functions to read, modify and create ELF files.

> **ibdw**
> dwarf, dwfl and dwelf functions to read DWARF, find separate debuginfo, symbols and inspect process state.

> **libasm**
> asm and disasm functions to assemble and disassamble instructions (partial support for i686 and BPF instructions only).

Apart from its direct use, `elfutils` can be linked to `backward-cpp` (an already existing wrap) to allow the display of demangled stack traces with inline source code.

Some notes:

- `elfutils` requires the `m4` binary to generate some definition files. Currently we directly check for `m4` using `find_program`, though we could be smarter in the future and fallback to using the `m4` wrap. I am not litterate enough in meson to do that properly for now, but that could be a welcomed addition in the future.

- Wrapping autotools projects requires using the `unstable-external_project` feature, or I may be mistaken.

### DIFF
--- a/subprojects/packagefiles/elfutils/meson.build
+++ b/subprojects/packagefiles/elfutils/meson.build
@@ -1,0 +1,21 @@
+project('elfutils', 'c', meson_version: '>=0.56.0')
+
+# elfutils generates definition files using `m4`
+# TODO: we could be smarter here and handle cases where the `m4`
+# binary is setup as a subproject
+m4 = find_program('m4', required: true)
+
+mod = import('unstable-external_project')
+
+proj = mod.add_project('configure',
+  configure_options: [
+    '--prefix=@PREFIX@',
+    '--libdir=@PREFIX@/@LIBDIR@',
+    '--includedir=@PREFIX@/@INCLUDEDIR@',
+    '--disable-debuginfod',
+  ],
+)
+
+asm = proj.dependency('asm')
+dw = proj.dependency('dw')
+elf = proj.dependency('elf')
--- a/releases.json
+++ b/releases.json
@@ -553,6 +553,16 @@
       "3.3.4-1"
     ]
   },
+  "elfutils": {
+    "dependency_names": [
+      "asm",
+      "dw",
+      "elf"
+    ],
+    "versions": [
+      "0.190-1"
+    ]
+  },
   "emilk-loguru": {
     "dependency_names": [
       "loguru"
--- a/subprojects/elfutils.wrap
+++ b/subprojects/elfutils.wrap
@@ -1,0 +1,12 @@
+[wrap-file]
+directory = elfutils-0.190
+source_url = https://sourceware.org/elfutils/ftp/0.190/elfutils-0.190.tar.bz2
+source_filename = elfutils-0.190.tar.bz2
+source_hash = 8e00a3a9b5f04bc1dc273ae86281d2d26ed412020b391ffcc23198f10231d692
+patch_directory = elfutils
+
+[provide]
+asm = asm
+dw = dw
+elf = elf
+elfutils = asm dw elf


ao2:
This is cool, thanks @NewbiZ would it make sense to specify a `lib` prefix for the libraries or is that against meson conventions?

Thanks, Antonio

neheb:
hrm I don't think this would be allowed.

@eli-schwartz @bgilbert ?

subprojects/packagefiles/elfutils/meson.build
```diff
@@ -0,0 +1,21 @@
+project('elfutils', 'c', meson_version: '>=0.56.0')
+
+# elfutils generates definition files using `m4`
+# TODO: we could be smarter here and handle cases where the `m4`
+# binary is setup as a subproject
+m4 = find_program('m4', required: true)
+
+mod = import('unstable-external_project')
+
+proj = mod.add_project('configure',
```
eli-schwartz, line 10: The wrapdb collects "native" ports, i.e. you should transcribe the Makefile.am rules into meson `library()` etc. rather than using the configure script.

The unstable-external_project module exists exactly so that projects which don't have wraps can be "quick and dirty" used in a meson project, but by the same extension there's really no reason such a wrapped-configure needs a WrapDB to centralize the build files for it.

And people using the WrapDB generally expect the result to be a subproject that nicely parallelizes, supports find_program overrides (which external_project is incapable of doing), exports `*-uninstalled.pc` files, and works cross platform where possible. Granted, cross platform isn't too much of a concern for a dependency that parses and manipulates Unix ELF files, but for many other projects that *could* be used on Windows, it would be quite surprising to have a WrapDB entry only build on windows inside of a mingw shell.

eli-schwartz (changes requested):

eli-schwartz:
> Apart from its direct use, `elfutils` can be linked to `backward-cpp` (an already existing wrap) to allow the display of demangled stack traces with inline source code.
> 
> Some notes:
> 
>  * `elfutils` requires the `m4` binary to generate some definition files. Currently we directly check for `m4` using `find_program`, though we could be smarter in the future and fallback to using the `m4` wrap. I am not litterate enough in meson to do that properly for now, but that could be a welcomed addition in the future.

These points are extra relevant: already existing wraps can provide e.g.
- `meson.override_find_program('m4', m4_prog)`
- `meson.ovrride_dependency('awkward-cpp', awkward_cpp_dep)`

And a native meson wrap for elfutils would find those overrides via the wrap system. But a wrapped Autotools external_project cannot find either one, and can only locate a system dependency. Meson can't inject its own results into `AC_CHECK_PROG([M4], [m4])` and actually can't even guarantee that the configure script is a real GNU one as opposed to a custom "implements the same interface as GNU" one.

You need a native wrap in order to make this all work smoothly.

subprojects/packagefiles/elfutils/meson.build
tobiasdiez: @eli-schwartz I very much understand the general sentiment of this policy and would love to have meson-based wraps for everything. But quite often one has an "external_project"-based wrap that is non-trivial. For example, because the project doesn't support out-of-tree builds or the embedding into meson needs other types of patches. In those situations, it would be convenient to be able to share and reuse those wraps, even though the overall experience is certainly not on par with a complete meson-based wrap. And even in the case the `meson.build` is a trivial wrapper around "unstable-external_project" it's still nice to not have too look up the exact url, calculate the hash etc.

Would it be possible to accept autoconf wraps, maybe prefixed with "external" or similar to emphasize that they are not "proper" meson wraps?

PR #1412, opened 2024-02-20
abseil: add 'enable_fuzzer' option

jmillan:
Enabling fuzzer on a project using abseil requires abseil itself to be compiled with the fuzzer options.

By creating a yieliding option abseil will take the value when the parent project enables it.


subprojects/packagefiles/abseil-cpp/meson.build
```diff
@@ -13,6 +13,18 @@ cpp = meson.get_compiler('cpp')
...
+cpp_args=[]
+link_args=[]
+
+if get_option('enable_fuzzer')
+  cpp_args += [
+    '-fsanitize=address,fuzzer',
+  ]
+  link_args += [
+    '-fsanitize=address,fuzzer',
+  ]
```
jpakkane, line 25: This code is reinventing functionality that is already in Meson. You should be able to achieve this by setting `b_sanitize=address,fuzzer'.

dcbaker: I hate to be the bearer of bad news... but no, you can't because https://github.com/mesonbuild/meson/pull/12648 has never landed and `fuzzer` isn't in our list of known good sanitizers. Maybe you want to weigh in on the design issue holding that PR up? :)

ibc: As an alternative, generic and simpler proposal, may I know what is wrong with adding extra_cpp_args and extra_link_args in meson_options.txt so the user can fine control them in very specific scenarios such as fuzzer (which only works on Linux+clang)?

jmillan: > As an alternative, generic and simpler proposal

this is basically what I did, but the current PR is more explicit which makes IMO more sense.

jmillan: So basically this PR mimics https://github.com/mesonbuild/meson/pull/12648 by using an option which is `enable_fuzzer`. This provides with a way to run a fuzzer on abseil and other projects depending on it..

I'm happy to remove this commit once the PR is merged.

jmillan: Indeed I've just tried using `b_sanitize=address` and it works.

I'm closing this PR.

PR #1414, opened 2024-02-21
ci: fix meson installs for macOS workflows

benoit-pierre:
Cf. https://github.com/benoit-pierre/wrapdb/actions/runs/7993113654/job/21828196022#step:5:18
```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try brew install
    xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-brew-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.
    
    If you wish to install a non-brew packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -160,6 +160,8 @@ jobs:
           find /usr/local/bin -lname '*/Library/Frameworks/Python.framework/*' -delete
           sudo rm -rf /Library/Frameworks/Python.framework/
           brew install --force python3 && brew unlink python3 && brew unlink python3 && brew link --overwrite python3
+          # Work around PEP 668 nonsense…
+          find /usr/local/Cellar/python* -name EXTERNALLY-MANAGED -print0 | xargs -0 rm -vf
       - name: Install packages
         run: |
           brew install ninja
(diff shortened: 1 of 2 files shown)


eli-schwartz (changes requested):
Thanks. Please include a comment in the yml about why we are doing this, including an assertion that the flag name is incorrect, the PEP breaks python and the flag is needed to *fix* python.

eli-schwartz:
It would probably be better to remove the EXTERNALLY-MANAGED file entirely... that's what meson.git does. Not a blocker though.

benoit-pierre:
Done.

neheb:
would be great to also update the Alpine workflow to fix a similar issue and switch to edge instead of an old alpine version.

eli-schwartz (approved):

benoit-pierre:
@neheb: edge or latest stable?

neheb:
hrm maybe just removing

```
branch: v3.18
```

is best. it defaults to latest-stable.

PR #1416, opened 2024-02-21
giflib: update to 5.2.2

benoit-pierre:
Not sure what to do about the Windows failures (VisualStudio): using `GifAsciiTable8x8 DATA` in the def file should work (at least it does when cross-compiling with MinGW)…


neheb:
I asked ChatGPT. It suggested replacing DATA with CONST.

benoit-pierre:
```
/usr/lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld: cannot export CONST: symbol not defined
```
For reference:
https://learn.microsoft.com/en-us/cpp/build/reference/exports?view=msvc-170

neheb:
hmmm next idea:

```
-Dextern="extern __declspec(dllexport)"
```

instead of the def file. Not ideal but could work.

neheb:
actually, try taking the def file from https://github.com/microsoft/vcpkg/blob/master/ports/giflib/exports.def

benoit-pierre:
According to the doc:

> All functions exported by giflib now have DGif, EGif, or Gif as a name prefix.

So I don't see what `FreeLastSavedImage` and `openbsd_reallocarray` are doing on the exports list.

The other missing functions should IMHO be exported, including `GifQuantizeBuffer`, which is actually used by many application (the Arch Linux package includes [this patch](https://gitlab.archlinux.org/archlinux/packaging/packages/giflib/-/blob/main/giflib-5.1.9-fix-missing-quantize-API-symbols.patch?ref_type=heads)).

Anyway, thanks for the link, I'll give the alternative syntax a whirl.

benoit-pierre:
No dice, again, works with MinGW, but not VisualStudio…

neheb:
looking at the actual code,

DGifDecreaseImageCounter
DGifGetGifVersion
DGifGetImageHeader

are not marked as extern in header files.

neheb:
From NEWS:

```
The undocumented and deprecated GifQuantizeBuffer() entry point
has been moved to the util library to reduce libgif size and attack
surface. Applications needing this function are couraged to link the
util library or make their own copy.

The following obsolete utility programs are no longer installed:
gifecho, giffilter, gifinto, gifsponge. These were either installed in
error or have been obsolesced by modern image-transformation tools
like ImageMagick convert. They may be removed entirely in a future
release.
```

That patch is a bit wrong.

Those executables should probably not be built by default.

benoit-pierre:
> looking at the actual code,
> 
> DGifDecreaseImageCounter DGifGetGifVersion DGifGetImageHeader
> 
> are not marked as extern in header files.

`DGifDecreaseImageCounter` is not in the header, so it goes away. But the other are still declared (as "legacy").

benoit-pierre:
> That patch is a bit wrong.
> 
> Those executables should probably not be built by default.

`gifecho`, for example, is only built for running the tests (it's indeed not installed).

benoit-pierre:
Same with the other programs.

Regarding `GifQuantizeBuffer`, see https://sourceforge.net/p/giflib/bugs/142/.

benoit-pierre:
Anyway, I'll just drop those tests that can't be build on Windows (and remove `GifAsciiTable8x8` from the exports) for now.

benoit-pierre:
Note: `gifcolor` needs `GifAsciiTable8x8` too.

benoit-pierre:
OK, should be good to go now.

PR #1420, opened 2024-02-24
libxml2: Fix missing libreadline header checks

amyspark:
Hi,

This is a PR to fix an issue I experienced when applying the wrap in macOS. libreadline is present there, but the original would (correctly) reject it because it also checks for the presence of two header files.

See https://github.com/GNOME/libxml2/blob/a581f65194212f183dcbe77da44657d477a4758d/configure.ac#L883-L897


amyspark:
cc @nirbheek

PR #1421, opened 2024-02-24
flac: Fix misapplication of /arch:AVX2 to the whole of libFLAC

amyspark:
Hi,

This is a PR to narrow the application of `/arch:AVX2` to libFLAC. With the current version, the library would cause a runtime crash in boxes whose CPU did not support AVX.

To fix this, I've pulled the affected files into a separate static library, and then `link_whole` it together with the rest. This mimics, to an extent, what upstream does: https://github.com/xiph/flac/blob/72787c3fee472809fb3f819558c7176066f0ee44/src/libFLAC/CMakeLists.txt#L22

cc @nirbheek


subprojects/packagefiles/flac/src/libFLAC/meson.build
```diff
@@ -4,10 +4,25 @@ else
   extra_args = ['-DFLAC__NO_DLL']
 endif
 
+avx2_args = []
+
 if cc.get_argument_syntax() == 'msvc' and cdata.get('WITH_AVX')
-  extra_args += '/arch:AVX2'
+  avx2_args += '/arch:AVX2'
```
neheb, line 10: hrm so this is MSVC only. Does AVX2 support work witth GCC/Clang?

amyspark: It employs [function multiversioning](https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/Function-Multiversioning.html) for non-MSVC: https://github.com/xiph/flac/blob/72787c3fee472809fb3f819558c7176066f0ee44/src/libFLAC/include/private/cpu.h#L96

neheb: at runtime. interesting.

PR #1423, opened 2024-02-26
lexy: new wrap

Totto16:
This is a mostly header only project from [here](https://github.com/foonathan/lexy)

It ports most features from cmake, but not all of these below:
- benchmarks: only some, not needed for a wrap
- tests: 2 tests are some complicated ones, that generate many things via cmake, they are not required imo, to use this as wrap

But it's usable since it ports everything needed


Totto16:
I noticed, that the wrap for doctest, which is only a clone of the upstream [meson.build](https://github.com/doctest/doctest/blob/master/meson.build#L3) adds the dir "/doctest" to the include directories, but the [(https://github.com/doctest/doctest/blob/master/CMakeLists.txt#L53) adds the root ("/") of the project as include directory, so this projects uses doctests includes like this: 
```cpp
#include <doctest/doctest.h>
```
 which is perfectly valid with CMake, but meson fails, since it expects: 
```cpp
#include <doctest.h>
```

The best approach imo is to fix this upstream, but are there any other options, to fix this in a good way, if that wouldn't be possible (which it should be, but who know, until this [PR](todo) is accepted and a new release includes the fix?) 
We can't pacth (diff_files field) the upstream meson.build file here... 🤔

Totto16:
I noticed, that the wrap for doctest, which is only a clone of the upstream [meson.build](https://github.com/doctest/doctest/blob/master/meson.build#L3) adds the dir "/doctest" to the include directories, but the [CMake Defintion](https://github.com/doctest/doctest/blob/master/CMakeLists.txt#L53) adds the root ("/") of the project as include directory, so this projects uses doctests includes like this: 
```cpp
#include <doctest/doctest.h>
```
 which is perfectly valid with CMake, but meson fails, since it expects: 
```cpp
#include <doctest.h>
```

The best approach imo is to fix this upstream 8in doctest), but are there any other options, to fix this in a good way, if that wouldn't be possible (see https://github.com/doctest/doctest/issues/792)
We can't patch (diff_files field) the upstream meson.build file here... 🤔 

Fixing it in the lexy upstream isn't possible that easily, due to usage of cmake there, and I think, the discussion would be the same as in https://github.com/doctest/doctest/issues/792 🤔 

@eli-schwartz Any opinion on this, since you are the creator of https://github.com/doctest/doctest/issues/670

PR #1425, opened 2024-02-26
fuse: add libudev-zero to Alpine CI

neheb:


jpakkane:
> subprojects/libfuse-fuse-2.9.9/util/meson.build:33:9: ERROR: Dependency "udev" not found, tried pkgconfig and cmake

neheb:
yeah this is wrong.

PR #1429, opened 2024-02-28
libopenjp2: update to 2.5.2

bgilbert:
OpenJPEG no longer supports C89 compilers, so `stdint.h` and `inttypes.h` are now mandatory.  Since it doesn't depend on GNU extensions, switch to `c99` to avoid a warning on Meson &ge; 1.3.0.

Upstream now includes `-DOPJ_STATIC` in the pkg-config `Cflags.private`.  The Meson `pkgconfig` module doesn't provide a way to do that, so conditionally include it in `Cflags` when we've enabled the define.


neheb:
Build all ignores warnings, so there’s that.

bgilbert:
I've switched to `c99` outright.  The library portion was already built that way.

PR #1433, opened 2024-03-02
libopenjp2: remove dead code

bgilbert:
Almost two thirds of the `libopenjp2` wrap is commented out or unreachable. Remove this code and the corresponding options.  Most of the commented-out code is unconverted CMake directives from the original commit 847561a11f in April 2020 and hasn't been maintained for new releases.

Drop build directives for the Java binding, which have some uncommented lines but still contain CMakeisms and can't have ever worked.

Some of the other options currently don't build successfully, but leave those alone for now.


jpakkane:
Nice cleanup.

PR #1436, opened 2024-03-02
bring godot to wrapdb 

v-for-vandal:
Hi!
This is part request-for-comments, part PR. I have a wrapdb files for building godot gdextension module with meson. And I would like to commit it to the wrapdb, however there are a few complications:
1. Quite some time ago I found on the internet a wrap file for godot-3. A couple years later godot 4.2 was released and I updated/tuned the wrap files for it. However, I could not find the original author of this file nor can I find again the source of original file. As such, I have no idea what is the license for it. And I can't leave credit for original author :(  What is correct thing to do in this situation ?
2. I can't verify that wrap file will work on every architecture out here. I can only verify basic amd64 setup. As such, there is a possibility of problems on other platforms.
3. Being a game engine, godot maintains several version in active use. This one is for 4.2 . Say, tomorrow another person comes with wrapfile for godot 4.1 . How does one make both wrapfiles available  to users ?

This wrap files themselves are a bit work-in-progress, I will need to support install/not install option, and add a bit more documentation, but for a first draft it will suffice.


subprojects/packagefiles/godot-cpp/meson.build
```diff
@@ -0,0 +1,975 @@
...
+  default_options: ['cpp_std=c++17'])
+
+# Alas, godot has a lot of these
+add_project_arguments([
+  '-Wno-unused-parameter',
+  '-Wno-extra-semi',
+  '-Wno-gnu-anonymous-struct',
+  '-Wno-nested-anon-types',
+
+  ], language: 'cpp', )
```
jpakkane, line 14: You do not need brackets here.

jpakkane:
> This one is for 4.2 . Say, tomorrow another person comes with wrapfile for godot 4.1 . How does one make both wrapfiles available to users

The common approach is to have a name-suffixed dependency name. If 4.1 and 4.2 are not compatible with each other and both need to "live" at the same time, then the dep names could be `godot41` and `godot42` or something similar. If upstream has a naming convention, follow that, otherwise anything sensible should be fine.

v-for-vandal:
@jpakkane Hi. I think PR is done.
The only remaining question is about license - as I mentioned, I initially took meson.build and meson-generator-binding somewhere from internet. Considering that one is python 3-liner and second one was straightforward simple meson.build, and that I modified them both, I hope that is enough.

PR #1438, opened 2024-03-04
fmt: User proper SPDX license, which is `MIT WITH fmt-exception`

dcbaker:
It's MIT, with a special exception. Fortunately SPDX has an identifier
for the fmt-exception. Importantly, it is not one of the BSD licenses,
it's the similarly permissive MIT license.


dcbaker:
Derp, FMT is MIT with an exception, I was looking at google test...

PR #1443, opened 2024-03-08
glib: update to 2.80.0

bgilbert:


bgilbert:
Upstream now uses a construct which causes warnings on 1.3.0+, but where a replacement construct isn't available until 1.4.0 (https://github.com/mesonbuild/meson/issues/5273).  According to https://github.com/mesonbuild/meson/issues/5273#issuecomment-1852043352, this is fine, because warnings won't break anything.  :man_shrugging:

PR #1445, opened 2024-03-11
fix(type_safe): /Wd4800 -> /wd4800

judemille:
the capitalization is important here, apparently.
MSVC continues to confuse me sometimes.


eli-schwartz:
Commit message is confusing and misleading and doesn't really explain what's being changed. :p

judemille:
Fair enough lol. I can amend it.

PR #1462, opened 2024-03-21
libdwarf: new wrap

judemille:
have to manually replace a couple meson.build files. davea42/libdwarf-code#234 will remove that need.

otherwise, pretty standard.


meator:
Hey, is this PR still active? If not, I'd be willing to take over. It looks like https://github.com/davea42/libdwarf-code/pull/234 is already a part of a released version of libdwarf, so this wrap could become a "regular wrap" without a vendored meson build system.

PR #1466, opened 2024-03-27
curl: update to 8.7.1

### LINKED ISSUE #1454: Add ability to enable WebSocket support for curl (and maybe other experimental features?)
I know the initial PR (https://github.com/mesonbuild/wrapdb/pull/1149) said "advanced / experimental options are not supported", but I was wondering if these could be added so we can use them in our code, at our own risk, and off by default of course.

I was looking to make the changes locally myself in order to test, but the [`meson.build` for curl](https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/curl/meson.build) is pretty big and I don't know exactly where the changes would fit.

benoit-pierre:
Additionally, add Web Sockets support (experimental, disabled by default). Closes #1454.


benoit-pierre:
Looks like the build in broken on platforms without `getsockname` (e.g. Windows), Cf.: https://github.com/curl/curl/issues/13210.

benoit-pierre:
Let's wait for 8.7.2.

PR #1469, opened 2024-03-28
wayland-protocols: update to 1.35

MonterraByte:


MonterraByte:
Updated to v1.35.

(Note that the CI failures on Windows and MacOS are expected.)

PR #1471, opened 2024-04-01
libscpi: new wrap

jakergrossman:


subprojects/packagefiles/libscpi/examples/meson.build
```diff
@@ -0,0 +1,45 @@
+common_args = [
+  '-Wextra',
```
jpakkane, line 2: Do not add this by hand, it should come from the warning level setting.

jakergrossman:
Does this repository squash PRs? I couldn't find the answer.

jpakkane:
A single commit is preferred unless the MR changes a lot of things.

jakergrossman:
- Removed examples from wrap
- Removed hand-added compiler flags

subprojects/packagefiles/libscpi/libscpi/meson.build
```diff
@@ -0,0 +1,29 @@
...
+install_headers(libscpi_hdrs, subdir : 'scpi')
+
+pkg_mod = import('pkgconfig')
+pkg_mod.generate(
+  name : 'libscpi',
+  filebase : 'libscpi',
+  description : 'Open Source SCPI device library',
+  subdirs : 'scpi',
+  libraries : libscpi_lib,
+  version : '2.3',
```
benoit-pierre, line 26: `version` default to the project version if not specified.

subprojects/packagefiles/libscpi/libscpi/meson.build
```diff
@@ -0,0 +1,29 @@
...
+
+# Make this library usable from the system's
+# package manager.
+subdir('inc/scpi')
+install_headers(libscpi_hdrs, subdir : 'scpi')
+
+pkg_mod = import('pkgconfig')
+pkg_mod.generate(
+  name : 'libscpi',
+  filebase : 'libscpi',
```
benoit-pierre, line 22: `filebase` is not needed (same as `name`).

subprojects/packagefiles/libscpi/libscpi/meson.build
```diff
@@ -0,0 +1,29 @@
+subdir('src')
+
+libscpi_lib = both_libraries('scpi',
```
benoit-pierre, line 3: Why not `library`? (And let the user decide with `--default-library=…`)

benoit-pierre: For the tests?

jakergrossman: Yes, for the tests. When I tried to use a shared library, the tests could not compile because they refer to internal symbols that are not exported in the SO

subprojects/packagefiles/libscpi/libscpi/meson.build
jakergrossman: ack

subprojects/packagefiles/libscpi/libscpi/meson.build
jakergrossman: ack

subprojects/packagefiles/libscpi/libscpi/meson.build
jakergrossman: Is there a portable way to say "export all symbols?"

benoit-pierre: How are symbols exported? You're not passing any specific arguments for building a static or shared variant.

benoit-pierre: Right, `__attribute__((visibility ("hidden")))` is hiding some symbols needed for the tests.

What I would do: always build an internal static library (`libscpi_static`), and use it for the tests and to build the exported library (using `library(…, link_with: libscpi_static, …)`:

```diff
--- a/libscpi/meson.build
+++ b/libscpi/meson.build
@@ -1,9 +1,19 @@
 subdir('src')
 
-libscpi_lib = both_libraries('scpi',
+libscpi_inc = include_directories('inc')
+
+libscpi_static = static_library(
+  'scpi_static',
   libscpi_srcs,
-  include_directories : 'inc',
-  install : true,
+  include_directories: libscpi_inc,
+)
+
+libscpi_lib = library(
+  'scpi',
+  libscpi_srcs,
+  link_with: libscpi_static,
+  include_directories: libscpi_inc,
+  install: true,
 )
 
 libscpi_dep = declare_dependency(
@@ -23,7 +33,6 @@
   description : 'Open Source SCPI device library',
   subdirs : 'scpi',
   libraries : libscpi_lib,
-  version : '2.3',
 )
 
 subdir('test')
--- a/libscpi/test/meson.build
+++ b/libscpi/test/meson.build
@@ -12,13 +12,13 @@
   'test_scpi_utils',
 ]
 
-libscpi_static = libscpi_lib.get_static_lib()
+# libscpi_static = libscpi_lib.get_static_lib()
 
 foreach t : test_names
   test_exe = executable(t, t+'.c',
     dependencies : [ cunit_dep ],
     link_with : libscpi_static,
-    include_directories : '../inc',
+    include_directories : libscpi_inc,
   )
   test(t, test_exe)
 endforeach
```

benoit-pierre:
> * Removed examples from wrap

Why? Assuming they only the use the public API, those are a good way to check the resulting shared library or DLL export all the required symbols.

subprojects/packagefiles/libscpi/libscpi/meson.build
neheb: there was a different way to do this. export_all_objects or something.

neheb: https://mesonbuild.com/Reference-manual_returned_build_tgt.html#build_tgtextract_all_objects

benoit-pierre: Yep, forgot about that.

jakergrossman:
> > * Removed examples from wrap
> 
> Why? Assuming they only the use the public API, those are a good way to check the resulting shared library or DLL export all the required symbols.

I think I was having issues with the TCP examples, but I just added them back guarded them with a header check for 'netinet/in.h' for those two examples.

subprojects/packagefiles/libscpi/libscpi/meson.build
jakergrossman: Yes, that works great, added.

jakergrossman:
Tested with the following in the root of `wrapdb`:

```
rm -rf builddir &&
meson subprojects purge --confirm &&
meson setup builddir/ -Dwraps=libscpi --default-library=shared &&
meson compile -C builddir -v &&
meson test -C builddir
```

PR #1474, opened 2024-04-07
libnpupnp: update to 6.1.1

neheb:


neheb:
I'll update once new version gets released. Autotools goes away (good riddance).

PR #1476, opened 2024-04-09
ci: run macOS builds on both arm64 and x86_64

bgilbert:
The `macos-latest` runner image has historically been x86_64, but it's [currently migrating](https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/) from `macos-12` to `macos-14` which [runs on arm64 hardware](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/).  It seems useful to get test coverage on both architectures, so arrange that for now.  Presumably we'll eventually have to drop macOS x86_64 tests when the `macos-12` image goes out of support.

Homebrew installs into `/opt/homebrew` on arm64 Macs.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -145,10 +145,13 @@ jobs:
 
   MacOS:
     if: github.event_name != 'schedule' || github.repository == 'mesonbuild/wrapdb'
-    runs-on: macos-latest
+    runs-on: ${{ matrix.platform == 'arm64' && 'macos-14' || 'macos-12' }}
     env:
       TEST_BUILD_ALL: 1
       TEST_FATAL_WARNINGS: ${{ github.event.inputs.fatal_warnings }}
+    strategy:
+      matrix:
+        platform: ['arm64', 'x86_64']
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build_all.yml
```diff
@@ -158,11 +161,15 @@ jobs:
...
+          case "${{ matrix.platform }}" in
+          arm64) homebrew=/opt/homebrew ;;
+          x86_64) homebrew=/usr/local ;;
+          esac
+          find $homebrew/bin -lname '*/Library/Frameworks/Python.framework/*' -delete
           sudo rm -rf /Library/Frameworks/Python.framework/
           brew install --force python3 && brew unlink python3 && brew unlink python3 && brew link --overwrite python3
           # Work around PEP 668 nonsense…
-          find /usr/local/Cellar/python* -name EXTERNALLY-MANAGED -print0 | xargs -0 rm -vf
+          find $homebrew/Cellar/python* -name EXTERNALLY-MANAGED -print0 | xargs -0 rm -vf
```
benoit-pierre, line 172: Can't we get the install location from homebrew (`brew --prefix` maybe?)

eli-schwartz: Probably. In meson.git's CI we add `$(brew --prefix llvm)/bin` to PATH, for example.

bgilbert: Ah yes, good call.  Fixed.

.github/workflows/build_all.yml
```diff
@@ -145,10 +145,13 @@ jobs:
 
   MacOS:
     if: github.event_name != 'schedule' || github.repository == 'mesonbuild/wrapdb'
-    runs-on: macos-latest
+    runs-on: ${{ matrix.platform == 'arm64' && 'macos-14' || 'macos-12' }}
```
neheb, line 148: hmm? Would be simpler to use 14 unconditionally.

neheb:
Did you run this in your actions instance? Would be great to know what to expect.

.github/workflows/build_all.yml
bgilbert: AFAICT there's no x86_64 for 14.

bgilbert:
> Did you run this in your actions instance? Would be great to know what to expect.

Yes, both workflows.  The builds ran to completion, though I didn't compare build failures between x86_64 and arm64, since any new failures seem like problems with individual wraps.

neheb:
Just saw it.

```
      1 2024-04-09T21:10:08.5332840Z 5 skipped: arduinocore-avr, google-woff2, graphite2, libgpiod, mocklibc
      2 2024-04-09T21:10:08.5333890Z 18 failed: backward-cpp, blueprint-compiler, cairo, curl, glib, glib-networking, grpc, imguizmo, libffi, libffmpegthumbna        iler, liblangtag, libpsl, libsndfile, opus, orcus, sdl2_image, sdl2_mixer, unit-system

```

some of those failures are quite interesting...

PR #1477, opened 2024-04-10
libarchive: update to 3.7.3

benoit-pierre:
Additionally:
- fix `-Diconv=enabled` support: error out when not available
- add some missing checks


subprojects/packagefiles/libarchive/meson.build
```diff
@@ -276,62 +286,86 @@ else
   endif
 endif
 
-if meson.version().version_compare('>= 0.60')
-  iconv_dep = dependency('iconv', disabler: true, required: get_option('iconv'))
+if not get_option('iconv').disabled() and meson.version().version_compare('>= 0.60')
```
dcbaker, line 289: This is turning `enabled()` into `auto()`, since this will automatically be disabled if the user has an old Meson version. Why don't we just bump the required version to 0.60, that's still pretty old at this point?

benoit-pierre: > This is turning enabled() into auto(), since this will automatically be disabled if the user has an old Meson version.

No, if `-Diconv=enabled` is used and the meson version is too old this will result in an error: `Feature iconv cannot be enabled`, instead of silently disabling iconv support (the previous behavior).

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -276,23 +276,34 @@ else
...
+  # Prefer libiconv if available (for `locale_charset`).
+  iconv_dep = dependency('iconv', disabler: true, method: 'system', required: false)
```
eli-schwartz, line 291: This doesn't really make sense. If libiconv is available, its header will override the builtin iconv header and cause the builtin version to fail to link. Hence meson will detect that builtin isn't available, and system is.

eli-schwartz (changes requested):

subprojects/packagefiles/libarchive/meson.build
benoit-pierre: So `dependency('iconv')` should prefer `libiconv`?

eli-schwartz: Yes. It will always return libiconv, if libiconv is available.

eli-schwartz: I seem to recall that when the initial update to use dependency() was made, I reviewed it and suggested a proper fallback be done, but the PR author didn't want to bother...

benoit-pierre: I'm on Arch Linux, libiconv is installed, this is what I get in the meson setup logs (using the previous wrapdb version):
```
Run-time dependency iconv found: YES
Running compile:
Working directory:  […]/libarchive/.b/native/meson-private/tmpx27f1who
Code:
 
        #define locale_charset meson_disable_define_of_locale_charset
        
        #include <limits.h>
        #undef locale_charset
        
        #ifdef __cplusplus
        extern "C"
        #endif
        char locale_charset (void);
        
        #if defined __stub_locale_charset || defined __stub___locale_charset
        fail fail fail this function is not going to work
        #endif
        
        int main(void) {
          return locale_charset ();
        }
-----------
Command line: `cc […]/libarchive/.b/native/meson-private/tmpx27f1who/testfile.c -o […]/libarchive/.b/native/meson-private/tmpx27f1who/output.exe -D_FILE_OFFSET_BITS=64 -O0` -> 1
stderr:
/usr/bin/ld: /tmp/ccvwIHDz.o: in function `main':
testfile.c:(.text+0x5): undefined reference to `locale_charset'
collect2: error: ld returned 1 exit status
```
and after changing to use the `system` method the `locale_charset` test is successful because this time `-liconv` is used:
```
Command line: `cc […]/libarchive/.b/native/meson-private/tmplpmz9_mh/testfile.c -o /home/bpierre/dev/libarchive/.b/native/meson-private/tmplpmz9_mh/output.exe -D_FILE_OFFSET_BITS=64 -O0 -liconv` -> 0
```

eli-schwartz: > That's not what I noticed locally: the `config.h` generated with meson does not `HAVE_LOCALE_CHARSET` set, unlike when using cmake. The `has_function` check is run without an iconv library, so it fails.

Can I see the log file for that?

> The meson documentation says libc support is preferred over libiconv, no?

The reason I wrote it to check libc first is because:
- it's more common and thus faster
- if libiconv is available as a shared library but not as a header (unlikely) then code will compile using the libc version but link to an unused -liconv, and the libiconv check will be "found" but not factually correct.

From my own experimentation, installing libiconv always causes the builtin check to report a compiler error when linking to libc, because iconv.h from libiconv comes first in the header search path and makes any code using the iconv functions compile into the symbol "libiconv_open" instead of "iconv_open". libc doesn't have that symbol, thus linking fails, whereas the system (-liconv) check succeeds because it *does* have libiconv_open as a symbol.

eli-schwartz: > I'm on Arch Linux, libiconv is installed, this is what I get in the meson setup logs (using the previous wrapdb version):

I need the full log file because the part I want to see is the part you didn't include.

benoit-pierre: There's this line in the PKGBUILD for libiconv:
```shell
mv -v "$pkgdir"/usr/include/{iconv.h,libiconv.h}
```
So no overridden `iconv.h` header.

> Can I see the log file for that?

See logs above.

eli-schwartz: > There's this line in the PKGBUILD for libiconv:
> 
> ```shell
> mv -v "$pkgdir"/usr/include/{iconv.h,libiconv.h}
> ```
> 
> So no overridden `iconv.h` header.
> 
> > Can I see the log file for that?
> 
> See logs above.

I don't understand how this is logically supposed to work since software isn't going to `#include <libiconv.h>`, that's why GNU libiconv installs the header the way it does.

subprojects/packagefiles/libarchive/meson.build
dcbaker: Okay, I see where you're handling that now. That seems like an odd place to handle it, instead of here.

subprojects/packagefiles/libarchive/meson.build
benoit-pierre: > I need the full log file because the part I want to see is the part you didn't include.

The test run for iconv is:
```c
Code:
 #include <iconv.h>

int main() {
    iconv_open("","");
}
```

And when using the `system` method:
```c
        #ifdef __has_include
         #if !__has_include("iconv.h")
          #error "Header 'iconv.h' could not be found"
         #endif
        #else
         #include <iconv.h>
        #endif
```
Which means `libiconv.h` is never used. And not used by libarchive either…

benoit-pierre: > I don't understand how this is logically supposed to work since software isn't going to #include <libiconv.h>, that's why GNU libiconv installs the header the way it does.

Yep, this seem clearly wrong.

benoit-pierre: In the mean time, I propose changing the code too:
```python
  iconv_dep = dependency('iconv', disabler: true, required: get_option('iconv'))
elif get_option('iconv').enabled()
  error('Feature iconv cannot be enabled')
else
  iconv_dep = dependency('', required: false)
endif
if iconv_dep.found()
  cdata.set('HAVE_ICONV', 1)
  cdata.set(
    'ICONV_CONST',
    cc.compiles(
      '''
    #include <iconv.h>
    size_t iconv(iconv_t, const char**, size_t*, char**, size_t*);
    int main(){}
    ''',
      dependencies: iconv_dep,
    ) ? 'const' : '',
  )
  deps += iconv_dep
else
  cdata.set('ICONV_CONST', '')
endif
```
So there's an error when `-Diconv=enabled` is used and meson < 0.60 is used, agreed?

eli-schwartz: Yes. The alternative is to somewhat painfully handroll a cc.links() test case to check for iconv support even on older mesons.

eli-schwartz: By the way, if you installed libiconv on Arch just to test support in the wrap for it, I would suggest you test the wrap in an Alpine Linux chroot instead. That uses musl, which has a libc version of iconv but commonly gets replaced by external libiconv available as an Alpine package and actually used in practice.

This case is known to work -- the current design of the meson dependency was guided by the need to get the meson testsuite passing on Alpine.

PR #1478, opened 2024-04-15
cpp-semver: new wrap

Totto16:
This is a header only library from [here](https://github.com/z4kn4fein/cpp-semver)

### DIFF
--- a/subprojects/packagefiles/cpp-semver/meson.build
+++ b/subprojects/packagefiles/cpp-semver/meson.build
@@ -1,0 +1,26 @@
+project(
+  'cpp-semver',
+  'cpp',
+  version: '0.3.3',
+  license: 'MIT',
+  meson_version: '>=1.2.0',
+  default_options: {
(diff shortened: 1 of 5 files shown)


subprojects/packagefiles/cpp-semver/meson.build
```diff
@@ -0,0 +1,24 @@
...
+
+install_subdir(
+  'include',
+  install_dir: get_option('includedir'),
+)
+
+cpp_semver_dep = declare_dependency(
+  include_directories: include_directories('include'),
+  version: meson.project_version(),
+)
```
dcbaker, line 20:
```suggestion
meson.override_dependency('cpp-semver', cpp_semver_dep)
```

Totto16: I always thought, that was bad practice to do that, but now I know otherwise, thnaks 👍🏼

PR #1480, opened 2024-04-20
gdk-pixbuf: update to 2.42.12

bgilbert:
Various built-in loaders are now disabled by default, so enable them in CI.

Add Python packaging module to MSYS because glib needs it.


neheb:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/wrapdb/wrapdb/_build/subprojects/glib-2.80.0/gio/glib-compile-resources'
```

wonder why...

bgilbert:
The `glib-compile-resources` error looks like a race condition that should be fixed by [this upstream PR](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/166).  The missing `packaging` package is because glib 2.80+ added a dependency on it; this also caused CI failures in #1443 but wasn't addressed there.  #1481 attempts to add support for wrap-specific Python dependencies.

eli-schwartz:
The upstream release is also badly broken due to https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/164

bgilbert:
Yup, was just discovering that.  The `others` option is new and wasn't added to CI.  I'll send an MR for the CI portion.

bgilbert:
Looks like the MSYS failure is inside the glib dependency.

PR #1481, opened 2024-04-21
Support installing Python packages in CI

bgilbert:
glib 2.80+ build-requires the Python `packaging` package, which isn't preinstalled in the MSYS runners.  In this specific case, we could install it from an MSYS package, but some Python dependencies might not be available from all platform package managers.  Support installing arbitrary Python packages from pip on all platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ meson compile -C builddir
     on MacOS CI runners.
   - `choco_packages`: (Optional) List of extra packages that will be installed
     on Windows CI runners.
+  - `python_packages`: (Optional) List of extra Python packages that will be
+    installed on all CI runners.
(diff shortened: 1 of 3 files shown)


neheb (approved):

bgilbert:
...actually it turns out that the problem was the MSYS runners, not the Visual Studio ones, and `packaging` is available from MSYS.  I still think this is useful, but it isn't strictly necessary for the original use case.  I'll update the PR accordingly.

eli-schwartz:
If there's no particular need for it right now I'd suggest sticking with msys2_packages -- it is anyways a better idea to use the native package manager as that is guaranteed to provide a prebuilt copy that doesn't need compiling from an sdist, regardless of whether the PyPI package provides wheels.

bgilbert:
Okay, sounds good.

PR #1482, opened 2024-04-21
libffi: fix Meson version warning

bgilbert:
Bump the minimum Meson version for calling `cc.preprocess(compile_args: 'x')` to 1.3.2 to fix a warning:

    subprojects/libffi-3.4.4/src/meson.build:48: WARNING: Project targets '>=0.64.0' but uses feature introduced in '1.3.2': compile_args arg in compiler.preprocess. compile_args were ignored before this version


jpakkane:
```
In file included from ../subprojects/libffi-3.4.4/testsuite/libffi.closures/stret_medium.c:10:
In file included from ../subprojects/libffi-3.4.4/testsuite/libffi.closures/ffitest.h:1:
../subprojects/libffi-3.4.4/testsuite/libffi.closures/../libffi.call/ffitest.h(156,9): warning: 'PRIu64' macro redefined [-Wmacro-redefined]
#define PRIu64 "I64u"
        ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\inttypes.h(174,9): note: previous definition is here
#define PRIu64       "llu"
```

Seems suspicious at least.

neheb:
The actual code:

```
/* MSVC kludge.  */
#if defined _MSC_VER
#if !defined(__cplusplus) || defined(__STDC_FORMAT_MACROS)
#define PRIuPTR "lu"
#define PRIu8 "u"
#define PRId8 "d"
#define PRIu64 "I64u"
#define PRId64 "I64d"
#endif
#endif
```

bgilbert:
That seems like an unrelated upstream bug, though?  The purpose of this PR is to fix the build with `--fatal-meson-warnings`.

PR #1484, opened 2024-04-21
zstd: don't add `-pthread` to static linking flags on Windows

bgilbert:
Meson always returns `-pthread` in `dependency('threads')` on non-MSVC compilers.  On Windows, zstd uses Windows threading primitives, so we don't need this.  Avoid adding `-pthread` to libzstd's link flags, either as a Meson subproject or via pkg-config `Libs.private`, so the application doesn't inadvertently depend on winpthreads.

Merged upstream in https://github.com/facebook/zstd/pull/3931.


eli-schwartz:
Thanks!

PR #1492, opened 2024-04-26
libarchive: update to 3.7.4

benoit-pierre:
The macOS build is fine on `x86_64`, but borked on `arm64`:
```
ld: library 'ssl' not found
```
Same issue when building curl (#1466).


neheb:
brew bug?

benoit-pierre:
> brew bug?

Not sure, looking into it.

> edit: does libarchive need libssl or just libcrypto?

Good catch, will amend to link with `libcrypto` only.

benoit-pierre:
Looks like it is a brew bug:
```bash
# the include directory is correct:
▸ pkg-config --cflags libcrypto
-I/opt/homebrew/Cellar/openssl@3/3.3.0/include
# but the libraries search path is not:
▸ pkg-config --libs libcrypto
-L/opt/homebrew/Cellar/openssl@3/3.3.0 -lcrypto
#
▸ ls -l /opt/homebrew/Cellar/openssl@3/3.3.0
total 1752
-rw-r--r--   1 runner  admin     990 Apr  9 12:12 AUTHORS.md
-rw-r--r--   1 runner  admin  782003 Apr  9 12:12 CHANGES.md
-rw-r--r--   1 runner  admin    1256 Apr 22 18:13 INSTALL_RECEIPT.json
-rw-r--r--   1 runner  admin   10175 Apr  9 12:12 LICENSE.txt
-rw-r--r--   1 runner  admin   84478 Apr  9 12:12 NEWS.md
-rw-r--r--   1 runner  admin    8182 Apr  9 12:12 README.md
drwxr-xr-x   4 runner  admin     128 Apr 22 18:13 bin
drwxr-xr-x   3 runner  admin      96 Apr  9 12:12 include
drwxr-xr-x  12 runner  admin     384 Apr 22 18:13 lib
drwxr-xr-x   4 runner  admin     128 Apr  9 12:12 share
```

benoit-pierre:
Reinstalling openssl fix it. ¯\\_(ツ)_/¯

.github/workflows/sanity_checks.yml
```diff
@@ -154,6 +154,12 @@ jobs:
           brew install ninja
           python3 -m pip install --pre meson
 
+      # The pkg-config file for openssl is borked: the libraries search
```
neheb, line 157: I think disable openssl in ci_config.json with this comment.

benoit-pierre: Rather than fix it for everybody?

benoit-pierre: If we wait a couple of days, maybe it won't be an issue anymore: https://github.com/actions/runner-images/issues/9755#issuecomment-2081579531.

neheb: sure. I don't know if editing sanity_checks.yml for a specific Homebrew bug is the correct solution here.

benoit-pierre: So it's still not fixed…

I disagree that disabling openssl in `ci_config.json` is the right approach. It's a dependency of at least 8 wraps, IMHO fixing it in `sanity_checks.yml` so all wraps benefit from it is the better solution.

Alternatively, ignore the macOS arm64 failure and merge.

neheb: Yeah why not. It's experimental anyways.

PR #1494, opened 2024-04-27
oatpp, oatpp-test: update to 1.3.1

Totto16:
This updates `oatpp` and `oatpp-test` to the latest upstream release.

This fixes a few issue with this wrap, we had earlier, like the incorrect tests on windows (see #1031)


subprojects/packagefiles/oatpp/meson.build
```diff
@@ -91,6 +92,13 @@ if is_msvc_or_similar
 endif
 
 
+if cpp.get_id() == 'clang' or cpp.get_id() == 'gcc'
```
neheb, line 95: Could use get_supported_arguments

PR #1495, opened 2024-04-29
tree-sitter: fix meson.build

Gottox:
this fixes a c&p mistake as well as setting the version for the project so that the pkg-config file is logically equal to the one shipped by tree-sitter.


jpakkane:
You need to add a new release to releases.json.

PR #1498, opened 2024-04-30
Fix soundtouch pkg-config include paths

neodesys:
When built with autotools or cmake, the libsoundtouch project defines the path '-I${includedir}/soundtouch' in its .pc file. This is the same path that is also used in Ubuntu and Fedora Linux distributions packages. This patch adds this path to the .pc file generated by the meson project.


tp-m:
cc @amyspark @nirbheek

dcbaker:
Does upstream only put the soundtouch/ subdir in, or does it do both? We should match what upstream does

neodesys:
The libsoundtouch project only specifies the -I${includedir}/soundtouch path and nothing more. But I think it may be fair to specify both paths in case that some projects have been using this wrap file with the previous wrong path.

dcbaker:
Maybe others have a different opinion, but I don't think we should keep doing something that upstream doesn't do. The point of these is to match their behavior, not invent our own.

neodesys:
:+1: if the default behavior is to always keep the upstream project default, we should keep to it. I've updated the PR.

dcbaker (approved):

PR #1503, opened 2024-05-06
fftw3: Add new package

antoniovazquezblanco:


subprojects/fftw3.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = fftw-3.3.10
+source_url = http://fftw.org/fftw-3.3.10.tar.gz
+source_filename = fftw-3.3.10.tar.gz
+source_hash = 56c932549852cddcfafdab3820b0200c7742675be92179e59e6215b340e26467
+patch_directory = fftw3
+method = cmake
+
```
eli-schwartz, line 8: Can't use cmake, the wrapdb is for native meson builds.

eli-schwartz (changes requested):

tools/sanity_checks.py
```diff
@@ -36,6 +36,9 @@
         'extract.mk',
         'rewrite.mk',
     ],
+    'fftw3': [
+        'CMakeLists.txt',
```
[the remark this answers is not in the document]
antoniovazquezblanco: This is to fix what fftw devels have just efficiently merged in https://github.com/FFTW/fftw3/pull/354 but will probably take a long long time to get released. It fixes VS clang-cl builds.

subprojects/fftw3.wrap
antoniovazquezblanco: Unfortunatelly, fftw uses magic to generate (yes, you heard it right) and build their code. I understand the "meson only" build policy but I will not be able to provide meson build system support in a reasonable amount of time. I will mark this PR as a draft and will update my contribution when I find time to tackle the bigger job...

Thank you!

PR #1504, opened 2024-05-06
volk: Add wrap

antoniovazquezblanco:


subprojects/volk.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = volk-3.1.2
+source_url = https://github.com/gnuradio/volk/releases/download/v3.1.2/volk-3.1.2.tar.xz
+source_filename = volk-3.1.2.tar.xz
+source_hash = ce62f645c2ee3ef8b38f923de78ef4df9a1b7696d618f57781955803ee29bab3
+method = cmake
```
eli-schwartz, line 6: The wrapdb can't use cmake, it has to be a native meson build.

eli-schwartz (changes requested):

antoniovazquezblanco:
@eli-schwartz, I have a doubt. This build requires Python Mako. I was able to provide that dependency via `ci_config.json` but many package managers do not provide python packages. How can I provide that in Windows (via chocolatey) or in Mac OS (via brew)?

Thanks

neheb:
I think @bgilbert worked on a commit adding pip package support.

bgilbert:
Yes, it was #1481.  I could resurrect that PR if desired.

antoniovazquezblanco:
It will take me a long while until I find the time to redo this in meson. I believe there is zero rush in reviving that. It was to know if this was thought or not.

Thanks

PR #1505, opened 2024-05-06
Only permit native Meson wraps.

jpakkane:


eli-schwartz:
You beat me to it. :D

PR #1506, opened 2024-05-06
luajit: add version 2.1.1713773202

benoit-pierre:


manipuladordedados:
Will meson_options.txt not include LUAJIT_ENABLE_LUA52COMPAT?

benoit-pierre:
> Will meson_options.txt not include LUAJIT_ENABLE_LUA52COMPAT?

It's not used in the project that depends on LuaJIT for which the initial meson port was done, which is why it's missing. Let me look at adding it.

subprojects/packagefiles/luajit/meson.build
```diff
@@ -0,0 +1,201 @@
...
+if get_option('lua52compat')
+  TARGET_XCFLAGS += '-DLUAJIT_ENABLE_LUA52COMPAT'
+endif
+
+use_fno_stack_protector = cc.has_argument('-fno-stack-protector')
+if use_fno_stack_protector
+  TARGET_XCFLAGS += '-fno-stack-protector'
+endif
+
+use_fomit_frame_pointer = cc.has_argument('-fomit-frame-pointer')
```
neheb, line 160: Instead of this, you could add it directly with get_supported_arguments.

benoit-pierre: I also use `use_fno_stack_protector` (and `use_fomit_frame_pointer`) later in the summary.

neheb: Ah OK

benoit-pierre: OK changed to:
```python
TARGET_XCFLAGS += cc.get_supported_arguments(
  '-fno-stack-protector',
  '-fomit-frame-pointer',
)

[…]

summary(
  {
    'use -fno-stack-protector'              : '-fno-stack-protector' in TARGET_XCFLAGS,
    'use -fomit-frame-pointer'              : '-fomit-frame-pointer' in TARGET_XCFLAGS,
[…]
```

manipuladordedados:
Looks good to me.

subprojects/packagefiles/luajit/src/meson.build
```diff
@@ -0,0 +1,274 @@
...
+  'luajit' + luajit_exe_extension,
+  pointing_to: luajit_exe_basename,
+  install_dir: get_option('bindir'),
+  install_tag: 'runtime',
+)
+
+install_data(
+  '../COPYRIGHT',
+  install_dir: get_option('datadir') / 'licenses/luajit',
+  install_tag: 'doc',
```
eli-schwartz, line 239: This can be handled by the license_files argument to project() albeit raises the meson version requirement.

Benefit being that it's tied into the builtin licensedir option.

subprojects/packagefiles/luajit/src/meson.build
```diff
@@ -0,0 +1,274 @@
...
+  variables: {
+    'INSTALL_CMOD': install_cmod,
+    'INSTALL_LMOD': install_lmod,
+  },
+)
+meson.override_dependency('luajit', luajit_dep)
+
+if get_option('default_library') == 'static'
+  if host_machine.system() == 'windows'
+    error('building a static LuaJIT library is not supported on Windows')
```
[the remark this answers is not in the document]
kasper93: Also why is it unsupported? Actually this is the target I would be most interested in, static luajit.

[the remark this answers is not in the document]
benoit-pierre: This is what the official Makefile has to say:
```make
  # Mixed mode is not supported on Windows. And static mode doesn't work well.
  # C modules cannot be loaded, because they bind to lua51.dll.
```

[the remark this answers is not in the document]
benoit-pierre: Here are the official supported modes:
```make
##############################################################################
################################  BUILD MODE  ################################
##############################################################################
# The default build mode is mixed mode on POSIX. On Windows this is the same
# as dynamic mode.
#
# Mixed mode creates a static + dynamic library and a statically linked luajit.
BUILDMODE= mixed
#
# Static mode creates a static library and a statically linked luajit.
#BUILDMODE= static
#
# Dynamic mode creates a dynamic library and a dynamically linked luajit.
# Note: this executable will only run when the library is installed!
#BUILDMODE= dynamic
```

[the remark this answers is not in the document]
kasper93: I would change to `warning` if anything. I agree `mixed`(default_library=both) mode on Windows may/will not work. But dynamic(shared) or static works ok.

MSYS2 ships static luajit https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-luajit/PKGBUILD

I've also used staticky linked version. So I would keep it as an option with know limitations of Windows static linking.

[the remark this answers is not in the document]
benoit-pierre: OK, so how about:
- if you're not building the luajit executable, use whichever library mode you like (including on Windows)
- otherwise on Windows:
  - `both`: unsupported, error
  - `shared`: OK
  - `static`: warning

?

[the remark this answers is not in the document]
kasper93: Ok for me.

subprojects/packagefiles/luajit/src/meson.build
benoit-pierre: `1.1.0`, maybe a little to fresh?

benoit-pierre: I tweaked the code to use `licensedir` on `meson >= 1.1.0` instead.

subprojects/packagefiles/luajit/src/meson.build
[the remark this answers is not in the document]
benoit-pierre: Done, please review.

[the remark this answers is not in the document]
kasper93: I've tested in my pipeline, builds and works correctly. Warning is little bit ominous, because except external modules there shouldn't be problems. Nevertheless it is fine.

subprojects/packagefiles/luajit/src/meson.build
```diff
@@ -0,0 +1,342 @@
...
+    _val = get_option(_subdir) / 'lua' / abiver
+  endif
+  set_variable(_var, _val)
+endforeach
+
+luajit_h = custom_target(
+  'luajit.h',
+  capture: true,
+  command: [
+    'sed',
```
kasper93, line 27: Would be nice to avoid this external dependency. It may not always be available.

subprojects/packagefiles/luajit/src/meson.build
```diff
@@ -0,0 +1,342 @@
...
+  lj_vm_out = custom_target(
+    'lj_vm.S',
+    output: 'lj_vm.S',
+    command: [
+      buildvm_exe,
+      '-m', LJVM_MODE,
+      '-o', '@OUTPUT@',
+    ],
+  )
+endif
```
kasper93, line 64: The only different is file extension. To make it more readable, you could just set the name up front.

```suggestion
lj_vm = host_machine.system() == 'windows' ? 'lj_vm.o' : 'lj_vm.S'
lj_vm_out = custom_target(
  lj_vm ,
  output: lj_vm ,
  command: [
    buildvm_exe,
    '-m', LJVM_MODE,
    '-o', '@OUTPUT@',
  ],
)
```

benoit-pierre: Thanks, amended.

subprojects/packagefiles/luajit/src/meson.build
benoit-pierre: There's also a dependency on `grep` through `unwind_check.sh`, so you need a Unix like system.

The alternative would be to try switching to Python for both cases. I don't think it's worth it.

kasper93:
Any news about it, remaining points to resolve?

PR #1507, opened 2024-05-09
libopenjp2: fix Linux s390x build

bgilbert:
We weren't checking for a big-endian host.


neheb:
should we add CI for this?

bgilbert:
The way to do it would be to cross-compile for s390x (which is really the only remaining non-legacy big-endian architecture) and use an `exe_wrapper` to run tests with QEMU user mode emulation.  We couldn't skip the tests, since endianness problems are unlikely to cause compile failures, only test failures (and only if the tests are wired up).  This would _significantly_ slow down CI runs for projects with slow tests, all for a use case that I think basically no one cares about.

So I don't think CI is worth it, but I can pursue it if folks want.

neheb:
it's very simple to add. the Alpine Linux action supports s390x directly.

But yeah. In my experience, it's really slow. Maybe limit to sanity_checks.py? ping @eli-schwartz

bgilbert:
Also maybe worth noting that CI wouldn't have caught this problem, since the `libopenjp2` wrap doesn't have any tests.

PR #1509, opened 2024-05-09
glib: update to 2.80.2

bgilbert:


bgilbert:
Test failures:

```
stdout: 1: UNKNOWN: Windows Subsystem for Linux has no installed distributions.
stdout: 2: UNKNOWN: Distributions can be installed by visiting the Microsoft Store:
stdout: 3: UNKNOWN: https://aka.ms/wslstore
```

I'm not sure if this is a CI misconfiguration or a test bug.

bgilbert:
Is it expected that we fix these test failures downstream?

bgilbert:
Since glib's Meson build system is upstream, I don't think these test failures are actionable here.  Is there anything specific needed before this can merge?

PR #1514, opened 2024-05-14
libxml2: update to 2.12.7

bgilbert:


neheb:
As an aside, libxml2 has a meson build now. Unfortunately, it needs https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/244

PR #1516, opened 2024-05-14
google-benchmark: update to 1.8.3

### LINKED ISSUE #1311: google-benchmark.wrap benchmark test
In the meson build of google-benchmark.wrap at the end of the file
there is the source of the test/meson.build

**subdir('test')**

this brings the problem that each time is used google benchmark also the test listed in test/meson.build
are built and executed.

is it possible to add a meson_options.txt in the google-benchmark
in order to choose if to import or not the  test/meson.build?

I could do a possible fix for a pull request  but I cannot create branch in this project.

marktsuchida:
Compared to 1.7.1-2, this
- Additionally provides 'benchmark-main' which includes the provided `main()`
- Adds the option 'tests' to disable tests (Closes #1311)
- Adds more of the tests

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1003,9 +1003,11 @@
   },
   "google-benchmark": {
     "dependency_names": [
-      "benchmark"
+      "benchmark",
+      "benchmark-main"
     ],
     "versions": [
+      "1.8.3-1",
       "1.7.1-2",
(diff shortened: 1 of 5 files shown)


marktsuchida:
(Rebased on master and force-pushed.)

PR #1517, opened 2024-05-14
libupnp: update to 1.14.19

neheb:
Use some loops instead of set10.


neheb:
Ok so arm64 CI is using

```
/opt/homebrew/bin/pkg-config)
```

and the x86 one is using

```
/usr/local/bin/pkg-config
```

is the latter missing on arm64?

neheb:
ping @benoit-pierre 

x86 is using system openssl and arm64 is using broken homebrew openssl...

benoit-pierre:
Yep, since #1492 was merged without the fix. Which I assume was the intent, option 2: ignore the error on macOS amd64, no?

neheb:
Yeah. I'm just curious why the difference.

benoit-pierre:
What difference?

neheb:
the different pkg-config binaries. The /opt one seems to prefer homebrew packages that are keg-only.

PR #1519, opened 2024-05-15
jansson: fix format

neheb:
PRId64 doesn't work because of missing header.


neheb:
meh macOS warns. Don't feel like working around it.

PR #1520, opened 2024-05-20
cjson: update to 1.7.18

blmaier:


neheb:
```
22/22 misc_tests       FAIL            2.91s   (exit status 3221225477 or signal 3221225349 SIGinvalid)
```

IIRC this is a library problem. I bet this will work if default_library == static. I wonder what the issue is though...

blmaier:
I would guess it's an issue because of [this new commit](https://github.com/DaveGamble/cJSON/commit/5b502cdbfb21fbe5f6cf9ffbd2b96e4281a741e6) that does an intentional use-after-free.

neheb:
That's ifdeffed to valgrind.

benoit-pierre:
`#ifndef ENABLE_VALGRIND`, so enabled in our case, no?

neheb:
Ah I can’t read.

bgilbert:
Obsoleted by #1657.  Sorry, didn't notice the duplicate.

PR #1521, opened 2024-05-22
catch2: update to 3.6.0

stephanlachnit:


stephanlachnit:
Alpine seems broken independent of this change

benoit-pierre:
It looks line the Alpine job works again, can you close and reopen to trigger a check?

PR #1522, opened 2024-05-22
CI improvements

benoit-pierre:
Example workflow run: [Sanity checks](https://github.com/benoit-pierre/wrapdb/actions/runs/9198634428).

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -375,6 +375,13 @@ class TestReleases(unittest.TestCase):
             install_packages(['sudo', 'apt-get', '-y', 'install', '--no-install-recommends'], debian_packages)
         elif brew_packages and is_macos():
             install_packages(['brew', 'install', '--quiet'], brew_packages)
+            if is_ci():
+                # Ensure binaries from keg-only formulas are available (e.g. bison).
+                out = subprocess.check_output(['brew', '--prefix'] + brew_packages)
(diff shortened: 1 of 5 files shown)


benoit-pierre:
As per discussion in #819, updated to also improve brew package installs.

tools/sanity_checks.py
```diff
@@ -361,52 +364,46 @@ def check_new_release(self, name: str, builddir: str = '_build', deps=None, prog
...
-                subprocess.check_call(['brew', 'install'] + brew_packages)
-            else:
-                s = ', '.join(brew_packages)
-                print(f'The following packages could be required: {s}')
+                # Ensure binaries from keg-only formulas are available (e.g. bison).
+                out = subprocess.check_output(['brew', '--prefix'] + brew_packages)
+                for prefix in out.decode().split('\n'):
+                    bindir = Path(prefix) / 'bin'
+                    if bindir.exists():
+                        meson_env['PATH'] = str(bindir) + ':' + meson_env['PATH']
```
jpakkane, line 384: This means that `brew` installed binaries will silently override system ones.  Is that something we truly want? In theory this could be used to hijack commands, but in practice it would mean that `brew` is already compromised so they can get in via other means.

benoit-pierre: Yeah, formulas are ruby code, no?

neheb (approved):

PR #1524, opened 2024-05-22
Adding the Fossil Logic C API

dreamer-coding-555:
Just adding the Fossil Logic C API, guide me if I missed something.


subprojects/fscl-xalgorithms-c.wrap
```diff
@@ -0,0 +1,12 @@
...
+# website: https://fossillogic.com/
+# ======================
+[wrap-file]
+directory = fscl-xalgorithms-c-2.4.0
+source_url = https://github.com/fossil-lib/fscl-xalgorithms-c/archive/refs/tags/v2.4.0.tar.gz
+source_filename = fscl-xalgorithms-c-2.4.0.tar.gz
+source_hash = d72cc6e67c685df9b3b4691243c0e024abd47c4570dde812403297794b2fb813
+
+[provide]
+fscl-xalgorithms-c = fscl_xalgorithms_c_dep
```
eli-schwartz, line 12: Git is saying these files are badly formatted!

eli-schwartz (changes requested):
In general, a PR with multiple commits, the latter repairing errors in the former, should be squashed by committing the fixes like this:

```
git commit --amend
git push --force
```

git rebase --interactive also allows you to mark some commits as "fixup" instead of "pick" which can be used to retroactively do the same thing.

Also I'm interested in why the checksums were originally wrong...

ci_config.json
```diff
@@ -192,90 +192,70 @@
   },
   "fscl-xalgorithms-c": {
     "fatal_warnings": true,
-    "build_options": [
-      "with_test=enabled"
-    ],
+    "skip_tests": true,
```
benoit-pierre, line 195: You need to prefix your options with the project name (since it's used as a subproject):
```json
    "build_options": [
      "fscl-xalgorithms-c:with_test=enabled"
    ],
```

dreamer-coding-555:
> Also I'm interested in why the checksums were originally wrong...

Aparently used sha224sum insted of sha256sum tool as suggested probably should have double checked that.

ci_config.json
```diff
@@ -190,6 +190,76 @@
       "libxkbcommon-dev"
     ]
   },
+  "fscl-xalgorithms-c": {
+    "fatal_warnings": true,
```
benoit-pierre, line 194: Unnecessary: it already defaults to true.

ci_config.json
dreamer-coding-555: how do I skip microsoft compiler? Havent gotten around to making it work with MSVC yet

ci_config.json
```diff
@@ -190,6 +190,76 @@
...
+    "skip_tests": true,
```
benoit-pierre, line 195: Don't skip the tests, fix your options!

ci_config.json
```diff
@@ -190,6 +190,76 @@
...
+    "build_on": {
+      "windows": false
+    }
```
benoit-pierre, line 198: That's does do what you think it does. Fix your build. (This is to mark a project as not supported on Windows, and assumes you raise an error in your `meson.build`)

ci_config.json
dreamer-coding-555: applied format for enabling my test cases

ci_config.json
dreamer-coding-555: Assuming that's the norm removed extra flag fatal_warnings

ci_config.json
dreamer-coding-555: Removed. This is a platform independent project(s)

benoit-pierre:
For the `python` issue with `fscl-xcore-c`, use `find_program('python3')` (Cf. [documentation](https://mesonbuild.com/Reference-manual_functions.html#find_program)).

benoit-pierre:
Yes, same syntax as in `ci_config.json`: `-Dsubproject:option=value`.

dreamer-coding-555:
Is it possible to set up a similar sanity-checking system for the sake of saving time when creating these wrap files?

https://github.com/fossil-lib/sanity_check

eli-schwartz:
I'm not sure I understand what you're trying to do with that repo.

Isn't this already what tools/sanity_checks.py from this repository does?

dreamer-coding-555:
The idea would be to have a redundant layer for testing new upstream wraps before considering adding to the WrapDB project. But that's off-topic so let's see, can I use just the python3 command for the test generator step in my projects or am I required to use find_program? How can I skip MSVC?

eli-schwartz:
> The idea would be to have a redundant layer for testing new upstream wraps before considering adding to the WrapDB project.

I run tools/sanity_checks.py from a terminal shell all the time. That's how I test wraps before considering adding them to the WrapDB.

> But that's off-topic so let's see, can I use just the python3 command for the test generator step in my projects or am I required to use find_program?

This is a misunderstanding of what commands are, what meson's `find_program()` is, and how the CPython project distributes itself.

Your repository uses the `python` executable, which doesn't always exist, there are a number of ecosystems that only provide `python3`.

Switching from "python" to the version with a 3 at the end, fixes the problem.

(meson's `find_program()` will be internally run, if you don't manually specify it, on the first argument to run_command or custom_target if it is a string instead of a found/built executable/target.)

> How can I skip MSVC?

The error message indicates that the reason MSVC is aborting is because you're using an unofficial experimental-phase std, which implies... yeah.

dreamer-coding-555:
Just changed it for both the main and release branches and it's still complaining about the python executable being undetectable

benoit-pierre:
Because it's still using `https://github.com/fossil-lib/fscl-xmock-c/archive/refs/tags/v2.4.0.tar.gz`?

dreamer-coding-555:
Screenshot 2024-05-23 162904

benoit-pierre:
You have not made a new release, the wrap for `fscl-xmock-c` is still using v2.4.0, which does not have your last 2 commits, no?

dreamer-coding-555:
Shouldn't it point to the release branch? I prefer not to make a bunch of new releases just because of some small change for every issue. Shouldn't there be a checklist for the requirements of what to check when preparing a project for upstream?

dreamer-coding-555:
What am I testing against?

dreamer-coding-555:
Is it possible to have some of my projects depend on other projects? Algorithms and Structures depend on ToFu.

dreamer-coding-555:
#1531 just redoing this with new releases

PR #1525, opened 2024-05-23
Adds Pybind11 2.12.0 release

jeandet:


eli-schwartz:
Random nit: commit messages are usually imperative mood, so "add xxxx" rather than "adds xxxx".

PR #1536, opened 2024-05-28
nanoarrow: Add new package

WillAyd:
@paleolimbot

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -647,6 +647,20 @@
       "boost"
     ]
   },
+  "nanoarrow": {
+    "alpine_packages": [
+      "apache-arrow-dev"
+    ],
+    "brew_packages": [
+      "apache-arrow"
+    ],
+    "msys_packages": [
+      "arrow"
+    ],
+    "build_options": [
+      "nanoarrow:NANOARROW_BUILD_TESTS=true"
+    ]
+  },
   "nativefiledialog-extended": {
     "alpine_packages": [
       "gtk+3.0-dev"
--- a/releases.json
+++ b/releases.json
@@ -2311,6 +2311,14 @@
       "mujs"
     ]
   },
+  "nanoarrow": {
+    "dependency_names": [
+      "nanoarrow"
+    ],
+    "versions": [
+      "0.5.0-1"
+    ]
+  },
   "nativefiledialog-extended": {
     "dependency_names": [
       "nativefiledialog-extended"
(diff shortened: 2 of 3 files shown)


WillAyd:
Hmm we must be missing something for visual studio. Unfortunately not a platform I have access to, but judging by other PRs I think the issue will be related to what has been noticed in:

https://github.com/mesonbuild/wrapdb/pull/1387
https://github.com/mesonbuild/wrapdb/pull/1345

WillAyd:
That's my take too. Maybe related to some of the discussion/work in https://github.com/mesonbuild/meson/pull/10199 as well?

WillAyd:
Tried to add building the tests as part of CI but they require the arrow library, so we'd have a ways to go to get that to work with meson

ci_config.json
```diff
@@ -647,6 +647,17 @@
       "boost"
     ]
   },
+  "nanoarrow": {
+    "brew_packages": [
+      "apache-arrow",
```
eli-schwartz, line 652: Reminder that json sucks. :P

WillAyd:
Looks like Arrow is distributed via apt outside of the debian repository

https://arrow.apache.org/install/

eli-schwartz:
I think that for msys2 you can "just" say the package is called "arrow" and the package manager wrapper used will automatically attempt to install the matching toolchain package and figure out ucrt/clang64 etc by itself.

For Alpine, -dev packages exist so you apparently have to specifically install `apache-arrow-dev`.

For Ubuntu I think this currently sucks as is, we could I suppose extend ci_config.json to also support adding an apt repository.

eli-schwartz:
> if it's acceptable here to get the tests building/running on just one platform.

There is a high expectation the software works most places "as long as the dependencies can be installed goshdarnit" :D especially given the meson.build files are upstream, so I am okay merging it as is. Or at least as long as the trivially fixable stuff is hooked up, so, alpine and msys2, after which we should be able to say "yep, the wrapdb CI either shows this as passing or else fails because we just don't have support for setting it up, but the actual wrap is fine".

eli-schwartz:
Nice, all tests pass except Visual Studio and Ubuntu, which fail due to missing dependencies.

WillAyd:
Sweet thanks @eli-schwartz - excited to see this go in. Whenever the next nanoarrow release gets out we should have the visual studio issue fixed. I'll take a look at adding an apt repository for debian testing separately

PR #1538, opened 2024-05-30
libpng: fix build for ppc64le

classilla:
The current `meson.build` for `libpng` FTBFS on VSX-capable Power ISA systems because it does not include the VSX source files or set the proper compile flags. This fixes it for little-endian, where all little-endian 64-bit Power ISA systems are reliably known to have VSX support (big-endian would require separate changes that are not currently possible with what `host_machine` reports).


neheb:
releases.json needs bumping.

classilla:
Sorry about that, bumped.

jpakkane:
In the future please keep commits squashed to a single one.

classilla:
Thank you, will do.

PR #1545, opened 2024-06-11
pcre2: update to 10.44

bgilbert:


bgilbert:
The test failure is https://github.com/PCRE2Project/pcre2/issues/415, fixed by https://github.com/PCRE2Project/pcre2/pull/418.  AFAICT from those threads, the bug causes test fragility but is otherwise harmless.

PR #1547, opened 2024-06-12
blueprint-compiler: update to 0.14.0

v1993:


v1993:
Any idea what's going on with MacOS/Alpine failures? I can't seem to find any useful error messages in logs.

benoit-pierre:
Apparently no tests were ran, Cf. unittest documentation:

> The testRunner argument can either be a test runner class or an already created instance of it. By default main calls sys.exit() with an exit code indicating success (0) or failure (1) of the tests run. **An exit code of 5 indicates that no tests were run or skipped.**

benoit-pierre:
See https://github.com/python/cpython/issues/113661.

Anyway, no tests are not discovered is the underlying issue, because of this:
```meson
test('tests', py, args: ['-m', 'unittest'], workdir: meson.source_root())
```
And `meson.source_root()` will return the parent project source root:

> meson.source_root()
>
> Returns a string with the absolute path to the source root directory.
>
> This function will return the source root of the parent project if called from a subproject, which is usually not what you want. Try using meson.current_source_dir() or meson.project_source_root(). In the rare cases where the root of the main project is needed, use meson.global_source_root() that has the same behaviour but with a more explicit name.

So no tests are discovered.

Once the call is patched to use `meson.project_source_root()`, the tests fail as expected because of missing dependencies:
```
ERROR: tests.test_samples (unittest.loader._FailedTest.tests.test_samples)
ImportError: Failed to import test module: tests.test_samples
  File "/usr/lib/python3.11/unittest/loader.py", line 419, in _find_test_path
    module = self._get_module_from_name(name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/unittest/loader.py", line 362, in _get_module_from_name
    __import__(name)
  File "/wrapdb/subprojects/blueprint-compiler-v0.12.0/tests/test_samples.py", line 24, in <module>
    import gi
ModuleNotFoundError: No module named 'gi'


Ran 7 tests in 0.004s

FAILED (errors=3)
==============================================================================


Summary of Failures:

1/1 tests FAIL            0.08s   exit status 1

```

neheb:
Install py3-gobject3

benoit-pierre:
That still does not fix the underlying problem: https://gitlab.gnome.org/jwestman/blueprint-compiler/-/merge_requests/192.

neheb:
Yeah. Unfortunate.

bgilbert:
Updated to 0.16.0 in https://github.com/mesonbuild/wrapdb/pull/1889.

PR #1554, opened 2024-06-21
wayland: update to 1.23.0

bgilbert:


bgilbert:
Non-Linux tests are failing with:

```
Exception: Wrap wayland failed to configure due to bugs in the wrap, rather than due to being unsupported
```

In Alpine the test harness is apparently failing in `mkdtemp()` while working around the lack of an `XDG_RUNTIME_DIR` environment variable.

bgilbert:
I tried running CI against the wayland version currently in wrapdb and [got the same failures](https://github.com/bgilbert/wrapdb/actions/runs/10502616410), so this can be merged without adding any regressions.

PR #1556, opened 2024-06-21
Add nanobind

WillAyd:
cc @wjakob

### DIFF
--- a/subprojects/packagefiles/nanobind/meson.build
+++ b/subprojects/packagefiles/nanobind/meson.build
@@ -1,0 +1,81 @@
+project(
+    'nanobind',
+    'cpp',
+    version: '2.1.0',
+    meson_version: '>=0.53.0',
+    license: 'BSD-3-Clause',
+)
+
+py_mod = import('python')
+py = py_mod.find_installation()
+py_dep = py.dependency()
+
+robin_map_dep = dependency('robin-map')
+
+incdir = include_directories('include')
+
+# Arguments are derived from nanobind_build_library function in CMake configuration
+dep_compile_args = []
+dep_link_args = []
+
+if get_option('buildtype') == 'shared'
+  if host_machine.system() == 'darwin'
+    is_pypy = py_dep.get_variable('implementation_lower') == 'pypy'
+    if is_pypy
+      resp_file = meson.project_source_root() / 'cmake/darwin-ld-pypy.sym'
+    else
+      resp_file = meson.project_source_root() / 'cmake/darwin-ld-cpython.sym'
+    endif
+    add_project_link_arguments('-Wl,-dead_strip', '-Wl,x', '-Wl,-S', '-Wl,@' + resp_file)
+    dep_link_args += ['-Wl,-dead_strip', '-Wl,x', '-Wl,-S', '-Wl,@' + resp_file]
+  elif host_machine.system() != 'windows'
+    add_project_link_arguments('-Wl,-s', language: 'cpp')
+    dep_link_args += ['-Wl,-s']
+  endif
+
+  add_project_arguments('-DNB_BUILD', language: 'cpp')
+  dep_compile_args += ['-DNB_SHARED']
+elif host_machine.system() != 'windows' and host_machine.system() != 'darwin'
+  dep_compile_args += ['-ffunction-sections', '-fdata-sections']
+  dep_link_args += ['-Wl,--gc-sections']
+endif
+
+compiler = meson.get_compiler('cpp')
+if compiler.get_id() == 'msvc'
+  dep_compile_args += ['-D_CRT_SECURE_NO_WARNINGS']
+else
+  # Discussion in WrapDB PR #1556 as to whether or not this is necessary
+  # add_project_arguments('-fno-strict-aliasing', language: 'cpp')
+endif
+
+if get_option('buildtype') == 'release'
+  add_project_arguments('-DNB_COMPACT_ASSERTIONS', language: 'cpp')
+endif
+
+nanobind_lib = static_library(
+    'nanobind',
+    sources: [
+        'src/nb_internals.cpp',
+        'src/nb_func.cpp',
+        'src/nb_type.cpp',
+        'src/nb_enum.cpp',
+        'src/nb_ndarray.cpp',
+        'src/nb_static_property.cpp',
+        'src/common.cpp',
+        'src/error.cpp',
+        'src/trampoline.cpp',
+        'src/implicit.cpp',
+    ],
+    include_directories: [incdir],
+    dependencies: [py_dep, robin_map_dep],
+    gnu_symbol_visibility: 'hidden',
+)
+
+
+nanobind_dep = declare_dependency(
+    include_directories: [incdir],
+    link_with: [nanobind_lib],
+    dependencies: [py_dep, robin_map_dep],
+    compile_args: dep_compile_args,
+    link_args: dep_link_args,
+)
--- a/ci_config.json
+++ b/ci_config.json
@@ -685,6 +685,18 @@
       "nanoarrow:NANOARROW_BUILD_TESTS=true"
     ]
   },
+    "nanobind": {
+    "_comment": "Python fails on 32-bit MSVC because we don't have 32-bit Python",
+    "alpine_packages": [
+      "python3-dev"
+    ],
+    "msys_packages": [
+      "python"
+    ],
+    "build_options": [
+      "buildtype=release"
+    ]
+  },
   "nativefiledialog-extended": {
     "alpine_packages": [
       "gtk+3.0-dev"
--- a/releases.json
+++ b/releases.json
@@ -2380,6 +2380,14 @@
       "0.5.0-1"
     ]
   },
+  "nanobind": {
+    "dependency_names": [
+      "nanobind"
+    ],
+    "versions": [
+        "2.1.0-1"
+    ]
+  },
   "nativefiledialog-extended": {
     "dependency_names": [
       "nativefiledialog-extended"
--- a/subprojects/nanobind.wrap
+++ b/subprojects/nanobind.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = nanobind-2.1.0
+source_url = https://github.com/wjakob/nanobind/archive/refs/tags/v2.1.0.tar.gz
+source_filename = nanobind-2.1.0.tar.gz
+source_hash = c37c53c60ada5fe1c956e24bd4b83af669a2309bf952bd251f36a7d2fa3bacf0
+patch_directory = nanobind
+
+[provide]
+nanobind = nanobind_dep


subprojects/packagefiles/nanobind/meson.build
```diff
@@ -0,0 +1,39 @@
...
+        'src/nb_enum.cpp',
+        'src/nb_ndarray.cpp',
+        'src/nb_static_property.cpp',
+        'src/common.cpp',
+        'src/error.cpp',
+        'src/trampoline.cpp',
+        'src/implicit.cpp',
+    ],
+    include_directories: [incdir],
+    install: true,
```
eli-schwartz, line 31: nanobind's CMake files have an option whether to install it, do we also want that? If users can choose not to install it, we might want to use:

```meson
build_target(...,
    target_type: should_install ? 'library' : 'static_library',
)

File: subprojects/packagefiles/nanobind/meson.build [hunk 2]
```diff
@@ -0,0 +1,39 @@
+project(
+    'nanobind',
+    'cpp',
+    version: '2.0.0',
```
eli-schwartz, line 4: Add (for python module including `embed: ` kwarg support):

```meson
    meson_version: '>=0.53.0',
```

eli-schwartz (changes requested):
CI fails due to missing releases.json:

https://github.com/mesonbuild/wrapdb?tab=readme-ov-file#wrap-release-meta

Aside: I would suggest squashing fixup commits (`git commit --amend && git push --force` or alternatively, `git commit --fixup=HEAD` followed later by `git rebase -i --autosquash`).

subprojects/packagefiles/nanobind/meson.build
WillAyd: Nice catch. If I am reading the CMake configuration correctly it seems like it is only intended to be installed when not used as a subproject - so maybe I should just remove install altogether from this configuration?

eli-schwartz: Entirely possible. In that case, also change library to static_library, I think...

eli-schwartz:
IIRC releases.json has to be in alphabetical order, nanobind is right before nanoarrow but should instead be right after.

WillAyd:
That makes sense. Looks like some of the build failures have to do with the robin-map depdendency version as well - may need to update that as a precursor

wjakob:
@eli-schwartz @WillAyd : If I understand correctly, the goal of this PR is to build using pure Meson without depending on CMake? This would be awesome, but the recipe here looks somehow too easy to me.

Please take a look at some of the steps explained in the comment at the top of https://github.com/wjakob/nanobind/blob/master/src/nb_combined.cpp. And this is the "ground truth" reference for the extension build process: https://github.com/wjakob/nanobind/blob/master/cmake/nanobind-config.cmake

For example, does this wrapper

1. Avoid linking against a specific libpython.so/dylib? Basically, we don't want to bake a libpython path into the compiled shared library—the interpreter loading the extension will already have all of the libpython symbols.
2. Compute the extension suffix suitable for the desired Python ABI and stable/non-stable convention?
3. Set compilation flags required for correctness (`-fno-strict-aliasing`) in the static library part only?
4. Support stable ABI builds?
5. Set flags that are a good idea to obtain small binaries when making release builds ('-DNDEBUG' and '-DNB_COMPACT_ASSERTIONS', '-ffunction-sections/-fdata-sections/--gc-sections', various stripping flags -- see nanobind's CMake build system for details on these steps)
6. Use the chained fixups optimization for library builds on macOS? Nanobind ships a custom linker response file to enable this.
7. Support isolating extensions from others by specifying the `NB_DOMAIN` parameter? This is in general something that almost every extension should do to avoid cross-contamination.

Quite a lot of work has gone into the CMake interface produce nice output—I would be concerned about alternatives that produce subpar extensions by missing some of these steps. A good recipe that helped to reproduce the behavior in the Bazel build system (https://github.com/nicholasjng/nanobind-bazel) has been to run nanobind's CMake build sytem on a simple extension in verbose mode and then try to exactly reproduce all of the compiler and linker invocations with their flags on macOS+Windows+Linux (some subtle differences there as well).

I'm happy to provide further detail on anything with the caveat that I'll be traveling in July.

eli-schwartz:
> For example, does this wrapper
> 
> 1. Avoid linking against a specific libpython.so/dylib? Basically, we don't want to bake a libpython path into the compiled shared library—the interpreter loading the extension will already have all of the libpython symbols.
> 2. Compute the extension suffix suitable for the desired Python ABI and stable/non-stable convention?

These questions are related.

That's something every python module already has to handle. Setuptools handles this automatically, as does maturin. Meson handles it automatically too! CMake doesn't support it, which is why scikit-build and I assume nanobind as well, provide their own cmake modules that define functions which override cmake add_library.

Meson calls this `pymod.extension_module()`.

> 3. Set compilation flags required for correctness (`-fno-strict-aliasing`) in the static library part only?

This is a legacy of python 2 which violates the C/C++ language standards. It's not valid code if it has to compile with disabling the *mandatory* aliasing rules of the C++ language.

Since many people get aliasing wrong even though it breaks the language (and so did CPython 2.7!) compiler vendors offer a compiler extension that tells the compiler to cripple its codegen and produce really slow code that tiptoes around aliasing, adding redundant checks here there and everywhere to check at runtime whether the code actually followed the aliasing rules, and recover if it didn't follow the rules.

Aliasing violations can't come from CPython since CPython fixed that. They can still come from projects' own code, or from their nanobind dependency.


I think you're saying that nanobind needs this extra protection as the `cpp_args` kwarg to the library() call for nanobind itself? That can be added.

> 4. Support stable ABI builds?

Meson handles this automatically as part of `pymod.extension_module()`. If nanobind needs to be built with the limited API define then it should be manually calculated based on the `-Dpython.allow_limited_api=true` builtin option, as meson doesn't handle this for static libraries that use CPython and get linked into extension modules.

It may make sense to extend the meson language to handle static libraries here somehow. I'll think about this...

> 5. Set flags that are a good idea to obtain small binaries when making release builds ('-DNDEBUG' and '-DNB_COMPACT_ASSERTIONS', '-ffunction-sections/-fdata-sections/--gc-sections', various stripping flags -- see nanobind's CMake build system for details on these steps)

This can be added to the `declare_dependency()` function call in this PR. Meson does have a builtin `-Db_ndebug=true` option which should be respected, though.

> 6. Use the chained fixups optimization for library builds on macOS? Nanobind ships a custom linker response file to enable this.
> 7. Support isolating extensions from others by specifying the `NB_DOMAIN` parameter? This is in general something that almost every extension should do to avoid cross-contamination.

Those are interesting points. The linker response file can probably be handled just like point 5. I'm not sure how to handle point 7 (users could I suppose add that to their own extensions).

> Quite a lot of work has gone into the CMake interface produce nice output—I would be concerned about alternatives that produce subpar extensions by missing some of these steps. A good recipe that helped to reproduce the behavior in the Bazel build system (https://github.com/nicholasjng/nanobind-bazel) has been to run nanobind's CMake build sytem on a simple extension in verbose mode and then try to exactly reproduce all of the compiler and linker invocations with their flags on macOS+Windows+Linux (some subtle differences there as well).
> 
> I'm happy to provide further detail on anything with the caveat that I'll be traveling in July.

Thanks for your insights!

WillAyd:
> > 9\. Set flags that are a good idea to obtain small binaries when making release builds ('-DNDEBUG' and '-DNB_COMPACT_ASSERTIONS', '-ffunction-sections/-fdata-sections/--gc-sections', various stripping flags -- see nanobind's CMake build system for details on these steps)
> 
> This can be added to the `declare_dependency()` function call in this PR. Meson does have a builtin `-Db_ndebug=true` option which should be respected, though.

Does Meson not set most of these in the first place for a release build? I added -DNB_COMPACT_ASSERTIONS but from [previous testing](https://github.com/wjakob/nanobind/discussions/449#discussioncomment-9157841) the rest were automatically added - so I don't think need to try and manually add here?

WillAyd:
Ignore my previous comment - those flags probably came from the CMake wrap I tried before. Will take another look

wjakob:
Two quick responses to @eli-schwartz:

> CMake doesn't support it, which is why scikit-build and I assume nanobind as well, provide their own cmake modules that define functions which override cmake add_library.

Incorrect -- CMake handles it too (via `Python_add_library`). But the way that CMake does this on macOS (by specifying the legacy `-undefined dynamic_lookup` linker flag) is incompatible with the recently added chained fixups linker optimization, so nanobind uses a custom solution. There is lots of discussion about it [here](https://github.com/python/cpython/issues/97524#issuecomment-1458855301).

> Aliasing violations can't come from CPython since CPython fixed that. 

To my knowledge (and I would be really curious to be pointed to something saying otherwise) Core CPython C code to the cannot be compiled as regular C++ code. It heavily relies on macros that cast and reinterpret data structures that aren't related to each other through inheritance, which violates C++ type punning rules. The `-fno-strict-aliasing` is therefore required when compiling CPython-style C code in a C++ context. The reason why nanobind specifies this flag to the "libnanobind" utility library component is because it is written like regular CPython code, making plentiful use of such data structure reinterpretation that would be suspect in C++. But this flag isn't imposed on the remainder of the (user-provided) bindings, so aliasing analysis remains useful there.

thesamesam:
> To my knowledge (and I would be really curious to be pointed to something saying otherwise) Core CPython C code to the cannot be compiled as regular C++ code. It heavily relies on macros that cast and reinterpret data structures that aren't related to each other through inheritance, which violates C++ type punning rules.

I'm not sure about this. C of course has its own type punning rules. CPython 2.x _did_ have an infamous problem with it (https://peps.python.org/pep-3123/). I'm not aware of any issue in modern CPython.

numpy also has a bunch of C++ which is built without `-fno-strict-aliasing` (discussed at https://github.com/numpy/numpy/issues/25004) and nothing has gone wrong there.

Further, I'm not aware of any construct used in CPython which is legal under C aliasing rules but suddenly illegal under C++ rules. Could you point me to it?

SoapGentoo:
@wjakob please link where in the Python documentation C++ compilation requires `-fno-strict-aliasing`. C and C++ literally have pretty much the exact same strict aliasing rules. Specifically, the rules that [PEP 3123](https://peps.python.org/pep-3123/) relies are the same in C++, so your claims require actual evidence.

eli-schwartz:
I would use the string method `.version_compare('<3.12')`

WillAyd:
Any guidance on how to handle the MSVC errors about a debug build with a release Python interpreter?

> nanobind| WARNING: Using a debug build type with MSVC or an MSVC-compatible compiler
nanobind| when the Python interpreter is not also a debug build will almost
nanobind| certainly result in a failed build. Prefer using a release build
nanobind| type or a debug Python interpreter.

Should I just force a release build on msvc?

eli-schwartz:
Setting the release buildtype in ci_config.json is definitely a reasonable way to deal with this.

eli-schwartz:
The alpine failure is probably because alpine uses -dev packages just like Ubuntu.

WillAyd:
Have a feeling the VisualStudio CI builds need something else besides just Python installed for the import to work. I couldn't find a -dev package in choco - any ideas?

benoit-pierre:
After disabling fail-fastx for the VisualStudio matrix (cf. #1561), the x64 job [fails too](https://github.com/benoit-pierre/wrapdb/actions/runs/9651871546/job/26620727585).

.github/workflows/sanity_checks.yml
```diff
@@ -47,6 +47,7 @@ jobs:
   VisualStudio:
     runs-on: windows-latest
     strategy:
+      fail-fast: false
```
WillAyd, line 50: Temporarily including to make CI verification easier, as pointed out by @benoit-pierre

subprojects/packagefiles/nanobind/meson.build
```diff
@@ -0,0 +1,87 @@
...
+if get_option('buildtype') == 'shared'
+  # nanobind_link_options uses linker response files for either PyPy or
+  # CPython at this step - how do we support in Meson?
+  add_project_arguments('-DNB_BUILD', language: 'cpp')
+  dep_compile_args += ['-DNB_SHARED']
+
+  if host_machine.system() == 'darwin'
+    add_project_link_arguments('-Wl,-dead_strip', '-Wl,x', '-Wl,-S', language: 'cpp')
+  elif host_machine.system() != 'windows'
+    add_project_link_arguments('-Wl,-s', language: 'cpp')
```
eli-schwartz, line 37: Shouldn't these also be part of dep_link_args?

subprojects/packagefiles/nanobind/meson.build
```diff
@@ -0,0 +1,87 @@
...
+  if py_version.version_compare('>=3.12')
+    add_project_arguments('-DPy_LIMITED_API=0x030C0000', language: 'cpp')
+  endif
+endif
+
+if get_option('buildtype') == 'shared'
+  # nanobind_link_options uses linker response files for either PyPy or
+  # CPython at this step - how do we support in Meson?
+  add_project_arguments('-DNB_BUILD', language: 'cpp')
+  dep_compile_args += ['-DNB_SHARED']
```
eli-schwartz, line 32:
```
if is_pypy # how to check this?
    resp_file = meson.project_source_root() / 'cmake/darwin-ld-pypy.sym'
else
    resp_file = meson.project_source_root() / 'cmake/darwin-ld-cpython.sym'
fi
dep_link_args += ['-Wl,@' + resp_file]
```

@rgommers might remember if we've discussed how to detect pypy.

subprojects/packagefiles/nanobind/meson.build
```diff
@@ -0,0 +1,87 @@
...
+    add_project_link_arguments('-Wl,-dead_strip', '-Wl,x', '-Wl,-S', language: 'cpp')
+  elif host_machine.system() != 'windows'
+    add_project_link_arguments('-Wl,-s', language: 'cpp')
+  endif
+elif host_machine.system() != 'windows' and host_machine.system() != 'darwin'
+  dep_compile_args += ['-ffunction-sections', '-fdata-sections']
+  dep_link_args += ['-Wl,--gc-sections']
+endif
+
+# Here the CMake configuration sets PIC - does Meson handle this internally?
```
eli-schwartz, line 44: static_library (and library, to cover static ones) has the `pic: <bool>` option. By default it is unset, which means that the value of `-Db_staticpic` is used instead, so builders can make this choice. By default, the project option is true: all static libraries will build with PIC.

You can explicitly force this with `pic: true` as a kwarg to static_library to prevent anyone from overriding it on the command line. You can also explicitly force it off with `pic: false` if you're positive you don't want to build with PIC.

In general, static libraries that may later be used by shared libraries need this, disabling PIC is a niche use case.

subprojects/packagefiles/nanobind/meson.build
```diff
@@ -0,0 +1,87 @@
...
+  # Discussion in WrapDB PR #1556 as to whether or not this is necessary
+  # add_project_arguments('-fno-strict-aliasing', language: 'cpp')
+endif
+
+# Here the CMake configuration links against Python::SABIModule on Windows
+# if building an abi3 target - does Meson handle this internally?
+
+if get_option('buildtype') == 'release'
+  add_project_arguments('-DNDEBUG', '-DNB_COMPACT_ASSERTIONS', language: 'cpp')
+endif
```
eli-schwartz, line 59: `get_option('b_ndebug')` has 3 values:  true, false, if-release

Probably, projects should simply specify if-release for their own use. There's no need to manually pass -DNDEBUG IMO.

Opinions vary, apparently. :shrug:

WillAyd:
Seems like the current VisualStudio x64 failure is reported upstream and a bug with the latest msvc compiler version

https://github.com/wjakob/nanobind/issues/613

subprojects/packagefiles/nanobind/meson.build
WillAyd: The current CMake configuration sets these as PRIVATE on the target, so just copying that over. Happy to move this to the dependency interface if you think that makes more sense

benoit-pierre:
And for the x32 job: https://github.com/mesonbuild/wrapdb/pull/1562. With that get [similar errors](https://github.com/benoit-pierre/wrapdb/actions/runs/9652613958/job/26623122459).

subprojects/packagefiles/nanobind/meson.build
WillAyd: Ah nevermind - I take that back. I see it is called by both the internal function to build the nanobind library as well as the `nanobind_add_module` function that end users are expected to use.

I think what you are proposing makes sense

WillAyd:
Thanks @benoit-pierre . My assumption is those two jobs will continue to fail until a newer msvc release comes out (or we can try to test against an older compiler version)

subprojects/packagefiles/nanobind/meson.build
```diff
@@ -0,0 +1,91 @@
...
+compiler = meson.get_compiler('cpp')
+if compiler.get_id() == 'msvc'
+  dep_compile_args += ['-D_CRT_SECURE_NO_WARNINGS']
+else
+  # Discussion in WrapDB PR #1556 as to whether or not this is necessary
+  # add_project_arguments('-fno-strict-aliasing', language: 'cpp')
+endif
+
+# Here the CMake configuration links against Python::SABIModule on Windows
+# if building an abi3 target - does Meson handle this internally?
```
eli-schwartz, line 59: When a project uses:

```
py.extension_module(
    'foomod',
    'foomod.c',
    limited_api: '3.12',
)
```
meson links to python3.dll rather than python312.dll. I assume that's what "SABIModule" is.

WillAyd: Seems reasonable. I assume it would be more idiomatic to remove the code that defines Py_LIMITED_API in this configuration and just let Meson handle that

WillAyd:
Converted to draft to try out the nanobind commit that works around the MSVC compiler bug and get things green. Probably need to wait until the next nanobind release or MSVC fix, whichever comes first

However, I think the configuration itself is ready for any more review comments

WillAyd:
Is the MSVC bug affecting nanobind in CI a blocker for this? This is the upstream report:

https://developercommunity.visualstudio.com/t/Internal-compiler-error-w-nanobind-v19/10662883?sort=newest

If not and if there is no further feedback on this, I'd be happy to squash and skip MSVC builds for now to push through

eli-schwartz (approved):
Looks ready to me...

WillAyd:
When using that I notice that I have to do both:

```sh
meson wrap install robin-map
meson wrap install nanobind
```

Is it expected that the user has to specify this, or did I maybe miss something here that ensures when a user installs nanobind that robin-map gets pulled in as well?

eli-schwartz:
You aren't missing anything. `meson wrap install` doesn't know about recursive dependencies, and you might not necessarily care about them anyway -- some dependencies are harder than others, so e.g. people might not bother adding a wrap for zlib, but would add a wrap for curl which depends on zlib.

PR #1560, opened 2024-06-24
Add sparrow

WillAyd:
@JohanMabille

### DIFF
--- a/subprojects/packagefiles/sparrow/meson.build
+++ b/subprojects/packagefiles/sparrow/meson.build
@@ -1,0 +1,39 @@
+project(
+    'sparrow',
+    'cpp',
+    meson_version: '>=0.60.0',
+    version: '0.0.2',
+    license: 'Apache 2.0',
+    default_options: [
+        'cpp_std=c++20',
+    ]
+)
+
+incdir = include_directories('include')
+
+compile_args = []
+deps = []
+if get_option('use_date_polyfill')
+  compile_args += ['-DSPARROW_USE_DATE_POLYFILL']
+  deps += dependency('date')
+  deps += dependency('tz')
+endif
+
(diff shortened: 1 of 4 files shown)


subprojects/packagefiles/sparrow/meson.build
```diff
@@ -0,0 +1,50 @@
...
+    'include/sparrow/memory.hpp',
+    'include/sparrow/mp_utils.hpp',
+    'include/sparrow/sparrow_version.hpp',
+    'include/sparrow/typed_array.hpp',
+    'include/sparrow/variable_size_binary_layout.hpp',
+
+    'include/sparrow/details/3rdparty/float16_t.hpp',
+]
+
+install_headers(sparrow_headers, subdir: 'sparrow')
```
jpakkane, line 44: It's a lot less work to use `install_subdir` instead.

subprojects/packagefiles/sparrow/meson.build
```diff
@@ -0,0 +1,37 @@
...
+    'array_data_concepts.hpp',
+    'dictionary_encoded_layout.hpp',
+    'reference_wrapper_utils.hpp',
+]
+install_subdir(
+    'include/sparrow',
+    subdir: 'sparrow',
+    exclude_directories: ['details'],
+    exclude_files: exclude_files,
+)
```
benoit-pierre, line 31:
```suggestion
install_subdir(
    'include/sparrow',
    exclude_directories: ['details'],
    exclude_files: exclude_files,
    install_dir: get_option('includedir'),
    install_tag: 'devel',
    subdir: 'sparrow',
)
```
I think (please double check the files end up in the right location).

And add `meson_version: '>=0.60.0'` to the `project(…)` call.

PR #1561, opened 2024-06-24
ci: disable `fail-fast` option on VisualStudio matrix jobs

benoit-pierre:
Rational: the failure modes are not always the same, and you end up wasting more CI resources if you need to trigger multiple builds because you were not aware of different issues.

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -47,6 +47,7 @@ jobs:
   VisualStudio:
     runs-on: windows-latest
     strategy:
+      fail-fast: false
       matrix:
         platform: ['x64', 'x86']
     steps:


benoit-pierre:
Example: #1556, the x32 job fails because the python executable is not 32-bit, and the x64 is canceled even though it's (probably) okay.

neheb:
Would actions/setup-python fix?

benoit-pierre:
For the x86 job? I already thought about that, but forgot to specify the architecture, so let me try that again.

benoit-pierre:
Yep, [same error](https://github.com/benoit-pierre/wrapdb/actions/runs/9652613958/job/26623122459) as x64 now.

PR #1562, opened 2024-06-24
ci: setup a 32bits Python for the VisualStudio x32 jobs

benoit-pierre:
So building Python related stuff work.

Example: #1556.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -70,6 +70,14 @@ jobs:
         with:
           fetch-depth: 0
 
+      # Install a 32-bit Python so building related stuff work.
+      - name: Setup x86 Python
+        if: matrix.platform == 'x86'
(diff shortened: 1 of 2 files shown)


.github/workflows/build_all.yml
```diff
@@ -70,6 +70,14 @@ jobs:
         with:
           fetch-depth: 0
 
+      # Install a 32-bit Python so building related stuff work.
+      - name: Setup x86 Python
+        if: matrix.platform == 'x86'
```
neheb, line 75: Would make more sense to do on both IMO.

benoit-pierre: The x64 version already has a properly setup Python, and a different version: less homogeneous configurations mean more chances to catch bugs.

neheb: Fair enough,

PR #1563, opened 2024-06-26
libgpiod: Fix config file include and bump to v1.6.5

amboar:
Hello,

I've made a couple of small changes here along the lines of the PR title. It appears CI is unhappy with the result but I don't yet understand why. I'm after some feedback there, or if the changes look sensible, to integrate them anyway.

I haven't yet looked at what it would take to provide a wrap for v2 of the library, I'll look into that if I have a need for the new APIs.

Thanks!


benoit-pierre:
You'll have 2 split the PR in 2 (can't bump versions 2 times in the same PR).

And the issue with `gpiodcxx_dep` is because it's compiled without `-Dbindings=…`, so might try to tweak that in `ci_config.json`.

amboar:
Ah, thanks for the super quick feedback! Let me address those.

amboar:
Alright, I've split the changes. I'll close this PR and create another two. I'll create a PR for the fixes for 1.6.3-2 first, including the fix to `ci_config.json`, then once that's merged I'll rebase the bump to v1.6.5 and create the second PR.

PR #1564, opened 2024-06-26
libgpiod: Build definition fixes

amboar:
Disambiguate the path to the configuration header and configure the C++ bindings to target the host environment. In the process, address some technical debt in the CI configuration for libgpiod to enable the C++ binding.


benoit-pierre:
The confusingly named `build_on` does not mean that the CI should attempt or not to build for that particular platform, but that the build is expected to work (`true`) or fail with a specific error (`false`). See [liburing](https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/liburing/meson.build#L11) for example.

amboar:
Yeah I just discovered that. I was considering adding the following:

```diff
diff --git a/subprojects/packagefiles/libgpiod/meson.build b/subprojects/packagefiles/libgpiod/meson.build
index 2faf13315c04..74d14ff14126 100644
--- a/subprojects/packagefiles/libgpiod/meson.build
+++ b/subprojects/packagefiles/libgpiod/meson.build
@@ -10,6 +10,11 @@ project('libgpiod', 'c',
         license: 'LGPL-2.1-or-later',
         meson_version: '>=0.56.0')
 
+host_system = host_machine.system()
+if host_system != "linux"
+    error('libgpiod is not supported on @0@'.format(host_system))
+endif
+
 add_project_arguments('-D_GNU_SOURCE', language: 'c')
 
 pkg = import('pkgconfig')
```

... and it looks like that's what liburing is doing as well. Nice.

benoit-pierre:
IMHO it would better go add back `c_std=gnu99` to the `project(…)` call and just disable fatal warnings on the CI, rather than complexify the port.

And maybe fatal warnings should automatically be disabled for platforms where `build_on` is false.

benoit-pierre (approved):

PR #1566, opened 2024-06-26
cwalk: fix compilation when both shared and static is wanted

antoniovazquezblanco:
This is a follow up of https://github.com/mesonbuild/wrapdb/pull/1387

I am trying to improve on the already merged meson script by allowing compilation of both static and dynamic libs simultaneously with the proper arguments. This tryes to implement what @benoit-pierre suggested...
Also adding @likle from upstream...

From meson maintainers, would this be desirable or would a "def" file provide any other desirable behaviour?

Thanks


releases.json
```diff
@@ -513,7 +513,7 @@
       "cwalk"
     ],
     "versions": [
-      "1.2.9-1"
+      "1.2.9-2"
```
benoit-pierre, line 516: Don't patch the version, add a new one (first).

releases.json
```diff
@@ -513,7 +513,8 @@
       "cwalk"
     ],
     "versions": [
-      "1.2.9-1"
+      "1.2.9-1",
+      "1.2.9-2"
```
benoit-pierre, line 517: Newer versions first.

releases.json
antoniovazquezblanco: You are fast 😆

releases.json
antoniovazquezblanco: Thanks! :)

antoniovazquezblanco:
In MS compilers, this is producing a lot of warnings:
```
warning: 'cwk_path_change_segment' redeclared without 'dllimport' attribute: 'dllexport' attribute added [-Winconsistent-dllimport]
```

benoit-pierre:
> From meson maintainers, would this be desirable or would a "def" file provide any other desirable behaviour?

AFAIK, that's a Windows only solution, and `CWK_SHARED` also impact the build on other platforms:
```c
#if defined(_WIN32) || defined(__CYGWIN__)
#define CWK_EXPORT __declspec(dllexport)
#define CWK_IMPORT __declspec(dllimport)
#elif __GNUC__ >= 4
#define CWK_EXPORT __attribute__((visibility("default")))
#define CWK_IMPORT __attribute__((visibility("default")))
#else
#define CWK_EXPORT
#define CWK_IMPORT
#endif

#if defined(CWK_SHARED)
#if defined(CWK_EXPORTS)
#define CWK_PUBLIC CWK_EXPORT
#else
#define CWK_PUBLIC CWK_IMPORT
#endif
#else
#define CWK_PUBLIC
#endif
```

However, the `library(…)` call is missing `gnu_symbol_visibility: 'hidden'` to take full advantage of `__attribute__((visibility("default")))` when using clang/gcc.

benoit-pierre:
> redeclared without 'dllimport

That's because `-DCWK_EXPORTS` is needed too: from `CMakeLists.txt`, it should be unconditionally added do `c_args`.

Additionally, `CWK_SHARED` is part of the public interface:
```cmake
target_compile_definitions(cwalk PUBLIC CWK_SHARED)
```
So it should be added to the `compile_args` of `cwalk_dep`.

benoit-pierre:
You'll need to declare 2 dependencies, and use [`meson.override_dependency(…)`](https://mesonbuild.com/Reference-manual_functions.html#declare_dependency) with `static`, so the right compile arguments are used for both static and shared variants.

benoit-pierre:
And switch to this syntax in the wrap:
```ini
[provide]
dependency_names = cwalk
```

antoniovazquezblanco:
Finally found some time to get back to this but...

> You'll need to declare 2 dependencies, and use [meson.override_dependency(…)](https://mesonbuild.com/Reference-manual_functions.html#declare_dependency) with static, so the right compile arguments are used for both static and shared variants.

I do not know how to do this. Could you provide a detailed description please? Thanks @benoit-pierre !

benoit-pierre:
See https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/luajit/src/meson.build#L183 & https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/luajit/src/meson.build#L218.

PR #1574, opened 2024-07-04
zpp_bits: Add new wrap 4.4.23

QWERTIOX:
I'm not sure if meson_version: '>=1.0.1' should be so high but this is the one i'm using right now

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -3699,6 +3699,14 @@
       "1.2.8-2"
     ]
   },
+  "zpp_bits": {
+    "dependency_names": [
+      "zpp_bits"
+    ],
+    "versions": [
+      "4.4.23-1"
+    ]
+  },
   "zstd": {
     "dependency_names": [
       "libzstd"
--- a/subprojects/packagefiles/zpp_bits/include/meson.build
+++ b/subprojects/packagefiles/zpp_bits/include/meson.build
@@ -1,0 +1,3 @@
+copy = import('fs').copyfile('../zpp_bits.h')
+
+incdir = include_directories('.')
--- a/subprojects/packagefiles/zpp_bits/meson.build
+++ b/subprojects/packagefiles/zpp_bits/meson.build
@@ -1,0 +1,12 @@
+project(
+  'zpp_bits',
+  'cpp',
+  version: '4.4.23',
+  license: 'MIT',
+  meson_version: '>=0.64.0',
+  default_options : ['cpp_std=c++20']
+)
+
+subdir('include')
+
+zpp_bits_dep = declare_dependency(include_directories: incdir)
--- a/subprojects/zpp_bits.wrap
+++ b/subprojects/zpp_bits.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+source_url = https://github.com/eyalz800/zpp_bits/archive/refs/tags/v4.4.23.tar.gz
+directory = zpp_bits-4.4.23
+source_filename = zpp_bits-v4.4.23.tar.gz
+source_hash = ab21c6d12d9644da23e75eb798c98e5392037b4bd862d608878cc6dee169a389
+patch_directory = zpp_bits
+
+[provide]
+zpp_bits = zpp_bits_dep


QWERTIOX:
of course sanity checks pass on local machine...

benoit-pierre:
You forgot to update `releases.json`. 

As for the `meson_version: '>=1.0.1'` requirement, there's no reason for it.

QWERTIOX:
wait

benoit-pierre:
Why?

QWERTIOX:
because it was including junk like readme.md and could't find another way to move this header file or force including just header file

benoit-pierre:
If by included, you mean add the directory to the include flags, yes… So? You're afraid someone is going to use `#include <README.md>`?

And you could have just used [`fs.copyfile`](https://mesonbuild.com/Fs-module.html#copyfile) to copy the include to an `include` sub-directory at setup.

QWERTIOX:
i tried and it didn't see the include directory

benoit-pierre:
Please post the code you used.

QWERTIOX:
```
project(
  'zpp_bits',
  'cpp',
  version: '4.4.22',
  license: 'MIT',
)
fs = import('fs')
copy = fs.copyfile('zpp_bits.h', 'include')

incdir = include_directories('include')

zpp_bits_dep = declare_dependency(include_directories: include_directories())```

**Comment by benoit-pierre** [comment]
That's not how `fs.copyfile` work:
> 
    dest str: the name of the output file. If unset will be the basename of the src argument

Try something like this:
- `meson.build`
```meson
project(
  'zpp_bits',
  'cpp',
  version: '4.4.22',
  license: 'MIT',
)

subdir('include')

zpp_bits_dep = declare_dependency(include_directories: incdir)
```
- `include/meson.build`:
```meson
copy = import('fs').copyfile('../zpp_bits.h')

incdir = include_directories('.')
```

benoit-pierre:
(And go back to the original upstream source distribution).

QWERTIOX:
i think meson.build will still be included this way

QWERTIOX:
i mean it's in the same directory as header

benoit-pierre:
IHMO, you could have kept the original code. The real use case for `fs.copyfile` with respect to headers is to avoid making private headers accessible when they are mixed with public ones, particularly things like headers with generic names, like `config.h`.

> i mean it's in the same directory as header

So? You'll have multiple instances of that in the wraps provided (with makefiles, source files…). Unless someone is foolish enough to try to `include<meson.build>`, what is the harm exactly?

QWERTIOX:
maybe i'm perfectionist but i don't like that my tools reports that meson.build is one of the headers ready to be included

QWERTIOX:
but as you mentioned, other wraps do that so now i don't care

QWERTIOX:
of course i'm stupid - i forget to set c++ at least to c++20

releases.json
```diff
@@ -3686,4 +3694,4 @@
       "1.3.3-1"
     ]
   }
-}
+}
```
neheb, line 3697: uhhh

QWERTIOX: I don't know why it did that, maybe because it's second to last so maybe it's some problem with trailing comma that makes diff freak out

benoit-pierre: No, you removed the terminating newline.

QWERTIOX: Was that bad?

neheb: Yes.

QWERTIOX:
bruh... i didn't stage the rebase...

benoit-pierre:
More precisely, `release.json` was changed from Unix EOLs to DOS.

benoit-pierre:
What editor are you using, which is messing up all your files?

QWERTIOX:
Oh, i was thinking that i should use windows eol and my formatter deletes last eol

benoit-pierre:
No, Unix EOLs are better, and keep the terminating one. But more importantly, avoid reformatting existing file.

benoit-pierre:
Looking good, but note that [4.4.23](https://github.com/eyalz800/zpp_bits/releases/tag/v4.4.23) was released yesterday, so you might want to bump the version. ;)

benoit-pierre (approved):

QWERTIOX:
So should i create another pr or abandon this version

benoit-pierre:
Just amend the top commit.

QWERTIOX:
ofc eol was removed...

benoit-pierre (approved):

PR #1579, opened 2024-07-09
opus: update to 1.5.2

neheb:


neheb:
Known upstream issue: https://github.com/xiph/opus/pull/292

PR #1581, opened 2024-07-12
Add bzip2

blue42u:


subprojects/packagefiles/bzip2/meson.build
```diff
@@ -0,0 +1,109 @@
+project('bzip2', 'c', version: '1.0.8', meson_version: '>=1.0.0')
+
+
+add_project_arguments('-D_FILE_OFFSET_BITS=64', language: 'c')
```
benoit-pierre, line 4: This flag should already be automatically added by meson.

blue42u: TIL, thanks!

jpakkane:
Please keep your commits squashed to a single one.

PR #1583, opened 2024-07-17
Make imgui install with meson install

awest03:
Add `install: true` to library in meson.build


jpakkane:
You need to add a new entry to releases.json.

awest03:
> You need to add a new entry to releases.json.

Thanks, I didn't realise. It should be fixed now

PR #1596, opened 2024-07-26
Add libpfm

blue42u:
Libpfm is unique in that building for ARM only works when also building for Linux. This causes the build to fail for MacOS arm64 in CI. ~~https://github.com/mesonbuild/wrapdb/pull/1597 adds the logic necessary to handle this situation.~~

**EDIT:** To work around the above issue without changes to the CI, this wrap is designed to only support Linux. Attempts to use it on other platforms will result in an error.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -511,6 +511,14 @@
       "libpciaccess-dev"
     ]
   },
+  "libpfm": {
+    "_comment": "Wrap only supports Linux",
+    "build_on": {
+      "windows": false,
+      "msys2": false,
+      "darwin": false
+    }
+  },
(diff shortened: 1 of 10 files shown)


blue42u:
@jpakkane I've adjusted this wrap so it can pass CI by erroring out on any platform other than Linux (i.e. semi-workaround without #1597). Is this a reasonable solution?

jpakkane:
Why are you prefixing variable names with an underscore? It is nonstandard.

blue42u:
Force of habit coming from Python, there private members/functions/variables are prefixed with an underscore. New version pushed to remove them.

PR #1597, opened 2024-07-27
sanity_checks: Also disable builds by CPU family

blue42u:
In some cases the OS is not enough to determine if a build will succeed or fail, the CPU architecture may also change the result. This PR extends `tools/sanity_checks.py` to additionally consult the `host_machine.cpu_family()` (as determined by Meson), if either the system or CPU family is marked `false` the build is expected to fail, unless the composite `{system}-{cpu_family}` key is `true`. By example:

```json5
"wrap_name": {
  // Build expected to fail on Linux on any CPU, e.g. Linux + aarch64
  "build_on": {"linux": false},

  // Build expected to fail on x86_64 on any OS, e.g. Windows + x86_64
  "build_on": {"x86_64": false},

  // Build expected to fail on Linux on any CPU, and on x86_64 on any OS
  "build_on": {"linux": false, "x86_64": false},

  // Build expected to fail on Linux on any CPU, and on x86_64 on any OS,
  // except build expected to pass on Linux + x86_64
  "build_on": {"linux": false, "x86_64": false, "linux-x86_64": true},
}
```

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -335,19 +335,40 @@ class TestReleases(unittest.TestCase):
         self.assertTrue(version in source_url or version_ in source_url,
                         f'Version {version} not found in {source_url}')
 
+    @classmethod
+    def get_cpu_family(cls, builddir: str = '_build'):
+        if not hasattr(cls, "_cpu_family"):
+            options = ['-Dwraps=[]']
+            if Path(builddir, 'meson-private', 'cmd_line.txt').exists():
+                options.append('--wipe')
+            res = subprocess.run(['meson', 'setup', *options, builddir], check=True)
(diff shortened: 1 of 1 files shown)


jpakkane:
Can you post an example real world project that would need this?

blue42u:
There are definitely other libraries with architecture requirements, e.g. [`libunwind`](https://github.com/libunwind/libunwind) or the [`syscalls` Rust crate](https://github.com/jasonwhite/syscalls). Basically any library that uses straight assembly is going to have some restriction here.

That said, of the ones I have interest in developing a wrap for, libpfm is the only one that would fail WrapDB CI without this patch AFAICT. That isn't a very strong argument for this specific feature, I'm open to advice/alternative solutions.

PR #1599, opened 2024-07-30
bdwgc: provide dependency name used by many distros

piperswe:
The pkg-config file for Boehm GC is typically `bdw-gc.pc`, so adding this will allow defaulting to the system libgc and only falling back if not present.

### DIFF
--- a/subprojects/bdwgc.wrap
+++ b/subprojects/bdwgc.wrap
@@ -7,3 +7,4 @@ patch_directory = bdwgc
 
 [provide]
 gc = bdwgc_dep
+bdw-gc = bdwgc_dep


eli-schwartz:
> The pkg-config file for Boehm GC is typically `bdw-gc.pc`, so adding this will allow defaulting to the system libgc and only falling back if not present.

Actually, this appears to be a bug in #487 -- I don't understand why we ever added a "gc" dependency, as it doesn't exist upstream like that. The name "bdw-gc.pc" is the canonical name, it's distributed in https://github.com/ivmai/bdwgc/blob/master/bdw-gc.pc.in

Can you remove the old name and replace it with the new one, and fix the commit message to explain that this is a bug fix, not just "something that many distros do"?

neheb:
Done with https://github.com/mesonbuild/wrapdb/pull/2467

PR #1602, opened 2024-08-01
libjpeg-turbo: add Arm support

amyspark:
Hi all,

This PR adds armv7a and aarch64 to libjpeg-turbo's SIMD support list. I tested this with a SBC (Libre Computer's Le Potato, AML S903x) and with qemu-user-static for armv7a.

Let me know what you think.

cc @nirbheek


subprojects/packagefiles/libjpeg-turbo/simd/meson.build
```diff
@@ -58,6 +69,190 @@ if host_cpu in ['x86', 'x86_64']
...
+        mov MYVAR, #100
+        .unreq MYVAR
+        '''
+      endif
+      f = configure_file(
+        command: [python, '-c', 'import sys; print(sys.argv[1])', '@0@'.format(gastest)],
+        input: 'CMakeLists.txt',
+        output: 'gastest.S',
+        capture: true,
+      )
```
benoit-pierre, line 203: That `input: 'CMakeLists.txt',` argument looks like a mistake.

amyspark: No, this is intended. The `command` is what echoes the real contents of `gastest.S`. `configure_file`'s `input` argument is still required, so in order to generate the test file, I need to supply it.

benoit-pierre: The documentation is confusing. It seems to work fine without it (tested with meson 0.64, 1.2.0 & 1.5.1).

A comment explaining why there's a need to go through a generated file would be great (because the command line used by meson for `cc.compile` put the source file before the extra arguments, so `['-x', 'assembler-with-cpp']` has no impact). Which also means that those arguments can be removed, no?

bgilbert: Both fixed in #2171.  It turns out `-x assembler-with-cpp` was causing misdetections on Clang, which rejects those options at the end of the command line.

PR #1604, opened 2024-08-06
enlog: new wrap

detly:
Wrap for a logging library I wrote and want to share/use elsewhere. Will be interested to see if it passes on Windows first try.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -184,6 +184,14 @@
     ],
     "fatal_warnings": false
   },
+  "enlog": {
+    "build_options": [
+      "enlog:color=true",
+      "enlog:errno=auto",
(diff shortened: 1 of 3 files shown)


detly:
> Will be interested to see if it passes on Windows first try.

It didn't, but was still interesting.

detly:
Okay, this cannot pass tests on Windows + clang-cl, or on MSYS2 + clang, until [LLVM's #61916](https://github.com/llvm/llvm-project/issues/61916) is fixed, or I make my `difftest` thing allow variations. The direction of path separators is not material to regular usage of the library, just for reproducible builds and testing, so if you're okay with ignoring those specific failures, I am too.

PR #1607, opened 2024-08-10
add new wrap: "pegtl"

Totto16:
This is a header only library, so nothing special, the rest is taken from their CMake build definitions (example and test build setup)

Per default tests and examples are only built, if NOT in a subproject, but since the CI executes this as a subproject, I enabled those two there.

### DIFF
--- a/subprojects/packagefiles/pegtl/meson.build
+++ b/subprojects/packagefiles/pegtl/meson.build
@@ -1,0 +1,23 @@
+project(
+  'pegtl',
+  'cpp',
+  version: '3.2.7',
+  meson_version: '>=1.2.0',
+  default_options: {'cpp_std': 'c++17'},
+)
+
+pegtl_dep = declare_dependency(
+  include_directories: include_directories('include'),
+)
+
+meson.override_dependency('pegtl', pegtl_dep)
+
+is_subproject = meson.is_subproject()
+
+if get_option('tests').disable_auto_if(is_subproject).allowed()
+  subdir('src/test/pegtl')
+endif
+
+if get_option('examples').disable_auto_if(is_subproject).allowed()
+  subdir('src/example/pegtl')
+endif
--- a/subprojects/packagefiles/pegtl/src/example/pegtl/meson.build
+++ b/subprojects/packagefiles/pegtl/src/example/pegtl/meson.build
@@ -1,0 +1,63 @@
+example_files = [
+  'abnf2pegtl.cpp',
+  'analyze.cpp',
+  'calculator.cpp',
+  'chomsky_hierarchy.cpp',
+  'csv1.cpp',
+  'csv2.cpp',
+  'dynamic_match.cpp',
+  'expression.cpp',
+  'hello_world.cpp',
+  'indent_aware.cpp',
+  'iri.cpp',
+  'json_analyze.cpp',
+  'json_ast.cpp',
+  'json_build.cpp',
+  'json_count.cpp',
+  'json_coverage.cpp',
+  'json_parse.cpp',
+  'json_print_debug.cpp',
+  'json_print_names.cpp',
+  'json_trace.cpp',
+  'lua53_analyze.cpp',
+  'lua53_parse.cpp',
+  'modulus_match.cpp',
+  'parse_tree.cpp',
+  'parse_tree_user_state.cpp',
+  'peg2pegtl.cpp',
+  'proto3.cpp',
+  'recover.cpp',
+  's_expression.cpp',
+  'sum.cpp',
+  'symbol_table.cpp',
+  'token_input.cpp',
+  'unescape.cpp',
+  'uri.cpp',
+  'uri_print_debug.cpp',
+  'uri_print_names.cpp',
+  'uri_trace.cpp',
+]
+
+args = []
+cpp = meson.get_compiler('cpp')
+
+if cpp.get_argument_syntax() == 'msvc'
+  args += cpp.get_supported_arguments('/W4', '/WX', '/utf-8')
+else
+  args += cpp.get_supported_arguments('-pedantic', '-Wall', '-Wextra', '-Wshadow', '-Werror')
+endif
+
+fs = import('fs')
+
+foreach file : example_files
+
+  name = fs.replace_suffix(file, '')
+
+  example_exe = executable(
+    name,
+    files(file),
+    dependencies: pegtl_dep,
+    cpp_args: args,
+  )
+
+endforeach
--- a/subprojects/packagefiles/pegtl/src/test/pegtl/meson.build
+++ b/subprojects/packagefiles/pegtl/src/test/pegtl/meson.build
@@ -1,0 +1,163 @@
+test_files = [
+  'action_enable.cpp',
+  'action_match.cpp',
+  'actions_one.cpp',
+  'actions_three.cpp',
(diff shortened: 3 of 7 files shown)


subprojects/packagefiles/pegtl/meson.build
```diff
@@ -0,0 +1,24 @@
...
+  'pegtl',
+  'cpp',
+  version: '3.2.7',
+  meson_version: '>=1.2.0',
+  default_options: {'cpp_std': 'c++17'},
+)
+
+pegtl_dep = declare_dependency(
+  include_directories: include_directories('include'),
+  version: meson.project_version(),
```
benoit-pierre, line 11: `version` already defaults to the project version.

Totto16: Ok thanks, didn't know that 👍

subprojects/packagefiles/pegtl/src/example/pegtl/meson.build
```diff
@@ -0,0 +1,66 @@
...
+
+  name = fs.replace_suffix(file, '')
+
+  example_exe = executable(
+    name,
+    files(file),
+    dependencies: pegtl_dep,
+    override_options: {
+      'cpp_std': 'c++17',
+    },
```
benoit-pierre, line 62: Already the project default.

subprojects/packagefiles/pegtl/src/test/pegtl/meson.build
```diff
@@ -0,0 +1,159 @@
...
+    name + '_exe',
+    files(file),
+    dependencies: pegtl_dep,
+    override_options: {
+      'cpp_std': 'c++17',
+    },
+    cpp_args: args,
+  )
+
+  test(name + '_test', test_exe, workdir: meson.project_source_root())
```
benoit-pierre, line 157: I don't think adding the `_test` prefix to all test names is necessary.

subprojects/packagefiles/pegtl/src/test/pegtl/meson.build
```diff
@@ -0,0 +1,159 @@
...
+
+  name = fs.replace_suffix(file, '')
+
+  test_exe = executable(
+    name + '_exe',
+    files(file),
+    dependencies: pegtl_dep,
+    override_options: {
+      'cpp_std': 'c++17',
+    },
```
benoit-pierre, line 153: Same as above.

subprojects/packagefiles/pegtl/src/test/pegtl/meson.build
Totto16: Some time in the past you couldn't name them (executable and tests)the same way, so i named them
`<name>_test` and `<name>_exe`. 
Is it better to name one just `<name>` and is this guaranteed as of meson '>=1.2.0', since i set that in the project() call?

benoit-pierre: I'm not aware of such a limitation.

subprojects/packagefiles/pegtl/src/example/pegtl/meson.build
```diff
@@ -0,0 +1,63 @@
...
+  'uri.cpp',
+  'uri_print_debug.cpp',
+  'uri_print_names.cpp',
+  'uri_trace.cpp',
+]
+
+args = []
+cpp = meson.get_compiler('cpp')
+
+if cpp.get_id() == 'msvc'
```
benoit-pierre, line 44:
```suggestion
if cpp.get_argument_syntax() == 'msvc'
```

subprojects/packagefiles/pegtl/src/test/pegtl/meson.build
Totto16: It works, so I might remember something wrong, either way, I changed it 👍🏼

subprojects/packagefiles/pegtl/src/example/pegtl/meson.build
Totto16: I knew about that, but is seems, that it might be more complicated, since some CI fails on clang-cl, I'll investigate that and this if else later 👍🏼

Totto16: You are indeed correct, it now works on clang-cl too 👍🏼

Totto16:
@benoit-pierre This package doesn't explicitly support msys, but it compiles and configures successfully, but some tests fail, what is the best solution for that, fixing tests and supporting msys officially upstream is not possible that easily and not an option rn.

I disabled msys  via `ci_config.json` but now it complains, that 
```
Exception: Wrap pegtl successfully configured but was expected to fail
```

benoit-pierre:
Yes, that's not how `build_on` works. And from the looks of it, it's just one test, because of encoding issues. It seems like the test in question only account for MSVC on Windows:
```c++
#if defined( _MSC_VER )
#define TAO_PEGTL_TEST_FILENAME u"src/test/pegtl/file_äöü𝄞_data.txt"
#else
#define TAO_PEGTL_TEST_FILENAME "src/test/pegtl/file_äöü𝄞_data.txt"
#endif
```
You could just disable that test on Windows when the compiler is not `clang-cl` or `msvc`.

Totto16:
> ```c++
> #if defined( _MSC_VER )
> ```

On clang64 msys2  3 tests fail for the same reason, see e.g. [here](https://github.com/mesonbuild/wrapdb/actions/runs/10333226401/job/28605345845)

So I disabled those 3 on both msys platforms, as they deal with file names incorrectly

subprojects/packagefiles/pegtl/src/test/pegtl/meson.build
```diff
@@ -0,0 +1,168 @@
...
+  'uint32_general.cpp',
+  'uint64_general.cpp',
+  'uint8_general.cpp',
+  'utf16_general.cpp',
+  'utf32_general.cpp',
+  'utf8_general.cpp',
+  'visit.cpp',
+]
+
+# this tests don't work on msys since they handle file in a windwos style and do not acount for msys
```
benoit-pierre, line 129: I don't think that's quite right, I would have said:
> This tests only work on Windows when compiling with MSVC.

subprojects/packagefiles/pegtl/src/test/pegtl/meson.build
```diff
@@ -0,0 +1,168 @@
...
+]
+
+cpp = meson.get_compiler('cpp')
+
+# Detect if compiling in an MSYS2-based MinGW/Clang environment
+host_is_msys2 = host_machine.system() == 'windows' and cpp.get_argument_syntax() == 'gcc'
+
+if not host_is_msys2
+  test_files += special_test_files
+endif
```
benoit-pierre, line 143: In this instance, I would actually check the compiler id, as there are other compilers with a MSVC syntax that may not work (Cf. https://mesonbuild.com/Reference-tables.html):

```meson

# This tests only work on Windows when compiling with MSVC.
if host_machine.system() != 'windows' or cpp.get_id() in ['clang-cl', 'msvc']
  test_files += [
    'file_file.cpp',
    'file_mmap.cpp',
    'file_read.cpp',
  ]
endif
```

benoit-pierre (approved):

PR #1611, opened 2024-08-13
Add google-highway

WillAyd:
xref conversation with @jan-wassenberg in https://github.com/google/highway/issues/2294

This intentionally does not include any of the contrib packages that are part of the CMake config

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1068,6 +1068,14 @@
       "1.0.7-1"
     ]
   },
+  "google-highway": {
+    "dependency_names": [
+      "hwy"
+    ],
+    "versions": [
+      "1.2.0-1"
+    ]
+  },
   "google-snappy": {
     "dependency_names": [
       "snappy"
--- a/subprojects/google-highway.wrap
+++ b/subprojects/google-highway.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = highway-1.2.0
+source_url = https://github.com/google/highway/releases/download/1.2.0/highway-1.2.0.tar.gz
+source_filename = highway-1.2.0.tar.gz
+source_hash = 58e9d5d41d6573ad15245ad76aec53a69499ca7480c092d899c4424812ed906f
+patch_directory = google-highway
+
+[provide]
+hwy = hwy_dep
--- a/subprojects/packagefiles/google-highway/meson.build
+++ b/subprojects/packagefiles/google-highway/meson.build
@@ -1,0 +1,159 @@
+project(
+  'hwy',
+  'cpp',
+  version: '1.2.0',
+  meson_version: '>= 1.3.0',
+  license: 'Apache-2.0 OR BSD-3-Clause',
+)
+
+hwy_sources = files(
+    'hwy/abort.cc',
+    'hwy/aligned_allocator.cc',
+    'hwy/nanobenchmark.cc',
+    'hwy/per_target.cc',
+    'hwy/print.cc',
+    'hwy/targets.cc',
+    'hwy/timer.cc',
+)
+
+if host_machine.system() == 'windows'
+  add_project_arguments('-DNOMINMAX', language: 'cpp')
+endif
+
+hwy_lib = library(
+    'hwy',
+    hwy_sources,
+    cpp_static_args: ['-DHWY_STATIC_DEFINE'],
+    cpp_shared_args: ['-DHWY_SHARED_DEFINE', '-Dhwy_EXPORTS'],
+    gnu_symbol_visibility: 'hidden',
+)
+
+inc = include_directories('.')
+
+hwy_dep = declare_dependency(
+    link_with: hwy_lib,
+    include_directories: inc,
+)
+meson.override_dependency('hwy', hwy_dep)
+
+if not get_option('tests').disabled()
+  hwy_test_sources = [
+      'hwy/tests/test_util.cc',
+  ]
+
+  hwy_test_lib = library(
+      'hwy-test',
+      hwy_test_sources,
+      cpp_static_args: ['-DHWY_STATIC_DEFINE'],
+      cpp_shared_args: ['-DHWY_SHARED_DEFINE', '-Dhwy_test_EXPORTS'],
+      dependencies: [hwy_dep],
+      gnu_symbol_visibility: 'hidden',
+  )
+
+  hwy_tests = [
+    'hwy/contrib/algo/copy_test.cc',
+    'hwy/contrib/algo/find_test.cc',
+    'hwy/contrib/algo/transform_test.cc',
+    'hwy/abort_test.cc',
+    'hwy/aligned_allocator_test.cc',
+    'hwy/base_test.cc',
+    'hwy/bit_set_test.cc',
+    'hwy/highway_test.cc',
+    'hwy/nanobenchmark_test.cc',
+    'hwy/targets_test.cc',
+    'hwy/examples/skeleton_test.cc',
+    'hwy/tests/arithmetic_test.cc',
+    'hwy/tests/bit_permute_test.cc',
+    'hwy/tests/blockwise_shift_test.cc',
+    'hwy/tests/blockwise_test.cc',
+    'hwy/tests/cast_test.cc',
+    'hwy/tests/combine_test.cc',
+    'hwy/tests/compare_test.cc',
+    'hwy/tests/compress_test.cc',
+    'hwy/tests/convert_test.cc',
+    'hwy/tests/count_test.cc',
+    'hwy/tests/crypto_test.cc',
+    'hwy/tests/demote_test.cc',
+    'hwy/tests/div_test.cc',
+    'hwy/tests/dup128_vec_test.cc',
+    'hwy/tests/expand_test.cc',
+    'hwy/tests/float_test.cc',
+    'hwy/tests/foreach_vec_test.cc',
+    'hwy/tests/if_test.cc',
+    'hwy/tests/in_range_float_to_int_conv_test.cc',
+    'hwy/tests/interleaved_test.cc',
+    'hwy/tests/logical_test.cc',
+    'hwy/tests/mask_combine_test.cc',
+    'hwy/tests/mask_convert_test.cc',
+    'hwy/tests/mask_mem_test.cc',
+    'hwy/tests/mask_slide_test.cc',
+    'hwy/tests/mask_test.cc',
+    'hwy/tests/masked_arithmetic_test.cc',
+    'hwy/tests/masked_minmax_test.cc',
+    'hwy/tests/memory_test.cc',
+    'hwy/tests/minmax_test.cc',
+    'hwy/tests/mul_test.cc',
+    'hwy/tests/reduction_test.cc',
+    'hwy/tests/resize_test.cc',
+    'hwy/tests/reverse_test.cc',
+    'hwy/tests/saturated_test.cc',
+    'hwy/tests/shift_test.cc',
+    'hwy/tests/shuffle4_test.cc',
+    'hwy/tests/slide_up_down_test.cc',
+    'hwy/tests/sums_abs_diff_test.cc',
+    'hwy/tests/swizzle_block_test.cc',
+    'hwy/tests/swizzle_test.cc',
+    'hwy/tests/table_test.cc',
+    'hwy/tests/test_util_test.cc',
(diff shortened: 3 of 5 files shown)


WillAyd:
Not sure what to do with the current CI failures:

  - VisualStudio-clang-cl: maybe an upstream issue with the failing test?
  - MSYS2 (UCRT64) - `error: '_LIBCPP_VERSION' was not declared in this scope` - issue with the cpp compiler?
  - MSYS2 (CLANG64) - `error: undefined symbol: __truncsfbf2` - is this an issue with float16?

jan-wassenberg:
__truncsfbf2 was an [ABI breakage in llvm](https://github.com/llvm/llvm-project/issues/56854). This might indicate the compiler and compiler-rt packages shipped with msys2 were not matching.

The abort_test failure is:
"gtest-port.cc(859): error: Failed
Syntax error at index 16 in simple regular expression "Test Abort from [0-9]+ of abort_test.cc": '[' is unsupported."

We'll change this to use \d.

releases.json
```diff
@@ -1073,7 +1073,7 @@
       "highway"
     ],
     "versions": [
-      "1.2.0-1"
```
dcbaker: This list is purely additive, we never remove entries from it, with entries ordered from newest to oldest.

subprojects/packagefiles/google-highway/meson.build
```diff
@@ -1,7 +1,7 @@
 project(
   'highway',
   'cpp',
-  version: '1.2.0',
+  version: 'fa78bde6a8979031b4ed5a68da7d8cfa1d4255b8c41f49c335b4d7ea402fb769',
```
dcbaker, line 4: We generally do not package random git snapshots, unless the upstream project had decided not to make releases anymore. The documentation on the main git page suggests that highway is using semver, so I'm not sure why we'd move to a git snapshot.

dcbaker (changes requested):

subprojects/packagefiles/google-highway/meson.build
```diff
@@ -0,0 +1,155 @@
+project(
+  'highway',
+  'cpp',
+  version: '1.2.0',
+  license: 'Apache-2.0',
```
dcbaker, line 5: the documentation suggests this should be `Apache-2.0 OR BSD-3-Clause`

subprojects/packagefiles/google-highway/meson.build
```diff
@@ -0,0 +1,155 @@
...
+)
```
dcbaker, line 6: Please set a `meson_version` as well, I'd suggest `meson_version : '>= 1.3.0'`, see the following.

subprojects/packagefiles/google-highway/meson.build
```diff
@@ -0,0 +1,155 @@
...
+    'hwy/print.cc',
+    'hwy/targets.cc',
+    'hwy/timer.cc',
+)
+
+if host_machine.system() == 'windows'
+  add_project_arguments('-DNOMINMAX', language: 'cpp')
+endif
+is_shared = (get_option('default_library') != 'static')
+args = is_shared ? ['-DHWY_SHARED_DEFINE', '-Dhwy_EXPORTS'] : ['-DHWY_STATIC_DEFINE']
```
dcbaker, line 22: Don't do this, it breaks `default_library == 'both'`, instead add:
```
    cpp_static_args : ['-DHWY_STATIC_DEFINE'],
    cpp_shared_args : ['-DHWY_SHARED_DEFINE', '-Dhwy_EXPORTS'],
```

to the `lib_highway` below. This will do the right thing regardless of what `get_option('default_library')`

subprojects/packagefiles/google-highway/meson.build
```diff
@@ -0,0 +1,155 @@
...
+    link_with: lib_highway,
+    include_directories: inc,
+)
+
+if not get_option('tests').disabled()
+  highway_test_sources = [
+      'hwy/tests/hwy_gtest.h',
+      'hwy/tests/test_util-inl.h',
+      'hwy/tests/test_util.cc',
+      'hwy/tests/test_util.h',
```
dcbaker, line 43: Headers are not needed here, they'll just be dropped anyway.

subprojects/packagefiles/google-highway/meson.build
```diff
@@ -0,0 +1,155 @@
...
+  ]
+
+  test_args = is_shared ? ['-DHWY_SHARED_DEFINE', '-Dhwy_test_EXPORTS'] : ['-DHWY_STATIC_DEFINE']
```
dcbaker, line 46: Same thing about the shared vs static args here.

subprojects/google-highway.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = highway-1.2.0
+source_url = https://github.com/google/highway/releases/download/1.2.0/highway-1.2.0.tar.gz
+source_filename = highway-1.2.0.tar.gz
+source_hash = 58e9d5d41d6573ad15245ad76aec53a69499ca7480c092d899c4424812ed906f
+patch_directory = google-highway
+
+[provide]
+highway = google_highway_dep
```
dcbaker, line 9: This doesn't look like it matches what cmake does, which is provide `hwy-config`. This should match (as should the name in the meson.build and the releases.json)

subprojects/packagefiles/google-highway/meson.build
```diff
@@ -0,0 +1,155 @@
...
+    cpp_args: args,
+    gnu_symbol_visibility: 'hidden',
+)
+
+inc = include_directories('.')
+
+google_highway_dep = declare_dependency(
+    link_with: lib_highway,
+    include_directories: inc,
+)
```
dcbaker, line 36:
```meson
meson.override_dependency('hwy', google_highway_dep)
```

releases.json
WillAyd: Yep thanks! Sorry should mark this as draft - the latest commit is just checking to see if the issues with the first commit are resolved by upstream changes

dcbaker (changes requested):
I have some general review below.

Additionally, is there a chance that upstream would take the meson.build files directly? Our preference would be to have them upstream if possible, since that tends to give a better user experience and is generally less work than us having to maintain it out of tree.

subprojects/packagefiles/google-highway/meson.build
WillAyd: Is this the general pattern to always use `meson.override_dependency(...)` for internally declared dependencies? I assume this prevents meson from finding the dependency elsewhere on the system?

subprojects/packagefiles/google-highway/meson.build
WillAyd: Sounds good - was just trying to keep this in sync with the upstream CMake configuration, but assuming that's not a big deal

subprojects/packagefiles/google-highway/meson.build
dcbaker: This is only used for dependencies that you want to export, so in this case `hwy`, but if you have an internally used dependency you don't want to export (say if you wrapped the test helper library in a dependency), then you wouldn't do this.

Yes, it short circuits `dependency('hwy')` to return this dependency.

subprojects/packagefiles/google-highway/meson.build
```diff
@@ -0,0 +1,152 @@
...
+
+if host_machine.system() == 'windows'
+  add_project_arguments('-DNOMINMAX', language: 'cpp')
+endif
+
+hwy_lib = library(
+    'hwy',
+    hwy_sources,
+    cpp_static_args = ['-DHWY_STATIC_DEFINE'],
+    cpp_shared_args = ['-DHWY_SHARED_DEFINE', '-Dhwy_EXPORTS'],
```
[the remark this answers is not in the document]
WillAyd: Thanks!

WillAyd:
> Additionally, is there a chance that upstream would take the meson.build files directly? Our preference would be to have them upstream if possible, since that tends to give a better user experience and is generally less work than us having to maintain it out of tree.

I have no affiliation with google / google-highway so I'll leave that up to @jan-wassenberg

jan-wassenberg:
> Additionally, is there a chance that upstream would take the meson.build files directly? Our preference would be to have them upstream if possible, since that tends to give a better user experience and is generally less work than us having to maintain it out of tree.

hm, the interesting question is with what SLO. What we might be able to offer is to include it in our tree, and to make an effort to add new source files (in particular tests) to the meson build. But if it breaks, we won't be able to help :)

WillAyd:
@dcbaker I'm not super familiar with MSYS2 - do you know how to resolve this issue?

```sh
2024-08-27T17:50:06.4502660Z D:/a/_temp/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/as.exe: subprojects/highway-389fbc6c0e2b91dc1aba92709b14962344915f1d/cast_test.exe.p/hwy_tests_cast_test.cc.obj: too many sections (69839)
2024-08-27T17:50:06.4503089Z D:\a\_temp\msys64\tmp\ccSBPk3T.s: Assembler messages:
2024-08-27T17:50:06.4504829Z D:\a\_temp\msys64\tmp\ccSBPk3T.s: Fatal error: can't write 152 bytes to section .text of subprojects/highway-389fbc6c0e2b91dc1aba92709b14962344915f1d/cast_test.exe.p/hwy_tests_cast_test.cc.obj: 'file too big'
2024-08-27T17:50:06.4506749Z D:/a/_temp/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/as.exe: subprojects/highway-389fbc6c0e2b91dc1aba92709b14962344915f1d/cast_test.exe.p/hwy_tests_cast_test.cc.obj: too many sections (69839)
2024-08-27T17:50:06.4508135Z D:\a\_temp\msys64\tmp\ccSBPk3T.s: Fatal error: subprojects/highway-389fbc6c0e2b91dc1aba92709b14962344915f1d/cast_test.exe.p/hwy_tests_cast_test.cc.obj: file too big
2024-08-27T17:50:06.4508213Z 
```

dcbaker:
@WillAyd I'm not sure, but google suggests that `-Wa,-mbig-obj` might help, so adding something like:
```meson
add_project_link_arguments(cpp.get_supported_link_arguments(['-Wa,-mbig-obj']), language : 'cpp')
```
might help. If not I'll try to rope in one of the resident msys2 experts

WillAyd:
OK I think the meson build definition works for those now, but I am seeing this in the MSYS2 UCRT64 build:

```
2024-08-29T00:15:39.5808278Z Command line: `c++ D:/a/wrapdb/wrapdb/_build/meson-private/tmp6wjk2er3/testfile.cpp -o D:/a/wrapdb/wrapdb/_build/meson-private/tmp6wjk2er3/output.obj -c -D_FILE_OFFSET_BITS=64 -O0 -fpermissive` -> 1
2024-08-29T00:15:39.5809242Z stderr:
2024-08-29T00:15:39.5809752Z D:/a/wrapdb/wrapdb/_build/meson-private/tmp6wjk2er3/testfile.cpp: In function 'int main()':
2024-08-29T00:15:39.5810785Z D:/a/wrapdb/wrapdb/_build/meson-private/tmp6wjk2er3/testfile.cpp:6:17: error: '_LIBCPP_VERSION' was not declared in this scope
2024-08-29T00:15:39.5811482Z     6 |                 _LIBCPP_VERSION;
```

Which I don't think has anything to do with highway? Seems like an issue with the compiler in the environment?

Similarly, I don't think the CLANG64 build will be green on account of https://github.com/mesonbuild/wrapdb/pull/1611#issuecomment-2287906925

WillAyd:
Is there a way to skip the tests just for the windows / mysy platforms? I think the test failure in version 1.2.0 for windows is pretty pedantic, and the msys issue seems out of the control of the library

WillAyd:
This is being added upstream in highway so no need for this PR

jcapriot:
@WillAyd Do you still intend on adding a wrap file here (just without the patch)?

WillAyd:
Yea I think that makes sense. Happy to revive this PR, or, since you did the upstream, feel free to submit your own PR. 

You did the hard part so want to make sure you get the credit :-)

PR #1614, opened 2024-08-16
catch2: update to 3.7.0

stephanlachnit:


UnixY2K:
I think it could be related to catchorg/Catch2#2791

however the main issue seems to be related to `clang-cl` as the comment shows:
https://github.com/catchorg/Catch2/blob/fa306fc85eca7cc68a5362c503019fa823cbe411/tests/SelfTest/UsageTests/Misc.tests.cpp#L520

probably something like `#ifndef __clang__ ` could fix it in upstream?

Issue #1612, opened 2024-08-15
Delete vs_module_defs from cmocka?


gszy:
I’m not sure about the exact purpose of the `vs_module_defs` argument and the file it refers to, but here’s my problem with it: I have a C project with some cmocka-based unit tests. In a native Linux build the test binaries are linked to system `libcmocka.so.0` and work OK. In a MinGW-w64 build (where cmocka is a subproject) they’re linked to `cmocka.dll`, which does not exist, because Meson builds `libcmocka.dll`.

It turned out (thanks to `@elibrokeit`’s help on IRC) that the issue comes from the file that `vs_module_defs` refers to, where the library name is set to `cmocka.dll`. When I delete the argument, the test binaries are linked to `libcmocka.dll` and work OK. If you think the argument is not necessary, I can provide a PR.

eli-schwartz:
Well, I also said the library should be overridden to respect the name from the .def file. If you remove it entirely then a mingw toolchain will act like Unix and export every symbol from the code, and MSVC will export none of them -- thereby breaking the DLL.

That file lists the exact set of symbols that should be exported.

neheb:
Probably better to fix the name.

gszy:
In [the `library()` function documentation](https://mesonbuild.com/Reference-manual_functions.html#library) I’ve found parameter `name_prefix`, which could be set to empty string for Windows to always produce `cmocka.dll`, but I’m not sure it’s a good idea, especially since the description reads:

> […] By default this is lib on all platforms and compilers, except for MSVC shared libraries where it is omitted to follow convention […]

So it seems `libcmocka.dll` is the correct name for MinGW-w64 builds and `cmocka.dll` would be the correct name for MSVC builds. Perhaps I have to change something in my project? Why does it want to link with the MSVC library if I compile it with MinGW? I have pushed a minimal example to [Codeberg](https://codeberg.org/gs/meson-cmocka-test), could you have a look at it?

neheb:
def file says cmocka.dll. Which means you would have to get rid of lib for MinGW

gszy:
Oh, I see there are some precedents:
https://github.com/search?q=repo%3Amesonbuild%2Fwrapdb%20name_prefix&type=code
I’m preparing a PR.

PR #1615, opened 2024-08-16
cmocka: drop library name_prefix on Windows

gszy:
The current version of [cmocka](https://cmocka.org/) has source file `src/cmocka.def`, which provides the list of library symbols for MSVC builds. The library name in that file is simply `cmocka`, which results in projects using the library trying to link to `cmocka.dll`. When cross compiling with Meson and MinGW-w64 though, the actually built library file is `libcmocka.dll`.

For Windows builds, set [library](https://mesonbuild.com/Reference-manual_functions.html#library) `name_prefix` to empty string, so that Meson builds `cmocka.dll`, as expected by binaries linked to the library. Keep the default prefix for other operating systems.

Fixes: https://github.com/mesonbuild/wrapdb/issues/1612


eli-schwartz (changes requested):
Thanks. Please also bump the revision in releases.json so that we re-release a fixed wrap of cmocka.

neheb:
msys2 patches the def file to use the lib prefix: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-cmocka/002-dll-name.patch

Not worth doing so here.

neheb (approved):

PR #1617, opened 2024-08-17
Bump pango to 1.54.0

nirbheek:


nirbheek:
Hm, somehow icu is being picked up (via harfbuzz, which is a dep of pango) and it's failing to build only with clang/gcc on Windows. I believe this is because `test_releases()` is injecting icu.wrap into the build. So we need to fix that wrap too...

neheb:
Could avoid fixing the icu wrap by installing ICU on ci_config.json I think.

bgilbert:
https://github.com/mesonbuild/wrapdb/pull/1881 updates pango to 1.56.0.

bgilbert:
Superseded by https://github.com/mesonbuild/wrapdb/pull/1881.

PR #1555, opened 2024-06-21
Add tool to check for new upstream versions and autoupdate if possible

bgilbert:
Use upstream release data from Anitya ([release-monitoring.org](https://release-monitoring.org/)).  Anitya has a database of known software projects, a polling mechanism for identifying new releases, and a mechanism to map downstream package names to Anitya-tracked projects.

I've [populated Anitya](https://release-monitoring.org/distro/Meson%20WrapDB/) with WrapDB package names (click the names in the package name column), manually verifying in each case that the Anitya project matches the software actually shipped in WrapDB.  In cases where Anitya doesn't know about a project, I have mostly not added a new record, and am not sure whether or not we should add records in all such cases.

With Anitya populated, we can query a single HTTPS endpoint to get the current versions of all our projects it knows about.

Add a tool to use this data.  It can:

* List wraps, their versions in WrapDB, and their versions upstream.  Command-line options can select ports with or without official Meson support upstream, ports with or without updates available, and ports where Anitya doesn't know the current upstream version (because Anitya doesn't know about the project or doesn't know it's packaged in WrapDB).

    <details>
    <summary><tt>tools/versions.py list</tt> (click for sample run)</summary>
    <pre>
    abseil-cpp                     20230802.1  =>      20240116.2
    arduinocore-avr                     1.8.2  =>           1.8.6
    argparse                              3.0
    argtable3                           3.2.2
  * argu-parser                         0.0.1  =>|
    asio                               1.30.2
    backward-cpp                          1.6  =>|
    bdwgc                               8.2.2  =>           8.2.6
  * blueprint-compiler                 0.10.0  =>          0.12.0
    box2d                               2.4.1
    bshoshany-thread-pool               4.1.0  =>|
    c-ares                             1.24.0  =>          1.31.0
  * c-flags                             1.5.8  =>|
  * cairo                              1.18.0
    catch                               2.2.2  =>         2.13.10
  * catch2                              3.6.0
    catchorg-clara                      1.1.5  =>|
    centurion                           7.3.0
    cereal                              1.3.2
  * cexception                          1.3.3  =>|
  * cglm                                0.9.2  =>           0.9.4
    check                              0.15.2
    chipmunk                            7.0.3
    cjson                              1.7.17  =>          1.7.18
  * cli11                               2.4.1  =>           2.4.2
    cmark-gfm                   0.29.0.gfm.13
  * cmock                               2.5.3  =>|
    cmocka                              1.1.7
[...]

PR #1618, opened 2024-08-18
Remove autoupdate workflow

bgilbert:
This workflow was intended to automatically update wraps for projects that officially include Meson support upstream (see #1555).  The workflow requires some setup, however, and there doesn't seem to be much interest in running it within the `mesonbuild/wrapdb` repo.  As a result, the workflow has been [running and failing](https://github.com/mesonbuild/wrapdb/actions/workflows/autoupdate.yml) twice a week.  Remove the workflow from this repo, but leave the tool that it runs (`tools/versions.py`) for manual use.

I still think there's value in automatically updating these wraps, even if Meson isn't interested in hosting the workflow directly, so I'm planning to explore running it from a fork that isn't officially managed by the project.


bgilbert:
I've moved the [automatic update workflow](https://github.com/wrapdb-bot/autoupdate/actions/workflows/autoupdate.yml) to [this repo](https://github.com/wrapdb-bot/autoupdate).  To avoid spamming the repo, I've initially limited it to three open PRs, but will eventually lift that restriction if all goes well.

neheb:
Core developers don't care. Sorry for the inconvenience.

jpakkane:
> Core developers don't care. Sorry for the inconvenience.

It's not so much a case of not caring but of lack of time. For example I have spent all my time on the option refactor for _weeks_. There simply is not enough time to handle these kinds of issues. Sorry about that. It would be great if WrapDB had its own dedicated team of people who could focus on improving and running it in the way they deem best. Splitting the time of Meson devs between these two is not a great thing for anyone involved.

PR #1625, opened 2024-08-22
[libmicrohttpd] Correctly compute size of `timeval.tv_sec` on Windows.

mgautierfr:
On Windows, `timevale.tv_sec` is not a `time_t` but a long [1].
Libmicrohttpd check this value at runtime with assert. [2]

`prefix` is a copy of prefix used in upstream `configure.ac` plus the

   #ifdef _WIN32
   #include <zinsock.h>
   #endif

added.

[1] https://learn.microsoft.com/en-us/windows/win32/api/winsock/ns-winsock-timeval
[2] src/microhttpd/daemon.c, line 8210


mgautierfr:
I have added the release "0.9.77-3".

Is it possible to also have a "0.9.76-4" ? Simply add the version in `releases.json` ?

subprojects/packagefiles/libmicrohttpd/meson.build
```diff
@@ -118,7 +118,25 @@ foreach s : ['size_t', 'unsigned int', 'unsigned long long']
   cdata.set('SIZEOF_@0@'.format(s.underscorify().to_upper()), cc.sizeof(s))
 endforeach
 
-cdata.set('SIZEOF_STRUCT_TIMEVAL_TV_SEC', cc.sizeof('time_t', prefix: '#include <time.h>'))
+cdata.set(
```
neheb, line 121: this is garbage and will not work.

mgautierfr: Well, it works when compiling natively on Windows, Linux and MacOs.
And cross compiling from Linux 64bits to Linux 32 bits, arm (32 and 64 bits), Anroid (various archs).

And it is what is done by upstream libmicrohttpd autotools: https://github.com/Karlson2k/libmicrohttpd/blob/master/configure.ac#L2448-L2474

So can you explain a bit more "garbage" and "will not work" ?

kelson42: > this is garbage and will not work.

Extremly rude and inefficient feedback!

neheb: You ported an autotools check thinking meson works the same as autotools. Why do you think those macros for those headers are available?

Just switch the check from time_t to long.

mgautierfr: > Why do you think those macros for those headers are available?

Which macro are you speaking about ?

> Just switch the check from time_t to long.

`long` is not the right type outside of Windows.
My first test was indeed to use `long` when on Windows but it seems proper to me to test the size of `timeval.tv_sec` whatever is its real type to be platform independent.

benoit-pierre: > > Why do you think those macros for those headers are available?
> 
> Which macro are you speaking about ?

All those `HAVE_xxx`: they're not defined, so those `#ifdef HAVE_[…] #endif` blocks are dead code.

mgautierfr: Ok. New commit (to fixup) simply use `long` on Windows.

PR #1627, opened 2024-08-24
Box2d-v2.4.1 to Box2d-v3.0.1

AnkushRoy-code:
Upgrading box2d version that's it

There are 2 problems though:
1. I removed subprojects/packagefiles/box2d/unit-test. Reason:


subprojects/box2d.wrap
```diff
@@ -1,8 +1,8 @@
 [wrap-file]
-directory = box2d-2.4.1
-source_url = https://github.com/erincatto/box2d/archive/v2.4.1/box2d-2.4.1.tar.gz
-source_filename = box2d-2.4.1.tar.gz
-source_hash = d6b4650ff897ee1ead27cf77a5933ea197cbeef6705638dd181adc2e816b23c2
+directory = box2d-3.0.1
+source_url = https://github.com/AnkushRoy-code/box2d/archive/v3.0.1/box2d-3.0.1.tar.gz
```
AnkushRoy-code, line 3: This needs to be changed

AnkushRoy-code:
Oh no my include statements!! :/ no big deal

AnkushRoy-code:
I am not a windows user or have worked with multi threading so I don't fully understand why the test failed. I googled a lot and found that the feature trying to be used is not supported in Visual Studio (natively I guess). The author of box2d had this in his cmakelists.txt:
```
if (MSVC)
       ...
	# Atomics are still considered experimental in Visual Studio 17.8
	target_compile_options(box2d PRIVATE /experimental:c11atomics)
	...
	endif()
```

so. I don't know should I add this compile argument if the compiler is visual studio in my meson.build. Will this work? Or is there something else that is causing the error?

AnkushRoy-code:
I give up

AnkushRoy-code:
I’ve spent the ENTIRE day battling with Visual Studio and I’m absolutely at my wits' end! I wrestled with configurations all day, and nothing seems to work. My code runs perfectly in MSYS2 and even with clang-cl, but Visual Studio just refuses to cooperate. I’m beyond annoyed at this point.

I genuinely thought this would be a quick fix, but instead, I’ve wasted a whole day and now I’m staring down an exam tomorrow with zero prep. I can’t believe I let this stupid IDE mess up my schedule like this. I’m closing this pull request because I am DONE with Visual Studio. I don’t want to touch it ever again, and honestly, I don’t want to touch Windows either. This is just absurd!

PR #1628, opened 2024-08-25
Add tinyexpr mathematical expression parsing library

awest03:
Links:
- https://github.com/codeplea/tinyexpr
- https://codeplea.com/tinyexpr

There aren't any releases of the library, so I've numbered it `0.0.0`. The library isn't updated very often as it is mostly complete so I don't think this will be a problem going into the future.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -3391,6 +3391,14 @@
       "1.1.1-1"
     ]
   },
+  "tinyexpr": {
+    "dependency_names": [
+      "tinyexpr"
+    ],
+    "versions": [
+      "0.0.0-1"
+    ]
+  },
   "tinyfsm": {
     "dependency_names": [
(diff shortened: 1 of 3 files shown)


awest03:
The Sanity checks are failing because the version `0.0.0` isn't in the source url (because there is no release of this library)

jpakkane:
> There aren't any releases of the library,

WrapDB has a policy that it only accepts unmodified upstream releases.

awest03:
That's quite reasonable. The maintainer of tinyexpr doesn't seem to want to do a release when [asked a few years ago](https://github.com/codeplea/tinyexpr/issues/42). I'll close this now but I won't delete the branch in case someone outside of my projects wants to use it.

PR #1630, opened 2024-08-26
sanity_checks: add support for installing Python packages

benoit-pierre:
In preparation for Homebrew disabling `python-packaging`:

> python-packaging has been deprecated! It will be disabled on 2024-10-05.

Cf. https://github.com/Homebrew/homebrew-core/issues/157500

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -151,14 +151,18 @@ jobs:
           fetch-depth: 0
 
       - run: brew update
-      # github actions overwrites brew's python. Force it to reassert itself, by running in a separate step.
-      - name: unbreak python in github actions
-        run: |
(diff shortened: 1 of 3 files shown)


neheb:
MSYS2 deprecates too?

benoit-pierre:
No, but if you referring to the harfbuzz related changed in `ci_config.json`: there's a glib dependency, no point in having 2 different packages installed for the same requirement.

benoit-pierre:
Actually, I'm not use if `packaging` is also needed by `harfbuzz`, if if this was added just for glib.

benoit-pierre:
Never mind, it does not matter, it's needed by glib and need to be installed, including for the `VisualStudio (x86)` (where installing it is not part of the setup anymore).

PR #1637, opened 2024-08-28
add godot-4.3

v-for-vandal:
Updating godot wrap to godot-4.3


releases.json
```diff
@@ -1049,7 +1049,7 @@
       "godot_cpp"
     ],
     "versions": [
-      "4.2-1"
+      "4.3"
```
benoit-pierre, line 1052: Don't remove old versions: prepend the new one first.

benoit-pierre:
You need to update the projcet version in `meson.build` too.

releases.json
```diff
@@ -1049,6 +1049,7 @@
       "godot_cpp"
     ],
     "versions": [
+      "4.3",
```
benoit-pierre, line 1052:
```suggestion
      "4.3-1",
```

benoit-pierre: Don't forget the revision (bumped if the wrap is updated, but the project version is unchanged).

v-for-vandal: image

I don't think that is correct :( For example 4.2 and 4.2-1 are different branches (and they point to different commits). I can write "4.3-0", if you think it would be better.

benoit-pierre: It's the wrap revision, not the minor project version: `4.3-1` not `4.3.1`.

subprojects/packagefiles/godot-cpp/meson.build
```diff
@@ -1,7 +1,7 @@
 # Disclaimer: I've found this file on the internet at some point and could
 # not find it since. The original author is unknown to me :(
 
-project('godot-cpp', 'cpp', version: '4.2',
+project('godot-cpp', 'cpp', version: '4.3-0',
```
benoit-pierre, line 4: No, leave the project version unchanged:
```suggestion
project('godot-cpp', 'cpp', version: '4.3',
```

Project version != wrap release version.

v-for-vandal: sorry :(

Is there any way to ran ./tools/sanity_check locally only for godot ?

releases.json
v-for-vandal: Ok, got it.

v-for-vandal:
Ok, I did an extra check with my godot project and it seems everything is fine. No missing symbols, no unexpected segfaults.  I think you can merge the PR.

benoit-pierre (approved):

PR #1640, opened 2024-08-31
libsigcplusplus-3: init 3.6.0

judemille:
Name has the `-3` due to it being an ABI level. Tried to do `sigc++-3.0` to match the pkg-config name, but the sanity checks refuse.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1832,6 +1832,14 @@
       "3.6.4-1"
     ]
   },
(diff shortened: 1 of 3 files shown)


judemille:
Some minor patches have been done to the upstream meson.build to account for changes either not yet released (Meson minimum version) or not yet merged (cl-like compiler detection).

PR #1648, opened 2024-09-02
cairomm: init 1.18.0

judemille:
Patches to the Meson manifests are pending upstream.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -267,6 +267,14 @@
       "1.17.8-1"
     ]
   },
+  "cairomm": {
+    "dependency_names": [
+      "cairomm-1.16"
+    ],
+    "versions": [
+      "1.18.0-1"
+    ]
+  },
   "catch": {
     "versions": [
       "2.2.2-1",
--- a/subprojects/cairomm.wrap
+++ b/subprojects/cairomm.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = cairomm-1.18.0
+source_url = https://www.cairographics.org/releases/cairomm-1.18.0.tar.xz
+source_filename = cairomm-1.18.0.tar.xz
+source_hash = b81255394e3ea8e8aa887276d22afa8985fc8daef60692eb2407d23049f03cfb
+patch_directory = cairomm
+
+[provide]
+dependency_names = cairomm-1.16
--- a/subprojects/packagefiles/cairomm/data/meson.build
+++ b/subprojects/packagefiles/cairomm/data/meson.build
@@ -1,0 +1,61 @@
+# data
+
+# Input: install_prefix, install_libdir, install_datadir, install_includedir,
+#        install_pkgconfigdir, cairomm_pcname, cairomm_api_version,
+#        cairomm_requires, msvc14x_toolset_ver, cairo_dep
+# Output: -
+
+pkg_conf_data = configuration_data()
+pkg_conf_data.set('prefix', install_prefix)
+pkg_conf_data.set('exec_prefix', '${prefix}')
+pkg_conf_data.set('libdir', '${exec_prefix}' / install_libdir)
+pkg_conf_data.set('datarootdir', '${prefix}' / install_datadir)
+pkg_conf_data.set('datadir', '${datarootdir}')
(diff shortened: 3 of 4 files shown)


judemille:
There are build failures because of dependencies producing warnings. I have no idea what to do here.

eli-schwartz:
```
FAILED: subprojects/pixman-0.43.4/pixman/libpixman-mmx.a.p/pixman-mmx.c.obj 
"clang-cl" "-Isubprojects\pixman-0.43.4\pixman\libpixman-mmx.a.p" "-Isubprojects\pixman-0.43.4\pixman" "-I..\subprojects\pixman-0.43.4\pixman" "/MDd" "/nologo" "/showIncludes" "/utf-8" "/W2" "/clang:-std=gnu99" "/Od" "/Zi" "-Wdeclaration-after-statement" "-fno-strict-aliasing" "-Wundef" "-Wno-unused-local-typedefs" "-DHAVE_CONFIG_H" "-mmmx" "-Winline" "/Fdsubprojects\pixman-0.43.4\pixman\libpixman-mmx.a.p\pixman-mmx.c.pdb" /Fosubprojects/pixman-0.43.4/pixman/libpixman-mmx.a.p/pixman-mmx.c.obj "/c" ../subprojects/pixman-0.43.4/pixman/pixman-mmx.c
../subprojects/pixman-0.43.4/pixman/pixman-mmx.c(247,8): error: member reference base type '__m64' (vector of 1 'long long' value) is not a structure or union
  247 |     res.M64_MEMBER = x;
      |     ~~~^~~~~~~~~~~
../subprojects/pixman-0.43.4/pixman/pixman-mmx.c(262,21): error: member reference base type '__m64' (vector of 1 'long long' value) is not a structure or union
  262 |     uint64_t res = x.M64_MEMBER;
      |                    ~^~~~~~~~~~~
```

Not a warning. :P

The visualstudio issue is related to https://mesonbuild.com/Builtin-options.html#compiler-options
> Since *1.3.0* `c_std` and `cpp_std` options now accept a list of values. Projects that prefer GNU C, but can fallback to ISO C, can now set, for example, `default_options: 'c_std=gnu11,c11'`, and it will use `gnu11` when available, but fallback to c11 otherwise. It is an error only if none of the values are supported by the current compiler. Likewise, a project that can take benefit of `c++17` but can still build with `c++11` can set `default_options: 'cpp_std=c++17,c++11'`. This allows us to deprecate `gnuXX` values from the MSVC compiler. That means that `default_options: 'c_std=gnu11'` will now print a warning with MSVC but fallback to `c11`. No warning is printed if at least one of the values is valid, i.e. `default_options: 'c_std=gnu11,c11'`. In the future that deprecation warning will become an hard error because `c_std=gnu11` should mean GNU is required, for projects that cannot be built with MSVC for example.


Unfortunately fixing it would require bumping pixman's minimum version of meson to 1.3.0...

judemille:
> Not a warning. :P

Welp, I missed that in the output. Guess even turning fatal warnings off wouldn't be enough.

Any ideas on what to do here?

eli-schwartz:
Perhaps the codebase is incompatible with clang-cl. Do they wish to support it upstream? It may be able to be fixed there.

It is possible to set c_std as a build option to something better via ci_config.json, which seems worth it in this case. But it's possible it still won't end up being able to be compiled for Windows.

eli-schwartz:
Looks like the pixman CI failed when it was updated too, but no one investigated. It would be good to know if pixman configured with c99 at least builds on visualstudio, even if not on clang-cl.

judemille:
With no changes, pixman builds with my MSVS 2022 install.

I'm investigating making a patch now.

judemille:
@eli-schwartz I've contributed patches to pixman, cairomm, and cairo to resolve warnings and build issues. In the meantime until the next releases of those, what should I do?

eli-schwartz:
> In the meantime until the next releases of those, what should I do?

We yeet it in with broken Windows CI on the rationale that we have responded as best we can, and the clang-cl job is "best-effort" anyway. Hopefully after the next release of the various packages the CI will be clean.

PR #1655, opened 2024-09-05
glib-networking: update to 2.80.0

bgilbert:
Alternative to #1641.


bgilbert:
CI failures:

- MSVC x86 fails because the OpenSSL wrap doesn't have a pregenerated config for x86:

  ```
  openssl| Message: Generating OpenSSL configs...
  subprojects\openssl-3.0.8\meson.build:18:4: ERROR: Problem encountered: Generator only works on Linux, other platforms are not supported  
  ```

- macOS looks like a substantive failure:

  ```
  not ok /tls/openssl/connection/basic - GLib-FATAL-WARNING: poll(2) failed due to: Bad file descriptor.
  ```

  I've verified that 2.78.0 fails on macOS with the same error, so this is not a regression.

neheb:
Alternatively, this can use GnuTLS AFAIK.

I hate the OpenSSL wrap FWIW.

PR #1657, opened 2024-09-06
cjson: update to 1.7.18

bgilbert:


bgilbert:
MSVC test failures are expected according to a comment in `meson.build`.

jpakkane:
Then `ci-config.json` should be edited to make it not run the build on VS.

bgilbert:
AFAICT tests are failing because they're broken, not because they're fundamentally expected to fail, so removing the CI signal seems counterproductive.

In any case, `ci_config.json` doesn't do what you're asking for.  Setting `"build_on": {"windows": false}` still produces CI failures with this message:

```
Exception: Wrap cjson successfully configured but was expected to fail
```

PR #1662, opened 2024-09-07
ci: remove debug traces and update `build_all` python for macOS

benoit-pierre:
Follow up to #1630.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -167,14 +167,11 @@ jobs:
           fetch-depth: 0
 
       - run: brew update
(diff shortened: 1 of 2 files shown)


benoit-pierre:
Is the Strawberry Perl workaround for VisualStudio not needed for `build_all` too?

neheb:
Guess not.

neheb:
I see EXTERNALLY-MANAGED is gone for macOS. Could this work for Alpine as well?

benoit-pierre:
No, since chroot is used.

PR #1603, opened 2024-08-04
sdl2: update to 2.30.6 + misc improvements

benoit-pierre:
https://github.com/libsdl-org/SDL/releases/tag/release-2.30.6

PR #1663, opened 2024-09-07
sdl2: fix regression in header lookup

Totto16:
In https://github.com/mesonbuild/wrapdb/pull/1603/ the header lookup was changed, before it, we used `cc.check_header` (https://github.com/mesonbuild/wrapdb/pull/1603/files#diff-ce9d055aa1e27dc2674a4d13fe96cf1dc0cc0031de8e7269e18b0d62518b0719L568) but after that we use `cc.has_header` (https://github.com/mesonbuild/wrapdb/pull/1603/files#diff-ce9d055aa1e27dc2674a4d13fe96cf1dc0cc0031de8e7269e18b0d62518b0719R762)

This can lead to a wrong sdl_config.h in some situations, an example:

When cross compiling for android (aarch64), the compilation fails with the following error: 

```
 .../toolchains/android-ndk-r27b/toolchains/llvm/prebuilt/linux-x86_64/lib/clang/18/include/immintrin.h:14:2: error: "This header is only meant to be used on x86 and x64 architecture"
   14 | #error "This header is only meant to be used on x86 and x64 architecture"
```

(See e.g. this CI run https://github.com/OpenBrickProtocolFoundation/oopetris/actions/runs/10747059046/job/29808986656?pr=178#step:8:11627 )

If we look closer, this is due to the config, since it detected, that we have `immintrin.h`, sets `HAVE_IMMINTRIN_H` to true and then SDL uses it. By using `cc.check_header` it works, since the header doesn't pass this, but passes`cc.has_header` .


benoit-pierre:
IMHO, it would have been better to special case only the check for `immintrin.h`, since `has_header` is much faster than `check_header`.

PR #1678, opened 2024-09-09
Provide nlohmann/json dependency with multiple headers

vid512:
meson.build of nlohmann/json contains additional dependency with multiple headers, instead of single one. This PR allows wrapDB users to use it.

Multiple headers are useful eg. if you want to forward-declare nlohmann::json object  (see `json_fwd.hpp`).

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -2469,9 +2469,11 @@
   },
   "nlohmann_json": {
     "dependency_names": [
-      "nlohmann_json"
+      "nlohmann_json",
+      "nlohmann_json_multiple_headers"
     ],
     "versions": [
+      "3.11.3-2",
       "3.11.3-1",
       "3.11.2-1",
       "3.10.5-1",
--- a/subprojects/nlohmann_json.wrap
+++ b/subprojects/nlohmann_json.wrap
@@ -8,3 +8,4 @@ source_hash = a22461d13119ac5c78f205d3df1db13403e58ce1bb1794edc9313677313f4a9d
 
 [provide]
 nlohmann_json = nlohmann_json_dep
+nlohmann_json_multiple_headers = nlohmann_json_multiple_headers


eli-schwartz:
IIRC nlohmann_json is never installed this way and the dependency lookup won't ever work for the general case. Not sure this is a good idea...

vid512:
What do you mean "this way"? Via wrapdb?

I just tested it: 
- installed dependency wrap it via `meson wrap install nlohmann_json`, 
- manually edited `subprojects/nlohmann_json.wrap` as in https://github.com/mesonbuild/wrapdb/pull/1678/commits/c1b781c106f29c9a43913b7b50edbb07084e853e, to simulate this PR
- included `dependency('nlohmann_json_multiple_headers')` in meson.build
- now cpp file with `#include <nlohmann/json_fwd.hpp>` and forward-declared `nlohmann::json` compiles just fine

dcbaker:
Me either. I'm looking, and I would rather see if upstream would take an update to their Meson. I'm opening a PR right now for that purpose. (I use this project downstream, so I have some interest in making this better)

dcbaker:
https://github.com/nlohmann/json/pull/4452

vid512:
OK I see my mistake now. For whatever reason I thought json_fwd.hpp is only available with multiple headers, I completely missed that json_fwd.hpp is in single_include as well. Yes, this is pointless. At least it prompted update upstream. :)

eli-schwartz:
The other factor here is that the usual name is `pkg-config --cflags nlohmann_json` or for cmake, `find_package(nlohmann_json)`.

Implementing a new dependency name would mean it only worked as a wrap, but not as a system package, which is a naming wart and probably a leaky internal implementation detail. And implementation details could probably be used by projects that really want them, via `subproject().get_variable()`.

It would generally only matter for projects that don't care about the forward declarations, but rather want to overly specify whether they use amalgamated headers.

PR #1685, opened 2024-09-14
pcre2: forbid `jit_sealloc` on Android and don't require it by default

bgilbert:
The non-NetBSD version of this allocator requires `secure_getenv()`, which doesn't exist in Bionic.  Disable it on Android.  Prevents build error:
    
    /opt/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android31-clang  -o subprojects/pcre2-10.44/pcre2grep subprojects/pcre2-10.44/pcre2grep.p/src_pcre2grep.c.o -Wl,--as-needed -Wl,--no-undefined -pie subprojects/pcre2-10.44/libpcre2-8.a -pthread
    ld: error: undefined symbol: secure_getenv
    >>> referenced by sljitProtExecAllocatorPosix.c:73 (../subprojects/pcre2-10.44/src/sljit/allocator_src/sljitProtExecAllocatorPosix.c:73)
    >>>               src_pcre2_jit_compile.c.o:(create_tempfile) in archive subprojects/pcre2-10.44/libpcre2-8.a

On platforms that don't support `jit_sealloc`, disable it by default rather than forcing the superproject to override the option.


bgilbert:
The test failure is expected (https://github.com/mesonbuild/wrapdb/pull/1545#issuecomment-2161745749).

neheb:
Wonder if CI is available for this.

PR #1696, opened 2024-09-15
add imgui-docking

tmayoff:
The commit history get really messed up with the last [PR](https://github.com/mesonbuild/wrapdb/pull/1227).

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -412,6 +412,14 @@
       "libvulkan-dev"
     ]
   },
+  "imgui-docking": {
+    "debian_packages": [
+      "liballegro5-dev",
+      "libglfw3-dev",
+      "libsdl2-dev",
+      "libvulkan-dev"
+    ]
+  },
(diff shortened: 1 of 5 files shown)


benoit-pierre:
You could have rebased and force-pushed.

tmayoff:
I tried a couple of times, but I'm not going to spend time messing with it. This was easier and faster

tmayoff:
Thanks @benoit-pierre for the help. This should be good to merge, the failure is the same as the non-docking branch

AstroOrbis:
If we could update this to 1.91.3 to match the `imgui` dependency or update both to use 1.91.5 and 1.91.5-docking respectively that'd be amazing, as using a local patch isn't always a viable/maintainable option. LGTM!

tmayoff:
I'll bump the version once I get a chance closer to the weekend, otherwise someone else can do it before then now that this is merged

PR #1697, opened 2024-09-16
mockcpp: add `mockcpp` C/C++ test framework (git-master)

brendan-simon-indt:
Add the `mockcpp` C/C++ test framework.

mockcpp is a jmock-like generic C/C++ Mock Framework, which doesn't use complex template technique which will result in very heavy compiling overhead.

https://github.com/sinojelly/mockcpp

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -2372,6 +2372,11 @@
       "3.0.6-1"
     ]
   },
+  "mockcpp": {
+    "versions": [
+      "git-master"
+    ]
+  },
   "mocklibc": {
     "versions": [
       "1.0-2"
--- a/subprojects/mockcpp.wrap
+++ b/subprojects/mockcpp.wrap
@@ -1,0 +1,10 @@
+[wrap-git]
+method = cmake
+; directory = mockcpp-head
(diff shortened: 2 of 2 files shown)


bgilbert:
In general, software in WrapDB must be an official release with a version number, not a Git branch.  (Also, the project hasn't had any Git commits in almost two years; is it still maintained?)

brendan-simon-indt:
I'm not sure if it counts as a release, but there was tag on 26 June 2021.  There have been various commits since, but nothing recently.  I have asked the maintainer if they would create a release.

The wrap file worked for me, in the sense that MSYS2 UCRT environment was able to download and compile the `mockcpp` project, and I was able to compile a unit test (with `Catch2` via wrap DB) that linked against `mockcpp` and produce an executable that ran as expected.

I trialed `mockcpp` with `Catch2` for a work project.  I was happy with the initial test results.
I only submitted the PR to help others easily use `meson` and `mockcpp` if they felt inclined.

Maybe wrap DB can have some kind of experimental or development versioning scheme to give access to some projects, knowing that these may not be guaranteed to work.

PR #1705, opened 2024-09-19
taocpp-json: add taoJSON version 1.0.0-beta.14

rherilier:


rherilier:
I opened #1706 to handle dashes in version strings

releases.json
```diff
@@ -3436,6 +3436,14 @@
       "1.12-1"
     ]
   },
+  "taocpp-json": {
+    "dependency_names": [
+      "taocpp_json"
```
eli-schwartz, line 3441: The upstream dependency name is taocpp-json, with a dash not an underscore!

subprojects/packagefiles/taocpp-json/meson.build
```diff
@@ -0,0 +1,16 @@
...
+    default_options: [
+        'cpp_std=c++17',
+    ]
+)
+
+taocpp_json_inc = include_directories('include')
+
+taocpp_json_dep = declare_dependency(
+    include_directories: [taocpp_json_inc]
+)
```
eli-schwartz, line 16: The upstream CMakeLists.txt depends on PEGTL, why doesn't this?

eli-schwartz (changes requested):
What about hooking up the tests in src/test/json ?

releases.json
rherilier: my bad, I mimic how`abseil-cpp` was defined to make taocpp-json's files. So, I thought that dependency were the symbol defined in the *provide* section of wrap file...

There are many other entries with dependencies in the usual meaning too.

So, I looked in`releases.json` and found that many entries have no *dependency_names* node, (such *catch* or *mocklibc*). Does it imply that the wrap name is implicitly used as a dependency?

Can you help me clarifying this point?

subprojects/packagefiles/taocpp-json/meson.build
rherilier: taocpp-json is header-only, PEGTL is only required for tests.

releases.json
eli-schwartz: > my bad, I mimic how`abseil-cpp` was defined to make taocpp-json's files. So, I thought that dependency were the symbol defined in the _provide_ section of wrap file...

They are, indeed, the symbol defined in the provide section of the wrap file.'

For example, abseil-cpp provides the dependency symbol:

/usr/lib64/pkgconfig/absl_base.pc

and hence uses "absl_base" in its provide section.

Similarly, taocpp-json provides taocpp-json-config.cmake, and hence should list "taocpp-json" in the wrap releases, to annotate the fact that the provide section has been fixed to provide "taocpp-json".

> and found that many entries have no _dependency_names_ node, (such _catch_ or _mocklibc_).

They are very old and buggy.

subprojects/packagefiles/taocpp-json/meson.build
eli-schwartz: What does "header-only" have to do with it? How is someone expected to utilize taocpp-json if PEGTL is not installed and checked for?

releases.json
rherilier: thanks you for the clarification

subprojects/packagefiles/taocpp-json/meson.build
rherilier: not my day... all my assumptions are based on the Debian's package which has not been updated for 4 years, and at that time, taoJSON embedded PEGTL.

rherilier:
> What about hooking up the tests in src/test/json ?

I can add it if it is mandatory. If so, do I have to add examples and performance programs too?

PR #1708, opened 2024-09-21
libtiff: update to 4.7.0; support 12-bit JPEG

bgilbert:
Autodetect dual 8/12-bit JPEG support in libjpeg-turbo 3.0+.

Also fix dual 8/12-bit JPEG support in the libjpeg-turbo wrap.


subprojects/packagefiles/libtiff/meson.build
```diff
@@ -75,6 +76,12 @@ cdata.set('HAVE_JBG_NEWLEN', jbig_dep.found())
 cdata.set10('HAVE_DECL_OPTARG', cc.has_header_symbol('unistd.h', 'optarg'))
 cdata.set10('HAVE_GETOPT', cc.has_header_symbol('getopt.h', 'getopt'))
 
+if cc.has_function('jpeg12_read_scanlines', dependencies: jpeg_dep)
```
benoit-pierre, line 79: Won't this fail when the dependency is a wrap?

bgilbert: Yup, good catch.

benoit-pierre: I think after `>=2.1.91`, support can be assumed to be present (Cf. https://github.com/libjpeg-turbo/libjpeg-turbo/commit/e8b40f3c2ba187ba95c13c3e8ce21c8534256df7), so maybe only use `cc.has_function` on older versions?

benoit-pierre: Yep see https://github.com/libjpeg-turbo/libjpeg-turbo/commit/b5a9ef64ea299423b7fb46c47f4a70697a442dd4.

bgilbert: 2.1.9x were betas, so I've special-cased internal dependencies to check for 3.0.0 and up.

PR #1710, opened 2024-09-23
zlib-ng: add

bgilbert:
Based on `zlib.wrap` and zlib-ng's CMake build.

The wrap omits these upstream features:

- Support for building without optimizations (`WITH_OPTIM=OFF`), with `-march=native` (`WITH_NATIVE_INSTRUCTIONS=ON`), without new strategies, without runtime CPU detection, or with reduced memory requirements
- Support for building as C99
- Configurable API function name prefix
- Forcible override of optimization levels greater than 2, enablement of `-fno-semantic-interposition`, and use of software floating point on 32-bit ARM.  It's not the build system's job to override the user's settings.  We do forcibly disable LTO to prevent it potentially merging blocks built for different CPU subarches.  (GCC, at least, won't do that in any case.)
- Defining `HAVE_SYMVER` (optional, and currently a no-op)
- Defining `ZLIB_DEBUG` in debug builds.  Wrap users probably don't want this.
- Naming the static build `zlibstatic` on Windows
- Maintainer warnings
- Sanitizers, fuzzers, benchmarks, coverage, and most tests

Also, the exact build logic differs from upstream in a number of small details.

The wrap always uses the dependency name `zlib-ng`, even when building in zlib-compat mode.  Meson will not allow zlib and zlib-ng to share the dependency name `zlib` within WrapDB.  Parent projects can invoke `meson.override_dependency('zlib', ...)` if desired.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -3834,6 +3834,14 @@
       "1.2.8-2"
     ]
   },
+  "zlib-ng": {
+    "dependency_names": [
+      "zlib-ng"
+    ],
+    "versions": [
+      "2.2.2-1"
+    ]
+  },
   "zpp_bits": {
     "dependency_names": [
       "zpp_bits"
--- a/subprojects/packagefiles/zlib-ng/LICENSE.build
+++ b/subprojects/packagefiles/zlib-ng/LICENSE.build
@@ -1,0 +1,19 @@
+(C) 1995-2024 Jean-loup Gailly and Mark Adler
+
+This software is provided 'as-is', without any express or implied
+warranty. In no event will the authors be held liable for any damages
+arising from the use of this software.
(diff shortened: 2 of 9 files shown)


okias:
Thank you for this PR!

dcbaker:
> The wrap always uses the dependency name zlib-ng, even when building in zlib-compat mode. Meson will not allow zlib and zlib-ng to share the dependency name zlib within WrapDB. Parent projects can invoke meson.override_dependency('zlib', ...) if desired.

This seems like a place that the wrap infrastructure could be enhanced, there are likely more cases where a dependency can be provided by multiple upstreams. Not saying you have to do this, just noting it.

PR #1716, opened 2024-09-25
json-c: update to 0.18

bgilbert:
Set library version to match upstream.  Avoid Meson warning about `c_std=gnu99` on MSVC.


subprojects/packagefiles/json-c/meson.build
```diff
@@ -81,6 +83,7 @@ libjson_c = library(
   sources,
   include_directories: inc,
   install: true,
+  version: '5.4.0',
```
jpakkane, line 86: For anyone else looking into this. `5.4.0` _is_ the correct version number, or at least the one used in the CMake files.

PR #1718, opened 2024-09-25
Convert subprojects to `c_std=gnu99,c99`

bgilbert:
On MSVC, Meson &ge; 1.3.0 warns that `c_std=gnu99` should be spelled `c_std=gnu99,c99`, but older Meson rejects the latter.  Condition `c_std` on the Meson version.


bgilbert:
We can't do this for `physfs` yet because of https://github.com/mesonbuild/meson/issues/13639.

bgilbert:
Meson could really use a future-feature warning to make it more obvious that the conditional syntax is required.  Issue filed as https://github.com/mesonbuild/meson/issues/13713.

bgilbert:
Is anything blocking this?  It seems like it'd be relatively uncontroversial.

PR #1719, opened 2024-09-25
libuv: update to 1.49.1

bgilbert:
Fix build on Meson < 1.3.0.


bgilbert:
UCRT64 should be fixed by https://github.com/libuv/libuv/pull/4586 and macOS arm64 is tracked upstream in https://github.com/libuv/libuv/issues/4574.  We can't disable individual tests from the Meson config because we'd also need to modify `test/test-list.h`.

neheb:
@jpakkane would be a good idea to give @bgilbert merge access.

PR #1725, opened 2024-10-04
sqlite3: remove shell from library

eerii:
The `shell.c` file is responsible for the sqlite command line tool, which shouldn't be a part of the sqlite library. It includes a `main` function that causes duplicate symbol errors when linking statically. It wasn't originally present in the other repo (https://github.com/mesonbuild/sqlite3/blob/3.34.1/meson.build)


bgilbert:
Change LGTM.  Looks like `shell.c` was added in #781, without explanation.

Please add `3.46.1-2` to `releases.json`.

jpakkane:

eerii:
Sorry about that! It should be added now

PR #1726, opened 2024-10-04
Update nanobind from 2.1.0 to 2.2.0

WillAyd:
FYI @wjakob


wjakob:
Are further changes needed to enable the free-threading support? (There were some changes in the CMake build system related to this)

WillAyd:
@wjakob that's a good callout and sorry I missed here. I think just need to detect when the Python distribution is free threaded and define the appropriate nanobind macro

PR #1730, opened 2024-10-04
tools/versions: update URLs with dashed or underscored version

bgilbert:
Some projects use tag or tarball names with dashes or underscores as the version component separator.


bgilbert:
Anyone want to take a look?

bgilbert:
@neheb, any objection to merging?

PR #1741, opened 2024-10-08
nanoarrow: Update to 0.6.0

paleolimbot:
Just checking the nanoarrow 0.6.0 release candidate for issues prior to release!

cc @WillAyd


releases.json
```diff
@@ -2464,6 +2464,7 @@
       "nanoarrow"
     ],
     "versions": [
+      "0.6.0",
```
WillAyd, line 2467: I think need to append -1 to the end of this

WillAyd:
I could be misreading but I think that error is strictly due to how we are mixing the RC tag with the git hash in this PR, and probably not something with our Meson configuration.

To test the RC you might be able to replace `0.6.0-1` temporarily with `33d2c8b973d8f8f424e02ac92ddeaace2a92f8dd-1`

WillAyd:
Ah sorry I was looking at the wrong error message. When I run locally (after the pulling tags as described in https://github.com/mesonbuild/wrapdb/issues/1742)  I do see the issue is with the Arrow dependency. I'll take a closer look

```sh
Called: `/usr/bin/pkg-config --modversion arrow` -> 1
stderr:
Package arrow was not found in the pkg-config search path.
Perhaps you should add the directory containing `arrow.pc'
to the PKG_CONFIG_PATH environment variable
Package 'arrow', required by 'virtual:world', not found
-----------
CMake binary for host machine is cached.

Determining dependency 'arrow' with CMake executable '/home/willayd/mambaforge/envs/pantab-dev/bin/cmake'
Try CMake generator: auto
Calling CMake (['/home/willayd/mambaforge/envs/pantab-dev/bin/cmake']) in /home/willayd/clones/wrapdb/_build/meson-private/cmake_arrow with:
  - "-DNAME=arrow"
  - "-DARCHS=i386-linux-gnu;x86_64-linux-gnu"
  - "-DVERSION="
  - "-DCOMPS="
  - "-DSTATIC=OFF"
  - "--trace-expand"
  - "--trace-format=json-v1"
  - "--no-warn-unused-cli"
  - "--trace-redirect=cmake_trace.txt"
  - "-DCMAKE_TOOLCHAIN_FILE=/home/willayd/clones/wrapdb/_build/meson-private/cmake_arrow/CMakeMesonToolchainFile.cmake"
  - "."
WARNING: Could not find and exact match for the CMake dependency arrow.

However, Meson found the following partial matches:

    ['Arrow::arrow_shared']

Using imported is recommended, since this approach is less error prone
and better supported by Meson. Consider explicitly specifying one of
these in the dependency call with:

    dependency('arrow', modules: ['arrow::<name>', ...])
```

WillAyd:
Yea sorry - that comment was ill-advised. Actually I think if you just change the version as described to match the git hash you will see more green.

When we merged this originally not all of the checks were green anyway (Ubuntu, and the VisualStudio checks were in a red state). I'm guessing there might be something missing with the Arrow pkgconfig resolution on those platforms, but I don't think that is a blocker

eli-schwartz:
> I'm guessing there might be something missing with the Arrow pkgconfig resolution on those platforms, but I don't think that is a blocker

Yeah, I believe that is because Debian/Ubuntu still do not package apache-arrow at all, so while you could install it in ci_config.json for alpine and brew you could not do so for debian. There is nothing really to do about this, the failing CI does not indicate an actionable issue nor an issue in the wrap itself.

eli-schwartz:
Probably also a good idea to squash and force push. ;)

eli-schwartz:
Thanks for the ping. The macOS issue is bizarre (it appears to spend 4 hours installing everything from python to LLVM to boost to cmake to grpc *from source* and then bombs out merging the results?)

PR #1752, opened 2024-10-17
spdlog: install headers and pkg-config file

stephanlachnit:
Same reasoning as in https://github.com/mesonbuild/wrapdb/pull/1746


sebastianstricker:
I just want to mention that i created https://github.com/mesonbuild/wrapdb/pull/1907 since this change broke my current pipeline.

It requires elevated privileges to install headers system wide or the prefix needs to be pointed to a dummy location. Tough i did not try the latter, as it seems like a hacky fix, not something longterm.

Anyway, i think it does not make sense to always install, as pointed out in my PR. I am no expert with meson, but I did not find a way to disable installation on a subproject basis from the main projects meson.build file. Which would arguably just be the best option.

I am open for getting enlightened in case i missed the point or how to avoid my troubles with it. I see you applied the same change to multiple wraps and thought I'd write this to discuss.

PR #1756, opened 2024-10-20
Add wrap files for Kraken-Engine-0.0.1

durkisneer1:


durkisneer1:
I've been forgotten 😔

jpakkane:
You need to rebase against current master to make the error go away.

durkisneer1:
Got it, thanks. There were no errors with the sanity checks on my local system with VS and MingW so hopefully the next workflow pass doesn't also fail.

jpakkane:
You used a merge. Those are not allowed because they screw up history. Convert it to a rebase instead. Thanks.

Issue #1758, opened 2024-10-23
nanobind: `limited_api` not propagated to nanobind static library build


lgarrison:
Hi @WillAyd,

Thank you for the excellent nanobind wrapper! I'm running into an issue when compiling limited API extensions that you might have some insight into. If I have a `meson.build` snippet like this:

```meson
nanobind_dep = dependency('nanobind', static: true)
py.extension_module(
  'my_module_name',
  sources: ['src/example.cpp'],
  dependencies: [nanobind_dep],
  install: true,
  limited_api : '3.12'
)
```

The compiler output shows that `src/example.cpp` is built with `-DPy_LIMITED_API=0x030c0000`, but nanobind itself is not. For example:

```
[1/13] ccache c++ -Imy_module_name.abi3.so.p -I. -I.. -I../subprojects/nanobind-2.2.0/include -I../subprojects/robin-map-1.3.0/include -I/install/include/python3.12 -I/mnt/home/lgarrison/.local/share/uv/python/cpython-3.12.7-linux-x86_64-gnu/include/python3.12 -fvisibility=hidden -fvisibility-inlines-hidden -fdiagnostics-color=always -D_GLIBCXX_ASSERTIONS=1 -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c++17 -O0 -g -fPIC -ffunction-sections -fdata-sections -DPy_LIMITED_API=0x030c0000 -MD -MQ my_module_name.abi3.so.p/src_example.cpp.o -MF my_module_name.abi3.so.p/src_example.cpp.o.d -o my_module_name.abi3.so.p/src_example.cpp.o -c ../src/example.cpp
[...]

PR #1760, opened 2024-10-24
nanobind: compile nanobind lib with `-DNB_FREE_THREADED` with nogil python

lgarrison:
Also export the flag to downstream extensions.

I tested this locally, and it seemed to have the correct behavior. Specifically, both the nanobind library and the Python extension were compiled with `-DNB_FREE_THREADED` under free-threaded Python, and neither were with "normal" Python.

As discussed in https://github.com/mesonbuild/wrapdb/issues/1758#issuecomment-2433447758.


lgarrison:
CC @WillAyd

subprojects/packagefiles/nanobind/meson.build
```diff
@@ -52,6 +52,13 @@ if get_option('buildtype') == 'release'
   add_project_arguments('-DNB_COMPACT_ASSERTIONS', language: 'cpp')
 endif
 
+freethreaded = py.get_variable('Py_GIL_DISABLED', 0) != 0
+if freethreaded
+  freethreaded_def = '-DNB_FREE_THREADED'
+  add_project_arguments(freethreaded_def, language : 'cpp')
+  dep_compile_args += [freethreaded_def]
```
WillAyd, line 59: Does this need to be passed through the dependency? I was under the impression it was only needed internally by nanobind

WillAyd (commented):
Thanks!

subprojects/packagefiles/nanobind/meson.build
lgarrison: This was a guess based on looking at the nanobind source. The main header, `nanobind/nanobind.h`, transitively includes headers like `nb_types.h` that make use of `NB_FREE_THREADED`, e.g. [here](https://github.com/wjakob/nanobind/blob/fd22b8c457cd291db2774a5526b5be39376212cd/include/nanobind/nb_types.h#L829)

WillAyd: Ah interesting...I think that makes sense then

WillAyd (approved):
thanks @lgarrison !

PR #1763, opened 2024-10-27
monocypher: new with v4.0.2-1

rtgiskard:


rtgiskard:
According to an upstream [PR](https://github.com/LoupVaillant/Monocypher/pull/268), it should build on windows, but not sure what's the problem with VisualStudio ..

neheb:
You need to provide a .def file with all the symbols.

rtgiskard:
> You need to provide a .def file with all the symbols.

Should that be handled by meson automatically? After all, it builds well on other platform

rtgiskard:
Any recomendation on the case, failed to skip build_on windows with the ci_config.json. Personally speaking, I don't use windows and not really care whether it build on windows or not.

PR #1765, opened 2024-10-30
add constraint src to chipmunk2d build

johnalanbrook:
Noticed when using chipmunk2d for my own project that uses the constraints, they were not present in the meson build.


jpakkane:
You need to add a new release to `releases.json`.

PR #1767, opened 2024-10-30
New Wrap: cfitsio-4.5.0

kjmeagher:
This is a file wrap with a patch. It is mostly based on the
CMakeLists.txt provided in the source. But there were a few defines
in the autotolls version that werent in the cmake that I also added.

### DIFF
--- a/subprojects/packagefiles/cfitsio/meson.build
+++ b/subprojects/packagefiles/cfitsio/meson.build
@@ -10,8 +10,13 @@ SOVERSION = 10
 
 cc = meson.get_compiler('c')
 libm = cc.find_library('m', required: false)
-zlib = dependency('zlib', method: 'pkg-config', required:true, fallback: ['zlib', 'zlib_dep'])
-deps = [libm,zlib]
+zlib = dependency(
+  'zlib',
+  method: 'pkg-config',
+  required: true,
+  fallback: ['zlib', 'zlib_dep'],
+)
+deps = [libm, zlib]
 has_fortran = false
 
 if host_machine.system() == 'windows'
@@ -24,8 +29,8 @@ else
     required: get_option('fortran'),
   )
 
-  bzip2 = cc.find_library('bz2', required: false)
-  if not bzip2.found()
+  bzip2 = cc.find_library('bz2', required: get_option('bzip2'))
+  if not bzip2.found() and get_option('bzip2').allowed()
     bzip2_proj = subproject('bzip2')
     bzip2 = bzip2_proj.get_variable('bzip2_dep')
   endif
@@ -58,7 +63,7 @@ endif
 if cc.has_header('unistd.h')
   add_project_arguments('-DHAVE_UNISTD_H', language: 'c')
   if cc.has_header_symbol('unistd.h', 'ftruncate')
-     add_project_arguments('-DHAVE_FTRUNCATE', language: 'c')
+    add_project_arguments('-DHAVE_FTRUNCATE', language: 'c')
   endif
 else
   add_project_arguments('-DYY_NO_UNISTD_H', language: 'c')
@@ -182,11 +187,8 @@ libcfitsio = library(
   soversion: SOVERSION,
   install: true,
 )
-message('libcfitsio: '+libcfitsio.name()+' '+libcfitsio.full_path())
 
-cfitsio_dep = declare_dependency(
-  link_with: libcfitsio,
-)
+cfitsio_dep = declare_dependency(include_directories: '.', link_with: libcfitsio)
 
 pkg = import('pkgconfig')
 pkg.generate(
@@ -227,7 +229,7 @@ if get_option('utils')
     'funpack',
     'utilities/funpack.c',
     'utilities/fpackutil.c',
-    dependencies: [libm,cfitsio_dep],
+    dependencies: [libm, cfitsio_dep],
     link_with: libcfitsio,
     install: true,
   )


jpakkane:
> cfitsio| WARNING: Project does not target a minimum version but uses feature introduced in '1.1': meson.options file. Use meson_options.txt instead

You might want to just require a newer version of Meson instead.

subprojects/packagefiles/cfitsio/meson.build
```diff
@@ -0,0 +1,270 @@
...
+  native: false,
+  required: get_option('fortran'),
+)
+
+SOVERSION = 10
+
+cc = meson.get_compiler('c')
+libm = cc.find_library('m')
+
+if target_machine.system() != 'windows'
```
eli-schwartz, line 20: target_machine is used when the thing you are creating is a cross compiler that will, when built and installed and run, emit platform specific outputs for a platform that isn't the one it is installed for.

Probably you mean host_machine instead, as that is the machine that libcfitsio is being compiled to run on.

subprojects/packagefiles/cfitsio/meson.build
```diff
@@ -0,0 +1,270 @@
...
+  curl = dependency(
+    'libcurl',
+    required: get_option('curl'),
+    fallback: ['curl', 'libcurl'],
```
eli-schwartz, line 24: Fallback is probably not necessary since curl wraps can just `[provide] curl = curl_dep`.

subprojects/packagefiles/cfitsio/meson.build
```diff
@@ -0,0 +1,270 @@
...
+
+threads = dependency('threads', required: get_option('reentrant'))
+if threads.found()
+  add_project_arguments('-D_REENTRANT', language: 'c')
+endif
+
+zlib = dependency('zlib', fallback: ['zlib', 'zlib_dep'])
+
+if target_machine.system() != 'windows'
+  bzip2 = dependency(
```
eli-schwartz, line 39: bzip2 doesn't provide pkg-config files, please detect it with cc.find_library instead.

subprojects/packagefiles/cfitsio/meson.build
```diff
@@ -0,0 +1,270 @@
...
+)
+  add_project_arguments('-DHAVE_NET_SERVICES', language: 'c')
+  if cc.has_header_symbol('stdio.h', 'fmemopen')
+    add_project_arguments('-DHAVE_FMEMOPEN', language: 'c')
+  endif
+endif
+
+if get_option('sse2')
+  if cc.has_argument('-msse2')
+    add_project_arguments('-msse2', language: 'c')
```
eli-schwartz, line 64: This seems like an odd thing to have as a project option. Can't users just compile with -msse2 themselves, especially given its often the default for a given -march?

subprojects/packagefiles/cfitsio/meson.build
```diff
@@ -0,0 +1,270 @@
...
+  endif
+endif
+
+if cc.has_type('long long')
+  add_project_arguments('-DHAVE_LONGLONG', language: 'c')
+endif
+
+install_headers('fitsio.h', 'fitsio2.h', 'longnam.h')
+
+libcfitsio_la_SOURCES = [
```
eli-schwartz, line 112: The `_la` refers to "libtool archive" so I wonder how useful it is to have that in the name.

subprojects/packagefiles/cfitsio/meson.build
```diff
@@ -0,0 +1,270 @@
...
+if has_fortran
+  libcfitsio_la_SOURCES += [
+    'f77_wrap1.c',
+    'f77_wrap2.c',
+    'f77_wrap3.c',
+    'f77_wrap4.c',
+  ]
+endif
+
+libcfitsio = shared_library(
```
eli-schwartz, line 185: Does upstream really not support installing this as a static library e.g. using `--enable-static --disable-shared`? Usually libtool provides these options for free.

subprojects/packagefiles/cfitsio/meson.build
```diff
@@ -0,0 +1,270 @@
...
+  name: 'cifitsio',
+  description: 'FITS File Subroutine Library',
+  url: 'https://heasarc.gsfc.nasa.gov/fitsio/',
+)
+
+executable(
+  'fitscopy',
+  'utilities/fitscopy.c',
+  dependencies: cfitsio_dep,
+  install: get_option('utils'),
```
eli-schwartz, line 212: This option should be used in an "if" statement to control whether the executable() is defined -- not whether the install parameter is set.

subprojects/packagefiles/cfitsio/meson.options
```diff
@@ -0,0 +1,38 @@
...
+)
+option(
+    'reentrant',
+    type: 'feature',
+    description: 'Enable reentrant multithreading',
+)
+option(
+    'utils',
+    type: 'boolean',
+    description: 'Build fpack, funpack, fitscopy, and fitsverify executables',
```
eli-schwartz, line 14: ... and as we see here the option does indeed claim it will control building, not control installing.

eli-schwartz (changes requested):

subprojects/packagefiles/cfitsio/meson.build
kjmeagher: fixed

subprojects/packagefiles/cfitsio/meson.build
kjmeagher: removed

subprojects/packagefiles/cfitsio/meson.build
kjmeagher: both the platforms i developed this for (arch linux and macos) have pkg-config files for bzip2. but since they are not provided by up stream i will switch to find_library()

subprojects/packagefiles/cfitsio/meson.build
kjmeagher: I was copying from the autotools/cmake file, but you are right this should be removed

subprojects/packagefiles/cfitsio/meson.build
kjmeagher: removed

subprojects/packagefiles/cfitsio/meson.build
kjmeagher: this library is often used as a static library, switched to library()

PR #1772, opened 2024-11-04
wrapdb server static v1 api mirror and nginx config update

### LINKED ISSUE #1768: wrapdb.mesonbuild.com v1 API support
This is just a heads-up, so people know where to report issues and who to ping if anything comes up.

This only affects wraps with ancient Meson versions.

## What is happening

I plan on switching over wrapdb.mesonbuild.com to server v1 api requests (project list, project info, latest version, get wrap, get patch zip) to be served from static content instead of scripts and a local database.

This means substring search for wraps via substrings will no longer work, but that's user interactive and I think it's ok to retire that after 4 years. People will just have to browse the list on the website or the full project list returned.

Reason is to simplify the server setup (Hetzner will force-reboot the machine on Nov 8 and it's unclear if the v1 scripting setup will come back up on its own, and it seems a good idea anyway to get rid of that).

## Background

wrapdb.mesonbuild.com is a tiny VM on Hetzner that runs an nginx web server and handles API requests and wrap downloads for v1 API and v2 API.

In case of v2 API (introduced in v0.59 released on 18 Jul 2021) it's just redirects to github.

However, there's still support for the v1 API active which was used up to v0.58 (released on 2 May 2021).

Two bits of functionality, with a local database maintained on the server:

1. downloads of wraps and patches
2. remote API for listing available wraps and getting info on them

Mirror script and mirrored static content is here: https://github.com/tp-m/wrapdb-v1-api-mirror/

I have tested the new setup with the Meson v0.58 manual wrap tests, and it seems to work fine (using `/x1/` as api prefix before the `v1` redirect to static content goes live).

I have also run the mirror script against the existing v1 api (response from server scripts) and the redirect-to-static-content api variant on the server, and the mirrored files are identical in both cases.

tp-m:
See #1768 for background.

Content of this MR is basically from https://github.com/tp-m/wrapdb-v1-api-mirror/, seems better to have it right here in the repo next to the nginx config.

Closes #1768


eli-schwartz:
I'd prefer to avoid having all these files directly checked into git, though. e.g. we use GitHub Releases to store the ZIP files for all current releases because checking them into git tends to be rather suboptimal usage of git's object storage design.

tp-m:
Understood, but I'm just trying to document what's deployed right now on the server, so that it's easily possible to replicate the setup on a new machine if ever needed.

The size of the .zip files here is 977kB and that will not change going forward, since the v1 archive is frozen now.

If someone wants to make sure that **all** v1 zips (incl. sqlite) are available on github, so we can add redirects for them a la [this](https://github.com/mesonbuild/wrapdb/blob/master/nginx/default#L21) be my guest.

In the meantime I'm just documenting what's what, and seeing the fairly small size I don't think it's worth worrying about too much.

bgilbert:
Since these are derived artifacts of this repo, and not intended to change, it does seem awkward to commit them here as though they're the canonical editable copy.  Keeping them in a separate Git repo seems to make sense - maybe we should just add some documentation pointing to https://github.com/tp-m/wrapdb-v1-api-mirror/?

tp-m:
> Since these are derived artifacts of this repo, and not intended to change, it does seem awkward to commit them here as though they're the canonical editable copy. Keeping them in a separate Git repo seems to make sense - maybe we should just add some documentation pointing to https://github.com/tp-m/wrapdb-v1-api-mirror/?

Sure I can do that, just thought we might want to have everything together in one place.

tp-m:
Attaching it here as well then for posterity: [v1-static.tar.gz](https://github.com/user-attachments/files/17688595/v1-static.tar.gz)

tp-m:
Updated.

nginx/default
```diff
@@ -19,7 +16,8 @@ server {
   }
 
   # Redirect legacy v1 patch URLs.
-  # FIXME: This breaks download of sqlite that has not been imported to v2 because it is replaced by sqlite3
+  # FIXME: This breaks download of sqlite that has not been imported to v2 because it is replaced by sqlite3 (xclaesse)
+  # NOTE: no longer needed, we now serve v1 patch URLs directly from a static files mirror (in /var/www/v1-static, see below) (Tim, Nov 2024)
```
xclaesse, line 20: If it's no longer needed, just remove it :)

xclaesse: OTOH, why not just use that redirect? Files are already hosted on github, we don't need another static mirror, do we?

tp-m: As per my previous comment, if you want to make sure the files are all there including the old sqlite things I'm all for it, I'm just not planning to do that work, and I also suspect there might be some discussions otherwise surely you'd have just done it back when you added the comment.

tp-m: I don't think we should break things is what I'm saying, so if you want to enable the redirect make sure it won't break anything.

xclaesse: I ran this test script inside wrapdb-v1-api-mirror repo:
```python3
import glob
import requests

for i in glob.glob('**/get_zip', recursive=True):
    parts = i.split('/')
    name = parts[2]
    version = parts[3]
    revision = parts[4]
    url = f'https://github.com/mesonbuild/wrapdb/releases/download/{name}_{version}-{revision}/{name}_{version}-{revision}_patch.zip'
    response = requests.get(url)
    if response.status_code == 200:
        print(url, 'OK')
    else:
        print(url, 'not found')
```

Missing  URLs are:
https://github.com/mesonbuild/wrapdb/releases/download/libjpeg_9c-4/libjpeg_9c-4_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/libjpeg_9c-2/libjpeg_9c-2_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/libjpeg_9c-3/libjpeg_9c-3_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/libjpeg_9c-1/libjpeg_9c-1_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/libjpeg_9a-4/libjpeg_9a-4_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/libjpeg_9a-2/libjpeg_9a-2_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/libjpeg_9a-5/libjpeg_9a-5_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3310100-1/sqlite_3310100-1_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3320300-4/sqlite_3320300-4_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3320300-3/sqlite_3320300-3_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3320300-5/sqlite_3320300-5_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3250100-2/sqlite_3250100-2_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3250100-1/sqlite_3250100-1_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3230100-1/sqlite_3230100-1_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3080802-4/sqlite_3080802-4_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3080802-2/sqlite_3080802-2_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3080802-5/sqlite_3080802-5_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3140200-2/sqlite_3140200-2_patch.zip
https://github.com/mesonbuild/wrapdb/releases/download/sqlite_3140200-1/sqlite_3140200-1_patch.zip

IMHO it's fine to break queries and download of wraps, we just need to keep patch zips. Those 2 deps are trivial to update to v2, I don't think it's a big deal to break them, it's been deprecated for years.

tp-m: Re. "trivial to update" that's beside the point because people will be consuming third party source repositories/branches/tags that they can't easily update without switching to their own modified copies or overwriting the checkouts manually.

xclaesse:
- Print deprecation if a wrap still uses v1: https://github.com/mesonbuild/meson/pull/13903.
- Uploaded missing sqlite and libjpeg patches as github release assets, to fixes @eli-schwartz 's concern about using binaries inside git.
- Fix for wrapdb sanity check to ignore newly created assets: https://github.com/mesonbuild/wrapdb/pull/1780.
- With those changes, a simple redirect is enough to keep all v1 wraps working. `meson wrap search` with Meson <= 0.58 will be broken, I personally don't see that as a problem.

bgilbert:
Merging this since it reflects the actual deployed config.  I've filed https://github.com/mesonbuild/wrapdb/issues/2520 to track removal of the static mirror.

PR #1783, opened 2024-11-13
CI: use macos-latest

neheb:
12 is deprecated and it seems Homebrew removed bottles, forcing compilation of everything, slowing CI down massively.


bgilbert:
This implicitly drops macOS tests on x86_64, so that arch should be removed
from the macOS matrix to avoid running the same CI twice.

PR #1804, opened 2024-12-02
Define no asm source for libffi meson build

aaalloc:
Actually when compiling libffi with wrapdb patch, it gives 

```
subprojects/libffi-3.4.6/src/meson.build:89:10: ERROR: File wasm32/sysv.S does not exist.
```
and that's normal behavior because there is no sysv.S for wasm32 (https://github.com/libffi/libffi/tree/master/src/wasm32).

Current PR just put empty asm_sources when target is wasm32


releases.json
```diff
@@ -3952,4 +3953,4 @@
       "1.3.3-1"
     ]
   }
-}
+}
```
neheb, line 3956: what?

aaalloc: my editor added a newline at the end of the line

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

aaalloc: ok mb  done

eli-schwartz:
> any idea on why pipeline fails and sometime it has no problem ?

The test failure appears to be for previously uncovered CPU architectures -- we very recently added qemu based CI on Alpine, and that's what is failing. Hopefully @xclaesse can look at this, as the author of the libffi wrap.

We don't need to block this PR on that pipeline failure.

releases.json
eli-schwartz: Please squash changes into one logical change per commit -- don't make followup commits that back out of changes made in a previous commit.

You can do this with e.g. `git commit --amend && git push --force` to correct a mistake in the previous commit. See also `git rebase -i` to re-order and combine commits together via the "fixup" method.

eli-schwartz (changes requested):

releases.json
aaalloc: done!

neheb (approved):

bgilbert:
We already have a `3.4.6-2`, so this didn't actually end up bumping the libffi release.  Fix in https://github.com/mesonbuild/wrapdb/pull/1891.

PR #1807, opened 2024-12-04
Add some linting jobs for CI

dcbaker:
This is currently a draft since it only affects python files ATM. I'd like to also lint JSON and meson files.


subprojects/packagefiles/openblas/gen_install_headers.py
```diff
@@ -101,3 +102,4 @@ def main():
 
 if __name__ == "__main__":
     main()
+
```
eli-schwartz, line 105: Hmm?

subprojects/packagefiles/m4/checks/m4_test_runner.py
```diff
@@ -1,16 +1,15 @@
...
 
 # We want to print line endings normally, but each line should be a b'' string
 def clean(x):
     lines = []
-    for l in x.splitlines():
-        if b'examples' in l:
-            lines.append(str(l.replace(b'\\',b'/')))
+    for line in x.splitlines():
+        if b'examples' in line:
+            lines.append(str(line.replace(b'\\',b'/')))
```
eli-schwartz, line 10: This specific ruff "issue" is extremely opinionated and also wrong.

subprojects/packagefiles/libuv/link_file_in_build_dir.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
-import os, sys
+import os
+import sys
```
eli-schwartz, line 4: Ugh, please can we not!

dcbaker: Can we follow standard python coding guidelines? yes

subprojects/packagefiles/m4/checks/m4_test_runner.py
dcbaker: I get that it's opinionated, and in this case I'd agree waht was there before was fine, I think the second loop is more readaable with a `line` vs `l` though, since it's  a long loop and things like `line[len('MY_REALLY_LONG_STRING):]` are generally more readable than `l[len('MY_REALLY_LONG_STRING):]`

subprojects/packagefiles/openblas/gen_install_headers.py
eli-schwartz: Normally github's view of this would have indicated via some kind of red circle when the previous version didn't have a trailing end of line character, I thought. Meanwhile, the actual issue here is this adds a blank line. files ending with `main()` are a problem and files ending with `main()\n` are correct.

In fact, if I checkout the file locally, it doesn't appear to be missing the trailing end of line character. But if I checkout your PR, the file ends with `main()\n\n`, which is... surely wrong?

subprojects/packagefiles/libuv/link_file_in_build_dir.py
eli-schwartz: Nothing standard about this change, other than that black encourages it.

tools/utils.py
```diff
@@ -121,7 +121,7 @@ def is_linux() -> bool:
     return platform.system().lower() == 'linux'
 
 def is_windows() -> bool:
-    return platform.system().lower() == 'windows' and not "MSYSTEM" in os.environ
+    return platform.system().lower() == 'windows' and "MSYSTEM" not in os.environ
```
neheb, line 124: I can't tell is this is a bugfix.

subprojects/packagefiles/libuv/link_file_in_build_dir.py
trim21: This actually comes from pep8.

https://peps.python.org/pep-0008/#imports

eli-schwartz: The authors and maintainers of the cpython PEPs take every possible opportunity they possibly can to tell people that PEP 8 is ***not*** a standard for anything other than "the style for internal cpython project source code" and that people should stop putting them in a position to be the arbiters of python style, funnily enough.

And this particular rule is inconsistent and ugly so I can kind of see the point. :)

neheb:
Please rebase

Issue #1803, opened 2024-12-01
ninja still try to compile fmt.lib when only `fmt_header_only_dep`  is used


trim21:
for a meson config like this:

```meson
project(
    'app',
    'cpp',
    meson_version: '>= 1.3.0',
)

executable(
    'app',
    'main.cc',
    dependencies: [
        subproject('fmt').get_variable('fmt_header_only_dep'),
    ],
)
```

and after `meson setup build`, build.ninja looks like this:

```ninja

# Build rules for targets

build subprojects/fmt-11.0.2/fmt.dll.p/src_format.cc.obj: cpp_COMPILER ../subprojects/fmt-11.0.2/src/format.cc
 DEPFILE = "subprojects\fmt-11.0.2\fmt.dll.p\src_format.cc.obj.d"
 DEPFILE_UNQUOTED = subprojects\fmt-11.0.2\fmt.dll.p\src_format.cc.obj.d
 ARGS = "-Isubprojects\fmt-11.0.2\fmt.dll.p" "-Isubprojects\fmt-11.0.2" "-I..\subprojects\fmt-11.0.2" "-I..\subprojects\fmt-11.0.2\include" "-fdiagnostics-color=always" "-D_GLIBCXX_ASSERTIONS=1" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-std=c++14" "-O0" "-g" "-DFMT_LIB_EXPORT"

build subprojects/fmt-11.0.2/fmt.dll.p/src_os.cc.obj: cpp_COMPILER ../subprojects/fmt-11.0.2/src/os.cc
 DEPFILE = "subprojects\fmt-11.0.2\fmt.dll.p\src_os.cc.obj.d"
 DEPFILE_UNQUOTED = subprojects\fmt-11.0.2\fmt.dll.p\src_os.cc.obj.d
 ARGS = "-Isubprojects\fmt-11.0.2\fmt.dll.p" "-Isubprojects\fmt-11.0.2" "-I..\subprojects\fmt-11.0.2" "-I..\subprojects\fmt-11.0.2\include" "-fdiagnostics-color=always" "-D_GLIBCXX_ASSERTIONS=1" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-std=c++14" "-O0" "-g" "-DFMT_LIB_EXPORT"

[...]

PR #1811, opened 2024-12-06
fmt: add `header-only` option to make it truly header-only

trim21:
close https://github.com/mesonbuild/wrapdb/issues/1803

fmt has a header only mode which allow to be used without need to compile cc files.


subprojects/packagefiles/fmt/meson.build
```diff
@@ -13,7 +13,9 @@ fmt_lib = library(
   'src/os.cc',
   cpp_args: fmt_private_cpp_args,
   include_directories: 'include',
-  install: true,
+  build_by_default: false,
+  # so we do not build libfmt when using fmt_header_only
+  install: not meson.is_subproject(),
```
jpakkane, line 18: This breaks things for people who use this wrap and do want to build and use fmt as a shared library.

The common solution in these cases is to have a project option to select whether the dep should be built as a lib or is used as a header only lib.

subprojects/packagefiles/fmt/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option(
+  'enable-header-only',
```
jpakkane, line 2: We intentionally do not use 'enable-` prefix in option names as it is an autoconfism. `header-only` says the same thing more concisely.

jpakkane:
Yes.

PR #1813, opened 2024-12-07
tools: convert strings to f strings

neheb:
Done with find tools -name "*.py" -exec flynt -tc -tj -a '{}' \;

### DIFF
--- a/tools/create_release.py
+++ b/tools/create_release.py
@@ -125,7 +125,7 @@ class CreateRelease:
         response.raise_for_status()
         for r in response.json():
             if r['tag_name'] == self.tag:
-                self.upload_url = r['upload_url'].replace(u'{?name,label}','')
+                self.upload_url = r['upload_url'].replace('{?name,label}','')
                 print('Found release:', self.upload_url)
                 return
 
@@ -135,7 +135,7 @@ class CreateRelease:
         }
         response = requests.post(api, headers=headers, json=content)
         response.raise_for_status()
-        self.upload_url = response.json()['upload_url'].replace(u'{?name,label}','')
+        self.upload_url = response.json()['upload_url'].replace('{?name,label}','')
         print('Created release:', self.upload_url)
 
     def upload(self, path: Path, mimetype: str):
--- a/tools/create_release.py
+++ b/tools/create_release.py
@@ -44,7 +44,7 @@ class CreateRelease:
             self.create_wrap_file()
 
     def read_wrap(self):
-        filename = Path('subprojects', self.name + '.wrap')
+        filename = Path('subprojects', f'{self.name}.wrap')
         self.wrap = configparser.ConfigParser(interpolation=None)
         self.wrap.read(filename)
         self.wrap_section = self.wrap[self.wrap.sections()[0]]
@@ -86,7 +86,7 @@ class CreateRelease:
         base_name = Path(self.tempdir, f'{self.tag}_patch')
         shutil.make_archive(base_name.as_posix(), 'zip', root_dir=self.tempdir, base_dir=directory)
 
-        patch_filename = base_name.with_name(base_name.name + '.zip')
+        patch_filename = base_name.with_name(f'{base_name.name}.zip')
         self.upload(patch_filename, 'application/zip')
 
         h = hashlib.sha256()
@@ -101,7 +101,7 @@ class CreateRelease:
     def create_wrap_file(self):
         self.wrap_section['wrapdb_version'] = self.version
 
-        filename = Path(self.tempdir, self.name + '.wrap')
+        filename = Path(self.tempdir, f'{self.name}.wrap')
 
         # configparser write() adds multiple trailing newlines, collapse them
         buf = io.StringIO()
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -173,7 +173,7 @@ class TestReleases(unittest.TestCase):
             if name in {'sqlite', 'libjpeg'}:
                 continue
             self.assertIn(name, self.releases)
-            self.assertIn(version, self.releases[name]['versions'], f"for {name}")
+            self.assertIn(version, self.releases[name]['versions'], f'for {name}')
 
         # Verify keys are sorted
         self.assertEqual(sorted(self.releases.keys()), list(self.releases.keys()))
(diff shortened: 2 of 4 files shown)


tools/create_release.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, repo: T.Optional[str], token: T.Optional[str], tag: str):
             self.create_wrap_file()
 
     def read_wrap(self):
-        filename = Path('subprojects', self.name + '.wrap')
+        filename = Path('subprojects', f"{self.name}.wrap")
```
eli-schwartz, line 47: flynt appears to be converting all single quotes to double quotes for inexplicable reasons...

tools/create_release.py
```diff
@@ -101,13 +101,13 @@ def create_patch_zip(self):
...
 
-        filename = Path(self.tempdir, self.name + '.wrap')
+        filename = Path(self.tempdir, f"{self.name}.wrap")
 
         # configparser write() adds multiple trailing newlines, collapse them
         buf = io.StringIO()
         self.wrap.write(buf)
         buf.seek(0)
-        newbuf = buf.read().rstrip('\n') + '\n'
+        newbuf = f"{buf.read().rstrip('\\n')}\n"
```
eli-schwartz, line 110: This has very bad readability. Not all string concatenation should be converted to f-strings.

tools/sanity_checks.py
```diff
@@ -330,7 +330,7 @@ def check_source_url(self, name: str, wrap_section: configparser.SectionProxy, v
         if name == 'sqlite3':
             segs = version.split('.')
             assert(len(segs) == 3)
-            version = segs[0] + segs[1] + '0' + segs[2]
+            version = f"{segs[0] + segs[1]}0{segs[2]}"
```
eli-schwartz, line 333: Again, terrible readability.

tools/sanity_checks.py
```diff
@@ -372,8 +372,8 @@ def check_new_release(self, name: str, builddir: str = '_build', deps=None, prog
             deps = [d for d in deps if d not in skip_deps]
 
         options = ['-Dpython.install_env=auto', f'-Dwraps={name}']
-        options.append('-Ddepnames={}'.format(','.join(deps or [])))
-        options.append('-Dprognames={}'.format(','.join(progs or [])))
+        options.append(f"-Ddepnames={','.join(deps or [])}")
+        options.append(f"-Dprognames={','.join(progs or [])}")
```
eli-schwartz, line 376: Terrible readability...

tools/sanity_checks.py
```diff
@@ -403,13 +403,13 @@ def install_packages(kind, cmd, packages):
                 for prefix in out.decode().split('\n'):
                     bindir = Path(prefix) / 'bin'
                     if bindir.exists():
-                        meson_env['PATH'] = str(bindir) + ':' + meson_env['PATH']
+                        meson_env['PATH'] = f"{str(bindir)}:{meson_env['PATH']}"
```
eli-schwartz, line 406: Terrible readability...

eli-schwartz (changes requested):

tools/create_release.py
neheb: ruff does too. I don't know either.

tools/sanity_checks.py
neheb: really?

neheb:
added extra commit

tools/sanity_checks.py
```diff
@@ -436,16 +433,16 @@ def install_packages(kind, cmd, packages):
...
-                elif any(f"ERROR: {x}" in error for x in {'Dependency', 'Program', 'Pkg-config binary', 'CMake binary'}):
-                    if 'not found' in error:
-                        print('cannot verify in wrapdb due to missing dependency')
-                        return
-                elif 'ERROR: Could not execute Vala compiler: valac' in error:
+                if 'ERROR: Could not execute Vala compiler: valac' in error:
                     print('cannot verify in wrapdb due to missing dependency')
                     return
-                elif 'ERROR: failed to unpack archive with error: ' in error:
+                if 'ERROR: failed to unpack archive with error: ' in error:
```
eli-schwartz, line 439: Which pylint rule says to use repeated elif? Looks funny imo...

tools/create_release.py
eli-schwartz: Well, can you switch them back? ;)

By the way, ruff has

```toml
[format]
quote-style = "single"
```

No clue about flynt...

tools/sanity_checks.py
neheb: https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html

eli-schwartz: Huh, weird. Looks like meson's .pylintrc disables it explicitly. :)

neheb: Sounds like that should be undone.

neheb:
Anything else missing?

tools/create_release.py
neheb: interestingly, flynt does not always convert to "". I ran this somewhere else and it kept the ''.

tools/sanity_checks.py
```diff
@@ -194,7 +194,7 @@ def check_meson_version(self, name: str, version: str, patch_path: str, builddir
             json_file = Path(builddir) / "meson-info/intro-projectinfo.json"
             # don't check if the build was skipped
             if json_file.exists():
-                with open(json_file) as project_info_file:
+                with open(json_file, encoding="utf-8") as project_info_file:
```
eli-schwartz, line 197: quoting style

tools/sanity_checks.py
neheb: meson's pylintrc.

tools/sanity_checks.py
neheb: Fixed.

tools/sanity_checks.py
eli-schwartz: What's so great about this rule? An `elif` ladder is semantically useful to visually indicate that exactly one condition (or the final "else") must be triggered. The fact that they return from the function really shouldn't matter at all.

neheb: OTOH, less indentation is also useful.

eli-schwartz: It's literally the same indentation though? Having an if with an elif clause versus having two if clauses one after the other, will only textually differ in eliding the `el` part of `elif`, a two-byte non-indentation change.

It's entirely unrelated to whether `else` clauses are replaced with a non-indented hunk. An if/elif must be indented since it has to check a conditional.

neheb: it's basically the python equivalent of https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html

IMO not having the elif makes it easier to reason about. well, not in this example specifically but generally speaking.

neheb: Pushed an alternative version. Maybe you'll like this.

PR #1815, opened 2024-12-09
openblas: improvements to 0.3.28 patch files

rgommers:
These changes are what I needed to use an `openblas` subproject with SciPy. A few bits of minor maintenance (de-duplicating configure checks, silencing build warnings, fixing some typos), and a few bugs that were caught by the SciPy test suite, which is quite a bit more extensive than OpenBLAS's own tests.

Follows up on gh-1734.

Cc @HaoZeke for visibility.


subprojects/packagefiles/openblas/meson.build
```diff
@@ -597,53 +630,62 @@ template = f'@_msproot@/openblas_config_template.h'
...
+               name: 'openblas',
+               filebase: 'meson-openblas',
+               description: 'OpenBLAS via meson build',
+               version: f'@pkg_ver@_meson',
+               install_dir: pkg_install_dir,
+              )
+endif
 
-if build_testing
+if build_testing and not as_subproject
```
eli-schwartz, line 673: This is already disabled by default. Why prevent people from enabling tests when building as a subproject? Maybe they enabled both options because they want to test the dependencies compiled correctly too? :)

subprojects/packagefiles/openblas/lapack-netlib/BLAS/SRC/meson.build
```diff
@@ -276,9 +276,18 @@ if prec == 's' or build_single or build_all_prec
...
 # Create the blas library
-netlib_blas = library(_netlib_blas_name,
-               sources: _blas_netlib_srcs,
-               link_with: _allblas,
-               version: lapack_version,
-               soversion: lapack_major_version,
-               install: true)
+if as_subproject
+  # Avoid creation versioned .so symlinks. We can't pass empty string or list
+  # to the version/soversion keyword to make this more concise.
```
eli-schwartz, line 281: It's possible to pass additional `kwargs: ...` as a dictionary to a function, which allows conditionally passing arguments.

subprojects/packagefiles/openblas/meson.build
rgommers: Ah, not sure why I had to disable this, `build_testing` should default to `false` indeed.

subprojects/packagefiles/openblas/lapack-netlib/BLAS/SRC/meson.build
rgommers: Ah thanks, that was exactly what I was looking for.

Should `kwargs` be in the docs? https://mesonbuild.com/Reference-manual_functions.html#library and other functions don't have this keyword.

eli-schwartz: It is described at https://mesonbuild.com/Syntax.html#function-calls

This is baked into the function call syntax, individual functions aren't aware that such a keyword exists and indeed the type checking for `library('abc', 'abc.c', kwargs: {'c_args': '-DFOO=1'})` will literally believe that the function was called as `library('abc', 'abc.c', c_args: '-DFOO=1')`

subprojects/packagefiles/openblas/meson.build
rgommers: Addressed now, resolving.

rgommers:
I was confused for a moment about why Alpine is failing here while it was passing on gh-1734 - in addition to the Windows builds, which were failing there already. The reason is that the build configs have changed: the Alpine job in gh-1734 was using x86-64, while now there are four Alpine configs for x64, riscv64, ppc64le, armv7. The SIMD kernels for those platforms are not wired up at all, this wrap only supports x86-64 and arm64 right now. I'll update to emit a more informative error at the start of the build.

Other than that, can I just ignore the Alpine failure (for the reason as given in https://github.com/mesonbuild/wrapdb/pull/1734#issuecomment-2442857863) @eli-schwartz?

eli-schwartz:
Yes -- please do ignore and defer those new errors to be addressed at some future date.

rgommers:
Thanks. This should be good to go then. It passes the relevant CI jobs here, as well as the ones on @HaoZeke's repo (https://github.com/HaoZeke/OpenBLAS/pull/30), and the SciPy test suite in my local testing.

HaoZeke (approved):
Thanks @rgommers

PR #1820, opened 2024-12-12
libusb: support building for android

ColinKinloch:
Compile linux targets when the target system is android.

I'm not sure whether I should add support for the `ANDROID_OS` flag which currently exists on the libusb `master` branch: https://github.com/libusb/libusb/commit/de38189e8014fa393f4d8c1d9d3fdf5e2a95899d


jpakkane:

PR #1823, opened 2024-12-20
bshoshany-thread-pool: update to v5.0.0

bshoshany:
Starting with this version, the library is also available as a C++20 module. I tried to find if meson supports modules, but could only find [this issue](https://github.com/mesonbuild/meson/issues/5024) which is still open. Currently, in addition to copying the header file `include/BS_thread_pool.hpp` (which can still be included as a normal header file if C++20 modules are not available) with `install_headers`, the wrap also copies the module file, `modules/BS.thread_pool.cppm`. Does that make sense, or would the user not have access to that folder? Any help would be appreciated.


bshoshany:
Not sure why the Apline checks are failing, but this seems to be due to an issue with the `update-binfmts` command in the check itself, not with my PR.

neheb:
Please rebase.

bshoshany:
Done!

PR #1849, opened 2025-01-04
c-ares: upgrade to 1.34.4

trim21:
they [Reorganize their source tree](https://github.com/c-ares/c-ares/pull/822) so there are many file changes.

I don't know libtool, how should I upgrade this?

https://github.com/mesonbuild/wrapdb/blob/9e1f00996c3f13a2fc76837e209608593ca7b690/subprojects/packagefiles/c-ares/meson.build#L3-L6

for 1.34.4 it's `"21:3:19"`

https://github.com/c-ares/c-ares/blob/a531524a3d085fcd9a5e25d5f6cbdb953082c2b9/CMakeLists.txt#L33


neheb:
some of those header checks look like they can be replaced by a loop.

PR #1850, opened 2025-01-04
sqlite3: Add more build features, with configurable extensions

bgilbert:
Based on, and alternative to, #1847.  Rather than having a single config option that requires the superproject to manually configure all defines, as #1847 does, provide specific options for some useful cases:

- Disable some defines that increase strictness (which this PR enables by default)
- Enable or disable specific built-in SQLite extensions, affecting code size.  I've  omitted the ICU extension for now because it adds an external dependency.
- Enable or disable all extensions by default.  Specific extensions can still be configured individually, e.g. `-Dsqlite3:all-extensions=disabled -Dsqlite3:session=enabled`.  This is convenient for superprojects that want to future-proof against updated defaults.

Shared library size with GCC on Fedora 41:
| | debug | | release | |
| - | - | - | - | - |
| Current | 2169160 | | 1494408 | |
| New, `-Dsqlite3:all-extensions=disabled` | 2192192 | +1.1% | 1512112 | +1.2% |
| New, default (FTS3-5 only) | 2939344 | +35.5% | 1991048 | +33.2% |
| New, `-Dsqlite3:all-extensions=enabled` | 3283896 | +51.4% | 2241872 | +50.0% |


jpakkane:
@mar-v-in does this seem like a good solution to you?

mar-v-in:
I think this is fine:
- The default configuration largely matches popular distribution defaults, so it has a high likelihood of being a suitable fallback
- The configuration options make sense and provide most of the features one could imagine people want to use. There are probably a few switches not covered, but if someone needed them they could easily add them later.

PR #1853, opened 2025-01-04
Two fixes to the luajit wrap

arch1t3cht:
Two fixes to the luajit wrap, explained in their commit messages.


subprojects/packagefiles/luajit/meson.build
```diff
@@ -121,6 +121,8 @@ DASM_FLAGS += ['-D', f'VER=@LJ_ARCH_VERSION@']
 
 if host_machine.system() == 'darwin'
   unwind_external = true
+elif host_machine.system() == 'windows'
```
kasper93, line 124: Shouldn't this be `cc.get_id() == 'msvc'` to mimic original `msvcbuild.bat`. Otherwise we can use it. Or in fact we could make error of running `./unwind_check.sh` not critical and assume that `unwind_external` is false in this case, with a warning.

arch1t3cht: Good point, just matching `msvcbuild.bat` is not enough here, but I now also saw that the [Makefile](https://github.com/LuaJIT/LuaJIT/blob/f73e649a954b599fc184726c376476e7a5c439ca/src/Makefile#L311-L315) also disables `LUAJIT_UNWIND_EXTERNAL` whenever the target is Windows (no matter the compiler).

kasper93: Seems to be good then.

arch1t3cht:
The CI failure is unrelated: It fails because RISC-V is not supported by luajit, the RISC-V CI lane was added in ff119ca6018457d58b5a1cbf67683a701d6308c6 after the last commit to the luajit wrap.

arch1t3cht:
Pushed a commit that should detect failures due to unsupported architectures (as well as seems to be possible) and marks alpine as known unsupported.

To my knowledge there's no good way to obtain the compiler's error message in meson, so we cannot distinguish errors due to unsupported architectures from errors due to otherwise incorrect configurations, so this does have a slight risk of false positives in marking failures as "intended". The alternative (apart from making an additional python script) would be manually checking `host_machine.cpu_family`, but this would need to be manually updated when luajit starts supporting more architectures.

Alpine contains both a RISC-V lane and an x86 lane (and others), so the CI check will likely still fail since now Alpine x86 will succeed but is "expected to fail". As far as I can see, the possible settings in `ci_config.json` are not granular enough to signal which of the architectures will succeed, but I'm guessing that an unexpected success is better than an unexpected failure here. If not, please let me know how this should be handled.

arch1t3cht:
@jpakkane Is there anything else I can or should do for this? As explained above, I don't think there's a way to stop the CI from failing on Alpine, since luajit supports x86 but not RISC-V. If there is, please let me know.

PR #1862, opened 2025-01-09
zycore: new wrap

mochaaP:
https://github.com/zyantific/zycore-c/releases/v1.5.1

https://github.com/zyantific/zycore-c/pull/77


mochaaP:
working on a fix.

mochaaP:
the x86 check was fixed (in my branch, not tagged yet), no idea why clang-cl fails to link.

PR #1864, opened 2025-01-10
abseil-cpp: Expose more missing headers

ford-prefect:
This is enough for webrtc-audio-processing-2 to compile.


ford-prefect:
Thanks for the quick merge on the last one -- I missed a few and now tested this a bit better for the package this is being used in.

neheb:
know anything about that ppc failure?

ford-prefect:
> know anything about that ppc failure?

Looking at this [line](https://github.com/abseil/abseil-cpp/blob/master/absl/base/internal/unscaledcycleclock.cc#L87), it seems abseil only supports PPC with glibc and not musl, at least for now.

ford-prefect:
If we want to, we could set `ABSL_USE_UNSCALED_CYCLECLOCK` to 0 to avoid building this on PPC, but I'm not sure what effects that would have, but [it can't be too bad](https://android.googlesource.com/platform/external/abseil-cpp/+/HEAD/absl/base/internal/unscaledcycleclock_config.h#31).

neheb:
Sure. I highly doubt anyone uses WrapDB on PPC anyway.

ford-prefect:
Phew, looks like that got it to green.

subprojects/packagefiles/abseil-cpp/meson.build
```diff
@@ -13,14 +13,26 @@ cpp = meson.get_compiler('cpp')
...
 
+arch_cpp_flags = []
 hw_cpp_flags = []
 if host_machine.cpu_family() == 'x86_64'
   hw_cpp_flags += ['-maes', '-msse4.1']
 elif host_machine.cpu_family() == 'aarch64' and cpp.sizeof('void*') == 8
   hw_cpp_flags += ['-march=armv8-a+crypto']
 elif host_machine.cpu_family() == 'arm' and cpp.sizeof('void*') == 4
   hw_cpp_flags += ['-mfpu=neon']
+elif host_machine.cpu_family() == 'ppc' or host_machine.cpu_family() == 'ppc64'
```
dcbaker, line 24: `elif host_machine.cpu_family().startswith('ppc')` or `elif host_machine.cpu_family() in ['ppc', 'ppc64']`
For future reference.

PR #1865, opened 2025-01-10
quickjs-ng: new wrap

trim21:
Not ready for merge.

---

Add new wrap for quickjs-ng

cc @KaruroChori for some help and review

### DIFF
--- a/subprojects/packagefiles/quickjs-ng/meson.build
+++ b/subprojects/packagefiles/quickjs-ng/meson.build
@@ -109,4 +109,4 @@ if not meson.is_subproject()
 endif
 
 
-test('ctest', executable('ctest', 'ctest.c', dependencies: [quickjs_ng_dep]))
+test('ctest', executable('ctest', 'ctest.c', dependencies: [quickjs_ng_dep],build_by_default: false))
--- a/subprojects/packagefiles/quickjs-ng/meson.build
+++ b/subprojects/packagefiles/quickjs-ng/meson.build
@@ -84,7 +84,8 @@ endif
 
 add_project_arguments(c_defines, language: 'c')
 
-quickjs_lib = library(
+# TODO: support shared libs
+quickjs_lib = static_library(
   'quickjs',
   lib_src,
   dependencies: deps,
--- a/subprojects/packagefiles/quickjs-ng/meson.build
+++ b/subprojects/packagefiles/quickjs-ng/meson.build
@@ -94,10 +94,15 @@ endif
 add_project_arguments(c_defines, language: 'c')
 
 # TODO: support shared libs
-quickjs_lib = static_library('quickjs', lib_src, dependencies: deps)
+quickjs_lib = static_library(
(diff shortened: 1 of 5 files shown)


subprojects/packagefiles/quickjs-ng/meson.build
```diff
@@ -0,0 +1,106 @@
...
+  c_defines += ['-DWIN32_LEAN_AND_MEAN', '-D_WIN32_WINNT=0x0602']
+endif
+
+add_project_arguments(c_defines, language: 'c')
+
+# TODO: support shared libs
+quickjs_lib = static_library('quickjs', lib_src, dependencies: deps)
+
+quickjs_ng_dep = declare_dependency(
+  include_directories: include_directories('.'),
```
KaruroChori, line 100: I don't think this is a good idea. 
It exposes all headers, creating name collisions downstream.  
I encountered the same problem when adapting other projects like `tcc` or `mongoose`.  
The trick I used to avoid this, is like what I have done in here: https://github.com/lazy-eggplant/vs.http/tree/master/subprojects/packagefiles/mongoose
Basically I create a `meson-include` folder, and add symbolic links to the required headers in there.
However, this had a bug up until recently https://github.com/mesonbuild/meson/pull/14073 so it would not work with older versions :(.

KaruroChori: The only other alternatives I can see are
- to fork the project & better organize files so that we can have sane search paths
- keep the `.` and accepting things will break downstream when building, in scenarios which are quite obscure and hard to predict.
For context, a file `config.h` in tinycc was creating problems in one of my projects, but only on macos-14 with clang-19. I wasted way to many hours to figure that out.

KaruroChori: Yes, it is possible I guess. It makes the process of maintaining the patch more annoying than what it should be I guess. But it would work.

KaruroChori:
We might also generate the pkg for the library. Many which are distributed in wrapdb do.

mochaaP (changes requested):

subprojects/quickjs-ng.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = quickjs-0.8.0
+source_url = https://github.com/quickjs-ng/quickjs/archive/refs/tags/v0.8.0.tar.gz
+source_filename = quickjs-ng-v0.8.0.tar.gz
+source_hash = 7e60e1e0dcd07d25664331308a2f4aee2a88d60d85896e828d25df7c3d40204e
+patch_directory = quickjs-ng
+
+[provide]
+quickjs-ng = quickjs_ng_dep
```
mochaaP, line 9: use `meson.override_dependency` + `[provide] dependency_names`

trim21:
ref: https://github.com/quickjs-ng/quickjs/pull/902

PR #1870, opened 2025-01-13
stc: new wrap

mochaaP:
<https://github.com/stclib/STC/releases/tag/v5.0>

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -131,6 +131,7 @@ jobs:
           install: >-
             bison
             dos2unix
+            flex
             git
           pacboy: >-
             cc:p
--- a/subprojects/stc.wrap
+++ b/subprojects/stc.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = STC-4a76ff9bc961750b5d1932496f6b861e794f8178
+source_url = https://github.com/stclib/STC/archive/4a76ff9bc961750b5d1932496f6b861e794f8178/v5.0.tar.gz
+source_filename = v5.0.tar.gz
+source_hash = 9c8cd0a2e99c82d584cc106376b027183331ad07a5f602422ece95dcea35c2dd
+
+[provide]
+dependency_names = stc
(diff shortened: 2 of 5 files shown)


subprojects/stc.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = STC-5.0
+source_url = https://github.com/stclib/STC/archive/refs/tags/v5.0.tar.gz
+source_filename = v5.0.tar.gz
+source_hash = d61353852b9d7ef69b56fa32edcbc7934f2153385f3778536234201ceebcc950
+
+[provide]
+dependency_names = stc
```
eli-schwartz, line 8: Interestingly, it appears they provide a `declare_dependency()` but not an installable `import('pkgconfig').generate(stc_lib)`.

mochaaP: Thanks for the catch, I'll add it.

mochaaP:
@eli-schwartz `flex` is a base package in MSYS2: <https://packages.msys2.org/base/flex>
how can I install this without a platform prefix?

eli-schwartz:
https://www.msys2.org/docs/package-naming/#avoiding-writing-long-package-names

tools/sanity_checks.py seems to use pacboy as recommended there with the `:p` suffix, so I *think* you just say "flex"?

eli-schwartz:
@neheb is the one that set up the msys2 CI and probably knows best here.

neheb:
@mochaaP add flex here https://github.com/mesonbuild/wrapdb/blob/master/.github/workflows/sanity_checks.yml#L126

in build_all as well.

.github/workflows/build_all.yml
```diff
@@ -129,6 +129,7 @@ jobs:
         with:
           msystem: ${{matrix.platform}}
           install: >-
+            flex
```
neheb, line 132: alphabetize

mochaaP: done

mochaaP:
done

PR #1871, opened 2025-01-13
glib-networking: update to 2.80.1

bgilbert:
Use system OpenSSL on Alpine to avoid build failures on obscure arches.

Supersedes #1861.


bgilbert:
CI failures:

- MSVC x86 fails because the OpenSSL wrap doesn't have a pregenerated config for x86:

```
openssl| Message: Generating OpenSSL configs...
subprojects\openssl-3.0.8\meson.build:18:4: ERROR: Problem encountered: Generator only works on Linux, other platforms are not supported
```

- macOS looks like a substantive failure:

```
not ok /tls/openssl/connection/basic - GLib-FATAL-WARNING: poll(2) failed due to: Bad file descriptor.
```

- MSYS2 fails with:

```
not ok /tls/openssl/connection/client-auth-rehandshake - GLib-Net:ERROR:../subprojects/glib-networking-2.80.1/tls/tests/connection.c:272:on_rehandshake_finish: assertion failed (error == NULL): Peer does not support safe renegotiation (g-tls-error-quark, 1)
```

Compare https://github.com/mesonbuild/wrapdb/pull/1655#issuecomment-2332694229.

PR #1872, opened 2025-01-13
Add wrap files for mpark-patterns-0.3.0

adembudak:


subprojects/packagefiles/mpark-patterns/meson.build
```diff
@@ -0,0 +1,26 @@
...
+    'MPark.Patterns',
+    'cpp',
+    default_options: ['cpp_std=c++17'],
+    license: 'BSL-1.0',
+    version: '0.3.0',
+    meson_version: '>=1.6.1',
+)
+
+if meson.is_subproject() and meson.get_compiler('cpp').get_id() == 'msvc'
+    add_global_arguments('/Zc:__cplusplus', language: 'cpp')
```
jpakkane, line 11: You can't set global arguments when built as a subproject. You should use `add_project_arguments` instead.

subprojects/packagefiles/mpark-patterns/meson.options
```diff
@@ -0,0 +1 @@
+option ('tests', type: 'boolean', value: false, description: 'Build tests')
```
neheb, line 1: why not feature option?

subprojects/packagefiles/mpark-patterns/meson.build
dcbaker: Meson will automatically inject `/Zc:__cplusplus` if the version of MSVC supports it, this shouldn't be necessary anyway.

subprojects/packagefiles/mpark-patterns/meson.options
adembudak: Thanks, changed it to a feature option, but I'm not sure. Tests are not a feature of a library, aren't they?

jpakkane:
Please do not have merge commits in your MRs, only keep it up to date by rebasing. Those merge commits need to be removed before the MR can be accepted.

jpakkane:
There is still one merge commit.

PR #1874, opened 2025-01-14
luajit: port to OpenBSD

guijan:
It didn't take any work at all.
Tested by compiling the shared library and installing it to the system, neovim with lots of Lua plugins worked fine.


guijan:
Force-push because I forgot to update _releases.json_. Alpine failures don't seem to be caused by this PR according to https://github.com/mesonbuild/wrapdb/pull/1853#issuecomment-2571714892.

dcbaker (approved):

PR #1784, opened 2024-11-14
CI: test libatomic

neheb:

PR #1878, opened 2025-01-16
CI: add Ubuntu aarch64

bgilbert:
Until now, we've only tested aarch64 on macOS.

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -13,10 +13,13 @@ on:
 jobs:
   Ubuntu:
     if: github.event_name != 'schedule' || github.repository == 'mesonbuild/wrapdb'
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


neheb:
Alpine too.

bgilbert:
As of #1784, we run Alpine on armv7 and x86 but not aarch64 and x86_64.  For consistency, I think we should run it on both of aarch64 and x86_64, or neither.  WDYT?

bgilbert:
In any case, Alpine is waiting on https://github.com/jirutka/setup-alpine/pull/22, so let's get this part in for now.

neheb:
Oh? 32-bit arm is native now? That's real nice.

PR #1881, opened 2025-01-18
pango: update to 1.56.0

bgilbert:
Build errors:

1. MSYS builds throw a bunch of redefinition errors in the cairo dwrite code.  Not sure why this doesn't happen when building cairo directly.  Worked around with `cairo:dwrite=disabled`.

2. MSYS builds fail on:

   ```
   In file included from ../subprojects/sysprof/src/libsysprof-capture/sysprof-capture-reader.c:69:
   In file included from ../subprojects/sysprof/src/libsysprof-capture/sysprof-capture-reader.h:61:
   ../subprojects/sysprof/src/libsysprof-capture/sysprof-capture-types.h:63:11: fatal error: 'endian.h' file not found
   ```

   pango provides a valid `sysprof.wrap` that it doesn't use by default, but this causes glib to autodetect sysprof and fall back to the wrap.  In turn, libsysprof tries to use `endian.h`, which doesn't exist on Windows.  We don't see this in glib builds because glib's own `sysprof.wrap` doesn't list any provides and so is ignored.

    Worked around with `glib:sysprof=disabled`.

3. MSVC clang-cl fails on this:

   ```
   FAILED: subprojects/cairo-1.18.2/src/cairo-2.dll subprojects/cairo-1.18.2/src/cairo-2.pdb 
   "lld-link" @subprojects/cairo-1.18.2/src/cairo-2.dll.rsp
   lld-link: error: undefined symbol: __udivti3
   >>> referenced by D:\a\wrapdb\wrapdb\subprojects\cairo-1.18.2\src\cairo-wideint.c:390
   >>>               subprojects/cairo-1.18.2/src/cairo-2.dll.p/cairo-wideint.c.obj:(_cairo_uint128_divrem)
   lld-link: error: undefined symbol: __umodti3
   >>> referenced by D:\a\wrapdb\wrapdb\subprojects\cairo-1.18.2\src\cairo-wideint.c:391
   >>>               subprojects/cairo-1.18.2/src/cairo-2.dll.p/cairo-wideint.c.obj:(_cairo_uint128_divrem)
   ```

   Looks like this bug: https://github.com/llvm/llvm-project/issues/25679

   Not sure why cairo builds don't see this, and there's no evident workaround.


neheb:
I've seen that udiv issue too. I don't remember finding a good way to fix it.

bgilbert:
Should we land this then?

PR #1882, opened 2025-01-18
ci: add libunistring to libpsl

neheb:
Needed as ICU wrap seems to be broken.


neheb:
clang-cl's failure gets fixed with next version.

macOS failure is because pkgconfig there installs to a different directory.

jpakkane:
```
../subprojects/libpsl-0.21.5/tests/test-is-cookie-domain-acceptable.c(134,16): error: call to undeclared library function 'alloca' with type 'void *(unsigned long long)'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  134 |                         char *cmd = alloca(cmdsize);
      |                                     ^
../subprojects/libpsl-0.21.5/tests/test-is-cookie-domain-acceptable.c(134,16): note: include the header <stdlib.h> or explicitly provide a declaration for 'alloca'
```

Seems relevant.

neheb:
That’s fixed upstream, as mentioned

PR #1898, opened 2025-01-31
New Wrap: Kafel

manipuladordedados:


jpakkane:
Please keep the MR squashed to a single commit.

subprojects/packagefiles/kafel/meson.build
```diff
@@ -0,0 +1,73 @@
+project('kafel', 'c',
+  version: '2023.10.10',
+  default_options: [
+    'c_std=gnu11',
+  ])
+
+flex = find_program('flex', required: true)
+bison = find_program('bison', required: true)
```
jpakkane, line 8: You don't need `required: true`, it is the default.

PR #1903, opened 2025-02-02
libusb: support building for emscripten

darkvision77:
Added emscripten backend compilation.

`cpp_std=c++20` is required to build with emscripten.

How to compile:
```sh
# https://github.com/mesonbuild/meson/blob/master/cross/wasm.txt
meson setup --cross-file wasm.txt build -Ddefault_library=static
meson compile -C build
```


subprojects/packagefiles/libusb/meson.build
```diff
@@ -1,11 +1,14 @@
 project(
   'libusb',
-  'c',
+  'c', 'cpp',
```
neheb, line 3: It's probably better to use https://mesonbuild.com/Reference-manual_functions.html#add_languages

darkvision77: Yeah, agree. Fixed.

neheb:
I'm actually curious whether adding CI for emscripten is a good idea. I don't know how valuable it is.

From my experience, it blows up when needing to link to pthread.

darkvision77:
> I'm actually curious whether adding CI for emscripten is a good idea.

IMHO, CI for emscripten is a must have thing.
After all, if the original project supports builds with emscripten, then wrap should probably support them too.

darkvision77:
@neheb, what's the problem?

PR #1905, opened 2025-02-02
slirp: update to 4.9.0

bgilbert:


bgilbert:
[Upstream MR](https://gitlab.freedesktop.org/slirp/libslirp/-/merge_requests/150) adding `meson.override_dependency()` so we can stop caring about the name of upstream's dependency variable.

bgilbert:
MR landed.  We can change the wrap file for the next release.

PR #1907, opened 2025-02-03
spdlog wrap: Option to disable installation of headers.

sebastianstricker:
Added a meson option to specify whether or not to install headers.

Installing headers and pkg-config files is not always wanted, especially as spdlog is naturally used as a subproject. After compilation, one does not need the header files anymore, except when compiling spdlog as a shared library.

This patch adds an option to disable installation, but does not change the current default behaviour. The current CMAKE build file of spdlog currently contains this option as well, so it only makes sense to replicate it.

### DIFF
--- a/subprojects/packagefiles/spdlog/meson.build
+++ b/subprojects/packagefiles/spdlog/meson.build
@@ -51,15 +51,17 @@ else
 endif
 
 # install header and pkg-config file
-install_subdir('include/spdlog', install_dir: get_option('includedir'))
-pkg = import('pkgconfig')
-pkg.generate(
-  name: 'libspdlog',
-  filebase: 'spdlog',
-  description: 'Fast C++ logging library.',
-  url: 'https://github.com/gabime/spdlog',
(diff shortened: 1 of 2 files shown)


stephanlachnit:
If you are using spdlog as a static library, this should be solved by using

```sh
meson install --skip-subprojects spdlog
```

stephanlachnit:
Best case scenario would be a `skip_install_if_subproject` built-in option in Meson, which defaults to false, but can be set to true in `subproject()` and `dependency()`.

sebastianstricker:
> If you are using spdlog as a static library, this should be solved by using
> 
> ```shell
> meson install --skip-subprojects spdlog
> ```
> 
> Unfortunately, it is not possible to set this in Meson directly.
> 

Thanks for the reply. I was unaware so far of --skip-subprojects option. As you said, I expected to find that functionality in either subproject() or dependency(), as I'd ideally like to predefine the behaviour in the meson.build file, not the install command.

I am not sure if I this covers all use cases, however, I'll close this for now as it resolves my issue at least!

PR #1912, opened 2025-02-07
freeglut: fix broken wayland build

kusma:
The most recent commit, 8472c965 ("freeglut: add missing dependencies") broke when this was built for wayland, in two ways:

1. It dropped the xkbcommon dep, which is required on Wayland as well
2. It removed the XParseGeometry implementation, which is also required on Wayland.

The former was fixed simply by adding it back. The latter was fixed by simply always including the custom implementation. Previously, we'd *try* to find it, but it would always fail when building for Wayland.


jpakkane:
> src_fg_init.c.obj : error LNK2019: unresolved external symbol XParseGeometry referenced in function glutInit

This seems somewhat relevant.

kusma:
Yeah, I guess we need this on Windows too. I suppose Windows wasn't in CI during last update.

I'll fix it up.

kusma:
OK, that wasn't *quite* it, but I think I got it now.

PR #1916, opened 2025-02-11
zlib-ng: update to 2.2.4

bgilbert:


bgilbert:
Same clang armv7 SIGBUS as in 2.2.3: https://github.com/zlib-ng/zlib-ng/issues/1844

neheb:
interestingly enough I actually have an ARMv7 device running debian.

bgilbert:
Did you compile with GCC or clang?

neheb:
clang. GCC is fine. Posted GDB output in the other repository.

neheb:
I found the problem. Simplest fix is to disable NEON with clang + arm32 if possible.

PR #1917, opened 2025-02-11
New Wrap: Kafel 

manipuladordedados:


subprojects/packagefiles/kafel/meson.build
```diff
@@ -0,0 +1,72 @@
...
+  default_options: [
+    'c_std=gnu11',
+  ])
+
+flex = find_program('flex')
+bison = find_program('bison')
+pkgconf = import('pkgconfig')
+
+run_command(flex, '--outfile=lexer.c', 'src/lexer.l', check: true)
+run_command(bison, '-o=parser.c', 'src/parser.y', check: true)
```
dcbaker, line 12: You don't want to use `run_command` for this, this is what `custom_target` is for:
```meson
lexer = custom_target(
   command : [flex, '--outfile', '@OUTPUT0@', '--header-file='@OUTPUT1@', '@INPUT@'],
   input : 'src/lexer.l',
   output : ['@BASENAME@.c', '@BASENAME@.h'],
)
parser = custom_target(
  command : [bison, '-o', '@OUTPUTo@', '--defines=@OUTPUT1@', '@INPUT@'],
  input : 'src/parser.y',
  output : ['@BASENAME@.c', '@BASENAME@.h'],
)
```

subprojects/packagefiles/kafel/meson.build
```diff
@@ -0,0 +1,72 @@
...
+  'src/syscall.c',
+  'src/syscalls/amd64_syscalls.c',
+  'src/syscalls/i386_syscalls.c',
+  'src/syscalls/aarch64_syscalls.c',
+  'src/syscalls/mipso32_syscalls.c',
+  'src/syscalls/mips64_syscalls.c',
+  'src/syscalls/riscv64_syscalls.c',
+  'src/syscalls/arm_syscalls.c',
+  'lexer.c',
+  'parser.c'
```
dcbaker, line 32: Then you'll want to use the `parser` and `lexer` variables instead of these.

subprojects/packagefiles/kafel/meson.build
```diff
@@ -0,0 +1,72 @@
...
+)
+
+kafel_inc = include_directories('include', 'src')
+
+kafel_lib = library('kafel',
+  sources: kafel_sources,
+  include_directories: kafel_inc,
+  install: true
```
dcbaker, line 40:
```meson
  gnu_symbol_visibility : 'hidden',
  soversion : 1,
```
is also needed to match the makefile.

You'll need to double check that soversion though, and make sure it matches the makefile.

subprojects/packagefiles/kafel/meson.build
manipuladordedados: So, I had already tried it that way, but it didn't work, so I resorted to using run_command. Many people who use flex and bison seem to face the same issue and end up resorting to this method.

manipuladordedados: With this method it simply does not generate the .c files.

dcbaker: It won’t generate the c file if you don’t add the parser variable to the target, because custom targets are not built by default.

manipuladordedados: I am adding it, and it still doesn't work.

dcbaker: Here's an example of how to do it in real working code: https://github.com/dcbaker/meson-plus-plus/blob/main/src/frontend/meson.build

manipuladordedados: So, I had already tried using `custom_target` before, but without success.
It fails to generate the files at the exact moment they are needed, resulting in file not found errors. I attempted to fix this but was unable to resolve the issue.
However, after researching, I found an approach using `run_command`, which is used in similar contexts to mine.
Some examples:

- https://github.com/postgrespro/jsquery/blob/3073aebabdac58b082e95a63299216ca363feec6/meson.build#L22

- https://github.com/srbarton43/programmers_calculator/blob/bbed747521328c6b31168118c178d54fb19b3b9b/src/meson.build#L12

dcbaker: Neither of those projects are doing the right thing. There are two problems with the run_target approach you're taking,
1) it writes files into the source directory. Meson generally promises not to do this and it leads to issues, many cmake projects specifically go out of their way to turn off in-tree builds
2) the flex and bison files get written at configure time instead of build time, which means that their inputs are not properly tracked (so among other things ninja clean cant erase them), and to regenerate them a user needs to *manually* re-run meson, since Meson doesn't know those command write files

The critical thing here is that the output of the assigned variable from the custom_target needs to be in the sources list of the target that requires it:
```meson
x = custom_target(...)
library(
  'foo',
  'foo.c', 'lib..c', x,
)
```
That is how Meson knows that `libfoo` depends on the output of `x`, and properly orders it so that `x` is generated before any files in `libfoo` start compiling. the Meson++ example I pointed you to demonstrates how to write this as a `custom_target`, which is the right way to do this.

`run_target` isn't for codegen, it's for things like running linters or code formatters.

manipuladordedados: Yes, I tried this method as well, but the problem persists. The files are still being generated in another folder, causing the error.
```
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /usr/bin/ninja -C /home/walt/Desktop/Trabalho/kafel/emilua/build
ninja: Entering directory `/home/walt/Desktop/Trabalho/kafel/emilua/build'
[31/365] Compiling C object subprojects/kafel-20231004/libkafel.so.p/meson-generated_.._lexer.c.o
FAILED: subprojects/kafel-20231004/libkafel.so.p/meson-generated_.._lexer.c.o 
cc -Isubprojects/kafel-20231004/libkafel.so.p -Isubprojects/kafel-20231004 -I../subprojects/kafel-20231004 -I../subprojects/kafel-20231004/include -I../subprojects/kafel-20231004/src -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=gnu11 -O0 -g -fPIC -MD -MQ subprojects/kafel-20231004/libkafel.so.p/meson-generated_.._lexer.c.o -MF subprojects/kafel-20231004/libkafel.so.p/meson-generated_.._lexer.c.o.d -o subprojects/kafel-20231004/libkafel.so.p/meson-generated_.._lexer.c.o -c subprojects/kafel-20231004/lexer.c
cc1: fatal error: subprojects/kafel-20231004/lexer.c: No such file or directory
compilation terminated.
[34/365] Compiling C object subprojects/kafel-20231004/libkafel.so.p/src_kafel.c.o
FAILED: subprojects/kafel-20231004/libkafel.so.p/src_kafel.c.o 
cc -Isubprojects/kafel-20231004/libkafel.so.p -Isubprojects/kafel-20231004 -I../subprojects/kafel-20231004 -I../subprojects/kafel-20231004/include -I../subprojects/kafel-20231004/src -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=gnu11 -O0 -g -fPIC -MD -MQ subprojects/kafel-20231004/libkafel.so.p/src_kafel.c.o -MF subprojects/kafel-20231004/libkafel.so.p/src_kafel.c.o.d -o subprojects/kafel-20231004/libkafel.so.p/src_kafel.c.o -c ../subprojects/kafel-20231004/src/kafel.c
../subprojects/kafel-20231004/src/kafel.c:28:10: fatal error: lexer.h: No such file or directory
   28 | #include "lexer.h"
      |          ^~~~~~~~~
compilation terminated.
```

manipuladordedados: Any tips?

dcbaker: Can you show me what your updated meson.build looks like? even just throwing it on pastebin or a github gist would be fine.

manipuladordedados: https://pastebin.com/SRhvc3nw

dcbaker: The meson.build looks correct, and builds fine for me with the kafel git master when run as a root project:
```
git clone --depth 1 https://github.com/google/kafel.git 
<download pastebin to meson.build>
meson setup builddir
ninja -C builddir
```

And manually inspecting the build.ninja shows exactly what I'd expect, every .c -> .o step in libkafel depends on lexer.h and parser.h.

But it fails as a subproject...

And I'm totally stumped *why*. Everything says that it's correct, but then the actual files are dumped into the root of the build directory, not into the project directory.

I'm running out of time for today. But I'll look at this more later (might be tomorrow my time)

eli-schwartz: parser.y:

```
%output "parser.c"
%defines "parser.h"
```

This overrides all command line arguments passed to the `bison` command. They will be silently ignored, no diagnostics shall be offered.

eli-schwartz: From the upstream Makefile:

```make
lexer.h lexer.c: lexer.l
	flex $<

parser.h parser.c: parser.y
	bison $<
```

So this is the classic badly written homebrew Makefile nonsense. They'd rather encode arguments into the source file and just... not do build commands in their Makefile.

eli-schwartz: This could be handled currently, by writing a script that is used to run bison/flex after `cd`ing to the output directory. Or by patching the upstream sources. There is no `workdir` argument for custom_target, though.

manipuladordedados: > The meson.build looks correct, and builds fine for me with the kafel git master when run as a root project:
> 
> ```
> git clone --depth 1 https://github.com/google/kafel.git 
> <download pastebin to meson.build>
> meson setup builddir
> ninja -C builddir
> ```
> 
> And manually inspecting the build.ninja shows exactly what I'd expect, every .c -> .o step in libkafel depends on lexer.h and parser.h.
> 
> But it fails as a subproject...
> 
> And I'm totally stumped _why_. Everything says that it's correct, but then the actual files are dumped into the root of the build directory, not into the project directory.
> 
> I'm running out of time for today. But I'll look at this more later (might be tomorrow my time)

Thanks, exactly that.

manipuladordedados: > parser.y:
> 
> ```
> %output "parser.c"
> %defines "parser.h"
> ```
> 
> This overrides all command line arguments passed to the `bison` command. They will be silently ignored, no diagnostics shall be offered.
> 
> Same for lexer.l:
> 
> ```
> %option outfile="lexer.c" header-file="lexer.h"
> ```

Damn, what now?
Do I have to edit them manually then? Is this allowed on Meson?

manipuladordedados: > This could be handled currently, by writing a script that is used to run bison/flex after `cd`ing to the output directory. Or by patching the upstream sources. There is no `workdir` argument for custom_target, though.

worked with:

``

eli-schwartz: Using `&&` in a custom_target command is officially a bug, it is not supposed to work but a part of the backend code that handles static libraries accidentally overrides it to make it work.

On the other hand:
> Damn, what now?
> Do I have to edit them manually then? Is this allowed on Meson?

subproject wraps support:

- diff_files: https://mesonbuild.com/Wrap-dependency-system-manual.html#diff-files
- copying the file over with a custom_target in the form of e.g. a sed or python script that deletes content dynamically
- submit a PR to upstream that fixes this in the long term

manipuladordedados: I used sed and fixed the issues, thanks!

eli-schwartz (changes requested):
I would suggest squashing these changes into one logical change per commit.

Also, please leave a comment in the meson.build file explaining why the sed is necessary. Considering how hard it was to debug, ensuring future readers of the wrap file understand what is going on seems like a good idea.

manipuladordedados:
> I would suggest squashing these changes into one logical change per commit.
> 
> Also, please leave a comment in the meson.build file explaining why the sed is necessary. Considering how hard it was to debug, ensuring future readers of the wrap file understand what is going on seems like a good idea.

Done.

subprojects/packagefiles/kafel/meson.build
manipuladordedados: Thank you, I fixed it!

manipuladordedados:
Could you help me with this CI error? @dcbaker @eli-schwartz

releases.json
```diff
@@ -1540,6 +1540,14 @@
       "1.8.4-1"
     ]
   },
+  "kafel": {
+    "dependency_names": [
+      "kafel"
+    ],
+    "versions": [
+      "2023.10.10"
```
mochaaP, line 1548: 2023.10.10-1

manipuladordedados: Thanks!

mochaaP (approved):

manipuladordedados:
The new error in the CI is:
`subprojects/kafel-20231004/meson.build:7:7: ERROR: Program 'flex' not found or not executable`.
I guess these containers don't have it installed?

mochaaP:
you need to declare it in ci_config.json.
example: image

manipuladordedados:
Hello!
I just realized the CI hasn’t kicked off for this PR yet. Any idea why?
Also, since this package deals with seccomp (which is Linux-only), maybe we could skip the CI for Windows and macOS?
Thanks!

mochaaP:
You can also configure this in ci_config.json

releases.json
```diff
@@ -1540,6 +1540,14 @@
       "1.8.4-1"
     ]
   },
+  "kafel": {
+    "dependency_names": [
+      "kafel"
+    ],
+    "versions": [
+      "2023.10.10-1"
```
dcbaker, line 1548: This doesn't match the archive name, which is 20231004

manipuladordedados: Done

releases.json
```diff
@@ -1540,6 +1540,14 @@
       "1.8.4-1"
     ]
   },
+  "kafel": {
+    "dependency_names": [
+      "kafel"
+    ],
```
dcbaker, line 1546: you need a `program_names` field here, I think.

dcbaker:
On the alpine systems it looks like you need to install the `bsd-compat-headers` package.

It may be worth adding a check for `sys/queue.h` in the meson.build like:
```meson
if not cc.has_header('sys/queue.h')
  error('sys/queue.h is required and cannot be found')
endif
```

releases.json
dcbaker: Looks like the CI is happy with what you have, so that's fine for me.

dcbaker:
I don't expect this to work for MacOS and Windows, so we should probably mark those as unsupported (it is a library for Linux syscaps after all).

It also looks like the version needs to not have the `.` in the name to make the sanity checker happy.

We're getting close!

manipuladordedados:
> I don't expect this to work for MacOS and Windows, so we should probably mark those as unsupported (it is a library for Linux syscaps after all).
> 
> It also looks like the version needs to not have the `.` in the name to make the sanity checker happy.
> 
> We're getting close!

Nicee, about making MacOS and Windows as unsupported, Its the same as the ci_config file?
```    "_comment": "Linux only",
    "build_on": {
      "darwin": false,
      "msys2": false,
      "windows": false

**Comment by dcbaker** [comment]
Looks right

**Comment by manipuladordedados** [comment]
> Looks right

I see, but is in another file? Its already like that in ci_config.
About the version, its already with "." no?
Thanks for you time!

**Comment by dcbaker** [comment]
I'm reaching the limit of my knowledge on the way the wrapdb ci works. @neheb?

**Comment by eli-schwartz** [comment]
The url issue is still not fixed so that would fail either way.

Regarding failures on configuring for e.g. Windows, there have been past issues where people used `build_on` to suppress platforms where the software SHOULD work but the wrap was incorrectly written, and rather than marking it as `broken` and merging it with failing CI under the principle:

> The perfect is the enemy of the good

they chose to mark it as disabled in `build_on`, thus preventing anyone from being aware that there were platforms which were still a TODO.

Naturally, since everyone loves and prefers technical limitations as a workaround for social problems (hahahaha but no really what can you do when people don't follow protocol) the CI was updated to treat `build_on` as an error if you use it without also making sure the wrapped `meson.build` errors out using an informative message (in particular the CI checks for `error()` that includes the term "unsupported" or "not supported" or "does not support").

We apologize for the inconvenience. 🫡

**Comment by manipuladordedados** [comment]
Oh... But what should I do now? How can I handle this CI problem?

**Comment by bgilbert** [comment]
You need to **both** set `build_on` in `ci_config.json` and make sure `meson.build` `error()`s out with an appropriate message on the unsupported platforms.

**Comment by manipuladordedados** [comment]
> You need to **both** set `build_on` in `ci_config.json` and make sure `meson.build` `error()`s out with an appropriate message on the unsupported platforms.

build_on was already present in ci_config.json, I improved the error message in the error() function in meson.build.
But it seems like the problem is something else, right?

**Comment by eli-schwartz** [comment]
> > You need to **both** set `build_on` in `ci_config.json` and make sure `meson.build` `error()`s out with an appropriate message on the unsupported platforms.
> 
> build_on was already present in ci_config.json, I improved the error message in the error() function in meson.build. But it seems like the problem is something else, right?

As I said above:


> also making sure the wrapped `meson.build` errors out using an informative message (in particular the CI checks for `error()` that includes the term "unsupported" or "not supported" or "does not support").

Note the specific terms to use. The current implementation is at https://github.com/mesonbuild/wrapdb/blob/1815cbe397fdc8cb0a245236899dfd499903eb7d/tools/sanity_checks.py#L433-L446

It is certainly possible to tweak that check based on discussion, but unless that code is changed, you'll need to specifically use one of the allowed terms.

**Comment by manipuladordedados** [comment]
> > > You need to **both** set `build_on` in `ci_config.json` and make sure `meson.build` `error()`s out with an appropriate message on the unsupported platforms.
> > 
> > 
> > build_on was already present in ci_config.json, I improved the error message in the error() function in meson.build. But it seems like the problem is something else, right?
> 
> As I said above:
> 
> > also making sure the wrapped `meson.build` errors out using an informative message (in particular the CI checks for `error()` that includes the term "unsupported" or "not supported" or "does not support").
> 
> Note the specific terms to use. The current implementation is at
> 
> https://github.com/mesonbuild/wrapdb/blob/1815cbe397fdc8cb0a245236899dfd499903eb7d/tools/sanity_checks.py#L433-L446
> 
> It is certainly possible to tweak that check based on discussion, but unless that code is changed, you'll need to specifically use one of the allowed terms.

Thank you for your patience and your time =)
I believe I understood and have already made the changes. Sorry for the mistakes, I’m a bit rusty.

**Comment by manipuladordedados** [comment]
@bgilbert Hey, could you help me out here? I've tried a few ways to skip the build on the unsupported architecture, but the CI still runs it and fails every time.

File: subprojects/packagefiles/kafel/meson.build [hunk 7]
```diff
@@ -0,0 +1,140 @@
...
+  default_options: ['c_std=gnu11'],
+)
+
+if host_machine.system() != 'linux'
+  error('kafel is not supported on this platform.')
+endif
+
+if host_machine.system() == 'linux' and host_machine.cpu_family() == 'ppc64le'
+  error('kafel does not support ppc64le')
+endif
```
bgilbert: The endianness is [not included](https://mesonbuild.com/Reference-tables.html#cpu-families) in the CPU family, and the magic words for this case are `unsupported architecture`.

```suggestion
if host_machine.cpu_family() == 'ppc64'
  error('unsupported architecture: ppc64')
endif
```

manipuladordedados: Thanks =)

subprojects/packagefiles/kafel/meson.build
```diff
@@ -0,0 +1,140 @@
+project(
+  'kafel',
+  'c',
+  version: '20231004',
+  license: 'Apache-2.0',
+  default_options: ['c_std=gnu11'],
+)
```
bgilbert, line 8:
```meson
  meson_version: '>=0.63.0',
```

subprojects/packagefiles/kafel/meson.build
```diff
@@ -0,0 +1,140 @@
...
+  input: 'src/lexer.l',
+  output: 'lexer.l.fixed',
+)
+
+lexer = custom_target(
+  'lexer',
+  command: [flex, '-o', '@OUTPUT0@', '--header-file=@OUTPUT1@', '@INPUT@'],
+  input: fix_lexer,
+  output: ['lexer.c', 'lexer.h'],
+  build_by_default: true,
```
bgilbert: I believe `build_by_default` shouldn't be necessary, since the lexer and parser outputs are included in `kafel_sources`.

dcbaker: Yeah, I think i suggested this as part of debugging the thing where they hard code the outputs in the files.

subprojects/packagefiles/kafel/meson.build
manipuladordedados: Thanks, done!

subprojects/packagefiles/kafel/meson.build
manipuladordedados: Thanks, done!

subprojects/packagefiles/kafel/meson.build
```diff
@@ -0,0 +1,140 @@
...
+  input: fix_lexer,
+  output: ['lexer.c', 'lexer.h'],
+)
+
+parser = custom_target(
+  'parser',
+  command: [bison, '-o', '@OUTPUT0@', '--defines=@OUTPUT1@', '@INPUT@'],
+  input: fix_parser,
+  output: ['parser.c', 'parser.h'],
+  build_by_default: true,
```
bgilbert: This one too.

bgilbert:
Thanks!

PR #1919, opened 2025-02-11
Add SDL3 wrap.

jpakkane:
Note that this is currently broken. It only does Linux. There is no configuration. It builds and links. It might work. I don't know as I did not test. Do not merge until it has at least Windows and macOS support.


subprojects/packagefiles/sdl3/meson.build
```diff
@@ -0,0 +1,252 @@
...
+# check token ['HAVE_INOTIFY']
+# check token ['HAVE_LIBUSB']
+# check token ['HAVE_O_CLOEXEC']
+]
+
+foreach f : check_functions
+  if cc.has_function(f.get(1), prefix : f.get(2), dependencies: [sdl_deps])
+    cdata.set(f.get(0), 1)
+  endif
+endforeach
```
eli-schwartz, line 202: Early Windows failure since dl can be a Windows emulated one which we happen to provide in the wrapdb. I suppose that isn't a general problem outside of the wrapdb CI... if https://github.com/dlfcn-win32/dlfcn-win32 isn't useful for SDL3 we can just use `allow_fallback: false` for it?

fr0zert:
Can you also create wrap for SDL_image SDL_shadercross libs?

jpakkane:
No-one commented so merging to work on other SDL3 libs. Assuming people will file bug reports (or preferably patches) if things don't work.

PR #1929, opened 2025-02-18
debug_assert: update to 1.3.4

shyun3:
https://github.com/foonathan/debug_assert/releases/tag/v1.3.4


subprojects/packagefiles/debug_assert/meson.build
```diff
@@ -23,11 +23,6 @@ if o_m_a != 'not overridden'
     debug_assert_ca += f'-DDEBUG_ASSERT_MARK_UNREACHABLE=@o_m_a@'
 endif
 
-o_a = get_option('override-assume')
-if o_a != 'not overridden'
-    debug_assert_ca += f'-DDEBUG_ASSERT_ASSUME=@o_a@'
-endif
-
```
shyun3, line 30: Support for `DEBUG_ASSERT_ASSUME` was removed in v1.3.3, but the CMake option is still available.

PR #1933, opened 2025-02-19
quickjs: new wrap

Gottox:
This pull request adds the `quickjs` library to the wrapdb.


Gottox:
Seems like qjs adds meson support upstream, closing.

ref: https://github.com/quickjs-ng/quickjs/pull/902

PR #1954, opened 2025-03-11
cxxopts: make install optional

shtrophic:
When using a header-only library such as cxxopts, it shouldn't be required for it to be installed to the system. I primarily oriented myself wrt. the `header-only` option of `fmt.wrap`.

### DIFF
--- a/subprojects/packagefiles/cxxopts/meson.build
+++ b/subprojects/packagefiles/cxxopts/meson.build
@@ -4,13 +4,17 @@ project('cxxopts',
     license : 'MIT'
 )
 
-install_headers('include/cxxopts.hpp')
+install = get_option('install')
 
-pkgc = import('pkgconfig')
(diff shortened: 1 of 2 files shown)


jpakkane:
`meson install` itself already has ways to not install parts of either by using  `--skip-subprojects` or `--tags`. It is better to use those than implement custom code that does the same thing in every project.

shtrophic:
...  or to have the ability to declare dependencies as private in the first place.
The issue I see here is that adding flags requires the user of the software to invoke meson correctly, while configuring the project in `meson.build` is something the developer can do as he intends.

stephanlachnit:
> The issue I see here is that adding flags requires the user of the software to invoke meson correctly, while configuring the project in `meson.build` is something the developer can do as he intends.

While I agree with your assessment here, I don't think there is a consensus to adding an install option each wrap, thus closing.

PR #1956, opened 2025-03-13
libcppgenerate: new wrap

meator:
This is a full port of libcppgenerate's CMake build system to Meson (not only a subproject-only project).

I have implemented several features to make this library more usable as a subproject/wrap:
1. when in a subproject, header files are not installed
2. when in a subproject, pkg-config files are neither installed nor generated
3. when in a subproject, static libraries are not installed (shared libraries are still installed if built, because they are required for the dependant project to run)

The `declare_dependency()` code is a bit ugly, because the CMake buildsystem provides two dependency names: `libcppgenerate` for shared library and `libcppgenerate-static` for static library. This style of link-dependent dependency separation isn't "natively" supported by Meson, so it has to be done more manually.

Since this is a complete port, I plan to ask upstream at some point whether they would be interested in integrating this build system themselves.

I am open to all feedback.


meator:
I need to change the working directory of a `custom_target()`. Because [`custom_target()` cannot do such things on its own](https://github.com/mesonbuild/meson/discussions/10390), I have to use an external program. The only portable (by portable I mostly mean Windows-compatible) way to do this is to provide a simple Python script which is roughly equivalent to

```
#!/bin/sh
cd "$1" && shift && exec "$@"
```

Wrapdb doesn't seem to allow such scripts. I do not know how to proceed, because I must change the working directory of the `custom_target()`. If anyone reading this has any tips, than please share them.

bgilbert:
Personally I think it'd be fine to add that Python script to `PER_PROJECT_PERMITTED_FILES`.

Issue #1937, opened 2025-02-20
Gtest Dependency ODR Violation


WillAyd:
There is a long upstream discussion on this in https://github.com/google/googletest/pull/4536, but the gist of it is that googletest is not meant to be used multiple times within a binary. Unfortunately, with the current WrapDB definition, it can be pretty trivial to run into an ODR violation:

```python
project(...)

gtest_dep = dependency('gtest')
test_helper_lib = library(
    'test-helper',
    ...
    dependencies: [gtest_dep],
)

text_exc = executable(
    'test-executable',
    ...
    dependencies: [gtest_dep],
    link_with: [test_helper_lib],
)
test('mytest', test_exc)
````

In such a configuration, ASAN will flag the ODR violation at runtime:

```
==168189==HINT: if you don't care about these errors you may set ASAN_OPTIONS=detect_odr_violation=0
SUMMARY: AddressSanitizer: odr-violation: global 'kElidedFramesMarker' at ../subprojects/googletest-1.15.2/googletest/src/gtest.cc:5014:19 in ...
==168189==ABORTING
```

Should the googletest wrap definition perhaps be changed to produce a shared library rather than offering up the source files?

jpakkane:
At least at the time upstream stance was that everyone should use sources directly rather than prebuilt libraries. Has that changed?

WillAyd:
I am not sure of the entire history but my interpretation of @higher-performance comment https://github.com/google/googletest/pull/4536#issuecomment-2173335752 is that we should be product a shared library

eli-schwartz:
FWIW multiple linux distros are quite happy to simply provide gtest as shared libraries installed to /usr/lib(64) and it seems to work out quite well when building other software that then depends on gtest for its testsuite.

jpakkane:
If that is the new recommend way then patches welcome I guess.

PR #1958, opened 2025-03-14
gtest: fix compilation under MSVC

neheb:
Turns out this path was not properly tested.

Changed handling to match CMake.


neheb:
ping @WillAyd

subprojects/packagefiles/gtest/googlemock/meson.build
```diff
@@ -13,39 +13,57 @@
...
+    )
+
+    gmock_main_dep = declare_dependency(
+        include_directories: [gtest_incdir, gmock_incdir],
+        link_with: gmock_main_lib,
+    )
 else
-    target_type = 'shared_library'
-endif
+    gmock_lib = library(
```
WillAyd, line 41: I think this still needs to be a shared_library (?). The original issue was that when used statically, it was trivial to produce an ODR violation with googletest. See also https://github.com/mesonbuild/wrapdb/issues/1937

neheb: testing now. https://github.com/google/googletest/blob/main/googlemock/CMakeLists.txt#L85 sets it as a static library, no?

WillAyd: I guess the question is if our goal is to match the CMake configuration or not. FWIW the CMake configuration itself is community-provided, i.e. not technically the official build system for gtest. I think based off of the discussion from the linked content the maintainers suggest that shared linkage is the way to go

Not an expert in this area however, so happy to be wrong

neheb: static build compiles and runs fine.

WillAyd: Yes that won't be a problem. The issue only appears at runtime in rare cases. 

The smallest reproducer I've been able to find I posted in https://github.com/google/googletest/pull/4536#issuecomment-2148422262 - is there an easy way to test this against that?

WillAyd: Ah ignore my comment above; the easiest reproducer I could find was in the issue already posted. So you tried this against that example and it ran fine? If so sure then I think it solves it, although it is entirely clear to me what would have changed between this and the original fix that prevents the ODR

PR #1961, opened 2025-03-17
Libcap: New Wrap

manipuladordedados:
https://sites.google.com/site/fullycapable/

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -514,6 +514,17 @@
       "windows": false
     }
   },
+  "libcap": {
+    "_comment": "- relies on Linux specific APIs",
+    "build_on": {
+      "darwin": false,
+      "msys2": false,
+      "windows": false
+    },
+    "debian_packages": [
+      "libattr1-dev"
+    ]
+  },
   "libdrm": {
     "_comment": "- relies on Linux and BSD specific APIs",
     "build_on": {
--- a/subprojects/packagefiles/libcap/meson.build
+++ b/subprojects/packagefiles/libcap/meson.build
@@ -1,0 +1,58 @@
+project(
+  'libcap',
+  'c',
+  version: '2.76',
+  license: 'BSD-3-Clause OR GPL-2.0-only',
+  default_options: ['c_std=gnu11'],
+)
+
+if host_machine.system() != 'linux'
+  error('libcap is not supported on this platform.')
+endif
+
+incdir = include_directories('libcap/include')
+
+cap_names_list_h = custom_target(
+  'cap_names.list.h',
+  output: 'cap_names.list.h',
+  input: 'libcap/include/uapi/linux/capability.h',
+  command: [find_program('gen_cap_names.py'), '@INPUT@', '@OUTPUT@'],
+  install: false,
+)
+
+_makenames = executable(
+  '_makenames',
+  sources: ['libcap/_makenames.c', cap_names_list_h],
+  include_directories: incdir,
+  native: true,
+)
+
+cap_names_h = custom_target(
+  'cap_names.h',
+  output: 'cap_names.h',
+  input: 'libcap/cap_names.header',
+  command: [_makenames, '@INPUT@'],
+  capture: true,
+  install: false,
+)
+
+libcap = library(
+  'cap',
+  sources: [
+    'libcap/cap_alloc.c',
+    'libcap/cap_extint.c',
+    'libcap/cap_file.c',
+    'libcap/cap_flag.c',
+    'libcap/cap_proc.c',
+    'libcap/cap_text.c',
+    'libcap/cap_syscalls.c',
+    cap_names_h,
+  ],
+  include_directories: incdir,
+  install: true,
+)
+
+libcap_dep = declare_dependency(
+  include_directories: incdir,
+  link_with: libcap,
+)
--- a/releases.json
+++ b/releases.json
@@ -1660,6 +1660,14 @@
       "3.6.2-1"
     ]
   },
+  "libcap": {
+    "dependency_names": [
+      "libcap"
+    ],
+    "versions": [
+      "2.76-1"
+    ]
+  },
   "libccp4c": {
     "dependency_names": [
       "libccp4c"
--- a/subprojects/libcap.wrap
+++ b/subprojects/libcap.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = libcap-2.76
+source_url = https://git.kernel.org/pub/scm/libs/libcap/libcap.git/snapshot/libcap-2.76.tar.gz
+source_filename = libcap-2.76.tar.gz
+source_hash = f6662862cb2727dcdbcd52a19c239eecb9d29c285c3768e9fb557cb027322034
+patch_directory = libcap
+
+[provide]
+libcap = libcap_dep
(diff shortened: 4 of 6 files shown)


yangm97:
Merge conflicts aside, is there anything holding this PR? 

I wanted to provide static builds on a meson project which is hanging on this specific dep.

Let me know if I can help testing this or something.

Thanks.

jpakkane:
```
In file included from ../subprojects/libcap-2.75/libcap/cap_syscalls.c:1:
../subprojects/libcap-2.75/libcap/libcap.h:28:10: fatal error: 'cap_names.h' file not found
   28 | #include "cap_names.h"
```

manipuladordedados:
> ```
> In file included from ../subprojects/libcap-2.75/libcap/cap_syscalls.c:1:
> ../subprojects/libcap-2.75/libcap/libcap.h:28:10: fatal error: 'cap_names.h' file not found
>    28 | #include "cap_names.h"
> ```

Fixed it!

subprojects/packagefiles/libcap/meson.build
```diff
@@ -0,0 +1,44 @@
+project('libcap', 'c',
+  version: '2.76',
+  license: 'BSD-3-Clause',
```
bgilbert, line 3:
```suggestion
  license: 'BSD-3-Clause OR GPL-2.0-only',
```

bgilbert:
We generally don't carry patches to upstream code.  Why does `_makenames` require a patch, and has it been reported upstream?

manipuladordedados:
> We generally don't carry patches to upstream code. Why does `_makenames` require a patch, and has it been reported upstream?

Hello!
The sh script is needed because the original _makenames.c depends on a file it’s supposed to generate (cap_names.list.h), creating a circular dependency. The script applies a patch to _makenames.c so it can build cleanly in Meson, which doesn’t handle circular dependencies automatically like the original Makefile does.

bgilbert:
That doesn't look correct.  `_makenames.c` generates `cap_names.h` but doesn't depend on it.  It does depend on `cap_names.list.h`, which is generated from `libcap/include/uapi/linux/capability.h` by `egrep` and `sed` commands in the Makefile.  Those commands might need to be reimplemented here as a small script (usually Python, but shell seems fine here since this wrap is Linux-only) but `_makenames.c` shouldn't need to be patched at all.

In any case, the `generate_cap_names` function in `generate_cap_names.sh` is never called, and `patched_makenames.c` produces a `cap_names.h` that doesn't actually contain any capability names.

manipuladordedados:
> That doesn't look correct. `_makenames.c` generates `cap_names.h` but doesn't depend on it. It does depend on `cap_names.list.h`, which is generated from `libcap/include/uapi/linux/capability.h` by `egrep` and `sed` commands in the Makefile. Those commands might need to be reimplemented here as a small script (usually Python, but shell seems fine here since this wrap is Linux-only) but `_makenames.c` shouldn't need to be patched at all.
> 
> In any case, the `generate_cap_names` function in `generate_cap_names.sh` is never called, and `patched_makenames.c` produces a `cap_names.h` that doesn't actually contain any capability names.

Thank you for your help. I was confused and made several mistakes along the way, but now it makes much more sense.
While everything works fine locally without issues, I'm still confused about the WrapDB restrictions. Why aren't external scripts allowed for dynamic file generation and manipulation, even when they're technically necessary for the build process?
I'd still like to understand this better, and any additional help would be greatly appreciated.

subprojects/packagefiles/libcap/meson.build
```diff
@@ -0,0 +1,49 @@
...
+    'buildtype=release']
+  )
+
+incdir = include_directories('libcap/include')
+
+cap_names_list_h = custom_target('cap_names.list.h',
+  output: 'cap_names.list.h',
+  input: 'libcap/include/uapi/linux/capability.h',
+  command: [
+    files('generator.sh'),
```
bgilbert: You can't call this `generator.sh` because that name has special significance in WrapDB.

subprojects/packagefiles/libcap/generator.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+exec grep -E '^#define\s+CAP_[A-Z0-9_]+\s+[0-9]+\s*$' "$1" | \
```
bgilbert: Nit: in a pipeline, `exec` doesn't do what it looks like it's doing, and it isn't really necessary here, so I'd drop.

subprojects/packagefiles/libcap/generator.sh
```diff
@@ -0,0 +1,8 @@
...
+    sed -e 's/^#define\s\+/{"/' \
+        -e 's/\s*$/},/' \
+        -e 's/\s\+/",/' \
+        -e 'y/ABCDEFGHIJKLMNOPQRSTUVWXYZ/abcdefghijklmnopqrstuvwxyz/' \
+    >> "$2"
```
bgilbert: Nit: I don't think append is ever what we really want here?

ci_config.json
```diff
@@ -514,6 +514,17 @@
       "windows": false
     }
   },
+  "libcap": {
+    "_comment": "- relies on Linux specific APIs",
+    "build_on": {
+      "darwin": false,
+      "msys2": false,
+      "windows": false
```
bgilbert, line 522: These want to see a corresponding `error()` call from the Meson code, whose text matches this test:

```python
if 'unsupported' in error or 'not supported' in error or 'does not support' in error:
```

manipuladordedados:
> CI will want you to run `meson format`.
> 
> > Why aren't external scripts allowed for dynamic file generation and manipulation, even when they're technically necessary for the build process?
> 
> We don't want to be in the business of shipping modified versions of upstream projects, so we don't allow patching of upstream sources. We do allow build scripts when needed.
> 
> The line between those is indeed a bit fuzzy. Patching an existing build tool instead of getting it fixed upstream would not have been good, but it's certainly fine to just copy some code out of the upstream Makefile. (Even if it needed to be translated to Python, which is essentially what we're doing with the Meson code anyway.) You can just add your script to `PER_PROJECT_PERMITTED_FILES`.

I really appreciate your help. Is everything looking good now?

subprojects/packagefiles/libcap/meson.build
```diff
@@ -0,0 +1,55 @@
+project(
+    'libcap',
+    'c',
+    version: '2.76',
+    license: 'BSD-3-Clause OR GPL-2.0-only',
+    default_options: ['c_std=gnu11', 'buildtype=release'],
```
bgilbert: We're dropping `buildtype` from `default_options` repo-wide; see #2208.

bgilbert (commented):
LGTM.  `meson format` will need to be rerun using the repo-specific config file; see the CI error message.

bgilbert:
Still needs to be reformatted properly.

manipuladordedados:
> Still needs to be reformatted properly.

Sorry, something was wrong with my editor.

bgilbert:
Thanks!

bgilbert:
`meson wrap install libcap` is currently failing on a 404.  https://github.com/mesonbuild/wrapdb/pull/2225 should fix it.

PR #1967, opened 2025-03-20
cpr: update to 1.11.2

Totto16:
There is nothing new in terms of build configuration or similar, since this is only a patch release.


jpakkane:
Test failure is unrelated.

PR #1968, opened 2025-03-20
curl: update to 8.12.1

Totto16:
This updates curl to 8.12.1

I used the diffs from [here](https://github.com/curl/curl/compare/curl-8_10_1...curl-8_12_1) to see all changes. I hope I caught all as it's 2 version updates.

There are some changes, that affect the code more, regarding extracting info from their 'Makefile.inc' files. 

Previously each test has a target called 'noinst_PROGRAMS', but now they changed the tests for 'libtests' to be named 'LIBTESTPROGS'. ~~I had to create two separate files for the extraction process, if anyone has a better idea, please feel free to tell me.~~


neheb:
ping @benoit-pierre

groengpx:
What is blocking this? The failed CI checks, I assume? The logs have expired, so I can't read them to investigate. Can I somehow help out?

Totto16:
> What is blocking this? The failed CI checks, I assume? The logs have expired, so I can't read them to investigate. Can I somehow help out?

Nothing really, just some review and I think after this long time some adjustments need to be made, to use the "new syntax" in .wrap files

Totto16:
@benoit-pierre this is no really ready, as all tests passes, and curl is already at 8.16.0 so we "need" to catch up 👀

groengpx:
Maybe just merge this, so we are at least somewhat more up to date, and then continue with 8.16.0?
Thanks for the fast response, btw.

Edit: ah, you meant 'now' instead of 'no'.  Quite the difference ;)

Totto16:
> Maybe just merge this, so we are at least somewhat more up to date, and then continue with 8.16.0? Thanks for the fast response, btw.

That was what I meant 😓  Huge jumps in versions are hard to port, so it's better to do in small incremental steps, and this would be the first one 😉

subprojects/packagefiles/curl/meson.build
```diff
@@ -785,10 +785,10 @@ foreach _provider : get_option('ldap-provider')
     _flag = 'OPENLDAP'
   elif _provider == 'win32'
     _dep = (host_machine.system() == 'windows' ? cc.find_library(
-      'wldap32',
+        'wldap32',
```
groengpx, line 788: Nitpick: what was wrong with the existing spacing here?
(also in various other places)

Totto16: We didn't have a format CI, now we have one and it complains about that, I just ran the formatter at the end

PR #1975, opened 2025-03-23
Feat: add format checks to the repo

### LINKED ISSUE #1969: [Discussion] format all build files in the same way
Since meson 1.5.0 added a `format` options, it would make sense IMO, to use that formatter (with eventual settings) to format all files, so that they are all consistent.

As far as I understand, the formatting of files is not really checked in PR's and also normal commits from Maintainers are not really formatted in a consistent style. It is not that important from a functionality perspective, but for general clean coding practices it is better to have a style, that is defined and that is used across the codebase consistently.

From a quick glance, it seems that most files are formatted with muon like styles, as I myself also used that for a long time, (and I also run it on save on new wraps I made).

I just ran this command, to see what this would look like: 

```bash
git ls-files --exclude='meson.build' --exclude='meson.options' --exclude='meson_options.txt' --ignored -c -z | xargs -0r meson format -i 
```

And this would be a huge commit (764 files changed) 😄  but it would be worth it IMO.

There could also be a meson format config `meson.format` or we could also use the ` --editor-config` switch of meson format.

What are your thoughts on this?

Totto16:
Fixes #1969

See also comments on #1969

This PR includes:
- A(minimal atm) `meson.format` (See [here](https://mesonbuild.com/Commands.html#format)) file
- It formats every `meson.build`, `meson_options.txt`, `meson.options` file
-  Adds another check step to the `sanity_checks.py` script

## Notes

The format check portion of the `sanity_checks` script gets only run on new releases and only on wraps with patches (in this repo) so no problem there (See concerns here https://github.com/mesonbuild/wrapdb/issues/1969#issuecomment-2744696887)

ALL relevant files were formatted in a separate commit, it was done using this command:
```bash
git ls-files --exclude='meson.build' --exclude='meson.options' --exclude='meson_options.txt' --ignored -c -z | xargs -0r meson format -i -c ./meson.format
```

## TODO

The formatting to use needs to be discussed, atm the config file only has one settings set:

```ini
use_editor_config=true
```

So it uses default settings.
See [here](https://mesonbuild.com/Commands.html#format) for all available options.


eli-schwartz (dismissed):
Avoid words such as "chore" in commit messages. The first ":"-delimited word in a commit message should be the affected area of the codebase, if applicable. See e.g. https://git-scm.com/docs/SubmittingPatches#describe-changes


Also:

> **chore**
> an unpleasant but necessary task.


I don't think that commit messages on the permanent record are the best place to describe the happiness level of the author when writing the patch, though I do sympathize with the sentiment.

> **feat**
> an achievement that requires great courage, skill, or strength.

This contrasts quite oddly with the former, and evokes images of a stage magician pulling a rabbit out of a hat. How do we determine which one commit is a pain to make and which one is a courageous act of great acclaim? What if different people have a different view on which one is which? For example, I'd argue that adding a config file that causes everything to be beautiful is the acclaimed feat, whereas applying automated changes is the unpleasant drudge work that you'd rather not be doing.

bgilbert:
> The formatting of ALL files is not that good of an idea (suggested here [#1969 (comment)](https://github.com/mesonbuild/wrapdb/issues/1969#issuecomment-2743859719)) as the CI complains about all of them with the message `Subprojects files changed but no new release added into releases.json`

Yeah, I think we should find a way to do it anyway.  I don't know the release tooling very well, so I'm not sure if it would break something if we ignore the CI on this one.  But otherwise we're going to be spending years dealing with unreviewable PRs that change  two lines, redo whitespace, and move punctuation around all in the same commit.

eli-schwartz:
The purpose of CI is to be a helpful heads up that something might be wrong and reviewers should factor this into their review as an early-warning sign.

Reviewers should feel ***strongly*** encouraged to use their best judgment about when a CI check doesn't make sense to be concerned about. The alternative is:
- not merging perfectly reasonable, otherwise acceptable PRs
- engaging in complex and laborious painful agonizing attempts to add some kind of pointless CI trigger to teach the CI to... not run on a certain change.

Obviously if your goal is that CI should ***not run*** on certain changes, and adding modifications to the CI to not run is an acceptable possibility, but hard or annoying to code, then ignoring the CI is equally acceptable save for the fact that historical statuses will show a red X and in the unlikely event someone looks at that old PR and asks why it was merged with failing CI, you have the downside of having to have this discussion about why it's not a problem.

And having discussions about why something isn't a problem is the number one most dangerous FOSS risk and we must automate away the possibility of having to have such discussions at all cost. Actually I lie. Nobody cares. :D

Having "always green" CI is more important when the CI is checking for defects in the code and all future PRs will be equally broken in CI once a defect is introduced into the code. That is not the case here for two reasons:
- the wrapdb CI only runs on the subcomponent being modified, so CI failures don't affect other subcomponents, and we sometimes introduce stricter CI, so any failures need to be considered in the context of *manually* testing the previous revision of that subcomponent, and CI can't actually express that, only reviewers can
- the CI check under discussion doesn't check for code defects at all, it checks to see if you've forgotten yo update the version counter. The code is valid before and after, and the CI is checking the diff hunk.


Note: this response is not about the change itself, only a meta-commentary about the change process.

(I don't have very strong opinions about this PR. I've been bitten before by automatic enforced formatting, in particular formatters that apply a 100% canonicalization rather than applying fixits, but I don't have experience with the meson formatter. Left to my own devices I'd include the config file as a benefit to people making changes, but rather than checking it in CI I'd expect people to have their editor set up to automatically apply it and commit however much of the resulting changes they find useful.)

bgilbert:
I've confirmed by inspection that `create_release.py` will not reupload an existing release if the latter's source is changed without updating `releases.json`.

> Left to my own devices I'd include the config file as a benefit to people making changes, but rather than checking it in CI I'd expect people to have their editor set up to automatically apply it and commit however much of the resulting changes they find useful.

They won't, though.  The regular committers might, but this repo also gets drive-by PRs.

While I don't feel suuuuuuper-strongly about it, I'm in favor of enforcing `meson format`.  I don't love all of the choices the formatter makes, with or without configuration, but this repo has a lot of build files and a decent amount of inconsistency and there's value in not having to think about formatting anymore.

tools/sanity_checks.py
```diff
@@ -140,7 +140,7 @@
         'process-zconf.py',
     ],
 }
-NO_TABS_FILES = ['meson.build', 'meson_options.txt']
+NO_TABS_FILES = ['meson.build', 'meson_options.txt', 'meson.options']
```
eli-schwartz, line 144: I'm slightly skeptical about this just because using this will immediately imply a minimum meson version of at least 1.1.0, which is a bit restrictive. It's not even available in Debian stable, let alone debian oldstable.

It's a very easy limitation to avoid, and avoiding it makes individual wraps more useful to a wider range of user.

On the other hand, it's possible a wrap legitimately needs newer versions -- various rust improvements, fs.relative_to(), dependency('atomic') since 1.7.0 (!!) -- so I don't know. I guess it's fine to allow, even if in most cases we'd still refrain from using it by default...

Totto16: This just adds the new options file (of which there a 6 in the repo atm (reflex, cpp-semver, mpark-patterns, pegtl, glpk, godot-cpp)) to the files, that are checked for tabs. This list just lists file names, that when encountered are checked.

tools/sanity_checks.py
```diff
@@ -481,6 +501,9 @@ def check_files(self, subproject: str, patch_path: Path) -> None:
         if not_permitted:
             not_permitted_str = ', '.join([str(f) for f in not_permitted])
             self.fail(f'Not permitted files found: {not_permitted_str}')
+        if unformatted:
+            unformatted_str = ', '.join([str(f) for f in unformatted])
+            self.fail(f'Not formatted files found: {unformatted_str}')
```
bgilbert, line 506: For contributors who don't know about `meson format`, can we make sure the error message has clear guidance about how to fix the problem?

bgilbert (commented):
Minor: I would favor adding `kwargs_force_multiline=true`, since IMO it makes function calls more legible.  Interested in other opinions though.

Totto16:
> Minor: I would favor adding `kwargs_force_multiline=true`, since IMO it makes function calls more legible. Interested in other opinions though.

I have no opinion on this, although it looks nicer with that setting. In the end the maintainers of this repo should agree on a style to use. 🤷🏼‍♂️

bgilbert (commented):
Two minor comments; LGTM otherwise.  I'll plan to merge this in a few days if nobody objects.

(For the record: I have not hand-inspected the last commit, but have verified that it matches a `meson format` run and will do so again before merging.)

tools/sanity_checks.py
```diff
@@ -481,6 +501,11 @@ def check_files(self, subproject: str, patch_path: Path) -> None:
         if not_permitted:
             not_permitted_str = ', '.join([str(f) for f in not_permitted])
             self.fail(f'Not permitted files found: {not_permitted_str}')
+        if unformatted:
+            unformatted_str = ', '.join([str(f) for f in unformatted])
```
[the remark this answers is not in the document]
Totto16: I agree with that view, but I find it better to have a "human readable" message, with commas, and the whole command after that, so I changed the message to

```python
            unformatted_str = ', '.join([str(f) for f in unformatted])
            unformatted_files_for_command = ' '.join([f"\"{str(f)}\"" for f in unformatted])
            self.fail(
                f'''Not formatted files found: {unformatted_str}
Run the following command, to format these files:
meson format --inplace {unformatted_files_for_command}''')
```


this has a few advantages, the command can just be copied, without manually assembling it, and also whitespace are handled correctly

this would look like this, if I run it locally:

```bash
AssertionError: Not formatted files found: subprojects/packagefiles/openssl/meson_options.txt, subprojects/packagefiles/openssl/meson.build
Run the following command, to format these files:
meson format --inplace "subprojects/packagefiles/openssl/meson_options.txt" "subprojects/packagefiles/openssl/meson.build"
```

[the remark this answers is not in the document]
bgilbert: Sure, works for me.

Optional improvement: `shlex.quote()` output would be more legible and a bit safer (on non-Windows) than hand-wrapping in `\"`.

bgilbert:
I just noticed that this indents options files to four spaces rather than two.  I'm merging this anyway to make forward progress, and will post a followup PR.

PR #1976, opened 2025-03-23
libwebp: update to 1.5.0

neheb:
Add PIC to static libraries to fix compilation with aarch64.

ping @benoit-pierre


jpakkane:
Thinking about this more, it should probably set `b_staticpic` properly rather than hardcoding it in every single project.

eli-schwartz:
Am I missing some context here?

Also as far as setting library kwargs vs project options, if for some reason libwebp intrinsically needs PIC for codegen then per target kwargs is correct. If this is just about linking libwebp into an application's shared framework then nothing should be changed in the wrapdb at all.

neheb:
The context is build all failure on aarch64

eli-schwartz:
Okay, I've gone ahead and taken a much closer look and I understand it a lot better now.

The context is that these are internal static_library()s which are used solely for linking into a shared library exported by that project.

Maybe a more correct way to describe this would be:
```meson
pic: get_option('default_library') != 'static' or get_option('b_staticpic'),
```

neheb:
what's wrong with unconditional pic?

eli-schwartz:
It's not necessarily *wrong*. For security reasons, distros prefer to use it everywhere. But it's not "needed", unless the code ends up in a shared library.

PIC has a small performance hit. Distros accept that tradeoff in the name of security and by arguing that the performance hit is quite minor. There are people out there though that like to turn off all the security flags they can, chasing after performance.

It's certainly easier (simpler, cleaner implementation) to get a correctly building shared library if you simply toggle PIC on

eli-schwartz (dismissed):
I don't see how this solves the problem, which is that the error log I saw was for sharpyuv/meson.build -- which has a static library too for simd support (neon specific in fact!)

bgilbert:
`b_staticpic` is documented to default to `true`.  If so, then while there's potentially a bug if the user overrides the option, how is it causing failures in wrapdb CI?

eli-schwartz:
That is a very good question and I've been slightly scratching my head over it too...

PR #1979, opened 2025-03-23
libjpeg-turbo: fix compilation under aarch64

neheb:
PIC is needed for static libraries.

ping @bgilbert


bgilbert:
All the `static_library()` calls should use the same `pic:` expression, either matching the existing `jpeg12` and `jpeg16` ones, or probably the fancier one in https://github.com/mesonbuild/wrapdb/pull/1976#issuecomment-2746621269.  I'd argue the x86-specific static library should include that logic too, even if it doesn't actually matter.

neheb:
know why it's failing?

bgilbert:
Hmm, no idea.  Looks like it was doing that before, though, so I think this is okay to merge.

PR #1981, opened 2025-03-26
libxslt: update to 1.1.43

neheb:


neheb:
uhh, what?

subprojects/packagefiles/libxslt/meson.build
```diff
@@ -97,6 +97,15 @@ endif
 
 libxml2_dep = dependency('libxml-2.0', version: '>=2.9.10')
 with_modules = libxml2_dep.get_variable(pkgconfig: 'modules', internal: 'modules', default_value: '0').to_int() != 0
+if host_machine.system() == 'cygwin'
+  config.set('MODULE_EXTENSION', '.cygdll')
```
bgilbert, line 101: Upstream's Autotools has this as just `.dll`.

subprojects/packagefiles/libxslt/meson.build
```diff
@@ -97,6 +97,15 @@ endif
...
+elif host_machine.system() == 'darwin'
+  config.set('MODULE_EXTENSION', '.bundle')
```
bgilbert, line 103: Shouldn't this be `.dylib`?

subprojects/packagefiles/libxslt/meson.build
neheb: Sounds like an upstream problem

subprojects/packagefiles/libxslt/meson.build
neheb: I have no idea. This was copy/paste from another wrap (htslib)

subprojects/packagefiles/libxslt/meson.build
bgilbert: While I don't have convenient access to a Cygwin system, I can't find any evidence that Meson builds Cygwin shared libraries with a `.cygdll` extension.  Why do you think this shouldn't be `.dll`?

subprojects/packagefiles/libxslt/meson.build
bgilbert: `.bundle` applies to _directories_ with a particular layout, not to files.  On macOS, libtool at least thinks modules (runtime-loadable shared libraries) should be `.so` and other dynamic libraries should be `.dylib`.  Upstream libxslt's Autoconf selects `.so` but I'm not completely clear what we should do here.

On the other hand, upstream 1.1.43 disables plugins by default, with the future intent to remove them.

subprojects/packagefiles/libxslt/meson.build
neheb: I have one. No cygdll extension, just dll. Removed.

subprojects/packagefiles/libxslt/meson.build
neheb: Removed. This is based on cmake conf btw. I have no idea what cmake does.

subprojects/packagefiles/libxslt/meson.build
neheb: Wonder if this is outdated: https://github.com/samtools/htslib/blob/master/configure.ac#L273

bgilbert: I suppose `dlopen()` and equivalent functions don't actually care about the file extension of the specified file, so arbitrary extensions will work if a library decides to use them.

PR #1982, opened 2025-03-26
Add SDL3_image.

jpakkane:


durkisneer1:
Is this being held until it's supported on Windows like the SDL3 wrap was?

PR #1988, opened 2025-03-27
minizip-ng: update to 4.0.9

bgilbert:
Drop unused `ng` option.


bgilbert:
This PR enables tests in CI for the first time.  MSVC complains about `test_main.cc` marking `main()` as `dllimport`, but 4.0.7 did that too.

neheb:
this issue could be avoided by not building test_main.cc and depending on gmock_main instead.

bgilbert:
Apparently `meson setup --fatal-meson-warnings` is now failing builds for CMake problems?

```
WARNING: CMake reported that the package bzip2 was not found, even though Meson's preliminary check succeeded.
```

Added in 1.7.1+, https://github.com/mesonbuild/meson/pull/14216.

bgilbert:
> this issue could be avoided by not building test_main.cc and depending on gmock_main instead.

Yes, that fixed it, thanks.

PR #1989, opened 2025-03-27
gdbm: update to 1.25

bgilbert:


bgilbert:
gdbm 1.25 introduces a mandatory dependency on `clock_nanosleep()`, which is not available on macOS.  I've reported this upstream and will leave the PR open as a placeholder.

bgilbert:
Fixed upstream in [ed0a86534568](https://git.savannah.gnu.org/cgit/gdbm.git/commit/?id=ed0a865345681982ea02c6159c0f3d7702c928a1).

neheb (approved):

PR #1992, opened 2025-03-27
sdl3: fix cross-compilation from UNIX to Windows

okvik:
When trying to cross-compile the `sdl3` wrap for Windows (from Linux) I encountered an issue with some system libraries supposedly missing:

```
$ meson setup build -Dwraps=sdl3 --cross-file zig-x86_64-windows.txt 
    ...
subprojects/SDL3-3.2.4/meson.build:256:17: ERROR: C shared or static library 'Imm32' not found
```

It turns out the problem is simply the case-sensitivity of UNIX file names and the fix is making the library names lower-case, which is how the `sdl2` wrap names these system libraries already.


neheb:
it's lowercase in Windows too:

```
C:\Windows\system32/imm32.dll
```

It's probably MSDN brain damage that uses mixed case.

PR #1999, opened 2025-03-28
Add farmhash.wrap

raghaven447:


subprojects/farmhash.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = farmhash-master
+source_url = https://github.com/google/farmhash/archive/refs/heads/master.zip
```
jpakkane, line 3: No master, only releases are accepted.

raghaven447: Changed as per the suggestion

subprojects/farmhash.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = farmhash-1.1
+source_url = https://github.com/raghaven447/farmhash/archive/refs/tags/v1.1.tar.gz
```
bgilbert, line 3: This is a personal fork of an archived repo that was last updated 6 years ago.  Is there a commonly-used downstream fork?

PR #2000, opened 2025-03-28
Add gemmlowp wrap

raghaven447:


subprojects/gemmlowp.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = gemmlowp-master
+source_url = https://github.com/google/gemmlowp/archive/refs/heads/master.zip
+source_filename = gemmlowp-master.zip
```
jpakkane, line 4: This is not acceptable. WrapDB ships only tagged releases, not trunk.

raghaven447: changed as per the suggestion

bgilbert: It appears that you've forked the project to your own personal repo and then tagged a release, which doesn't address the intent of the original concern.

PR #2006, opened 2025-03-31
Add fft2d wrap

raghaven447:


subprojects/fft2d.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = fft2d-1.0
+source_url = https://github.com/raghaven447/fft2d/archive/refs/tags/v1.0.tar.gz
```
jpakkane, line 3: This needs to point to the upstream release.

PR #2007, opened 2025-03-31
Add xla wrap

raghaven447:


subprojects/xla.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = xla-1.0
+source_url = https://github.com/raghaven447/xla/archive/refs/tags/v1.0.tar.gz
```
jpakkane, line 3: This is a personal fork. That is not acceptable, the urls need to point to unaltered upstream releases.

PR #2009, opened 2025-03-31
Add FP16 wrap

raghaven447:


subprojects/fp16.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = FP16-1.0
+source_url = https://github.com/raghaven447/FP16/archive/refs/tags/v1.0.tar.gz
```
bgilbert, line 3: This is a personal fork of [Maratyszcza/FP16](https://github.com/Maratyszcza/FP16).  Upstream has decided not to tag releases (https://github.com/Maratyszcza/FP16/issues/3, https://github.com/Maratyszcza/FP16/issues/19, https://github.com/Maratyszcza/FP16/issues/33) so I think WrapDB policy will not allow it to be packaged.

PR #2010, opened 2025-04-02
Indent options files to two spaces

bgilbert:
https://github.com/mesonbuild/wrapdb/pull/1975 indented them to four spaces, which is inconsistent with the settings for `meson.build`.


eli-schwartz:
IMHO 4 spaces is a nice default and maybe we should just be consistent about that. It's also the same one python uses.

bgilbert:
The Meson docs seem to consistently use two spaces for `meson.build` example code.  While the docs don't seem to have an opinion for options files, it feels inconsistent to have a different convention there.

eli-schwartz:
Yes but I am opinionated and think the docs made a mistake. :P

bgilbert:
I'm actually not a fan of two-space indents, but there's some value in being minimally confusing.

Since the indentation part is controversial, I'll break out the fixup commits into a separate PR.

jpakkane:
Two spaces was my mistake. Four would be better because that is what everybody else uses by default.

Totto16:
The .editorconfig says:

``ìni
[*.py]
indent_size = 4

[meson.build]
indent_size = 2
```

so adjust it there too

**Comment by dcbaker** [comment]
I'm a big fan of two space indent and not a fan of 4 space indent, and I have at least one practical reason for liking it:
```meson
if (really_long conditional
    that goes more than one line)
    "What this is unclear!"
endif

if (long
    line 2)
  "this is more readable"
endif
```

neheb (approved):

PR #2016, opened 2025-04-02
add raylib wrap

Totto16:
This is by no means feature complete, I just converted the Makefile  + cmake build files to this.

Feel free to suggest changes :)

Linux compilation tested, (Window will be tested by CI) Macos and others (cross compile for example) not


QWERTIOX:
Git didn't allow me to commit directly to this pr https://github.com/Totto16/wrapdb/pull/3 @Totto16 did you allow commits from maintainers?

Totto16:
> Git didn't allow me to commit directly to this pr @Totto16 did you allow commits from maintainers? My fixes [Totto16#3](https://github.com/Totto16/wrapdb/pull/3)

I did allow edit by maintainers, but since you are no maintainer of this repo, you can't push to the branch.

I'll look into your PR into this branch-

Thanks for helping out, It worked locally for me, but while trying to fix it for windows I added an error and then I didn't have enough time to fix it again. 😓

ivanrg99:
any updates on this?

Totto16:
> any updates on this?

We should ask potential reviewers (maintainers), as I use this wrap already privately, it works for me ™️ , but it also compiles everywhere, so it just needs to be reviewed 😓

subprojects/packagefiles/raylib/meson.build
```diff
@@ -0,0 +1,94 @@
+project(
+  'raylib',
+  'c',
+  version: '5.5',
+  license: 'zlib',
+  default_options: ['c_std=none'],
+  meson_version: '>=1.1.0',
```
bgilbert: It's not worth a 1.1.0 minimum solely for `meson.options`.  Let's call it `meson_options.txt`.

subprojects/packagefiles/raylib/meson.build
```diff
@@ -0,0 +1,94 @@
+project(
+  'raylib',
+  'c',
+  version: '5.5',
+  license: 'zlib',
+  default_options: ['c_std=none'],
```
bgilbert: To confirm, we really need to specify no `-std`, rather than accepting whatever `-std` the parent project selects?

subprojects/packagefiles/raylib/examples/meson.build
```diff
@@ -0,0 +1,261 @@
...
+      #this is needed, as this example doesn't include glad correctly
+      specific_compile_args += '-DGLAD_GLAPI_EXPORT'
+      #this is needed, as the hard copy of glad.h inside the examples/external folderincorrectly uses 'WIN32' instead of '_WIN32' as detection for windows
+      if host_machine.system() == 'windows'
+        specific_compile_args += '-DWIN32'
+      endif
+    endif
+
+
+    executable(
```
bgilbert, line 230: These aren't installed, aren't `override_find_program`med (and should not be), and aren't stored into a variable for use by the parent project, so is there a use case for building them at all?  The wrap could be simplified a lot by not building the examples.

subprojects/packagefiles/raylib/meson.build
```diff
@@ -0,0 +1,94 @@
...
+is_shared_lib = get_option('default_library') == 'shared'
+
+if is_shared_lib
+  private_compile_args += '-DBUILD_LIBTYPE_SHARED'
+  compile_args += '-DUSE_LIBTYPE_SHARED'
+endif
+
+
+if target_platform == 'DRM'
+  # without EGL_NO_X11 eglplatform.h tears Xlib.h in which tears X.h in
```
bgilbert, line 57: `tears`?

subprojects/raylib.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = raylib-5.5
+source_url = https://github.com/raysan5/raylib/archive/refs/tags/5.5.tar.gz
+source_filename = raylib-5.5.tar.gz
+source_hash = aea98ecf5bc5c5e0b789a76de0083a21a70457050ea4cc2aec7566935f5e258e
+patch_directory = raylib
+
+[provide]
+raylib = raylib_dep
```
bgilbert:
```suggestion
dependency_names = raylib
```

subprojects/packagefiles/raylib/src/meson.build
```diff
@@ -0,0 +1,332 @@
...
+  deps += dependency('gbm')
+  deps += dependency('threads')
+  deps += cc.find_library('m')
+  deps += dependency('dl')
+  deps += cc.find_library('rt')
+
+  need_atomic_lib = not cc.has_function(
+    'atomic_flag_clear',
+    prefix: '#include <stdatomic.h>',
+    dependencies: [cc.find_library('c')],
```
bgilbert: Explicitly depending on the C library is redundant, no?

bgilbert (changes requested):
There's already a `glfw` wrap the parent project can install if needed, so we shouldn't support the embedded GLFW at all.  It adds a lot of complexity here and reimplements Meson code maintained elsewhere, and also using private copies of library code isn't great.

subprojects/packagefiles/raylib/meson.build
Totto16: Sure 👍

subprojects/packagefiles/raylib/examples/meson.build
Totto16: The 'examples' option is off by default, and the reason for them being built is, that they provide "tests" as this project has no tests, and without them, we would never know, if the rest works correctly on every platform. Examples are built in CI, to test if everything works as expected.

subprojects/packagefiles/raylib/src/meson.build
Totto16: Yes, indeed

subprojects/packagefiles/raylib/meson.build
Totto16: Not my words, the comment was copy pasted from here https://github.com/raysan5/raylib/blob/ec06f9be371abff097c2426e8a39c358051ff397/src/Makefile#L433

Totto16: And he meant https://www.merriam-webster.com/dictionary/tear

subprojects/packagefiles/raylib/meson.build
Totto16: I am not sure, as the makefile and cmake file upstream just use nothing, i think specifying nothing would work I suppose, so we just use the default or the parent projetc version

Totto16:
> There's already a `glfw` wrap the parent project can install if needed, so we shouldn't support the embedded GLFW at all. It adds a lot of complexity here and reimplements Meson code maintained elsewhere, and also using private copies of library code isn't great.

I agree and I removed it, but I left it in for commit #1, so that in the future, if this would be needed, it would be in the git history :)

subprojects/packagefiles/raylib/meson.build
Totto16: I just checked again, using the Makefile it uses 'c99', so I used that too

ci_config.json
```diff
@@ -1252,6 +1252,26 @@
...
+  "raylib": {
+    "alpine_packages": [
+      "libxi-dev",
+      "libxkbcommon-dev",
+      "mesa-dev",
+      "wayland-dev"
+    ],
+    "debian_packages": [
+      "libxi-dev",
+      "libxkbcommon-dev",
```
neheb, line 1363: some of these have wraps. Any reason not to use?

Totto16: just faster compiler time in the CI, glfw could also be added to this, but since I had it as internal dependency, and raylib compiled it too, I didn't add it at first, and didn't visit this afterwards. I also think, more basic dependencies, like libxkbcommon are fine to install here, but I have nothing against also building the wrap there 🤷🏼‍♂️

Totto16:
@neheb or @bgilbert  

Any plans to merge this in the future, are any changes from me needed, or any notes on https://github.com/mesonbuild/wrapdb/pull/2016#discussion_r2463172603 

Just checking after some time 👀

QWERTIOX:
Hey, will be this merged?

PR #2019, opened 2025-04-03
inih: update to r59

bgilbert:


neheb:
https://github.com/benhoyt/inih/blob/master/tests/meson.build

looks off. shouldn't runtest.sh be ran through find_program?

neheb:
Upstream needs MSVC CI too.

bgilbert:
The CI failures here are test failures only, and should be fixed by https://github.com/benhoyt/inih/pull/192 (thanks @neheb!) so I think this is good to merge.

PR #2039, opened 2025-04-10
libkqueue: don't leak private headers to dep

neheb:


subprojects/packagefiles/libkqueue/meson.build
```diff
@@ -144,7 +144,8 @@ libkqueue = library(
   dependencies: kqueue_deps,
 )
 
+dep_inc = include_directories('include')
 libkqueue_dep = declare_dependency(
   link_with: libkqueue,
-  include_directories: kqueue_inc,
+  include_directories: dep_inc,
```
dcbaker, line 150: This could just be `include_directories: 'include'`

dcbaker:
With our without my suggestion: r-b

subprojects/packagefiles/libkqueue/meson.build
neheb: sure, OTOH this was a pattern for compatibility with 0.49.

PR #2044, opened 2025-04-14
fontconfig: update to 2.16.2

bgilbert:
Prevent several packages from requiring iconv in CI.  The current version of libxml2 disables autodetection for its deps, and we don't have iconv on Windows.


neheb:
Would be interesting to port https://github.com/win-iconv/win-iconv here.

PR #2048, opened 2025-04-17
add win-iconv wrap

Totto16:
See https://github.com/mesonbuild/wrapdb/pull/2044#issuecomment-2802862225 😄 

The latest release is quite old, the diff between then and now has a few changes :(
[see here](https://github.com/win-iconv/win-iconv/compare/v0.0.8...master)
(ref https://github.com/win-iconv/win-iconv/issues/43)
Edit: 0.0.9 released, I will add that later, once this PR is merged

Note: after this PR we can remove all occurrences of `libxml2:iconv=disabled` in the ci_config, as `iconv` can be found on Windows.

~~Question: Why is the ci_config settings not working? Why are non-windows ci jobs not skipped?~~
I should just read the entire log 🤦🏼‍♂️

### DIFF
--- a/subprojects/packagefiles/win-iconv/meson.build
+++ b/subprojects/packagefiles/win-iconv/meson.build
@@ -1,0 +1,60 @@
+project(
+  'win-iconv',
+  'c',
+  version: '0.0.8',
+  meson_version: '>=0.54.0',
+)
+
+if host_machine.system() != 'windows'
+  error(
+    'win-iconv is only supported on windows, and not supported on ' + host_machine.system(),
+  )
+endif
+
+default_libiconv_dll = get_option('default_libiconv_dll')
+
+c_args = [
+  '-D_CRT_SECURE_NO_WARNINGS',
+  '-DUSE_LIBICONV_DLL',
+  '-DDEFAULT_LIBICONV_DLL="' + default_libiconv_dll + '"',
+]
+
+c_lib_args = []
+
+if get_option('default_library') == 'both'
+  error('default_library=both is not supported')
+elif get_option('default_library') == 'shared'
+  c_lib_args += '-DMAKE_DLL'
+endif
+
+src_files = files('win_iconv.c')
+
+iconv_lib = library(
+  'iconv',
+  src_files,
+  c_args: c_args + c_lib_args,
+  vs_module_defs: 'iconv.def',
+  install: true,
+)
+
+win_iconv_dep = declare_dependency(
+  include_directories: include_directories('.'),
+  link_with: iconv_lib,
+)
+
+meson.override_dependency('iconv', win_iconv_dep)
+
+win_iconv = executable(
+  'win_iconv',
+  src_files,
+  c_args: c_args + ['-DMAKE_EXE'],
+  install: true,
+)
+meson.override_find_program('win_iconv', win_iconv)
+
+install_headers('iconv.h')
+
+if get_option('tests')
+  win_iconv_test_exe = executable('win_iconv_test', files('win_iconv_test.c'))
+  test('win_iconv_test', win_iconv_test_exe)
+endif
--- a/subprojects/packagefiles/win-iconv/meson_options.txt
+++ b/subprojects/packagefiles/win-iconv/meson_options.txt
@@ -1,0 +1,13 @@
+option(
+    'tests',
+    type: 'boolean',
+    value: true,
+    description: 'build tests',
+)
+
+option(
+    'default_libiconv_dll',
+    type: 'string',
+    value: '',
+    description: 'comma-separated list of iconv DLLs to try loading at runtime, e.g. "iconv.dll,libiconv.dll"',
+)
--- a/subprojects/win-iconv.wrap
+++ b/subprojects/win-iconv.wrap
@@ -1,0 +1,10 @@
+[wrap-file]
+directory = win-iconv-0.0.8
+source_url = https://github.com/win-iconv/win-iconv/archive/refs/tags/v0.0.8.tar.gz
+source_filename = win-iconv-0.0.8.tar.gz
+source_hash = 23adea990a8303c6e69e32a64a30171efcb1b73824a1c2da1bbf576b0ae7c520
+patch_directory = win-iconv
+
+[provide]
+dependency_names = iconv
+program_names = win_iconv
--- a/ci_config.json
+++ b/ci_config.json
@@ -1308,6 +1308,14 @@
       "wayland:dtd_validation=false"
     ]
   },
+  "win-iconv": {
+    "_comment": "- Requires Windows",
+    "build_on": {
+      "alpine": false,
+      "darwin": false,
+      "linux": false
+    }
+  },
   "x-plane-sdk": {
     "build_options": [
       "x-plane-sdk:cpp=enabled"
--- a/releases.json
+++ b/releases.json
@@ -4064,6 +4064,17 @@
       "0.8.2-1"
     ]
   },
+  "win-iconv": {
+    "dependency_names": [
+      "iconv"
+    ],
+    "program_names": [
+      "win_iconv"
+    ],
+    "versions": [
+      "0.0.8-1"
+    ]
+  },
   "wren": {
     "dependency_names": [
       "wren"


neheb:
```
Exception: Wrap win-iconv failed to configure due to bugs in the wrap, rather than due to being unsupported
```

neheb:
you need to change the commit message to having one of

```
if 'unsupported' in error or 'not supported' in error or 'does not support' in error:
```

subprojects/packagefiles/win-iconv/meson.options
```diff
@@ -0,0 +1,14 @@
+
```
neheb, line 1: I don't think this should start with a newline

subprojects/packagefiles/win-iconv/meson.build
```diff
@@ -0,0 +1,67 @@
...
+  meson_version: '>=1.1.0',
+)
+
+if host_machine.system() != 'windows'
+  error('win-iconv is only supported on windows, and not supported on '
+    + host_machine.system())
+endif
+
+cc = meson.get_compiler('c')
+cc_is_msvc = cc.get_id() in ['clang-cl', 'msvc']
```
neheb, line 14: could avoid with get_argument_syntax

subprojects/packagefiles/win-iconv/meson.build
```diff
@@ -0,0 +1,66 @@
...
+  error('win-iconv is only supported on windows, and not supported on '
+    + host_machine.system())
+endif
+
+cc = meson.get_compiler('c')
+
+c_args = []
+
+if cc.get_argument_syntax() == 'msvc'
+  c_args += '/D_CRT_SECURE_NO_WARNINGS'
```
bgilbert, line 18: Meson will auto-convert `-D`-style defines.

subprojects/packagefiles/win-iconv/meson.build
```diff
@@ -0,0 +1,66 @@
...
+if host_machine.system() != 'windows'
+  error('win-iconv is only supported on windows, and not supported on '
+    + host_machine.system())
+endif
+
+cc = meson.get_compiler('c')
+
+c_args = []
+
+if cc.get_argument_syntax() == 'msvc'
```
bgilbert, line 17: I believe the define is needed with Windows headers, regardless of the compiler.  I'd suggest making the define unconditional since we only support Windows anyway.

subprojects/packagefiles/win-iconv/meson.build
```diff
@@ -0,0 +1,66 @@
...
+
+win_iconv = executable(
+  'win_iconv',
+  src_files,
+  c_args: c_args + ['-DMAKE_EXE'],
+  install: true,
+)
+meson.override_find_program('win_iconv', win_iconv)
+
+install_headers('iconv.h')
```
bgilbert, line 61: Upstream also installs `localcharset.h`.

subprojects/packagefiles/win-iconv/meson.build
```diff
@@ -0,0 +1,66 @@
...
+c_args = []
+
+if cc.get_argument_syntax() == 'msvc'
+  c_args += '/D_CRT_SECURE_NO_WARNINGS'
+endif
+
+libiconv_dll_opt = get_option('libiconv_dll')
+
+if libiconv_dll_opt != ''
+  c_args += '-DUSE_LIBICONV_DLL'
```
bgilbert, line 24: Upstream defines this unconditionally, which lets win-iconv respect the `WINICONV_LIBICONV_DLL` environment variable even if `DEFAULT_LIBICONV_DLL` isn't defined.  Let's define this by default, matching upstream.  If you want, you could add a Meson option to omit it.

subprojects/packagefiles/win-iconv/meson.options
```diff
@@ -0,0 +1,13 @@
...
+    type: 'boolean',
+    value: true,
+    description: 'build tests',
+)
+
+option(
+    'libiconv_dll',
+    type: 'string',
+    value: '',
+    description: 'comma separated list (e.g. "iconv.dll,libiconv.dll"), this is used to get win32 dlls at runtime',
```
bgilbert, line 12: This doesn't seem very clear.  Maybe something like `comma-separated list of iconv DLLs to try loading at runtime, e.g. "iconv.dll,libiconv.dll"`?

subprojects/win-iconv.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = win-iconv-0.0.8
+source_url = https://github.com/win-iconv/win-iconv/archive/refs/tags/v0.0.8.tar.gz
+source_filename = win-iconv-0.0.8.tar.gz
+source_hash = 23adea990a8303c6e69e32a64a30171efcb1b73824a1c2da1bbf576b0ae7c520
+patch_directory = win-iconv
+
+[provide]
+iconv = win_iconv_dep
```
bgilbert, line 9: Minor: since `meson.build` unconditionally calls `meson.override_dependency()`, this can be written as

```
dependency_names = iconv
```

subprojects/packagefiles/win-iconv/meson.build
```diff
@@ -0,0 +1,66 @@
+project(
+  'win-iconv',
+  'c',
+  version: '0.0.8',
+  meson_version: '>=1.1.0',
```
bgilbert, line 5: This seems overly restrictive.  Does this wrap require any features from newer Meson versions?

bgilbert (commented):
Looks good generally.  Thanks for doing this!

subprojects/packagefiles/win-iconv/meson.build
neheb: In other wraps, there's a trick to use the /D syntax with get_supported_arguments to apply this define only on MSVC compilers. _CRT_SECURE_NO_WARNING is only valid there.

neheb: yeah.

```
/usr/x86_64-w64-mingw32/include$ grep _CRT_SECURE_NO_WARNINGS *.h
_mingw_mac.h:  "This function or variable may be unsafe, use _CRT_SECURE_NO_WARNINGS to disable deprecation"
_mingw_mac.h:#if !defined (_CRT_SECURE_NO_WARNINGS) || (_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES == 0)
```

bgilbert: MinGW headers also use `_CRT_SECURE_NO_WARNINGS` to silence warnings, and in that case the compiler and the argument syntax are GCC.

neheb: hrm looking at it more carefully, you might be right. Interestingly even though there's a define to print a deprecation message, I haven't seen it in practice. I wonder why...

subprojects/packagefiles/win-iconv/meson.build
Totto16: I got a warning, since I didn't specify an explicit version `meson.options` as options files is only supported since 1.1.0 😓

subprojects/packagefiles/win-iconv/meson.build
Totto16: I just copied it from the original Makefile, but they might have done it incorrectly 👍🏼

subprojects/packagefiles/win-iconv/meson.build
bgilbert: Yeah, it's better to call it `meson_options.txt` and let the wrap be used on systems with older Meson.

subprojects/packagefiles/win-iconv/meson.build
Totto16: I changed it to `-D` as it can be also used under MSYS2, see the other comment https://github.com/mesonbuild/wrapdb/pull/2048#discussion_r2051030852

subprojects/win-iconv.wrap
Totto16: yes, but then we need another meson version check, as `meson.override_dependency()` was introduced in 0.60.0 ish (I have to look it up, but it was before 1.0)

subprojects/packagefiles/win-iconv/meson.options
Totto16: I just copied the upstream description and added a clarification, but it is still not very clear, so I'll change it 👍🏼

subprojects/win-iconv.wrap
Totto16: It was introduced in `>=0.54.0` So I'll use that as meson version, which seems reasonable to me, since that is a really old version, which everybody should be able to use 👍🏼

subprojects/packagefiles/win-iconv/meson.build
Totto16: that is only present since 0.0.9, check out the 0.0.8 tag, or my branch in https://github.com/Totto16/wrapdb/tree/udpate_iconv_to_0.0.9 for that version, I tackel that version, after this PR is closed, as stated in the initial description

subprojects/packagefiles/win-iconv/meson_options.txt
```diff
@@ -0,0 +1,13 @@
...
+    type: 'boolean',
+    value: true,
+    description: 'build tests',
+)
+
+option(
+    'libiconv_dll',
+    type: 'string',
+    value: '',
+    description: 'comma-separated list of iconv DLLs to try loading at runtime, e.g. "iconv.dll,libiconv.dll',
```
bgilbert, line 12: Nit: missing trailing `"`.

subprojects/packagefiles/win-iconv/meson_options.txt
```diff
@@ -0,0 +1,13 @@
+option(
+    'tests',
+    type: 'boolean',
+    value: true,
+    description: 'build tests',
+)
+
+option(
+    'libiconv_dll',
```
bgilbert, line 9: Hmm, maybe `default_libiconv_dll` would be clearer?

subprojects/packagefiles/win-iconv/meson.build
bgilbert: Whoops, good call.

PR #2049, opened 2025-04-17
oatpp, oatpp-test: update to 1.3.1 

Totto16:
This updates `oatpp` and `oatpp-test` to the latest upstream release.

This fixes a few issue with this wrap, we had earlier, like the incorrect tests on windows (see #1031)


jpakkane:
```
stderr:
pure virtual method called
terminate called without an active exception
```

Totto16:
I think upstream doesnt support riscv, and back when this was introduced, there was no riscv CI 🤔

Totto16:
I changed some things, so that the ci_config also allows excluding certain architectures, feel free to suggest or ways on how to solve this in a better way.

But riscv is currently not yet officially supported, which I think might be the case for many packages, but this one actually fails, the reason is probably strange, but I didn't bother investigating.

Totto16:
> Most projects shouldn't care about the CPU architecture at all, some will have architecture-specific optimizations but should fall back to generic code, and only a few specialized projects will be targeted to specific architectures. I'm not sure it makes sense to add an architecture filter to CI, since in most cases arch-specific failures reflect a genuine bug and the filter will encourage the bug to be ignored.

I agree with you, but the "problem" with that is, that meson wraps can't add patches, the upstream doesn't make releases that often and also don't support riscv officially. So I am not sure how and if they are gonna fix that bug. 

I agree, that most of the time that is not really needed, but there might be some cases, where e.g. 32 bit platforms are not supported anymore, this might come up in the future. I'll remove that stuff from this PR, and try to debug the error.

If the error doesn't exist with the official buildsystem (who knows 🤷🏼‍♂️), as I couldn't find any issue or PR regarding that issue, this build definition has an error, otherwise there is little we / I can do, to make the tests pass.  

So if the error exists, upstream doesn't care (at least atm) about riscv, as there are, at least atm, not that many users, could this PR be merged despite the riscv tests fail?  I don't love that, but blocking all other architectures, just because one architecture is not supported, and we test that in the CI is also not that great of an option. 

And I also don't love that, as nowhere (Except this PR) that thing is documented, I could leave the riscv detection in the code, so that the CI hard fails, but all in all there are no great options IMO 🤷🏼‍♂️

bgilbert:
My preference, and I think the general approach in this repo, is to omit the explicit riscv detection from `meson.build`, and merge with the riscv tests failing.  Since upstream doesn't explicitly refuse to support riscv (as far as I know), we shouldn't enforce that in `meson.build`.

It's not unusual to merge PRs here with some CI failures, either for harmless issues (e.g. tests that are known to be broken) or for real problems which are not regressions.  As you say, there are problems that our own policies prevent us from fixing downstream.

If you're able to reproduce the riscv failure with the upstream build system, then it'd be good to report it to upstream.  But that isn't a requirement for merging here.

Totto16:
I confirmed, that the issue also exists, when build with the upstream build system. So I reverted every change regarding the ci config. I might eventually file an issue upstream, but they don't even have a CI and I never use riscv 🤷🏼‍♂️ 

So this should be mergable with the test for riscv failing

subprojects/packagefiles/oatpp/meson.build
```diff
@@ -1,23 +1,21 @@
-
 project(
   'oatpp',
   'cpp',
   default_options: ['default_library=static', 'cpp_std=c++11'],
-  version: '1.3.0',
+  version: '1.3.1',
   license: 'Apache-2.0',
+  meson_version: '>=1.3.0',
```
bgilbert, line 7: Why is this needed now?

subprojects/packagefiles/oatpp/meson.build
```diff
@@ -92,7 +86,15 @@ if is_msvc_or_similar
   )
 endif
 
-
+proj_flags = cpp.get_supported_arguments(
+  '-Wno-unused-parameter',
+  '-Wno-pessimizing-move',
+  '-Wno-unknown-warning-option',
```
bgilbert, line 92: I'm okay with matching upstream here, but FWIW in Meson all dubious compiler options should be wrapped in `get_supported_arguments()` or similar, so this shouldn't be needed.

Totto16:
Oh shit, it already got merged 😓

Totto16:
Should I do a `-2` revision, with the changes regarding your reviews @bgilbert

bgilbert:
Yeah, please do.

subprojects/packagefiles/oatpp/src/meson.build
```diff
@@ -42,6 +42,11 @@ endforeach
 
 
 foreach oatpp_module : oatpp_modules
+  if (oatpp_module == 'oatpp-tests') and not allowed
```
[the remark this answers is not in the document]
Totto16: Addressed in #2053

subprojects/packagefiles/oatpp/meson.build
Totto16: Addressed in #2053

PR #2052, opened 2025-04-18
update win-iconv to v0.0.9

Totto16:
follow up of #2048


subprojects/packagefiles/win-iconv/meson.build
```diff
@@ -27,7 +27,7 @@ elif get_option('default_library') == 'shared'
   c_lib_args += '-DMAKE_DLL'
 endif
 
-src_files = files('win_iconv.c')
+src_files = files('iconv.h', 'localcharset.h', 'win_iconv.c')
```
bgilbert, line 30: Headers [don't need to be included](https://mesonbuild.com/FAQ.html#do-i-need-to-add-my-headers-to-the-sources-list-like-in-autotools) in library/executable sources.

PR #2053, opened 2025-04-18
Fix remaining reviews of #2049

Totto16:
#2049 was merged too early, a few review comments need to be addressed

## Reviews

https://github.com/mesonbuild/wrapdb/pull/2049#discussion_r2051132560:

That was indeed an error, I looked at the upstream changes and I didn't understand what I did either, I correctly reflected the upstream changes now, see
[here](https://github.com/oatpp/oatpp/compare/1.3.0...1.3.1#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR28) for those

https://github.com/mesonbuild/wrapdb/pull/2049#discussion_r2051122607

restricting it to `>=1.3.0` wasn't really needed, I removed a minimum required version, as there are no features, that require a certain version (checked locally and in the CI with `--fatal-meson-warnings`), and also renamed `meson.options` to `meson_options.txt`


Totto16:
riscv CI failure is ok, as discussed in #2049

subprojects/packagefiles/oatpp/src/meson.build
```diff
@@ -28,39 +28,27 @@ test_module_inc_dirs += include_directories('.')
...
+endif
 
-  conf = configuration_data()
-  conf.set_quoted('OATPP_THIS_MODULE_NAME', 'oatpp')  ## name of the module (also name of folders in installation dirs)
-  conf.set_quoted('OATPP_THIS_MODULE_VERSION', meson.project_version())  ## version of the module (also sufix of folders in installation dirs)
-  conf.set_quoted('OATPP_THIS_MODULE_LIBRARIES', oatpp_modules_flat)  ## list of libraries to find when find_package is called
-  conf.set_quoted('OATPP_THIS_MODULE_TARGETS', oatpp_modules_flat)  ## list of targets to install
-  conf.set_quoted('OATPP_THIS_MODULE_DIRECTORIES', oatpp_modules_flat)  ## list of directories to install
+conf = configuration_data()
+conf.set_quoted('OATPP_THIS_MODULE_NAME', oatpp_module)  ## name of the module (also name of folders in installation dirs)
```
bgilbert, line 40: I'm not sure all of the quoted values should actually be quoted.  The generated file doesn't look right:

```cmake
if(NOT TARGET oatpp::"oatpp")
    include("${CMAKE_CURRENT_LIST_DIR}/"oatpp"Targets.cmake")
endif()

set_and_check("oatpp"_INCLUDE_DIRS "${PACKAGE_PREFIX_DIR}/include/oatpp-/"oatpp"/")
set_and_check("oatpp"_LIBRARIES_DIRS "${PACKAGE_PREFIX_DIR}//oatpp-/")

set("oatpp"_LIBRARIES )
set(OATPP_BASE_DIR "${PACKAGE_PREFIX_DIR}/include/oatpp-/")
```

But also, do we actually need to generate the `.cmake` file?  This wrap is the only one in WrapDB that uses `cmake.configure_package_config_file()` (or `cmake.write_basic_package_version_file()`) and generally we don't expect wraps to build libraries that can be ingested natively by CMake.  The fact that the generated file is called `oatppConfig.cmakeConfig.cmake` suggests that no one has been using it.

Totto16: Yeah, I am no regular cmake user, and don't know everything there, this seems wrong. I just tried to do the same as the upstream build files, but it is indeed not needed. I'll remove it 👍🏼

bgilbert:
Thanks!

Looking at the riscv logs, the message

```
pure virtual method called
terminate called without an active exception
```

seems likely to be a consequence of the timeout on the previous line:

```
|2025-04-22 14:04:16 1745330656635914| ASSERT[FAILED]:"ClientServerTestRunner: Error. Timeout." && elapsed < timeout
```

so I think this is probably the result of slow CPU emulation and there isn't actually an upstream bug.

PR #2054, opened 2025-04-19
sanity_checks: auto-disable fatal warnings if CMake warning is in logs

bgilbert:
New Meson releases sometimes add dubious warnings that break existing packages built with `--fatal-meson-warnings`, which is the WrapDB default. Add a mechanism to detect when setup fails on a warning that matches a regex and automatically rerun setup without `--fatal-meson-warnings`.  This is a large hammer, but allows us to continue making forward progress in this repo while fixes percolate into Meson.

For now, match the new warning in 1.7.1+ that triggers if CMake is used to detect a dependency when the `Find*.cmake` file is shipped with CMake itself and the dependency is not available.  This shouldn't be fatal, and in fact it triggers even for optional dependencies.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -806,11 +806,9 @@
     "skip_tests": true
   },
   "minizip-ng": {
-    "_comment": "Fatal warnings disabled because CMake is unhappy: https://github.com/mesonbuild/meson/pull/14216",
     "build_options": [
       "minizip-ng:tests=enabled"
-    ],
-    "fatal_warnings": false
+    ]
   },
   "mpdecimal": {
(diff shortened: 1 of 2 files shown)


bgilbert:
Warning added in https://github.com/mesonbuild/meson/pull/14216.  There's some post-merge discussion there.

bgilbert:
Proposed Meson fix in https://github.com/mesonbuild/meson/pull/14496.

bgilbert:
I'd like to get this landed to unblock CI.  Anyone want to take a look?

neheb (approved):

PR #2038, opened 2025-04-10
spdlog: opinions

bgilbert:
Drop support for the bundled copy of fmt.  For cases where fmt isn't already available, fmt is in WrapDB.  Don't use a random copy shipped with spdlog.

Also compile the library by default, and adjust the soname to match upstream.

PR #2059, opened 2025-04-21
spdlog: install library

stephanlachnit:
Fixup for https://github.com/mesonbuild/wrapdb/pull/2038: installations fail since the shared library is not installed


bgilbert:
Yup, thanks!

PR #2063, opened 2025-04-23
sdl3_image patch: adding missing 'IMG.c' to img_src files

robertomorrison0:
'IMG.c' was missing from the img_src files, so some functions like `IMG_Load` were missing


bgilbert:
You'll need to add a revision bump to `releases.json`.

robertomorrison0:
I realized that this needs more work, as it won't work as expected and platform support is not there.
I'll open another pull request, when this is ready.

PR #2066, opened 2025-04-25
SDL3_image: update

robertomorrison0:
I based this on the SDL2_image wrap and was only able to test it on Arch Linux.
Functions like 'IMG_Load' are available now.


releases.json
```diff
@@ -3496,7 +3496,7 @@
       "sdl3_image"
     ],
     "versions": [
-      "3.2.4-1"
+      "3.2.4-2"
     ]
```
jpakkane, line 3500: You need to add a new entry at the top. Do not delete the old one.

robertomorrison0: Thanks, I removed the merge commit and added the old version entry back in

jpakkane: JSON does not support having a comma at the end of an array.

robertomorrison0: Sorry for all this errors, I usually use json-c, where that is not a problem. Thank you for your patience.

robertomorrison0: As I just fixed some problems in the existing version, I didn't expect it was not compatible with the sdl3 version, that is used in the wrapdb. Should I downgrade this? I'm currently looking into updating the sdl3 wrap too, but I don't know, if/when I have the time too finish that.

PR #2067, opened 2025-04-26
Sdl3: Update to 3.2.10

robertomorrison0:
Moving to most recent release, as the sdl3_image wrap (3.2.4) depends on sdl3 >= 3.2.6


subprojects/packagefiles/sdl3/meson.build
```diff
@@ -278,9 +291,9 @@ elif host_machine.system() == 'windows'
   cdata.set('SDL_XINPUT', 1)
   cdata.set('SDL_TIMER_WINDOWS', 1)
   sdl_deps += cc.find_library('version')
-  sdl_deps += cc.find_library('imm32')
-  sdl_deps += cc.find_library('setupapi')
-  sdl_deps += cc.find_library('winmm')
+  sdl_deps += cc.find_library('Imm32')
+  sdl_deps += cc.find_library('SetupAPI')
+  sdl_deps += cc.find_library('Winmm')
```
HeliumAnt, line 296: these should stay lowercase, otherwise breaks crosscompile, see #1992

neheb: it's lowercase on windows too. That mixed case is an MSDN documentation bug.

jpakkane:
Looks good, please squash to a single commit.

PR #2069, opened 2025-04-27
Add atomic_queue wrap

stephanlachnit:
https://github.com/max0x7ba/atomic_queue


stephanlachnit:
MacOS tests fail due to a header included which uses linux pthread functions. Should be solved via https://github.com/max0x7ba/atomic_queue/pull/80

PR #2072, opened 2025-04-28
backward: Add dependency to just grab header

WillAyd:
backward-cpp distributes a a header-only library in addition to the module. The header-only library can be useful as a standalone dependency, especially on systems where the module may be difficult to compile (i.e. Windows)

https://github.com/bombela/backward-cpp?tab=readme-ov-file#installation

### DIFF
--- a/subprojects/backward-cpp.wrap
+++ b/subprojects/backward-cpp.wrap
@@ -8,3 +8,4 @@ patch_directory = backward-cpp
 
 [provide]
 backward-cpp = backward_dep
+backward-cpp-interface = backward_interface_dep
--- a/releases.json
+++ b/releases.json
@@ -192,9 +192,11 @@
   },
   "backward-cpp": {
     "dependency_names": [
-      "backward-cpp"
+      "backward-cpp",
+      "backward-cpp-interface"
     ],
     "versions": [
+      "1.6-6",
       "1.6-5",
       "1.6-4",
       "1.6-3",
--- a/subprojects/packagefiles/backward-cpp/meson.build
+++ b/subprojects/packagefiles/backward-cpp/meson.build
@@ -25,35 +25,23 @@ dwarf = dependency(
 )
 
 deps = []
-
+args = []
 # stack walking
 if libunwind.found()
   deps += libunwind
-  add_project_arguments(
-    '-DBACKWARD_HAS_LIBUNWIND=1',
-    language: 'cpp',
-  )
+  args += ['-DBACKWARD_HAS_LIBUNWIND=1']
 endif
 
 # stack detail extraction
 if dw.found()
   deps += dw
-  add_project_arguments(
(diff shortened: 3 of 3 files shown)


WillAyd:
That is a surprising amount of failures, although I don't believe any of them are related to this change (?)

dcbaker:
A couple of things I see. I think you'll need to include the `-DBACKTRACE_HAS_N` declare_dependency for the interface. which likely means taking the stuff out of the `add_project_rguments(...)`, sticking it in an array, and then adding it to the dependencies as necessary.

I'm not sure if backtrace has been updated since alphine was added. This is annoying, because alpine is MUSL based, and doesn't have an execinfo.h or libexecinfo, so we really need to have a check in the Meson that does something like (untested):
```meson
has_backtrace = cpp.has_header('execinfo.h') and cpp.has_header_symbol('execinfo.h', 'backtrace')
add_project_arguments('-DBACKWARD_HAS_BACKTRACE=@0@'.format(has_backtrace.to_int(), language : 'cpp')
```
to deal with this.

neheb:
Last I looked into this fixing the execinfo.h issue requires upstream changes/

dcbaker:
I just looked at the header to see that you could set `BACKTRACE_HAS_BACKTRACE=0`, but I'm not surprised to hear that path isn't well tested and doesn't work.

WillAyd:
Just rebased on top of https://github.com/mesonbuild/wrapdb/pull/2129

In the process I simplified this and took out any of the macro changes, since 2129 mostly dealt with those. I think there's still a gap with the `BACKWARD_HAS_BACKTRACE` macro and detection, but hoping we can leave that to another PR

subprojects/backward-cpp.wrap
```diff
@@ -8,3 +8,4 @@ patch_directory = backward-cpp
 
 [provide]
 backward-cpp = backward_dep
+backward-cpp-interface = backward_cpp_interface_dep
```
bgilbert: Nit: could call the variable `backward_interface_dep` for consistency with the other variable (and, yes, inconsistency with the provide name), since this is technically part of the wrap's API.

bgilbert (commented):
I think this still needs to propagate the defines into the `declare_dependency()` `compile_args` as mentioned in the first paragraph of https://github.com/mesonbuild/wrapdb/pull/2072#issuecomment-2839710221.

subprojects/backward-cpp.wrap
WillAyd: done!

bgilbert:
The new Alpine failures are because the `suicide` test is no longer receiving the defines, since it doesn't depend on `backward_dep`.  It'll need `args` passed explicitly.

WillAyd:
Done! I believe the existing Alpine failures exist on main

PR #2077, opened 2025-04-30
liblzma: gettext functions aren't used

ColinKinloch:
~I think this is a typo, setting `HAS_GETTEXT` and `HAS_DCGETTEXT` both based on whether `dcgettext` is available, however I could be wrong.~

I was a little to focused on fixing my immediate issue when writing https://github.com/mesonbuild/wrapdb/pull/2065 it didn't occur to me how odd it looked.


bgilbert:
So it turns out that upstream defines these as a side effect of calling `AM_GNU_GETTEXT()`, but doesn't actually use either define.  While I think this PR is okay as-is, it's a bit awkward to carry configuration code that doesn't actually do anything.  Alternatively, we could drop these defines entirely.

neheb:
I don't remember if this was my fault. I believe I was originally trying to match the autoconf config. If this is really unused then that's better.

PR #2111, opened 2025-05-15
Remove unnecessary subproject variables inside meson.build

QWERTIOX:
I'm not sure if it's wright, sanity checks passes - original commit that added this variable was later removed https://github.com/mesonbuild/wrapdb/commit/db21541a5afcaa5daed94c3264c1b12fc4eeb37d

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@ project(
 wraps = get_option('wraps')
 
 foreach w : wraps
-  sub = subproject(w)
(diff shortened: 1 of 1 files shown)


bgilbert:
For the record, use of the variable was removed in a7472a8e9a286c744416462b423333b2b332cbe5, a day after it was added in f182996e436a8c62dcab8c25c90b7e7915e885d7.

PR #2113, opened 2025-05-19
Add Yaml-cpp wrap

j-carl:


subprojects/packagefiles/yaml-cpp/0001-missing-header-emitterutils.cpp.patch
```diff
@@ -0,0 +1,10 @@
+diff --git i/src/emitterutils.cpp w/src/emitterutils.cpp
```
eli-schwartz, line 1: This is not `git format-patch` output, why not? The provenance of this change aligns with https://github.com/jbeder/yaml-cpp/pull/1310 so you should be documenting that...

You should probably just ask upstream to release a new version though.

eli-schwartz (changes requested):

subprojects/packagefiles/yaml-cpp/0001-missing-header-emitterutils.cpp.patch
j-carl: > This is not git format-patch output, why not? The provenance of this change aligns with https://github.com/jbeder/yaml-cpp/pull/1310 so you should be documenting that...

I was not aware that I should use `git format-patch`, so I generated just a plain patch from `git diff`.
I changed the file.

> You should probably just ask upstream to release a new version though.

There are already couple issues open which are asking for a new release: https://github.com/jbeder/yaml-cpp/issues/1360, https://github.com/jbeder/yaml-cpp/issues/1347, https://github.com/jbeder/yaml-cpp/issues/1335

Until that happens I want to release the library with one patch, that allows to build it.

j-carl:
Is it possible to have a wrap with `diff_files`?

jpakkane:
No. The policy is to only add build files, no code changes. We don't want (or even can) take on the maintenance burden those bring.

PR #2115, opened 2025-05-19
azure-sdk-for-cpp: New wrap

WillAyd:


WillAyd:
This is another third party dependency of the Apache Arrow project. This implementation does not cover the entire Azure SDK, but only the components required by Apache Arrow

ci_config.json
```diff
@@ -43,6 +43,18 @@
       "boost"
     ]
   },
+  "azure-sdk-for-cpp": {
+    "build_on": {
+      "msys2": false,
+      "windows": false
```
WillAyd, line 49: There seems to be some kind of issue with missing symbols on Windows. I can't tell if it has to do with the export macros or maybe even some missing Windows libraries that are needed.

For now I've just punted on it, in hopes that someone with more Windows knowledge and interest can fix this at a later date

PR #2118, opened 2025-05-21
freeglut: require GL/glu.h

kusma:
Freeglut requires GL/glu.h, which isn't always installed. Make sure we check for it, so we can get a meaningful error message early on instead of just failing to compile.

Because not all platforms has glu as a dependency, we also need to fall back to searching for the library. This search is copied from Mesa's demos project.


kusma:
Seems like the MacOS build issue is a github-actions issue: https://github.com/actions/runner-images/issues/9966

kusma:
Yeah, testing this locally works fine...

releases.json
```diff
@@ -931,6 +931,7 @@
       "glut"
     ],
     "versions": [
+      "3.4.0-4",
```
paulmenzel, line 934: Could be a separate commit?

paulmenzel (approved):
Thank you for coming up with a fix for my issue https://gitlab.freedesktop.org/mesa/demos/-/issues/50. The diff looks good to me.

releases.json
kusma: It doesn't seem like that's the norm from the commit history, is it?

.github/workflows/sanity_checks.yml
```diff
@@ -188,6 +188,11 @@ jobs:
           brew install ninja
           python3 -m pip install --pre meson
 
+      - name: Work around runner-images issue 9966
+        run: |
+          brew unlink python@3.11
+          brew link --overwrite python@3.11
```
bgilbert, line 194: I've been using this code for a while:

```bash
# https://github.com/actions/runner-images/issues/11545
# https://github.com/q3aiml/ledger/commit/f53b35ae
brew list -1 | grep python@ | while read formula; do
    brew unlink $formula
    brew link --overwrite $formula
done
```

subprojects/packagefiles/freeglut/meson.build
```diff
@@ -67,12 +67,29 @@ freeglut_sources = files(
   'src/fg_window.c',
 )
 
+dep_glu = dependency(
+  'glu',
+  required: false,
+)
+# GLU is part of OpenGL.Framework
+if not dep_glu.found() and host_machine.system() != 'darwin'
```
bgilbert, line 75: We don't require the dependency on macOS, then?

kusma: What do you mean? Just because GLU is part of OpenGL.Framework doesn't mean Freeglut isn't useful on macOS, no?

kusma: Did you perhaps mix up GLU and GLUT?

bgilbert: Why does this check include `and host_machine.system() != 'darwin'`?  The `dependency()` call specifies `required: false`, and on macOS we don't do the followup `cc.find_library()` call with implicit `required: true`, so the wrap ends up requiring GLU except on macOS.

.github/workflows/sanity_checks.yml
```diff
@@ -196,6 +196,10 @@ jobs:
           python3 -m pip install license-expression
           python3 -m pip install --pre meson
 
+      - name: Work around runner-images issue 9966
+        run: |
+          brew list -1 | grep python | while read formula; do brew unlink $formula; brew link --overwrite $formula; done
```
bgilbert: This seems a bit long as a one-liner.  Any reason not to keep the original formatting?

subprojects/packagefiles/freeglut/meson.build
kusma: GLU is always required to build FreeGLUT, on all platforms.

I don't have a mac system at hand to test with, but my understanding is that on macOS, dep_glu will not be found, but because it doesn't have `disabler: true`, including it in `freeglut_deps` is benign. And on macOS, GLU will be available through `dependency('gl')`, which is already part of  `freeglut_deps`.

So I'm not really sure what you think the problem is here. GLU will be found, just through another dependency on macOS.

.github/workflows/sanity_checks.yml
kusma: Just because that's what it seemed others were doing. I don't know YAML and formatting details well enough to know if the multi-line version would work or not without testing both.

I suppose I can try and see if it helps?

subprojects/packagefiles/freeglut/meson.build
bgilbert: > And on macOS, GLU will be available through `dependency('gl')`

That answers it, thanks.  Might be good to mention that in a comment, since it's not clear by inspection.

kusma: OK, finally got around to updating the comment, sorry for the delay!

.github/workflows/sanity_checks.yml
kusma: updated, hope it works :)

kusma:
The alpine failures seems unrelated to this MR...

neheb:
cherry picked into https://github.com/mesonbuild/wrapdb/pull/2466

PR #2123, opened 2025-05-22
sqlitecpp: update from 3.3.2 to 3.3.3

bgilbert:


neheb:
another problem with gtest: https://github.com/SRombauts/SQLiteCpp/pull/504

PR #2125, opened 2025-05-22
Include root project directory in dependencies

kompote:
Libwebsockets : 

The compilation process fails due to the inability from the parent project to locate the generated config headers, even though they are present in the build directory.

To resolve this issue, the include path dependency needs to be adjusted to incorporate the project root path, as the build directory mirrors the subproject structure.


subprojects/packagefiles/libwebsockets/meson.build
```diff
@@ -907,7 +907,9 @@ libwebsockets = library(
   ],
 )
 
-depinc = include_directories('include')
+depinc = []
+depinc += include_directories('include')
+depinc += include_directories('.')
```
bgilbert, line 912: It'd be a bit cleaner to do:
```suggestion
depinc = include_directories('include', '.')
```

neheb:
Fixed in https://github.com/mesonbuild/wrapdb/pull/2465

PR #2126, opened 2025-05-24
update hinnant-date to 3.0.3

Habbie:
upstream grew a new cmake option ENABLE_DATE_INSTALL (defaults to ON). I don't believe meson wrap users will care, so I did not add that to the options in packagefiles.


bgilbert:
The `project()` call in `meson.build` doesn't specify the project version, so technically there's nothing to update there, but it'd be good to add the version number there if you're up for it.

Habbie:
sure! 3.0.3 or 3.0.3-1 in there?

Habbie:
other packagefiles appear to not have the -1 suffix

PR #2127, opened 2025-05-24
MinifyCpp: add to wrapdb

QWERTIOX:


QWERTIOX:
* MSYS2 (CLANG64) I already knew of this bug but i couldn't reproduce it on another system until now
* VisualStudio (all) I have no idea why it can't link i don't use it myself

neheb:
```
LINK : fatal error LNK1181: cannot open input file 'subprojects\minifycpp\MinifyCpp.lib'
```

dllexport is needed.

QWERTIOX:
> ```
> LINK : fatal error LNK1181: cannot open input file 'subprojects\minifycpp\MinifyCpp.lib'
> ```
> 
> dllexport is needed.

Can you give me some reference link, the  only thing i found is https://mesonbuild.com/Reference-tables.html#msvc-__declsp and it doesn't say a lot to me

QWERTIOX:
Force pushed without staging...
I don't like how windows made my code look awful

neheb:
yeah it's not nice.

What's more, passing meson -Ddefault_library=static will throw a bunch of warnings saying dllexport and dllimport is wrong.

QWERTIOX:
> yeah it's not nice.
> 
> What's more, passing meson -Ddefault_library=static will throw a bunch of warnings saying dllexport and dllimport is wrong.

Should i add check in meson.build and use arguments to not add them when static linking?

neheb:
Sure. This is the CMake autogenerated thing that gets used:

```
#ifdef exiv2lib_STATIC
#  define EXIV2API
#  define EXIV2LIB_NO_EXPORT
#else
#  ifndef EXIV2API
#    ifdef exiv2lib_EXPORTS
        /* We are building this library */
#      define EXIV2API __attribute__((visibility("default")))
#    else
	/* We are using this library */
#      define EXIV2API __attribute__((visibility("default")))
#    endif
#  endif
#endif
```

For the exiv2 meson build, I handle this all in there and don't generate this file

https://github.com/Exiv2/exiv2/blob/main/meson.build#L10
https://github.com/Exiv2/exiv2/blob/main/src/meson.build#L114

QWERTIOX:
> For the exiv2 meson build, I handle this all in there and don't generate this file
> 
> https://github.com/Exiv2/exiv2/blob/main/meson.build#L10 https://github.com/Exiv2/exiv2/blob/main/src/meson.build#L114

Does this still work for default_library=both?

neheb:
I have no idea.

QWERTIOX:
> I have no idea.

To avoid issues in future i just allowed to set if build is static/shared in option

QWERTIOX:
It's good to merge @neheb

PR #2131, opened 2025-05-25
SDL3_image functions are currently failing to link when building with meson due to IMG.c missing

xoSauce:
IMG_Load(...), IMG_LoadTexture(...) since they are defined in IMG.c


bgilbert:
macOS is failing to link:

```
Undefined symbols for architecture arm64:
  "_IMG_Load", referenced from:
      _IMG_LoadTexture in src_IMG.c.o
  "_IMG_LoadGIF_IO", referenced from:
      _supported in src_IMG.c.o
  "_IMG_LoadTGA_IO", referenced from:
      _supported in src_IMG.c.o
  "_IMG_LoadTIF_IO", referenced from:
      _supported in src_IMG.c.o
  "_IMG_isGIF", referenced from:
      _supported in src_IMG.c.o
      _supported_anims in src_IMG.c.o
  "_IMG_isTIF", referenced from:
      _supported in src_IMG.c.o
```

I have no idea why.

neheb:
Might be able to reproduce on Linux with lld instead of bfd.

Speaking of which, no such CI is present. I wonder if the alpine ones should be modified to do so.

tux314159:
Hello! Recently encountered this, on Apple you need to include the source file IMG_ImageIO.m, enable objc and link in the apple libraries (see how the SDL3 wrap does it). I also have a working patch I can send when I get back to a computer if you'd like.

neheb:
Yes

PR #2136, opened 2025-05-30
libgrapheme: new wrap

Gottox:


subprojects/packagefiles/libgrapheme/benchmark/meson.build
```diff
@@ -0,0 +1,28 @@
...
+  'utf8-decode.c',
+  'word.c',
+)
+
+libutf8proc_dep = dependency('libutf8proc')
+
+bench_c_flags = ['-Wno-overlength-strings']
+
+foreach bench_src : bench_sources
+  exe_name = bench_src.full_path().split('/')[-1].split('.')[0]
```
jpakkane: This fails because on Windows the separator is a backslash. It's better to list the executable names and then form the source file by appending `.c` to it.

Gottox: Fixed. There's another issue with windows systems: The generator executables in `gen/` expect to run in the project root. `custom_target` has no option to set the cwd, so I'm forced to build a small wrapper script, that changes the directory. Are there any tips how to handle that?

bgilbert: You can write the wrapper in Python, wrap it in `find_program()` or `files()` when invoking it from `custom_target()`, and add it to `PER_PROJECT_PERMITTED_FILES`.

subprojects/packagefiles/libgrapheme/meson.build
```diff
@@ -0,0 +1,75 @@
...
+  'libgrapheme',
+  'c',
+  version: '2.0.2',
+  default_options: [
+    'c_std=c99',
+    'warning_level=3',
+    'b_lundef=false',
+    'buildtype=minsize',
+  ],
+  meson_version: '>=1.4.0',
```
bgilbert: At least according to Meson's warnings (which aren't always complete) this can be `>=0.57.0`.

subprojects/packagefiles/libgrapheme/benchmark/meson.build
Gottox: The last part was missing, thanks! :)

Gottox:
libgrapheme seems to be broken on Windows systems at the moment.

bgilbert:
Yeah, looks like that's been fixed in Git, so I think it's okay to ignore the Windows failures for this release.  Meanwhile, it'd be good to define `_CRT_SECURE_NO_WARNINGS` on Windows for future use.

subprojects/packagefiles/libgrapheme/meson.build
```diff
@@ -0,0 +1,78 @@
...
+  'c',
+  version: '2.0.2',
+  default_options: [
+    'c_std=c99',
+    'warning_level=3',
+    'b_lundef=false',
+    'buildtype=minsize',
+  ],
+  meson_version: '>=0.57.0',
+)
```
bgilbert, line 13: License please.

subprojects/packagefiles/libgrapheme/meson.build
```diff
@@ -0,0 +1,78 @@
...
+  install: true,
+  version: version,
+  soversion: soversion,
+  c_args: lib_c_args,
+  link_args: lib_link_args,
+)
+
+install_headers(
+  'grapheme.h',
+  subdir: '.',
```
bgilbert: Is this line redundant?

subprojects/packagefiles/libgrapheme/benchmark/meson.build
```diff
@@ -0,0 +1,21 @@
...
+
+bench_c_flags = ['-Wno-overlength-strings']
+
+foreach exe_name : benches
+  src = exe_name + '.c'
+  exe = executable(
+    'bench-' + exe_name,
+    [src, gen_headers, 'util.c'],
+    dependencies: [libutf8proc_dep, libgrapheme_dep],
+    install: false,
```
bgilbert: Nit: this is the default.

subprojects/packagefiles/libgrapheme/test/meson.build
```diff
@@ -0,0 +1,27 @@
...
+  'case',
+  'character',
+  'line',
+  'sentence',
+  'utf8-decode',
+  'utf8-encode',
+  'word',
+]
+
+test_c_flags = ['-Wno-overlength-strings']
```
bgilbert:
```suggestion
test_c_flags = cc.get_supported_arguments('-Wno-overlength-strings')
```

subprojects/packagefiles/libgrapheme/meson.build
```diff
@@ -0,0 +1,78 @@
...
+soversion = '@0@.@1@'.format(version_split[0], version_split[1])
+lib_c_args = []
+lib_link_args = []
+
+# To produce the same binary as with the Makefile approach, we need to
+# add the -ffreestanding and -nostdlib flags
+if get_option('b_lundef') == false and cc.has_argument('-ffreestanding') and host_machine.system() != 'darwin'
+
+  lib_c_args += ['-ffreestanding']
+  lib_link_args += ['-nostdlib']
```
bgilbert, line 42: These are unusual arguments for a desktop build.  I don't fully understand the consequences, though GCC docs say that `-nostdlib` usually wants at least `-lgcc`.  Willing to defer to upstream here, just wondering if this could cause issues.

Gottox: I'm adding it as an option.

subprojects/packagefiles/libgrapheme/benchmark/meson.build
bgilbert: Note that there are a few more of these.

Gottox:
Also, I added an entry to `ci_config.json` to not test on windows for now.

bgilbert:
That won't do what you want; it expects to see a "not supported" message in the logs.  Since we're expecting the next release to work, you can revert the `ci_config` change and we'll merge with failing CI.

bgilbert:
There's another `-Wno-overlength-strings` that needs guarding while you're at it.

bgilbert:
(`ci_config` is still there in the last push, if that wasn't what you intended.)

Gottox:
Thanks for the catch, yea, forgot to `git add` the file.

bgilbert:
Thanks!

PR #2141, opened 2025-06-02
ci: `apt-get update` before installing packages

### LINKED ISSUE #2139: Website is not updating
Actions to update the website fails:
https://github.com/mesonbuild/wrapdb/actions/runs/15380110654/job/43269711254
https://github.com/mesonbuild/wrapdb/actions/runs/15361488180/job/43229150317
https://github.com/mesonbuild/wrapdb/actions/runs/15360896105/job/43227889348
https://github.com/mesonbuild/wrapdb/actions/runs/15338405003/job/43159863123
https://github.com/mesonbuild/wrapdb/actions/runs/15338402139/job/43159852181
https://github.com/mesonbuild/wrapdb/actions/runs/15330497346/job/43135827121
Last that passed:
https://github.com/mesonbuild/wrapdb/actions/runs/15259705144/job/42915080341

bgilbert:
Don't fail to create a release or update the website if any dependency packages have aged out of the mirror since the GitHub runner image was built.  The first case only affects packages with generators (OpenSSL), so have `create_release.py` do it when needed rather than unconditionally doing it from the job YAML.

Fixes #2139.

### DIFF
--- a/tools/create_release.py
+++ b/tools/create_release.py
@@ -70,6 +70,7 @@ class CreateRelease:
             debian_packages = ci.get(self.name, {}).get('debian_packages', [])
             if debian_packages and is_debianlike():
                 if is_ci():
+                    subprocess.check_call(['sudo', 'apt-get', 'update'])
                     subprocess.check_call(['sudo', 'apt-get', '-y', 'install'] + debian_packages)
                 else:
                     s = ', '.join(debian_packages)


QWERTIOX:
Does we really need to install build deps every time we want to update the website? Wasn't be easier to create separate docker image with everything installed?

bgilbert:
We could, but then we'd need another workflow to periodically update the container image, which hardly seems worth the complexity.  The job doesn't run very often, and the Ubuntu packages are fetched from a local mirror inside Azure, so there isn't a lot of resource consumption.  (The PyPI packages are fetched directly from PyPI, unfortunately, but we could configure `setup-python` to enable caching to reduce the load on PyPI.)

bgilbert:
`setup-python` doesn't have a way to directly specify the cache key, though, which makes it awkward to maintain distinct caches in multiple jobs.  It might be better to invoke `actions/cache` directly.

bgilbert:
PyPI packages are fronted by Fastly (at no cost to PyPI) so presumably we're not causing a lot of load there either.  Regular CI jobs only fetch Meson (~1 MB) and the website update job looks like it fetches about 10 MB.  I don't think it's worth adding additional caching for now.

bgilbert:
I'll merge this tomorrow to fix the website build unless there are objections.

eli-schwartz:
> Does we really need to install build deps every time we want to update the website? Wasn't be easier to create separate docker image with everything installed?

Right, we do this for the Meson CI but that's because:
- it can take quite a bit of time to build those containers. Especially since we have to compile parts of it from source.
- we need to be able to tell when failures are caused by changes to meson code, as opposed to changes in the distro -- in the latter case, the container just doesn't get built that week.

I don't think it's worth it here for apt packages alone. Using the GitHub cache is faster than reaching out all the way to fastly, so it has some value -- and also means we can cache locally compiled wheels in case PyPI only has an sdist.

eli-schwartz (approved):

eli-schwartz:
Note: wheezy.template and hotdoc both get built from sdist in the most recently passing website update. So caching pip would allow reusing the precompiled wheels for both.

PR #2143, opened 2025-06-04
libserialport: Add version 0.1.2-1.

antoniovazquezblanco:


subprojects/packagefiles/libserialport/meson.build
```diff
@@ -0,0 +1,48 @@
+project(
+  'libserialport',
+  'c',
+  version: '0.1.2',
+  license: 'LGPL3+',
```
bgilbert: SPDX:

```suggestion
  license: 'LGPL-3.0-or-later',
```

subprojects/packagefiles/libserialport/meson.build
```diff
@@ -0,0 +1,48 @@
...
+  libserialport_cflags += '-DLIBSERIALPORT_ATBUILD'
+endif
+
+libserialport_lib = library(
+  'libserialport',
+  libserialport_sources,
+  c_args: libserialport_cflags,
+  link_args: libserialport_ldflags,
+  include_directories: libserialport_includes,
+  version: meson.project_version(),
```
bgilbert: Upstream uses a libtool version distinct from the project version, i.e. `SP_LIB_VERSION_STRING` is `1:0:1` which corresponds to 0.1.0.

subprojects/packagefiles/libserialport/config.h
bgilbert, line 1: This file should be generated with `configure_file()`, not hardcoded in the repo.  Not every define in the upstream `config.h.in` is actually used, but some are and we should perform the appropriate feature tests.

antoniovazquezblanco: Thanks @bgilbert for the pointers. Currently looking into this. Will try to update soon.

Thanks! :)

antoniovazquezblanco: Now compiling. Thanks for the pointers! :D

bgilbert: The feature checks are still missing, and they generally look as though they'd be straightforward to implement.  I think it'd be better to match upstream's checks as much as possible (especially since upstream doesn't have functional tests) unless you have a good argument for why they aren't necessary.

antoniovazquezblanco:
Got some time and got back to this. I hope it is a little bit better now.

subprojects/packagefiles/libserialport/meson.build
```diff
@@ -0,0 +1,176 @@
...
+pkg.generate(libserialport_lib)
+
+install_headers(libserialport_headers)
+
+libserialport_dep = declare_dependency(
+  include_directories: libserialport_includes,
+  link_with: libserialport_lib,
+)
+
+if meson.version().version_compare('>=0.54.0')
```
neheb: I'd just bump the minimum version. 0.49 is really old.

antoniovazquezblanco: Done!

antoniovazquezblanco:
Thanks! :D

PR #2145, opened 2025-06-07
rdkafka: update to 2.10.0

neheb:


bgilbert (approved):
Looks like there were a couple tests added, `0149-broker-same-host-port.c` and `0151-purge-brokers.c`, that we could add here.

PR #2147, opened 2025-06-07
add godot-4.4.1

v-for-vandal:
Update godot-cpp to version 4.4.1


v-for-vandal:
I checked it locally, seems to be working. I think it is safe to merge.
@bgilbert

subprojects/godot-cpp.wrap
```diff
@@ -1,8 +1,8 @@
 [wrap-file]
-directory = godot-cpp-godot-4.3-stable
-source_url = https://github.com/godotengine/godot-cpp/archive/refs/tags/godot-4.3-stable.tar.gz
-source_filename = godot-cpp-4.3.tar.gz
-source_hash = e6dfe8ee1ab3eeaadc9dba2dcb5ef39c4a99635986148fa893ce5c414153743f
+directory = godot-cpp-godot-4.4.1-stable
+source_url = https://github.com/godotengine/godot-cpp/archive/refs/tags/godot-4.4.1-stable.tar.gz
+source_filename = godot-4.4.1-stable.tar.gz
```
bgilbert: I intentionally changed this during the mass `source_filename` rename.  The `-stable` suffix is odd but might be okay if upstream called the tarball by this name.  But it doesn't; the file actually downloads as `godot-cpp-godot-4.4.1-stable.tar.gz`.  (Try it in a web browser.)  So let's stick with the `wrapname-version` format, `godot-cpp-4.4.1.tar.gz`; it's clearer.

subprojects/packagefiles/godot-cpp/meson.build
```diff
@@ -4,8 +4,9 @@
 project(
   'godot-cpp',
   'cpp',
-  version: '4.3',
+  version: '4.4.1',
   default_options: ['cpp_std=c++17'],
+  meson_version: '>=1.1',
```
bgilbert: Adding a `meson_version` is a good idea, but the 1.1 requirement is more restrictive than necessary.  It looks like this can be reduced to 0.58.0 if `meson.options` is renamed to `meson_options.txt`.

v-for-vandal: I can do it, if you want, but there seems to be no point. It doesn't look like anyone with old meson is using this wrap. It doesn't even seem that anyone except me is using it at all :(

bgilbert: Well, the value of putting the wrap into WrapDB is that others can use it if they want, so let's aim for the broader compat window.

bgilbert:
CI will want to see `godot-cpp-4.4.1.tar.gz`.  Also, please squash your changes.

subprojects/packagefiles/godot-cpp/meson.build
bgilbert: And thanks for maintaining it here, by the way.  This is a non-trivial wrap and &mdash; if nothing else! &mdash; it helps the ecosystem to have it available here for anyone to easily use in the future.

tools/sanity_checks.py
```diff
@@ -146,6 +146,7 @@
 SOURCE_FILENAME_PREFIXES = {
     'icu': 'icu4c',
     'libtomcrypt': 'crypt',
+    'godot-cpp' : 'godot',
```
bgilbert: This isn't needed anymore.

subprojects/packagefiles/godot-cpp/meson.build
```diff
@@ -4,8 +4,10 @@
 project(
   'godot-cpp',
   'cpp',
-  version: '4.3',
+  license = 'MIT',
```
bgilbert:
```suggestion
  license: 'MIT',
```

tools/sanity_checks.py
v-for-vandal: fixed

subprojects/packagefiles/godot-cpp/meson.build
v-for-vandal: fixed

v-for-vandal:
I would like to leave a small instruction on how to update this package, because at the moment, every 5-6 months I come back to update it and have to remember how it all works. What is a good place to do it?  meson.build? godot.wrap?

bgilbert:
`meson.build`, I think.

v-for-vandal:
@bgilbert all done, I believe :)

bgilbert:
Thanks!

PR #2149, opened 2025-06-08
Add legibility requirements for `source_filename`

bgilbert:
There are no technical requirements on `source_filename` other than the file extension, so wraps have historically used a variety of different filename formats.  Intuitively the filename should match upstream's name for its source archive.  However, many archives are autogenerated by Git forges, and while e.g. GitHub specifies a filename in the `Content-Disposition` HTTP response header, that name isn't reflected in `source_url`.  This has caused wraps to use filenames like `v1.7.18.tar.gz`, which has two problems:

- Anyone looking at the `packagecache` directory won't know what that file is, or whether it's still needed.
- If two subprojects have different archives called `v1.0.0.tar.gz`, the second one will fail to unpack on a hash mismatch.

Require `source_filename` to be legible.  `directory` (the name of the top-level directory inside the source archive) provides a good, though not perfect, guide to what the archive should be called.  Instead of adding complex rules to permit various existing cases (I tried that first), make the rules fairly restrictive and fix wraps as needed.  Specifically, require `source_filename` minus file extensions to match one of the following templates:

- `<directory>`, the common and expected case.
- `<directory>-<version>`, for Codeberg archives, whose `directory` doesn't include the version.
- `<wrap-name>-<version>`, for cases where `directory` is not legible.  For example, godot-cpp fetches a tag called `godot-4.3-stable` from a GitHub repo called `godot-cpp`, causing GitHub to produce a `directory` called `godot-cpp-godot-4.3-stable`.
- An explicit override defined in `SOURCE_FILENAME_PREFIXES`, for when upstream specifies a filename but that filename includes something unexpected.


subprojects/emilk-loguru.wrap
```diff
@@ -1,7 +1,7 @@
 [wrap-file]
 directory = loguru-2.1.0
 source_url = https://github.com/emilk/loguru/archive/v2.1.0.tar.gz
-source_filename = loguru_2.1.0.tar.gz
+source_filename = loguru-2.1.0.tar.gz
```
eli-schwartz, line 4: I'm fine with these renamings (no real benefit other than easier linting, but also no real harm), but do we really need to bump the wrap revision for cases where the name was already unambiguous?

bgilbert:
Okay, I've kept the revision bumps where the project name was missing or actively incorrect, and dropped them otherwise.

PR #2150, opened 2025-06-08
Update nanoarrow to 0.7.0

paleolimbot:
Like https://github.com/mesonbuild/wrapdb/pull/1741 ...this is a check before the release of nanoarrow 0.7.0.


releases.json
```diff
@@ -2708,7 +2708,8 @@
     ],
     "versions": [
       "0.6.0-1",
-      "0.5.0-1"
+      "0.5.0-1",
+      "fd2263bbeb7c6d330fbfaf556fa1997aa1091179-1"
```
WillAyd: Based off of the CI failures you might need to put this first in the list instead of at the end

ci_config.json
```diff
@@ -894,20 +894,11 @@
...
-    ],
     "build_options": [
-      "nanoarrow:tests=true",
-      "nanoarrow:ipc=true",
-      "nanoarrow:device=true",
-      "nanoarrow:testing=true"
+      "nanoarrow:tests=enabled",
+      "nanoarrow:ipc=enabled",
+      "nanoarrow:device=enabled",
+      "nanoarrow:testing=enabled"
```
WillAyd, line 951: Since we are installing apache arrow, we may also want to enabled `nanoarrow:tests_with_arrow=enabled` here.

Whenever https://github.com/apache/arrow-nanoarrow/pull/763 happens we could simplify this with just `nanoarrow:auto_features=enabled`, but I don't think that is a blocker for this release

paleolimbot: I removed the part where arrow is installed since the install itself fails on some platforms (the tests with arrow are more like integration tests and I don't think they're necessary here).

releases.json
paleolimbot: Ah thanks! I got a different error about the order of the list when I put it at the start 😢  (but all the tests still run)

WillAyd:
Strange - I wonder if the following warning in the culprit:

```
LINK : warning LNK4098: defaultlib 'MSVCRTD' conflicts with use of other libs; use /NODEFAULTLIB:library
```

WillAyd:
I wonder if the following code is suspicious:

```
        flatcc_dep = dependency(
            'flatcc',
            default_options: ['default_library=static'],
        )
```

In the main project call's default_options we specify `buildtype=release` but when pulling in this dependency we don't say anything about the build type. So maybe Meson is not merging them and creating a debug build of flatcc to be linked into the ipc target

WillAyd:
Ah...so in our CI we require meson<1.8.0, where a lot of updates to subproject object propogation were made, but we only enforce that during CI dependency installation and not in the meson.build configuration itself. It looks like the wrapdb is using 1.8.2, so there's also a gap in our coverage from that

WillAyd:
For debugging it might be helpful to temporarily add verbose messaging so we can pinpoint where the debug library is getting pulled in:

```
modified   tools/sanity_checks.py
@@ -491,7 +491,7 @@ class TestReleases(unittest.TestCase):
                         print('cannot verify in wrapdb due to missing dependency')
                         return
             raise Exception(f'Wrap {name} failed to configure due to bugs in the wrap, rather than due to being unsupported')
-        subprocess.check_call(['meson', 'compile', '-C', builddir], env=meson_env)
+        subprocess.check_call(['meson', 'compile', '-C', builddir, '--verbose'], env=meson_env)
         if not ci.get('skip_tests', False):
             test_options = ci.get('test_options', [])
             if self.timeout_multiplier != 1:
```

WillAyd:
I wonder if there's something with Meson's default_options handling that we are missing (potentially so with 1.8x)

@eli-schwartz in our arrow-nanoarrow project we set a `default_option` of `buildtype=release`. When compiling the arrow-nanoarrow project directly, it appears that that buildtype is propogated to all subprojects. 

However, looking at the verbose Ubuntu logs here, I see many of the subproject dependencies for nanoarrow are being built with `-O0 -g`, which I suppose makes sense if you consider nanoarrow just to be another subproject and not the main project you are compiling.

Is it a mistake for us to set `buildtype=release` in the nanoarrow config?

eli-schwartz:
The last release included an extremely large option refactoring that has had a number of regressions. There are probably more regressions.

If the code worked before and it's broken now then the first thing you should look at is probably what did meson do wrong. 

I think this is one of the known outstanding issues, for which a draft fix exists.

WillAyd:
Hard to say if it worked before or not. We weren't testing Windows previously (where mixing debug/release builds seems to be a failure), so its possible we've had this issue all along.

When building nanoarrow as the main project, I suppose it makes sense for the `default_option` of `buildtype` to propogate down to subprojects, and when I try with Meson 1.8.2 locally I still see that behavior. But I don't think that option should propogate to other subprojects if nanoarrow is built as one, right?

eli-schwartz:
To be clear, the failure is about b_vscrt, not `CXXFLAGS="-g"`, right?

WillAyd:
Not entirely sure. All of the tests fail on Windows, and in the logs there we see the warning of:

```
LINK : warning LNK4098: defaultlib 'MSVCRTD' conflicts with use of other libs; use /NODEFAULTLIB:library
```

which from my (possibly flawed) research indicates that we are mixing debug and non-debug libraries together, and Windows doesn't like that (?)

WillAyd:
I am hopeful some recent updates in nanoarrow will resolve the remaining failures. Here's a patch that should get us up to speed

```
From 3a7558019d109f0bb05dcb104ac2d405e8e8536f Mon Sep 17 00:00:00 2001
From: Will Ayd <william.ayd@icloud.com>
Date: Thu, 19 Jun 2025 13:14:50 -0400
Subject: [PATCH] bump to newer nanoarrow version

---
 releases.json              | 2 +-
 subprojects/nanoarrow.wrap | 8 ++++----
 2 files changed, 5 insertions(+), 5 deletions(-)

diff --git a/releases.json b/releases.json
index 4238402f..e9891228 100644
--- a/releases.json
+++ b/releases.json
@@ -2707,7 +2707,7 @@
       "nanoarrow"
     ],
     "versions": [
-      "14904f7933fbd31b5cd1761a33f12ae148aaaeac-1",
+      "5284a6c73ded1b94d4d2586023546e6edad1db7a-1",
       "0.6.0-1",
       "0.5.0-1"
     ]
diff --git a/subprojects/nanoarrow.wrap b/subprojects/nanoarrow.wrap
index 6aa1bb8f..2d5bfbd0 100644
--- a/subprojects/nanoarrow.wrap
+++ b/subprojects/nanoarrow.wrap
@@ -1,8 +1,8 @@
 [wrap-file]
-directory = arrow-nanoarrow-14904f7933fbd31b5cd1761a33f12ae148aaaeac
-source_url = https://github.com/paleolimbot/arrow-nanoarrow/archive/14904f7933fbd31b5cd1761a33f12ae148aaaeac.zip
-source_filename = arrow-nanoarrow-14904f7933fbd31b5cd1761a33f12ae148aaaeac.zip
-source_hash = 56fe777ee05b3fef9adaf29e172706fbc001a25782a53805d7f3e495fcbb704e
+directory = arrow-nanoarrow-5284a6c73ded1b94d4d2586023546e6edad1db7a
+source_url = https://github.com/paleolimbot/arrow-nanoarrow/archive/5284a6c73ded1b94d4d2586023546e6edad1db7a.tar.gz
+source_filename = arrow-nanoarrow-5284a6c73ded1b94d4d2586023546e6edad1db7a.tar.gz
+source_hash = 840527e565ffd808ac0944750c11996248b609efa3ecdfa7de8578982b9ebb98
 
 [provide]
 nanoarrow = nanoarrow_dep
-- 
2.43.0
```

bgilbert:
If nanoarrow is built as a subproject, its `default_options: ['buildtype=x']` is [supposed](https://mesonbuild.com/Reference-manual_functions.html#project_default_options) to be ignored entirely.  Meson 1.8+ incorrectly propagates the subproject's value through to `get_option('buildtype')` in the subproject (https://github.com/mesonbuild/meson/issues/14729), but that doesn't directly affect compiler options like `-O3` and `-g`.  However, `b_vscrt` defaults to `from_buildtype`, and I would guess that the incorrect `buildtype` value is thus producing your problem.

In any case, I notice that you're manually checking `get_option('buildtype')` elsewhere in the build config.  Not only can that be wrong in 1.8.x, it can be wrong in previous releases (https://github.com/mesonbuild/meson/issues/11645), so you may want to inspect a different option like `debug` instead.

bgilbert:
According to https://github.com/mesonbuild/meson/issues/14729#issuecomment-2990390082 `buildtype` is now supposed to be overridable in the subproject, though presumably its failure to affect the subproject cflags is a bug.

WillAyd:
Thanks @bgilbert - we already changed the `Dbuildtype=release` setting upstream. I'll take a look at the get_option('buildtype') == 'debug' checks as well

WillAyd (approved):

PR #1914, opened 2025-02-09
libffi: update to 3.4.8

bgilbert:

PR #2152, opened 2025-06-09
libffi: update from 3.4.8 to 3.5.0

bgilbert:


bgilbert:
Mostly the same CI failures as https://github.com/mesonbuild/wrapdb/pull/1914.  The MSVC ones are new, but 3.4.8 now also produces those failures, so it looks like a toolchain change rather than an upstream one.  This seems good to merge.

---

MSVC x64, MSVC x86, clang-cl:

```
334/334 libffi:closures / unwindtest_ffi_call         TIMEOUT        30.09s   exit status 1
```

macOS arm64:

```
481/720 libffi:closures / closure_loc_fn0 -O0              FAIL            0.01s   killed by signal 6 SIGABRT
482/720 libffi:closures / closure_loc_fn0 -O2              FAIL            0.01s   killed by signal 6 SIGABRT
```

Tons of Alpine riscv64 failures.

bgilbert:
Same failures in 3.5.1.

PR #2154, opened 2025-06-09
libarchive: update from 3.7.9 to 3.8.1

bgilbert:
Add support for Windows bcrypt library.


neheb:
isn't it mutually exclusive with openssl?

bgilbert:
Apparently not: https://github.com/libarchive/libarchive/pull/2538.  bcrypt doesn't support RIPEMD-160.

PR #2157, opened 2025-06-09
libusb: update from 1.0.28 to 1.0.29

bgilbert:


bgilbert:
Historically riscv64 was timing out.  Now that we have longer timeouts, tests are dying on SIGTERM.  Not sure why, but it seems unlikely to be a real regression and probably shouldn't block the update if it is.

PR #2158, opened 2025-06-10
Update Xorg libraries

bgilbert:


neheb:
All of these need some real love. I only added barebones versions to get rid of ci_config.json entries

neheb:
oh upstream already has meson.build files.

bgilbert:
Yeah, all of them except libxext have unreleased Meson configs in Git.

PR #2159, opened 2025-06-10
glbinding: update to 3.5.0

neheb:
Update standard to C++17 to fix gtest compilation.


neheb:
```
lld-link: error: undefined symbol: public: __cdecl glbinding::Boolean8::Boolean8(unsigned char)
```

weird.

neheb:
OK I think the problem is https://github.com/cginternals/glbinding/blob/master/source/glbinding/include/glbinding/gl/types.h#L13

I needs to be WIN32.

PR #2161, opened 2025-06-10
lmdb: update to 0.9.33

neheb:


neheb:
this looks like QEMU bugs.

subprojects/packagefiles/lmdb/libraries/liblmdb/meson.build
```diff
@@ -30,27 +22,31 @@ lmdb_dep = declare_dependency(
...
+if not cc.has_header(
+  'unistd.h',
+  required: get_option('programs'),
+)
+  subdir_done()
 endif
 
-if get_option('tests')
-  tests = ['mtest', 'mtest2', 'mtest3', 'mtest4', 'mtest5']
+progs = ['mdb_stat', 'mdb_copy', 'mdb_dump', 'mdb_load']
```
bgilbert: `mplay` is new, probably worth adding.

ci_config.json
```diff
@@ -827,13 +827,6 @@
       "libx11-dev"
     ]
   },
-  "lmdb": {
-    "_comment": "Tests and programs disabled by default",
-    "build_options": [
-      "lmdb:tests=true",
-      "lmdb:programs=true"
-    ]
-  },
```
bgilbert, line 836: Might be worth adding support for `skip_program_check`, like `skip_dependency_check`, so we don't have to manually ignore failures on Visual Studio.

subprojects/packagefiles/lmdb/libraries/liblmdb/meson.build
```diff
@@ -1,17 +1,9 @@
...
 
-src = ['mdb.c', 'midl.c']
+src = files('mdb.c', 'midl.c')
 
-c_args = ['-Wno-unused-parameter', '-Wbad-function-cast', '-Wuninitialized']
-add_project_arguments(
-  cc.get_supported_arguments(c_args),
-  language: 'c',
-)
+threads_dep = dependency('threads')
```
bgilbert: Since lmdb understands Windows threading, this should be skipped on Windows.  Otherwise MinGW will add a spurious pthreads dependency.

...oh no, Meson has a FAQ about this and it's wrong.

subprojects/packagefiles/lmdb/meson_options.txt
```diff
@@ -1,12 +1,5 @@
-option(
-    'tests',
```
bgilbert: Removing this option matches the upstream Makefile but I'm iffy on it.  In particular, the tests don't depend on `unistd.h`; do they work on Windows?  If so, it'd be nice to continue supporting them there, rather than using the same gate for both programs and tests.

bgilbert:
Emulated Alpine is getting ENOSYS, which indeed seems likely to be a qemu thing:

```
../subprojects/openldap-LMDB_0.9.33/libraries/liblmdb/mtest.c:50: mdb_env_open(env, "./testdb", MDB_FIXEDMAP , 0664): Function not implemented
```

subprojects/packagefiles/lmdb/meson_options.txt
neheb: readded

subprojects/packagefiles/lmdb/libraries/liblmdb/meson.build
neheb: on a slightly unrelated note, this meson format stuff is ugly.

```
@@ -3,7 +3,10 @@ cc = meson.get_compiler('c')
 src = files('mdb.c', 'midl.c')
 
-  threads_dep = dependency('', required: false)
+  threads_dep = dependency(
+    '',
+    required: false,
+  )
 else
   threads_dep = dependency('threads')
 endif
```

subprojects/packagefiles/lmdb/libraries/liblmdb/meson.build
neheb: OK

neheb:
This mplay thing is broken on Alpine:

```
../subprojects/openldap-LMDB_0.9.33/libraries/liblmdb/mplay.c:490:9: error: cannot assign to variable 'stdin' with const-qualified type 'FILE *const' (aka 'struct _IO_FILE *const')
  490 |                 stdin = fdopen(0, "r");
      |                 ~~~~~ ^
/usr/include/stdio.h:62:20: note: variable 'stdin' declared const here
   62 | extern FILE *const stdin;
      | ~~~~~~~~~~~~~~~~~~~^~~~~
../subprojects/openldap-LMDB_0.9.33/libraries/liblmdb/mplay.c:491:10: error: cannot assign to variable 'stdout' with const-qualified type 'FILE *const' (aka 'struct _IO_FILE *const')
  491 |                 stdout = fdopen(1, "w");
      |                 ~~~~~~ ^
/usr/include/stdio.h:63:20: note: variable 'stdout' declared const here
   63 | extern FILE *const stdout;
      | ~~~~~~~~~~~~~~~~~~~^~~~~~
```

using fdopen on stdin and stdout is...strange.

bgilbert:
Looking at [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/V2_chap02.html#tag_15_05_01), I think it's possible to quiesce `stdin`/`stdout` so the file descriptors can be safely replaced underneath them.  I'm not planning to do that work, so, [reported](https://bugs.openldap.org/show_bug.cgi?id=10355) upstream.

Also I had missed that `mplay` isn't in `IPROGS` so isn't built by default anyway.  Let's just take it out again, sorry for the trouble.

ci_config.json
bgilbert: https://github.com/mesonbuild/wrapdb/pull/2174

PR #2171, opened 2025-06-11
libjpeg-turbo: update from 3.1.0 to 3.1.1

bgilbert:


neheb:
So Alpine vs Ubuntu aarch64 is clang vs GCC. I see upstream uses some fp-contract stuff to switch the expected results.

bgilbert:
That fp-contract logic only affects tests that aren't implemented in Meson.

It turns out the Neon detection logic in the Meson port has always been broken, and was building incorrect library binaries with ARM GCC unless `-Dneon-intrinsics=enabled` was specified.  ARM Clang builds with Meson &ge; 1.2 apparently were unaffected because of another bug.  We only recently added aarch64 GCC to CI, which is why we didn't catch this.

PR #2173, opened 2025-06-11
Add svtjpegxs

xclaesse:


tp-m:
Nice!

Not sure if the SSE/AVX checks are sufficient like that. Compare to how opus checks for them, for example.

Wonder if the simd extension can be used here or if it's not a good fit.

xclaesse:
> Not sure if the SSE/AVX checks are sufficient like that. Compare to how opus checks for them, for example.

I admit I have no idea, but AFAIK that's what their cmake does. It's missing the check for corresponding msvc flag however.

> Wonder if the simd extension can be used here or if it's not a good fit.

As far as I can tell, those asm are not optional and only x86_64 is supported. We could be using simd when they provide multiple arches.

@eli-schwartz do you know if there is a way to tell the CI that only x86_64 is supported? ci_config.json can only disable platforms with build_on, not arches, right?

xclaesse:
> It's missing the check for corresponding msvc flag however.

Fixed that now.

eli-schwartz:
> @eli-schwartz do you know if there is a way to tell the CI that only x86_64 is supported? ci_config.json can only disable platforms with build_on, not arches, right?

I don't believe there's a way to do that yet. I think this is the first time the issue has come up...

xclaesse:
Ok, added "build_on_arch" for that.

subprojects/packagefiles/svtjpegxs/meson.build
```diff
@@ -0,0 +1,29 @@
...
+)
+
+cc = meson.get_compiler('c')
+if host_machine.cpu_family() == 'x86_64'
+  add_project_arguments(
+    '-DARCH_X86_64=1',
+    language: 'c',
+  )
+else
+  error('SVT-JPEG-XS requires x86_64 architecture')
```
eli-schwartz:
```suggestion
  error('unsupported architecture: SVT-JPEG-XS requires x86_64')
```

The sanity checks are a bit finicky here. They check for "unsupported" or "not supported".

xclaesse: oh, that's a much easier way to skip CI on those arch. Done.

xclaesse:
FInally got the CI to pass :)

subprojects/packagefiles/svtjpegxs/meson.build
eli-schwartz: I'd still make it a distinct commit.

subprojects/packagefiles/svtjpegxs/meson.build
```diff
@@ -0,0 +1,33 @@
...
+)
+
+cc = meson.get_compiler('c')
+if host_machine.cpu_family() == 'x86_64'
+  add_project_arguments(
+    '-DARCH_X86_64=1',
+    language: 'c',
+  )
+else
+  error('unsupported architecture: SVT-JPEG-XS requires x86_64')
```
tp-m, line 15: Bit confused by this. Are you saying you're only supporting x86_64 for now because that's the only thing you care about?

It should build fine on other architectures, with the C implementation.

In GStreamer cerbero we build it for all architectures I think, and we only enable the assembly for x86 + x86_64.

eli-schwartz: The upstream readme may or may not be accurate but it does say x86_64 is required, with zero qualifications. It says the unittests need AVX.

tp-m: > In GStreamer cerbero we build it for all architectures I think, and we only enable the assembly for x86 + x86_64.

That might not be quite right actually. I was only looking at the svt-jpeg-xs recipe, but it's only added as a dep in the  gst-plugins-bad recipe in case of x86_64.

xclaesse: > It should build fine on other architectures, with the C implementation.

I don't think they have C implementation, from what I see it is x86_64 only.

PR #2175, opened 2025-06-11
google-woff2: fix

bgilbert:
Upstream has no dllexport support, so build static libraries on Windows.


subprojects/packagefiles/google-woff2/meson.build
```diff
@@ -120,25 +155,31 @@ if meson.version().version_compare('>=0.46.0')
...
-  ],
-  version: meson.project_version(),
+libwoff2_encoder_src = files(
+  'src/font.cc',
+  'src/glyph.cc',
+  'src/normalize.cc',
+  'src/transform.cc',
+  'src/woff2_enc.cc',
 )
+if windows
```
neheb: could use build_target instead

bgilbert: Ooh, nice!  Didn't know that existed.

PR #2176, opened 2025-06-12
gtest: fix build on cygwin

na-trium-144:
on cygwin, gtest requires cpp_std=gnu++17, not cpp_std=c++17.
https://github.com/google/googletest/issues/813

however, we cannot set the default_options conditionally depending on the host_machine (https://github.com/mesonbuild/meson/issues/10065 ), so I think it's better to remove default_option for cpp_std and let it follow the options of the parent project.


bgilbert:
This construct works, and other wraps currently use it:

```meson
default_options: [meson.version().version_compare('>=1.3.0') ? 'cpp_std=gnu++17,c++17' : 'cpp_std=gnu++17'],
```

bgilbert:
Meson older than 1.3.0 will automatically fall back from `gnu++17` to `c++17` on MSVC, so `gnu++17` is safe there.  (Newer Meson will still fall back, but will complain because you aren't using the list form.)

bgilbert:
Thanks!

PR #2181, opened 2025-06-12
pcre2: add check whether builtin_mul_overflow links

na-trium-144:
In older versions of clang, `__builtin_mul_overflow` exists but fails to link.
So I added a check not only to see if it exists, but also to see if it actually links.
https://bugs.llvm.org/show_bug.cgi?id=16404


neheb:
funny. I think something similar happens with clang-cl builtins. I don't remember which package. opus maybe.

bgilbert:
Thanks!

CI failures are preexisting.

PR #2182, opened 2025-06-12
termbox2: new wrap

Gottox:


subprojects/packagefiles/termbox2/meson.build
```diff
@@ -0,0 +1,25 @@
+project(
```
bgilbert, line 1: License and ~registration~ minimum Meson version please.  Also probably a `default_options: ['c_std=c99']` to match upstream.

subprojects/termbox2.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = termbox2-2.5.0
+source_url = https://github.com/termbox/termbox2/archive/refs/tags/v2.5.0.tar.gz
+source_filename = termbox2-2.5.0.tar.gz
+source_hash = 1d7a9811060e3673be417019007acbf4e2575734b6f95baaf6e8aed27f06c65a
+patch_directory = termbox2
+
+[provide]
+termbox2 = termbox2_dep
```
bgilbert: Since we have `meson.override_dependency()` this can just be
```suggestion
dependency_names = termbox2
```

subprojects/packagefiles/termbox2/meson.build
bgilbert: Minimum Meson looks like 0.47.0.

Gottox: done!

* license added
* minimum version was reported by meson to be 0.54
* Added default options

subprojects/termbox2.wrap
Gottox: Done

bgilbert:
Thanks!

PR #2183, opened 2025-06-12
lvgl: Update to 9.3.0

livingsilver94:


bgilbert:
For the record, there's a _lot_ of bundled code here.  A bunch of it is somewhat obscure, but there's at least `lz4` and now `expat` that we have separate wraps for.  Ideally upstream would support unbundling (https://github.com/lvgl/lvgl/issues/8012) but we shouldn't piecemeal unbundle it ourselves and I understand why a library targeting embedded applications wouldn't care very much.

livingsilver94:
> I understand why a library targeting embedded applications wouldn't care very much.

It's a pointless discussion since LVGL should fix that, but that's debatable. They support DRM/KMS and Wayland as a target, the embedded excuse is no more :D

PR #2187, opened 2025-06-13
cmocka: remove unused variable

kimhmadsen:
Remove the variable `cfile` because it is unused.


bgilbert:
Passes in CI (at least as much as the previous PR did) with a revision bump.  Merging without one, since this isn't a functional change.

PR #2190, opened 2025-06-13
wayland-protocols: avoid CI failures for unsupported platforms

bgilbert:
Upstream has a custom error message for this.  Add it to the detection list.

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -511,9 +511,16 @@ class TestReleases(unittest.TestCase):
             if expect_working:
                 res.check_returncode()
             else:
-                if 'unsupported' in error or 'not supported' in error or 'does not support' in error:
-                    print('unsupported, as expected')
-                    return
+                for msg in [
+                    'unsupported',
+                    'not supported',
+                    'does not support',
+                    # wayland-protocols upstream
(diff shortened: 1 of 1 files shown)


tools/sanity_checks.py
```diff
@@ -511,9 +511,16 @@ def do_setup(builddir, options, meson_env):
...
-                    print('unsupported, as expected')
-                    return
+                for msg in [
+                    'unsupported',
+                    'not supported',
+                    'does not support',
+                    # wayland-protocols upstream
+                    'SFD_CLOEXEC is needed to compile Wayland libraries',
+                ]:
+                    if msg in error:
```
eli-schwartz, line 521: Sorry for not noticing this before.

This is hard to read here. Notice how the if statement is aligned with the list items. In general, consider instead;

```python
                for msg in ['unsupported',
                            'not supported',
                            'does not support',
                            # wayland-protocols upstream
                            'SFD_CLOEXEC is needed to compile Wayland libraries',
                            ]:
                    if msg in error:
```

(or keeping the ending ] on the same line as the last list element)

bgilbert: Sure: https://github.com/mesonbuild/wrapdb/pull/2196

PR #2194, opened 2025-06-15
Fix a bunch of license tags

bgilbert:
I've skipped a few where the correct SPDX expression wasn't immediately obvious.

Also add tags for a few projects.


bgilbert:
Dropped the revision bumps.  Really on the fence about whether to keep them; presumably no one is relying on the bad license values right now, but keeping them around just prolongs that.

PR #2200, opened 2025-06-17
au: new wrap

shyun3:
Bare minimum to provide library only, does not build tests or install.

### DIFF
--- a/subprojects/packagefiles/au/meson.build
+++ b/subprojects/packagefiles/au/meson.build
@@ -1,0 +1,11 @@
+project(
+    'au',
+    'cpp',
+    version: '0.4.1',
+    license: 'Apache-2.0',
+    default_options: ['cpp_std=c++14'],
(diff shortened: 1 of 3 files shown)


subprojects/packagefiles/au/meson.build
```diff
@@ -0,0 +1,13 @@
+project(
+  'au',
+  'cpp',
+  version: '0.4.1',
+  license: 'Apache-2.0',
+  default_options: ['cpp_std=c++14'],
+)
```
bgilbert, line 8: The `meson_version` requirement is `>=0.54.0`.

bgilbert (commented):
Please squash your commits.

subprojects/packagefiles/au/meson.build
shyun3: fixed

PR #2201, opened 2025-06-18
ci: report any `ci_config.json` stanzas as workflow annotations

bgilbert:
The annotations appear on the workflow summary screen and look like this by default:

image

and expanded:

image

Also, avoid spurious Homebrew annotations:

image


bgilbert:
It maybe isn't super-useful, but seemed nice.

CI failures are because of `master`, not this PR.

neheb (approved):

PR #2202, opened 2025-06-19
ci: determine minimum Meson version and annotate diff

bgilbert:
After sanity checks run, put `meson_version: '>=0'` into the `project()` call of any updated downstream wrap, rerun `meson setup`, collect the new-feature warnings, and add them to an annotation on the `meson_version` argument (or on `project()` if none).  Make the annotation a notice if the Meson dependency matches what we expect and a warning otherwise.  Also explicitly check for `c_std` or `cpp_std` in `default_options`, which requires Meson &ge; 0.63.

To avoid duplicate annotations we can only do this in one CI job, so pick Ubuntu x86_64.  That platform should be compatible with most wraps; only two have `"linux": false` in `ci_config.json`.

Don't fail CI if the Meson dependency doesn't match or if the check fails. Failures can occur due to rewriter limitations, and a wrap might need a newer Meson for platform-specific code, build options we don't test, or reasons unknown to `FeatureNew`.  For now this is just a review aid.

The resulting annotations look like this:

image

Also fix `default_options` checks when the `default_options` argument is not a list.

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -34,6 +34,14 @@ jobs:
         run: |
           ./tools/fake_tty.py ./tools/sanity_checks.py
 
+      - name: Report Meson version dependencies
+        if: matrix.platform == 'x86_64'
+        continue-on-error: true
+        env:
+          TEST_MESON_VERSION_DEPS: 1
+        run: |
+          ./tools/fake_tty.py ./tools/sanity_checks.py TestReleases.test_meson_version_deps
+
   Alpine:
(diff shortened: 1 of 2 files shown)


bgilbert:
CI failing because of the state of `master`.

bgilbert:
Did some refactoring and added a fix for the existing `default_options` lookup.  Also added an explicit check for `c_std`/`cpp_std` in `default_options`, which requires &ge; 0.63.0 but Meson doesn't warn about it.

@eli-schwartz, do you have some time to look at this?

neheb (approved):

PR #2203, opened 2025-06-19
gee: update from 0.20.6 to 0.20.8

bgilbert:


subprojects/packagefiles/gee/meson.build
```diff
@@ -2,7 +2,9 @@ project(
   'libgee',
   'c',
   'vala',
-  version: '0.20.6',
+  version: '0.20.8',
+  meson_version: '>=0.54.0',
```
neheb, line 6: I assume this is used for override dependency, which is handled alternatively as

```
if meson.version().version_compare('>=0.54.0')
```

I wonder if we should mandate a minimum meson_version. 0.54 is quite old.

bgilbert: Yeah, Debian oldstable (bullseye) has 0.56.2, Ubuntu jammy (22.04) has 0.61.2, RHEL 8 has 0.58.2.

I'd hesitate to remove compat conditionals that were already in the wrap, but there aren't any and no one has complained, so, inclined to omit.

PR #2206, opened 2025-06-19
Add Apache ORC wrap

WillAyd:
Meson support for Apache ORC was implemented in https://github.com/apache/orc/pull/2249 . We will have to wait for the next release to make this official, but I figured its worth testing a wrap entry anyway, in case any issues need to be patched before the next release

FYI @dongjoon-hyun

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -23,6 +23,14 @@
     "graphite2",
     "mocklibc"
   ],
+  "apache-orc": {
+    "test_options": [
+      "--timeout-multiplier=2"
+    ],
+    "alpine_packages": [
+      "tzdata"
+    ]
+  },
   "atomic_queue": {
     "_comment": "Tests require Boost, which is not available on WrapDB",
     "build_on": {
--- a/subprojects/apache-orc.wrap
+++ b/subprojects/apache-orc.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = orc-2.2.0
+source_url = https://www.apache.org/dyn/closer.lua?action=download&filename=orc/orc-2.2.0/orc-2.2.0.tar.gz
+source_fallback_url = https://github.com/apache/orc/archive/refs/tags/v2.2.0.tar.gz
+source_filename = orc-2.2.0.tar.gz
+source_hash = b15aca45a7e73ffbd1bbc36a78cd1422d41f07721092a25f43448e6e16f4763b
+
+[provide]
+orc = orc_dep
--- a/releases.json
+++ b/releases.json
@@ -132,6 +132,14 @@
       "20200225.2-1"
     ]
   },
+  "apache-orc": {
+    "dependency_names": [
+      "orc"
+    ],
+    "versions": [
+      "2.2.0-1"
+    ]
+  },
   "arduinocore-avr": {
     "versions": [
       "1.8.2-2",


tp-m:
Just want to point out that there is also [GStreamer's ORC (Optimized inner loop Runtime Compiler)](https://gitlab.freedesktop.org/gstreamer/orc/) which uses the same namespace for wrap file names and variable names and is a required dependency of the GStreamer/GTK/GNOME software stack.

dongjoon-hyun:
For the record, Apache ORC community starts to discuss to release Apache ORC 2.2.0 earlier than the pre-defined schedule (2026) to deliver this, @WillAyd . Please let us know if you need to change something in Apache ORC community further.
- https://lists.apache.org/thread/8wftl03kknm28c7f9rdxg2ckb7nn1q66

WillAyd:
Thanks @tp-m . I think we could also name this `apache-orc` to disambiguate

WillAyd:
OK it looks like https://github.com/apache/orc/issues/2238 upstream would ideally get resolved for the tests to pass here

eli-schwartz:
> Just want to point out that there is also [GStreamer's ORC (Optimized inner loop Runtime Compiler)](https://gitlab.freedesktop.org/gstreamer/orc/) which uses the same namespace for wrap file names and variable names and is a required dependency of the GStreamer/GTK/GNOME software stack.

gstreamer orc produces

/usr/lib64/liborc-0.4.so
/usr/lib64/pkgconfig/orc-0.4.pc

Apache orc produces:

/usr/lib64/liborc.so
/usr/lib64/cmake/orc/orcConfig.cmake

WillAyd:
@dongjoon-hyun what is the best channel to discuss the naming conflict with GStreamer's orc library?

subprojects/apache-orc.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = orc-2.2.0
+source_url = https://www.apache.org/dyn/closer.lua?action=download&filename=orc/orc-2.2.0/orc-2.2.0.tar.gz
```
WillAyd, line 3: I modeled this off the download link for the Apache Arrow project, but want to double check this is correct @dongjoon-hyun

WillAyd (commented):
Apache Orc 2.2.0 was released a few days back, so this is now ready for a real review.

To avoid any issues with GStreamer's ORC I changed the dependency and wrap name here to `apache-orc`.

eli-schwartz:
> To avoid any issues with GStreamer's ORC I changed the dependency and wrap name here to `apache-orc`.

There didn't used to be a dependency conflict with GStreamer's ORC. I pointed this out above.

The dependency name in the wrap file MUST be the same name as the pkg-config file / CMake `find_package()` name. It is such a big requirement that it's even the programmatic definition of the wrap provide field: "the system installed dependency name compatible with meson `dependency()` and CMake `find_package()`, that shall be overridden"

WillAyd:
Ah my mistake - I misinterpreted your comment as saying it could be problematic if the versioning policy for either library changed. I can revert back to the `orc` dependency name

eli-schwartz:
Ah, sorry for the confusion.

The naming policy is unlikely to ever change because projects which add a second copy of the version into the project name usually do it extremely deliberately -- the idea is to "support parallel installation of all versions", so they very much want to require you to specify the *name* using a versioned name.

If the naming policy does change, it would likely mean that Apache ORC and GStreamer ORC stop being able to be installed on the same Linux distro install, which would be kind of bad or at least unfortunate. Consequently the GStreamer project will step lightly here to avoid it.

And yes, the conflict would be significant -- the runtime library would have an overlapping name which can't really be solved. But as of now, the runtime library doesn't overlap. :)

WillAyd:
Makes sense. I think I've got the right balance here now then - the wrap itself is called `apache-orc` but the exposed dependency is still `orc` (as are the cmake / pkg-conf files)

subprojects/apache-orc.wrap
```diff
@@ -0,0 +1,9 @@
...
+source_fallback_url = https://github.com/apache/orc/archive/refs/tags/v2.2.0.tar.gz
+source_filename = orc-2.2.0.tar.gz
+source_hash = b15aca45a7e73ffbd1bbc36a78cd1422d41f07721092a25f43448e6e16f4763b
+
+[provide]
+orc = orc_dep
```
bgilbert, line 9: For future reference, upstream depends on a fairly new Meson, so it should use `meson.override_dependency()`, allowing this to change to `dependency_names = orc` and stop depending on upstream's internal variable name.

...Is upstream intentionally depending on a Meson that new?  For the "fallback for missing system library" use case, wraps are generally more useful if they support older Meson versions.

ci_config.json
```diff
@@ -23,6 +23,17 @@
     "graphite2",
     "mocklibc"
   ],
+  "apache-orc": {
+    "_comment": "Tests fail on Alpine. Windows symbol visibility broken",
+    "build_on": {
+      "alpine": false,
+      "msys2": false,
+      "windows": false
+    },
```
bgilbert: Are these still broken, and if so, is it expected that they'll be fixed?  `build_on` is generally meant for wraps that permanently won't support a platform, in which case the project's `meson.build` is expected to call `error()` on those platforms with an appropriate message.

WillAyd: Ah OK -  they are just broken now but can be fixed in the future. I'll remove this

subprojects/apache-orc.wrap
WillAyd: I don't think orc in particular needs 1.3.0, but the reason for that is other projects in the Apache space do actually require it, so I picked a consistent version across all projects. Certainly open to changing that

eli-schwartz: Using a consistent version across all projects isn't necessarily a bad thing, if e.g. it makes it easier for upstream to consistently test things.

The only downside of using a newer version than actually required is that people using meson versions that are even older, won't be able to use this wrap. That may or may not matter for any given use case. People stuck on ancient end of life operating systems that don't have python 3.7 might be stuck on meson ~0.60, but are they also using orc??? People on Debian oldstable (bookworm) have meson 1.0.1, which is just a bit too old, but... Are they really using orc as a wrap either? Maybe! Debian oldstable isn't that old, even as an active development platform. On the other hand, bookworm-backports has meson 1.7.0. Ubuntu 22.04 LTS has meson 0.61, but 24.04 LTS has meson 1.3.2 -- how far back do you want to go to support LTS versions?

The wrapdb has a soft policy "it's better to not restrict things for no reason because the point of the wrapdb is to allow people to use things on diverse systems". But it is just that -- a soft policy. For wraps that are intended to be upstream, following upstream support lifecycles is obviously far more important, and part of that means upstream decisions on how far to go in supporting old operating systems. "Latest Debian stable and Ubuntu LTS" is not a bad metric for that, and requiring meson 1.3.0 aligns with that metric.

That being said, if I personally were writing meson.build files to contribute to projects, I like to go the extra mile and determine the actual minimum version it will successfully compile with.

WillAyd: Cool thanks for that input. Happy to take a closer look upstream for the next release and see if that can be relaxed

ci_config.json
```diff
@@ -23,6 +23,12 @@
     "graphite2",
     "mocklibc"
   ],
+  "apache-orc": {
+    "_comment": "Tests fail on Alpine. Windows symbol visibility broken",
```
bgilbert: Nit: dangling comment.  This file tends to accumulate stale stuff as is, so I'd prefer we clean this up before merging.

bgilbert (approved):

bgilbert:
Can the Alpine failures be fixed by installing the `tzdata` package?

WillAyd:
Doesn't look like it unfortunately - there are other test errors beyond the availability of tzdata

bgilbert:
Looks like `/etc/localtime` would need to be configured explicitly, and we don't have an obvious mechanism for doing that.

PR #2207, opened 2025-06-19
boost-mp11: new wrap

shyun3:
https://github.com/boostorg/mp11

This is a header-only, standalone Boost library that does not have any other Boost dependencies.


bgilbert (approved):
Looks good to me.  I'll let this sit for a bit in case anyone else has opinions.

PR #2209, opened 2025-06-20
wren: make optional modules optional

lv-99-slime:


bgilbert:
Needs a `releases.json` update, and CI will want `meson_options.txt` reindented to four spaces.

lv-99-slime:
Ah, misunderstood it. Hopefully good now.

PR #2210, opened 2025-06-20
Remove subproject buildtype settings and use proper debug detection

### LINKED ISSUE #2208: Should setting -Dbuildtype=release at the project level be discouraged?
Going back to the discussion in https://github.com/mesonbuild/wrapdb/pull/2150#issuecomment-2971674074, I think we may want to discourage projects from setting the buildtype as part of their default options. When building a project as the main project it is harmless, but when used as a subproject, on Windows you will get failures when linking that subproject in to a main project who is built with a debug build type.

Assessing the current state of the wrapdb, there are only a few configurations that have this. I am happy to clean this up if we agree

WillAyd:
closes #2208


releases.json
```diff
@@ -1415,6 +1418,7 @@
       "gencmn"
     ],
     "versions": [
+      "77.1-1",
```
bgilbert:
```suggestion
      "77.1-2",
```

subprojects/packagefiles/stduuid/test/meson.build
```diff
@@ -15,7 +15,6 @@ if host_machine.system() == 'windows'
   ]
 else
   test_flags += ['-fexceptions']
-  test_override_options = ['buildtype=debugoptimized']
```
bgilbert, line 18: `test_override_options` isn't necessary now; we can just pass `override_options` in `executable()` below.

(This is a behavior change in any case, because the removed line isn't `+=`, but presumably that was a bug.)

subprojects/packagefiles/type_safe/meson.build
```diff
@@ -8,16 +8,7 @@ project(
 
 da_dep = dependency('debug_assert')
 
-b_ndebug = get_option('b_ndebug')
```
bgilbert: `b_ndebug` is the most relevant variable to use here, and we need `get_option('buildtype')` to emulate Meson's logic for the `if-release` case, so I think this one should be left as is (and its `releases.json` bump reverted).

bgilbert:
Thank you!

PR #2212, opened 2025-06-21
protobuf: do not use symlinks

meator:
Symlinks on Windows can be problematic. This PR replaces them with simple copies.

To make this change, it would be useful to know what even is the purpose of the symlink:

https://github.com/mesonbuild/wrapdb/blob/92719d5a4dfa6c03412eddd628322bc1e5c22016/subprojects/packagefiles/protobuf/src/meson.build#L2-L11

The wrap builds without it (and with all `proto_includes_link_dep` dependencies removed). If there is some elegant way to remove this entirely, it should be preferred.

What is the purpose of the symlink?


meator:
CI failure is not caused by this PR.

bgilbert:
It was added in https://github.com/mesonbuild/wrapdb/pull/1263 to address https://github.com/mesonbuild/wrapdb/issues/1259.

bgilbert:
Why not copy on Windows and symlink otherwise?

PR #2216, opened 2025-06-23
Remove a few more `buildtype` `default_options`

bgilbert:
cc @WillAyd


WillAyd:
Ah thanks for catching these!

neheb:
libgrapheme failures look like they can be worked around.

bgilbert:
Not all of them, I think?  They're already fixed upstream in any case.

PR #2219, opened 2025-06-24
mimalloc: add wrap for 3.1.5

stephanlachnit:
TODO:
- [ ] ~~Add redirect support on Windows~~ postponed
- [x] Test MinGW
- [x] Check if linkage works as expected

### DIFF
--- a/subprojects/packagefiles/mimalloc/meson.build
+++ b/subprojects/packagefiles/mimalloc/meson.build
@@ -1,0 +1,141 @@
+project(
+  'mimalloc',
+  'c',
+  meson_version: '>=0.46.0',
+  license: 'MIT',
+  version: '3.1.5',
+)
+
+mi_sources = []
+mi_deps = []
+mi_args = []
+mi_static_args = []
+mi_shared_args = []
+
+mi_sources += files(
+  'src/alloc-aligned.c',
+  'src/alloc-posix.c',
+  'src/alloc.c',
+  'src/arena-meta.c',
+  'src/arena.c',
+  'src/bitmap.c',
+  'src/heap.c',
+  'src/init.c',
+  'src/libc.c',
+  'src/options.c',
+  'src/os.c',
+  'src/page-map.c',
+  'src/page.c',
+  'src/prim/prim.c',
+  'src/random.c',
+  'src/stats.c',
+)
+
+# System libraries
+cc = meson.get_compiler('c')
+if host_machine.system() == 'windows'
+  psapi_dep = cc.find_library('psapi')
+  shell32_dep = cc.find_library('shell32')
+  user32_dep = cc.find_library('user32')
+  advapi32_dep = cc.find_library('advapi32')
+  bcrypt_dep = cc.find_library('bcrypt')
+  mi_deps += [psapi_dep, shell32_dep, user32_dep, advapi32_dep, bcrypt_dep]
+else
+  thread_dep = dependency('threads')
+  atomic_dep = cc.find_library(
+    'atomic',
+    required: false,
+  )
+  rt_dep = cc.find_library(
+    'rt',
+    required: false,
+  )
+  mi_deps += [thread_dep, atomic_dep, rt_dep]
+endif
+
+# Compile args
+mi_static_args += ['-DMI_STATIC_LIB']
+mi_shared_args += ['-DMI_SHARED_LIB', '-DMI_SHARED_LIB_EXPORT']
+if not get_option('debug')
+  mi_args += ['-DMI_BUILD_RELEASE']
+endif
+if get_option('local_dynamic_tls')
+  mi_args += cc.get_supported_arguments('-ftls-model=local-dynamic')
+else
+  mi_args += cc.get_supported_arguments('-ftls-model=initial-exec')
+endif
+
+# malloc override
+if get_option('override')
+  # On MacOS use MI_OSX_ZONE and MI_OSX_INTERPOSE
+  if host_machine.system() == 'darwin'
+    mi_sources += files('src/prim/osx/alloc-override-zone.c')
+    mi_args += ['-DMI_OSX_ZONE=1', '-DMI_OSX_INTERPOSE=1']
+  endif
+  # Disable built-in malloc
+  mi_args += cc.get_supported_arguments('-fno-builtin-malloc')
+  # Enable malloc override (except on static windows build)
+  mi_shared_args += ['-DMI_MALLOC_OVERRIDE']
+  if host_machine.system() != 'windows'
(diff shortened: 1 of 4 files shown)


stephanlachnit:
Works, on MinGW I get this compiler warning:

```
../subprojects/mimalloc-3.1.5/src/init.c:797:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
  797 | }
      | ^
../subprojects/mimalloc-3.1.5/src/init.c:797:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
../subprojects/mimalloc-3.1.5/src/init.c:797:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
../subprojects/mimalloc-3.1.5/src/init.c:797:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
../subprojects/mimalloc-3.1.5/src/init.c:797:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
../subprojects/mimalloc-3.1.5/src/init.c:797:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
```

But that doesn't seem to be critical.

subprojects/packagefiles/mimalloc/meson.build
```diff
@@ -0,0 +1,119 @@
...
+    'rt',
+    required: false,
+  )
+  mi_deps += [thread_dep, atomic_dep, rt_dep]
+endif
+
+# Compile args
+mi_static_args += ['-DMI_STATIC_LIB']
+mi_shared_args += ['-DMI_SHARED_LIB', '-DMI_SHARED_LIB_EXPORT']
+if get_option('buildtype') != 'debug'
```
bgilbert: `get_option('buildtype')` isn't reliable; see https://github.com/mesonbuild/meson/issues/11645.

```suggestion
if not get_option('debug')
```

subprojects/packagefiles/mimalloc/meson.build
```diff
@@ -0,0 +1,119 @@
...
+  mi_args += ['-DMI_BUILD_RELEASE']
+endif
+if get_option('local_dynamic_tls')
+  if cc.has_argument('-ftls-model=local-dynamic')
+    mi_args += ['-ftls-model=local-dynamic']
+  endif
```
bgilbert:
```suggestion
  mi_args += cc.get_supported_arguments('-ftls-model=local-dynamic')
```

`get_supported_arguments` requires Meson &ge; 0.43.0 (probably good to add a `meson_version` argument in `project()`).  That's a very old version, so seems fine to require it.

subprojects/packagefiles/mimalloc/meson_options.txt
```diff
@@ -0,0 +1,18 @@
+option(
+    'override',
+    type: 'boolean',
+    value: true,
```
bgilbert, line 4: This default seems a bit surprising: merely depending on mimalloc (including indirectly through another dependency) will change global behavior.

bgilbert (commented):
Upstream lists the 3.x series as beta, and we don't normally package those.  This is a new wrap, though, so maybe it's okay?

subprojects/packagefiles/mimalloc/meson.build
```diff
@@ -0,0 +1,119 @@
...
+  mi_args += mi_static_args
+elif get_option('default_library') == 'shared'
+  mi_args += mi_shared_args
+else
+  error('Building both static and shared libraries is not possible')
+endif
+
+# Library
+mi_inc = include_directories('include')
+mi_lib = library(
```
bgilbert, line 112: To match upstream, `soversion` should be the project major version and `version` should be the project `major.minor`.

subprojects/packagefiles/mimalloc/meson_options.txt
stephanlachnit: That's the opposite of surprising I'd say - the main point of this is that it is plug-and-play, using it explicitly is the exception.

stephanlachnit:
> Upstream lists the 3.x series as beta, and we don't normally package those. This is a new wrap, though, so maybe it's okay?

It's also the one available in the upcoming Debian stable, so I think it's fine.

subprojects/packagefiles/mimalloc/meson_options.txt
bgilbert: It may not be surprising for an allocator but it's surprising for a dependency.  In other circumstances, if I write `dependency('foo')`, the behavior of my program won't change unless I explicitly invoke foo.  More to the point, if my dependency `bar` writes `dependency('mimalloc')` without my knowledge, my project will be directly affected.  I'm not categorically opposed to this, but there's some value in requiring someone in the dependency chain to explicitly enable override.

stephanlachnit: Ah - maybe you misunderstood this. It doesn't override malloc with just `dependecy('mimalloc')`, you need to actually link it. Sure, if a dependency declares its dependency with mimalloc as public dependency then you might be affected but this has to be quite intentional.

stephanlachnit: Ah right, when linking dynamically it still affects others. Still, I think this is intended in >90% of the usecases - if you only want to overwrite malloc for your library you probably want to set this explicitly.

eli-schwartz:
> > Upstream lists the 3.x series as beta, and we don't normally package those. This is a new wrap, though, so maybe it's okay?
> 
> It's also the one available in the upcoming Debian stable, so I think it's fine.

It also means it's now quite challenging to add a downgraded wrap for the stable version that users might be hesitant to upgrade away from.

subprojects/packagefiles/mimalloc/meson.build
```diff
@@ -0,0 +1,141 @@
...
+# Windows redirect
+if host_machine.system() == 'windows'
+  if not get_option('win_redirect')
+    mi_args += ['-DMI_WIN_NOREDIRECT=1']
+  else
+    error('Windows redirect not implemented yet')
+  endif
+endif
+
+# Set shared and shared flags manually since Meson 1.3.0 is still too new
```
eli-schwartz, line 93:
```meson
if meson.version().version_compare('>=1.3.0')
    library(..., c_static_args: mi_static_args, c_shared_args: mi_shared_args)
else
    # current version of the code as merged
fi
```

It is *possible*.

stephanlachnit: True, but I would argue it's not very pretty and the use case is quite rare in this particular case.

PR #2234, opened 2025-06-30
json-c: remove threads dep and update config

neheb:


subprojects/packagefiles/json-c/config/meson.build
```diff
@@ -72,6 +70,10 @@ if conf.has('HAVE_STRINGS_H')
   check_function += 'strncasecmp'
 endif
 
+if cc.has_function('arc4random')
```
bgilbert: The loop below already calls `has_function` so the extra conditional isn't needed.

PR #2238, opened 2025-06-30
sdl2: update to 2.32.8

jgmdev:
Update to latest bug fix release of sdl2.


jgmdev:
Fixed warning of minimum required meson version by bumping to expected `meson_version: '>=0.63'`

bgilbert:
The check is still complaining because it expects to see `>=0.63.0`, but the `meson_version` argument is fine as is.

PR #2240, opened 2025-06-30
openal-soft: update to 1.24.3

neheb:


neheb:
```
0.63.0: c_std in subproject default_options, cpp_std in subproject default_options
```

what?

bgilbert:
Yeah, sanity checks explicitly check for [that](https://mesonbuild.com/Release-notes-for-0-63-0.html#compiler-options-can-be-set-per-subproject) because Meson doesn't.

bgilbert (approved):

PR #2242, opened 2025-07-02
xdelta3: add

bgilbert:
It isn't a library and it hasn't been maintained upstream since 2017, but I have a project whose test suite depends on it.

This can also live downstream if WrapDB isn't the right place for it.


bgilbert:
Will merge soon.

PR #2247, opened 2025-07-03
cppzmq: update to 4.11.0

stephanlachnit:


stephanlachnit:
Seems like this fails because the catch2 self tests are not compiling...

stephanlachnit:
Can we merge this since it is broken due to catch2 which was updated as well (e.g. https://github.com/mesonbuild/wrapdb/pull/2295)?

PR #2249, opened 2025-07-05
sdl3_ttf: add v3.2.2

KontrollFreek:
resubmitting because i moved around forks

### DIFF
--- a/subprojects/packagefiles/sdl3_ttf/meson.build
+++ b/subprojects/packagefiles/sdl3_ttf/meson.build
@@ -1,0 +1,107 @@
+project(
+  'sdl3_ttf',
+  'c',
+  license: 'zlib',
+  version: '3.2.2',
+  meson_version: '>=0.48.0',
+)
+
+parsed_project_version = meson.project_version().split('.')
+
+project_version = {
+  'major': parsed_project_version[0],
+  'minor': parsed_project_version[1],
+  'micro': parsed_project_version[2],
+}
+so_version = {
+  'major': '0',
+  'minor': project_version['minor'],
+  'micro': project_version['micro'],
+}
+dylib_compat_version = {
+  'minor': '0',
+  'micro': '0',
+}
+dylib_current_version = {
+  'micro': '0',
+}
+
+if project_version['minor'].to_int().is_even()
+  dylib_compat_version += {
+    'major': (100 * project_version['minor'].to_int() + 1).to_string(),
+  }
+  dylib_current_version += {
+    'minor': project_version['micro'],
+  }
+else
+  dylib_compat_version += {
+    'major': (100 * project_version['minor'].to_int() + project_version['micro'].to_int() + 1).to_string(),
+  }
+  dylib_current_version += {
+    'minor': '0',
+  }
+endif
+
+dylib_current_version += {
+  'major': dylib_compat_version['major'],
+}
+
+compiler = meson.get_compiler('c')
+
+project_arguments = [
+  'SDL_BUILD_MAJOR_VERSION=' + project_version['major'],
+  'SDL_BUILD_MINOR_VERSION=' + project_version['minor'],
+  'SDL_BUILD_MICRO_VERSION=' + project_version['micro'],
+]
+
+sdl_dep = dependency(
+  'sdl3',
+  version: '>=3.2.6',
+)
+
+freetype_dep = dependency('freetype2')
+thread_dep = dependency(
+  'threads',
+  required: false,
+)
+harfbuzz_dep = dependency(
+  'harfbuzz',
+  required: false,
+)
+plutosvg_dep = dependency(
+  'plutosvg',
+  required: false,
+)
+
+if harfbuzz_dep.found()
+  project_arguments += 'TTF_USE_HARFBUZZ=1'
+endif
+if plutosvg_dep.found()
+  project_arguments += 'TTF_USE_PLUTOSVG=1'
+endif
+
+foreach argument : project_arguments
+  add_project_arguments(
+    '-D' + argument,
+    language: 'c',
+  )
+endforeach
+
+subdir('src')
+ttf_lib = library(
+  'sdl3_ttf',
+  ttf_src,
+  include_directories: include_directories('include'),
+  dependencies: [sdl_dep, freetype_dep, thread_dep, harfbuzz_dep, plutosvg_dep],
+  soversion: so_version['major'],
+  version: so_version['major'] + '.' + so_version['minor'] + '.' + so_version['micro'],
+  darwin_versions: [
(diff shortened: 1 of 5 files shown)


KontrollFreek:
sdl3 is complaining about needing `libudev` for some reason? are packages not being automatically installed or am i missing something

KontrollFreek:
tested with `Ubuntu x86_64` using WSL, SDL just needs some extra deps to work properly.
if someone else wants to test, `sudo apt install libudev libpipewire-0.3-dev` installs everything needed for Ubuntu

z-bar:
> sdl3 is complaining about needing `libudev` for some reason? are packages not being automatically installed or am i missing something

They aren't; you need to modify the `ci_config.json` file for that, and `sdl3_ttf` will need the same packages as `sdl3` since the former depends on the latter.

KontrollFreek:
> They aren't; you need to modify the `ci_config.json` file for that, and `sdl3_ttf` will need the same packages as `sdl3` since the former depends on the latter.

Got it, will deal with that soon

KontrollFreek:
updated ci_config.json, any way i can rerun the checks?

z-bar:
> updated ci_config.json, any way i can rerun the checks?

Just commit and push the changes to your branch and it should rerun the checks automatically.

subprojects/packagefiles/sdl3_ttf/meson.build
```diff
@@ -0,0 +1,38 @@
...
+  required: false,
+)
+
+subdir('src')
+ttf_lib = library(
+  'sdl3_ttf',
+  ttf_src,
+  include_directories: 'include',
+  dependencies: [sdl_dep, freetype_dep, thread_dep, harfbuzz_dep, plutosvg_dep],
+)
```
bgilbert, line 102: Upstream sets a version and soversion so we probably should too.

subprojects/packagefiles/sdl3_ttf/meson.build
```diff
@@ -0,0 +1,38 @@
+project(
+  'sdl3_ttf',
+  'c',
+  license: 'zlib',
+  version: '3.2.2',
+)
```
bgilbert, line 7: Either needs `meson_version: '>=0.50.0'` or `include_directories:` arguments need to be wrapped in `include_directories()`.

KontrollFreek:
added all the suggested changes as well as added some macros defined in the original CMakeLists.txt

subprojects/packagefiles/sdl3_ttf/meson.build
bgilbert: Oh, and Darwin versions too.

subprojects/packagefiles/sdl3_ttf/meson.build
```diff
@@ -0,0 +1,77 @@
...
+endforeach
+
+subdir('src')
+ttf_lib = library(
+  'sdl3_ttf',
+  ttf_src,
+  include_directories: include_directories('include'),
+  dependencies: [sdl_dep, freetype_dep, thread_dep, harfbuzz_dep, plutosvg_dep],
+  soversion: '0',
+  version: version['major'] + '.' + version['minor'] + '.' + version['micro']
```
bgilbert: Upstream forces the major version to 0.  Once you've done that, the separate `soversion` argument isn't necessary since it'll be taken from the `version` major.

subprojects/packagefiles/sdl3_ttf/meson.build
```diff
@@ -0,0 +1,77 @@
+project(
+  'sdl3_ttf',
+  'c',
+  license: 'zlib',
+  version: '3.2.2',
+)
+
+version = {
```
bgilbert: This should be `meson.project_version().split('.')` rather than hardcoding the version twice.

subprojects/packagefiles/sdl3_ttf/meson.build
```diff
@@ -0,0 +1,77 @@
...
+
+compiler = meson.get_compiler('c')
+
+project_arguments = []
+
+is_glibc = compiler.has_header_symbol('stdlib.h', '__GLIBC__')
+sizeof_voidp = compiler.sizeof('void*')
+
+if is_glibc and sizeof_voidp == 4
+  project_arguments += '_FILE_OFFSET_BITS=64'
```
bgilbert: Meson sets this define automatically.

subprojects/packagefiles/sdl3_ttf/meson.build
KontrollFreek: sorry about that i read the source wrong

KontrollFreek:
i believe i fixed everything but im not entirely sure

KontrollFreek:
nevermind, adding dylib versions right now

KontrollFreek:
soversion, version, and darwin_versions are now added and supposedly working, i can't exactly figure out if it matches to the CMakeLists, but it seems to

subprojects/packagefiles/sdl3_ttf/meson.build
neheb: went down to 0.48.0

subprojects/packagefiles/sdl3_ttf/meson.build
```diff
@@ -0,0 +1,106 @@
...
+dylib_current_version = {
+  'micro': '0',
+}
+
+if project_version['minor'].to_int().is_even()
+  dylib_compat_version += {
+    'major': (100 * project_version['minor'].to_int() + 1).to_string(),
+  }
+  dylib_current_version += {
+    'minor': project_version['minor'],
```
bgilbert:
```suggestion
    'minor': project_version['micro'],
```

subprojects/packagefiles/sdl3_ttf/meson.build
```diff
@@ -0,0 +1,106 @@
+project(
+  'sdl3_ttf',
+  'c',
+  license: 'zlib',
+  version: '3.2.2',
+)
```
bgilbert, line 7: CI reports that we need `meson_version: '>=0.48.0'`.

bgilbert (commented):
Looks good, just two small fixes and then I think this can go in.

subprojects/packagefiles/sdl3_ttf/meson.build
KontrollFreek: micro keeps parity with the naming of the library, I figured it made more sense

bgilbert: We should match upstream's rules.  They know the relationship between the project version number and the ABI compatibility rules they intend to follow.

bgilbert:
Thanks!

PR #2250, opened 2025-07-05
glaze: add v5.5.2

KontrollFreek:
couldn't sanity check on my local copy but i tested with gcc on Windows 11 and Ubuntu 13 (WSL)

### DIFF
--- a/subprojects/packagefiles/glaze/meson.build
+++ b/subprojects/packagefiles/glaze/meson.build
@@ -1,0 +1,11 @@
+project(
+  'glaze',
+  'cpp',
+  license: 'MIT',
+  version: '5.5.2',
+)
+
+glaze_dep = declare_dependency(
+  include_directories: include_directories('include'),
+  compile_args: '-std=c++23',
+)
--- a/releases.json
+++ b/releases.json
@@ -1094,6 +1094,14 @@
       "5.2.1-1"
     ]
   },
+  "glaze": {
+    "dependency_names": [
+      "glaze"
+    ],
+    "versions": [
+      "5.5.2-1"
+    ]
+  },
   "glbinding": {
     "dependency_names": [
       "glbinding",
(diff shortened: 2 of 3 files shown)


KontrollFreek:
drafting because im running into some issues trying to use it locally

subprojects/packagefiles/glaze/meson.build
```diff
@@ -0,0 +1,11 @@
+project(
+  'glaze',
+  'cpp',
+  license: 'mit',
+  version: '5.5.2',
+)
```
bgilbert, line 6: `meson_version` should be `>=0.50.0`, or alternatively you can say `include_directories: include_directories('include')` below.

subprojects/packagefiles/glaze/meson.build
```diff
@@ -0,0 +1,11 @@
+project(
+  'glaze',
+  'cpp',
+  license: 'mit',
```
bgilbert: Minor:
```suggestion
  license: 'MIT',
```

subprojects/packagefiles/glaze/meson.build
```diff
@@ -0,0 +1,11 @@
+project(
+  'glaze',
+  'cpp',
+  license: 'mit',
+  version: '5.5.2',
+)
+
+glaze_dep = declare_dependency(
+  include_directories: 'include',
+  compile_args: '-std=c++23',
```
jpakkane, line 10: This is most definitely not allowed. It is both compiler specific and clashes with the builtin standard version option.

KontrollFreek: meson does not support c++23 when specifying a default version, and glaze will not build unless it is on c++23. also, not to mention all standard compilers (llvm, gcc, msvc, etc.) support this flag and the way it's formatted

neheb: He's saying having it a part of glaze_dep

bgilbert (changes requested):
Marking "request changes" for `-std` in `glaze_dep`.

PR #2258, opened 2025-07-10
eyalroz-printf: Add new wrap v6.2.0

crvux:


crvux:
`lead_directory_missing = true` it is because zipped dir name is `printf-<some_version>` which can very easily lead to a collision. So we use the directory is named the same as this package as "safety" layer

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
```diff
@@ -0,0 +1,151 @@
+cc = meson.get_compiler('c')
+## Checks related to the 'j', 'z' and 't' size modifiers
+sizeof_long = cc.sizeof('long')
+sizeof_long_long = cc.sizeof('long long')
+acceptable_jzt_type_sizes = [sizeof_long, sizeof_long_long]
+types = ['intmax_t', 'size_t', 'ptrdiff_t']
+foreach type : types
+  type_size = cc.sizeof(
+    type,
+    prefix: ['#include <stdint.h>'],
```
bgilbert: This would reduce the minimum Meson to 0.63.0:
```suggestion
    prefix: '#include <stdint.h>',
```

subprojects/packagefiles/eyalroz-printf/meson.build
```diff
@@ -0,0 +1,15 @@
...
+  meson_version: '>=1.0.0',
+  version: '6.2.0',
+  default_options: [
+    'b_staticpic=false',
+    'c_std=c99',
+    'default_library=static',  # For shared support we need to somehow provide putchar_ implementation
+  ],
+)
+
+subdir('printf-6.2.0')
```
bgilbert: This is slightly confusing and I don't think it's necessary.  We have other wraps where we've prefixed the wrap name but accepted upstream's somewhat generic directory name.  It has _a_ name in it, and a version.

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
```diff
@@ -0,0 +1,151 @@
...
+endforeach
+
+private_c_args = []
+public_c_args = []
+lib_config = {}
+if cc.get_argument_syntax() == 'gcc'
+  private_c_args += ['-Wall', '-Wextra', '-pedantic', '-Wconversion']
+elif cc.get_argument_syntax() == 'msvc'
+  private_c_args += '/W4'
+endif
```
bgilbert: Maybe just
```suggestion
private_c_args = cc.get_supported_arguments('-Wall', '-Wextra', '-pedantic', '-Wconversion', '/W4')
```

Also, for the record, these are pretty aggressive warning levels for use by the general public.  (It makes sense for project maintainers to use higher levels, but.)

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
```diff
@@ -0,0 +1,151 @@
...
+  compile_args: [public_c_args],
+)
+meson.override_dependency('eyalroz-printf', eyalroz_printf_dep)
+
+if get_option('tests')
+  subdir('test')
+endif
+
+# Prints the sizes of the different sections of the ELF file: text, dat, vss etc.
+size_prog = find_program('size')
```
bgilbert: `required` defaults to `true`.

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
```diff
@@ -0,0 +1,151 @@
...
+if size_prog.found()
+  custom_target(
+    'printf-sizes',
```
bgilbert: I think it's okay to have stuff like this if you want, but typically in wraps we don't reimplement the maintainer tools from the upstream project.

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
```diff
@@ -0,0 +1,151 @@
...
+    command: [size_prog, '-A', '-t', eyalroz_printf_lib],
+    capture: true,
+    output: 'printf_sizes.txt',
+  )
+endif
+
+# Produces lists of the symbols, and C++demangled symbols, inside the library
+nm_prog = find_program('nm')
```
bgilbert: The program at the found path isn't actually used in the shell commands.

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
```diff
@@ -0,0 +1,151 @@
...
+if nm_prog.found()
+  lib_path = eyalroz_printf_lib.full_path()
+  sh_cmd = 'nm --numeric-sort --print-size ' + lib_path + ' > @OUTDIR@/printf_symbols.txt'
+  filt_prog = find_program('c++filt')
+  if filt_prog.found()
+    sh_cmd += ' && nm --numeric-sort ' + lib_path + ' | c++filt > @OUTDIR@/printf_cpp_symbols.txt'
+  endif
+  custom_target(
+    'printf-symbols',
+    command: ['sh', '-c', sh_cmd],
```
bgilbert: Won't necessarily work on Windows.  At some level of complexity you'll need a Python script.  (Or you could remove this.)

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
crvux: I'm not entirely sure about the necessity of these targets. I use this library in bare-metal firmwares and it can be useful to see each symbol size. This is necessary to optimize the binary size. But it often happens that you need to look at all the symbols and not just for a one library.
I think if these targets have already been defined then let them remain :smile:

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
crvux: As far as i know powershell supports pipinng. If i additionally searched for powershell in the system (and then use it if found), could it work on Windows?

bgilbert: In principle, yes.  Personally I'd recommend the Python script for simplicity.

subprojects/packagefiles/eyalroz-printf/meson.build
crvux: On the other hand, the main directory called as the package and you don't need to look in the .wrap file if you need to patch something at the moment. Mb let's leave it like that?

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
crvux: Ok then if demangling is really needed i will try to add it in the next revisions via python script

subprojects/packagefiles/eyalroz-printf/meson.build
bgilbert: I won't veto it, but there's value in staying consistent with the other wraps in this repo.

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/test/meson.build
```diff
@@ -0,0 +1,95 @@
+tests_c_args = cc.get_supported_arguments(
```
crvux: Does the get_supported_arguments() support list of strings as an argument? In the documentation only variadic string

> The method accepts between 0 and infinity variadic arguments (arg...) of type [str](https://mesonbuild.com/Reference-manual_elementary_str.html)

subprojects/packagefiles/eyalroz-printf/meson.build
crvux: Changed this for consistency. If a collision occurs in the future, it will be easy to fix

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/test/meson.build
bgilbert: Yes, it's supported via [argument flattening](https://mesonbuild.com/Syntax.html#argument-flattening).

crvux:
@bgilbert are there any other shortcomings that need to be fixed?

crvux:
@bgilbert friendly ping

subprojects/packagefiles/eyalroz-printf/printf-6.2.0/meson.build
eli-schwartz: The more general issue here is that `meson setup -Dwarning_level=3` obsoletes this. If you just want the default to be to have all these warnings then it can be set in `project(default_options: ...)`.

If passing it to specific targets, there is still `override_options:`. And you can avoid doing any compiler checks or ifelse ladders.

subprojects/packagefiles/eyalroz-printf/test/meson.build
```diff
@@ -0,0 +1,75 @@
...
+  '-fdata-sections',
+  '-fverbose-asm',
+  '-Wunused-parameter',
+  '-ffat-lto-objects',
+  '-Wstrict-prototypes',
+  '/W4',  # msvc args
+]
+tests_c_args = cc.get_supported_arguments(
+  common_compiler_args,
+  '-std=c11',
```
eli-schwartz: What is the purpose of all these checks? `c_std=c11` in project default_options suffices.

-ffat-lto-objects makes *no sense*. It only has an effect on individual `*.o` files and exists in order to guarantee that `".a` archives are still usable after the compiler major version is upgraded, if and only if you also used `-flto`. For object files used to produce executables or shared libraries, it slows down the compilation to produce "fat" objects, which are never actually used in producing the executable or shared library (since the compiler uses the thin LTO section if the compiler version matches, which it always will inside a build directory, and throws out the fat section without even reading its contents).

And there is warning_level again! -Wall -Wextra -pedantic. It is joined by -g, which is already set by meson `-Ddebug=true`. Really, does the testsuite binary really need debug sections in order to correctly produce test results???

subprojects/packagefiles/eyalroz-printf/test/meson.build
```diff
@@ -0,0 +1,75 @@
...
+  '/std:c++11',
+)
+
+aliasing_exe = executable(
+  'aliasing',
+  'aliasing.c',
+  cfg_header_file,
+  c_args: [cc.get_supported_arguments('-fno-builtin-printf'), tests_c_args],
+  dependencies: [eyalroz_printf_dep],
+  include_directories: '..',
```
eli-schwartz, line 28: Use `build_by_default: false` for test executables. They don't need to be built when running `ninja && meson install`.

subprojects/packagefiles/eyalroz-printf/meson.build
```diff
@@ -0,0 +1,167 @@
...
+)
+eyalroz_printf_dep = declare_dependency(
+  include_directories: 'src',
+  link_with: eyalroz_printf_lib,
+  compile_args: [public_c_args],
+)
+meson.override_dependency('eyalroz-printf', eyalroz_printf_dep)
+
+if get_option('tests')
+  subdir('test')
```
eli-schwartz: ... Which also means you don't need an option here at all. You can unconditionally configure the test subdir, and it won't force users to build the test binaries. (Only running `meson test` will build them. And you can pass --suite to target only specific projects / subprojects.)

crvux:
@eli-schwartz I tried to port the logic with a minimum number of changes, that's why there are so many compiler flags in the tests 
https://github.com/eyalroz/printf/blob/master/test/CMakeLists.txt#L54
For now I still left some of the compiler warnings. Do you think it's worth removing?

I also had to comment out the build of one test (printf.test_suite), because ci run tests and one check is failed (in original cmake build is too). I hope this will be fixed in the next library release.

PR #2259, opened 2025-07-11
[argus] New wrap at v0.1.0

lucocozz:


bgilbert:
clang-cl needs `-D_CRT_SECURE_NO_WARNINGS` and MSVC needs

```
meson.version().version_compare('>=1.3.0') ? 'c_std=gnu11,c11' : 'c_std=gnu11'
```

assuming that C11 without GNU extensions is actually supported.

Also, if you're going to require Meson &ge; 1.1.0 anyway, you might as well rename `meson_options.txt` to `meson.options`.

None of those changes are required to merge this, since they're upstream changes.  But, FYI, since you're also the upstream.

lucocozz:
> clang-cl needs `-D_CRT_SECURE_NO_WARNINGS` and MSVC needs
> 
> ```
> meson.version().version_compare('>=1.3.0') ? 'c_std=gnu11,c11' : 'c_std=gnu11'
> ```
> 
> assuming that C11 without GNU extensions is actually supported.
> 
> Also, if you're going to require Meson ≥ 1.1.0 anyway, you might as well rename `meson_options.txt` to `meson.options`.
> 
> None of those changes are required to merge this, since they're upstream changes. But, FYI, since you're also the upstream.

I use C11 exstentions that not compatible with MSVC. I will add your suggestions in the next release in the case of possible improvement

bgilbert:
> I use C11 exstentions that not compatible with MSVC.

In that case, you can explicitly error out on MSVC.  If the `error()` call includes the string `unsupported`, `not supported`, or `does not support`, WrapDB CI can detect that the platform is not supported.

PR #2265, opened 2025-07-14
lua: update to 5.4.8

neheb:


subprojects/packagefiles/lua/meson.build
```diff
@@ -3,8 +3,8 @@ project(
   'c',
   license: 'MIT',
   meson_version: '>=0.49.2',
-  version: '5.4.6',
-  default_options: ['c_std=c99', 'warning_level=2'],
```
bgilbert, line 7: Looks like Lua only supports earlier than C99 if `LUA_USE_C89` is defined.

neheb: OK

bgilbert (approved):

PR #2266, opened 2025-07-14
c-ares: update to 1.34.5

neheb:


subprojects/packagefiles/c-ares/meson.build
```diff
@@ -1,14 +1,15 @@
 project(
   'c-ares',
   'c',
-  version: '1.34.4',
+  version: '1.34.5',
   license: 'MIT',
+  meson_version: '>=0.54.0',
```
bgilbert, line 6: Out of curiosity, why 0.54 and not 0.63?

neheb: Too new probably.

Ubuntu 22.04 comes with 0.61

neheb: OTOH properly removing cpp_std could be done with override_options probably.

neheb: wait a minute...this is a C library.

PR #2268, opened 2025-07-14
nng: update to 1.11

neheb:


neheb:
```
undefined reference to `__atomic_compare_exchange'
```

interesting...

neheb:
looks like x86 needs libatomic for 64-bit operations.

bgilbert (approved):

PR #2270, opened 2025-07-15
Add Apache Avro wrap

stephanlachnit:
Adds a wrap for the C++ bindings of [Apache Avro](https://github.com/apache/avro). Since it's only possible to build it without Boost using main, we need to wait until a new release.

TODO:

- [x] Get release tarball from https://dlcdn.apache.org/avro/ and adjust folder structure
- [x] Tests (optional)

/cc @WillAyd

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -274,6 +274,17 @@
       "0.9.10-1"
     ]
   },
+  "avro-cpp": {
+    "dependency_names": [
+      "avro-cpp"
+    ],
+    "program_names": [
+      "avrogencpp"
+    ],
+    "versions": [
+      "1.12.1-1"
+    ]
+  },
   "aws-c-common": {
     "dependency_names": [
       "aws-c-common"
--- a/releases.json
+++ b/releases.json
@@ -270,6 +270,17 @@
       "0.4.1-1"
     ]
   },
+  "avro-cpp": {
+    "dependency_names": [
+      "avro-cpp"
+    ],
+    "program_names": [
+      "avrogencpp"
+    ],
+    "versions": [
+      "1.12.2-1"
+    ]
+  },
   "aws-c-cal": {
     "dependency_names": [
       "aws-c-cal"
--- a/subprojects/avro-cpp.wrap
+++ b/subprojects/avro-cpp.wrap
@@ -1,0 +1,10 @@
+[wrap-file]
+directory = avro-cpp-1.12.1
+source_url = https://dlcdn.apache.org/avro/avro-1.12.1/cpp/avro-cpp-1.12.1.tar.gz
+source_filename = avro-cpp-1.12.1.tar.gz
+source_hash = 18a0d155905a4dab0c2bfd66c742358a7d969bcff58cf6f655bcf602879f4fe7
+patch_directory = avro-cpp
+
+[provide]
+dependency_names = avro-cpp
+program_names = avrogencpp
--- a/subprojects/avro-cpp.wrap
+++ b/subprojects/avro-cpp.wrap
@@ -4,6 +4,7 @@ source_url = https://dlcdn.apache.org/avro/avro-1.12.1/cpp/avro-cpp-1.12.1.tar.g
 source_filename = avro-cpp-1.12.1.tar.gz
 source_hash = 18a0d155905a4dab0c2bfd66c742358a7d969bcff58cf6f655bcf602879f4fe7
 patch_directory = avro-cpp
+diff_files = avro-cpp/fix_symbol_visibility.diff, avro-cpp/fix_datafile_32bit.diff
 
 [provide]
 dependency_names = avro-cpp
--- a/subprojects/avro-cpp.wrap
+++ b/subprojects/avro-cpp.wrap
@@ -1,0 +1,10 @@
+[wrap-file]
+directory = avro-cpp-1.12.2
+source_url = https://dlcdn.apache.org/avro/avro-1.12.2/cpp/avro-cpp-1.12.2.tar.gz
+source_filename = avro-cpp-1.12.2.tar.gz
+source_hash = 5e4cea5d9dc59bbc0e3e1967fd8f8a86b4cb7c1452c4b252561b8137e949f6fd
+patch_directory = avro-cpp
+
+[provide]
+dependency_names = avro-cpp
+program_names = avrogencpp
--- a/subprojects/packagefiles/avro-cpp/meson_options.txt
+++ b/subprojects/packagefiles/avro-cpp/meson_options.txt
@@ -1,0 +1,24 @@
+option(
+    'snappy',
+    type: 'feature',
+    value: 'auto',
+    description: 'Enable support for snappy codec',
+)
+option(
+    'zstd',
+    type: 'feature',
+    value: 'auto',
+    description: 'Enable support for zstd codec',
+)
+option(
+    'avrogencpp',
+    type: 'feature',
+    value: 'auto',
+    description: 'Build avrogencpp exectuable',
+)
+option(
+    'tests',
+    type: 'feature',
+    value: 'auto',
+    description: 'Build tests (requires avrogencpp)',
+)
--- a/ci_config.json
+++ b/ci_config.json
@@ -98,6 +98,26 @@
       "openssl"
     ]
   },
+  "avro-cpp": {
+    "build_options": [
+      "avro-cpp:snappy=enabled",
+      "avro-cpp:zstd=enabled",
+      "avro-cpp:avrogencpp=enabled"
+    ],
+    "_comment": "Tests require Boost, which is not available on WrapDB",
+    "alpine_packages": [
+      "boost-dev"
+    ],
+    "brew_packages": [
+      "boost"
+    ],
+    "debian_packages": [
+      "libboost-test-dev"
+    ],
+    "msys_packages": [
+      "boost"
+    ]
+  },
   "aws-c-common": {
     "build_options": [
       "aws-c-common:tests=enabled"
--- a/ci_config.json
+++ b/ci_config.json
@@ -100,6 +100,8 @@
   },
   "avro-cpp": {
     "build_options": [
+      "google-snappy:default_library=static",
+      "zstd:default_library=static",
       "avro-cpp:snappy=enabled",
       "avro-cpp:zstd=enabled",
       "avro-cpp:avrogencpp=enabled"
--- a/ci_config.json
+++ b/ci_config.json
@@ -99,6 +99,26 @@
       "openssl"
     ]
   },
+  "avro-cpp": {
+    "build_options": [
+      "avro-cpp:snappy=enabled",
+      "avro-cpp:zstd=enabled",
+      "avro-cpp:avrogencpp=enabled"
+    ],
+    "_comment": "Tests require Boost, which is not available on WrapDB",
+    "alpine_packages": [
+      "boost-dev"
+    ],
+    "brew_packages": [
+      "boost"
+    ],
+    "debian_packages": [
+      "libboost-test-dev"
+    ],
+    "msys_packages": [
+      "boost"
+    ]
+  },
   "aws-c-common": {
     "build_options": [
       "aws-c-common:tests=enabled"
--- a/ci_config.json
+++ b/ci_config.json
@@ -101,6 +101,8 @@
   },
   "avro-cpp": {
     "build_options": [
+      "google-snappy:default_library=static",
+      "zstd:default_library=static",
       "avro-cpp:snappy=enabled",
       "avro-cpp:zstd=enabled",
       "avro-cpp:avrogencpp=enabled"
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project(
   'wrapdb',
-  meson_version: '>=0.58.0',
+  meson_version: '>=0.63.0',
 )
 
 if (
--- a/subprojects/packagefiles/avro-cpp/fix_datafile_32bit.diff
+++ b/subprojects/packagefiles/avro-cpp/fix_datafile_32bit.diff
@@ -1,0 +1,13 @@
+diff --git a/impl/DataFile.cc b/impl/DataFile.cc
+index 5bf27c1ab..c9b9c8420 100644
+--- a/impl/DataFile.cc
++++ b/impl/DataFile.cc
+@@ -489,7 +489,7 @@ void DataFileReaderBase::readDataBlock() {
+         }
+ 
+         // Get the decompressed size
+-        size_t decompressed_size = ZSTD_getFrameContentSize(
++        unsigned long long decompressed_size = ZSTD_getFrameContentSize(
+             reinterpret_cast<const char *>(compressed_.data()), compressed_.size());
+         if (decompressed_size == ZSTD_CONTENTSIZE_ERROR) {
+             throw Exception("ZSTD: Not a valid compressed frame");
--- a/subprojects/packagefiles/avro-cpp/fix_symbol_visibility.diff
+++ b/subprojects/packagefiles/avro-cpp/fix_symbol_visibility.diff
@@ -1,0 +1,125 @@
+diff --git a/impl/json/JsonDom.hh b/impl/json/JsonDom.hh
+index 75728c6ba..b2be02b30 100644
+--- a/impl/json/JsonDom.hh
++++ b/impl/json/JsonDom.hh
+@@ -31,7 +31,7 @@
+ 
+ namespace avro {
+ 
+-class AVRO_DECL InputStream;
++class InputStream;
+ 
+ namespace json {
+ class Entity;
+@@ -59,7 +59,7 @@ enum class EntityType {
+     Obj
+ };
+ 
+-const char *typeToString(EntityType t);
++AVRO_DECL const char *typeToString(EntityType t);
+ 
+ inline std::ostream &operator<<(std::ostream &os, EntityType et) {
+     return os << typeToString(et);
+diff --git a/include/avro/Compiler.hh b/include/avro/Compiler.hh
+index bdcbb355e..7a97c4f4d 100644
+--- a/include/avro/Compiler.hh
++++ b/include/avro/Compiler.hh
+@@ -25,14 +25,14 @@
+ 
+ namespace avro {
+ 
+-class AVRO_DECL InputStream;
++class InputStream;
+ 
+ /// This class is used to implement an avro spec parser using a flex/bison
+ /// compiler.  In order for the lexer to be reentrant, this class provides a
+ /// lexer object for each parse.  The bison parser also uses this class to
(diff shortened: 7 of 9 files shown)


stephanlachnit:
Can we add `program_names` to a wrap for a program that might be disabled to be built or is that an issue?

subprojects/packagefiles/avro-cpp/lang/c++/meson.build
```diff
@@ -0,0 +1,106 @@
...
+  include_directories: avro_inc,
+  cpp_args: avro_flags,
+  dependencies: avro_deps,
+  version: meson.project_version(),
+  install: true,
+)
+
+install_subdir(
+  'include/avro',
+  install_dir: get_option('includedir'),
```
WillAyd: Hmm I don't think adding `install_dir` here is doing anything - you just want these files to go to the standard location (like /usr/local/include) right?

WillAyd (commented):
Cool - thanks for doing this!

subprojects/packagefiles/avro-cpp/lang/c++/meson.build
stephanlachnit: Yes, but `install_subdir` does have a default: https://mesonbuild.com/Reference-manual_functions.html#install_subdir_install_dir

WillAyd: Ah I see. In that case I think you need to do:

```python
install_subdir(
    'include/avro',
    install_dir: get_option('includedir') / 'avro',
    ...
)
```

At least that is what was done on wraps with a similar layout in the wrapdb (see the sparrow config as a reference).

There's a way to check locally if you cd into subprojects/avro-82a2bc8b034de34626e2ab8bf091234122474d50 , you can build the project there and try `meson install --dry-run` to confirm the installation paths are correct (I tried locally but the build was failing due to some const qualifier issue with fmt)

subprojects/packagefiles/avro-cpp/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option(
+    'build_executable',
+    type: 'boolean',
```
WillAyd: FYI on other Apache projects we have been using `feature` options instead of boolean. I don't think its a huge deal, but probably nice to stick with that pattern for consistency in the ecosystem

releases.json
```diff
@@ -208,6 +208,14 @@
       "0.4.1-1"
     ]
   },
+  "avro-cpp": {
```
WillAyd, line 273: You will want to add an entry here for `program_names` to house the avrogencpp program

subprojects/packagefiles/avro-cpp/meson_options.txt
WillAyd, line 1: Looks like CI will want you to reformat this and the main configuration file:

```shell
meson format --configuration meson.format --inplace subprojects/packagefiles/avro-cpp/meson_options.txt subprojects/packagefiles/avro-cpp/meson.build subprojects/packagefiles/avro-cpp/lang/c++/meson.build
```

subprojects/packagefiles/avro-cpp/lang/c++/meson.build
stephanlachnit: Seems to work for me:

```
Installing subdir subprojects/avro-*/lang/c++/include/avro to /tmp/usr/include/avro
Installing subprojects/avro-*/lang/c++/include/avro/AvroParse.hh to /tmp/usr/include/avro
```

WillAyd: OK cool - ignore what I said then. Thanks for checking

stephanlachnit: > At least that is what was done on wraps with a similar layout in the wrapdb (see the sparrow config as a reference).

Both seem to work, I guess Meson automatically strips the folder if they are identical

stephanlachnit:
@bgilbert any idea why the CI fails with this cryptic message:

```
Run ./tools/fake_tty.py ./tools/sanity_checks.py
  ./tools/fake_tty.py ./tools/sanity_checks.py
  shell: /usr/bin/bash -e {0}
  env:
    TEST_ANNOTATE_CONTEXT: yes
    TEST_SKIP_BUILD: yes
  File "/home/runner/work/wrapdb/wrapdb/./tools/fake_tty.py", line 6, in <module>
    sys.exit(os.waitstatus_to_exitcode(pty.spawn(sys.argv[1:])))
                                       ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/pty.py", line 195, in spawn
    os.execlp(argv[0], *argv)
  File "<frozen os>", line 563, in execlp
  File "<frozen os>", line 580, in execvp
  File "<frozen os>", line 603, in _execvpe
PermissionError: [Errno 13] Permission denied
Error: Process completed with exit code 1.
```

bgilbert:
> Can we add `program_names` to a wrap for a program that might be disabled

Yes, add a `skip_program_check` in `ci_config.json`.

> any idea why the CI fails with this cryptic message

The patch to `sanity_checks.py` removes its executable bit for some reason.

image

subprojects/packagefiles/avro-cpp/lang/c++/meson.build
```diff
@@ -0,0 +1,106 @@
...
+  install_tag: 'devel',
+)
+
+avro_dep = declare_dependency(
+  include_directories: avro_inc,
+  link_with: avro_lib,
+  dependencies: avro_public_deps,
+)
+
```
WillAyd: You will also want to add:

```meson
meson.override_dependency('avro', avro_dep)
```

releases.json
```diff
@@ -223,6 +223,17 @@
...
+    "dependency_names": [
+      "avro-cpp"
+    ],
+    "program_names": [
+      "avrogencpp"
+    ],
+    "versions": [
+      "1.12.99-1"
```
WillAyd: I think if you just include the hash here instead of the version you can work around the sanity check issue. So `82a2bc8b034de34626e2ab8bf091234122474d50-1`

subprojects/avro-cpp.wrap
```diff
@@ -0,0 +1,10 @@
+[wrap-file]
+directory = avro-82a2bc8b034de34626e2ab8bf091234122474d50
+source_url = https://github.com/apache/avro/archive/82a2bc8b034de34626e2ab8bf091234122474d50.tar.gz
+source_filename = arrow-82a2bc8b034de34626e2ab8bf091234122474d50.tar.gz
+source_hash = 1731c8aa96953caaa222ada25f8173a80d39d9ed7ffd181e62c25bf75076016d
+patch_directory = avro-cpp
+
+[provide]
+program_names = avrogencpp
+avro-cpp = avro_dep
```
WillAyd:
```suggestion
dependency_names = avro
```

(this pairs with the `meson.override_dependency('avro', avro_dep)` call)

subprojects/packagefiles/avro-cpp/lang/c++/meson.build
```diff
@@ -0,0 +1,106 @@
+avro_flags = []
+avro_deps = []
+avro_public_deps = []
+
+fmt_dep = dependency(
+  'fmt',
+  required: true,
+)
+avro_deps += fmt_dep
+avro_public_deps += fmt_dep
```
WillAyd: From a quick glance at the pkg-config template in avro, it also lists `@CODEC_PKG@` in the requires section, where CODEC_PKG is any of zlib, lzma and/or snappy. So two questions on that front:

1. Shouldn't the avro_public_deps include the compression dependencies?
2. Do we need to add lzma here as well? There exists a Meson wrapdb entry for it already

bgilbert: > Shouldn't the avro_public_deps include the compression dependencies?

Do clients of the library need to directly link with the compression dependencies or use their headers?  Some of the headers in `lang/c++/include/avro` `#include` headers from `fmt`, but I don't see any that `#include` compressor headers.

stephanlachnit: Yes, the codecs don't actually need to be linked, they are only used in this file: https://github.com/apache/avro/blob/557891c38c6c90eeaa43f1cb78ae8c3ae088009f/lang/c%2B%2B/impl/DataFile.cc

stephanlachnit:
@WillAyd how should the dependency be named? For avro-c the pkg-config file is named `avro-c`, but avro-cpp does not have a pkg-config file upstream. The CMake target is called `Avro` - should we go with that, or rather go with a more descriptive `avro-cpp`? Probably makes sense to add pkg-config support upstream then as well.

releases.json
stephanlachnit: I'm not quite sure what exactly will do the trick, since differences between the json and the meson build file will lead to some error at some point. Do you happen to know when there will be a new avro release? That might be the easiest...

WillAyd:
Hmm following the c pattern it sounds like avro-cpp is best. @eli-schwartz may have a suggestion here as well

eli-schwartz:
We definitely want to align with upstream. I agree that given the name "avro-c" it makes a decent amount of sense to use "avro-cpp" as well.

Avro upstream has only added CMake package config files for avro-cpp, not for avro-c, and it did that in January which hasn't been incorporated in a release yet. So it can be changed.

https://issues.apache.org/jira/browse/AVRO-3088 asked for "this way one could just use `find_package(avrocpp)` but it looks like unfortunately a decision was made to provide `find_package(Avro)` instead, which would mean it will be hard to have consistent naming.

WillAyd:
FYI @wgtmac - you may be interested in the discussion around the package name for Avro

wgtmac:
Its CMake project name is `project (Avro-cpp)`. So perhaps its package name and namespace should be `AvroCpp`? It is not too late to fix that before the next Avro release.

Proposed change: https://github.com/apache/avro/pull/3477

releases.json
WillAyd: Not sure. That's also probably a question for @wgtmac 

Although as far as the wrapdb is concerned, I think we will wait for the release before merging this. Any commit-hash inclusion here is just a workaround for now to trick CI

wgtmac: I'm not a maintainer of Apache Avro so I cannot make any promise here. Perhaps @Fokko @martin-g @thiru-mg can help with this?

wgtmac: BTW, iceberg-cpp is also depending on a hash tag of avro-cpp: https://github.com/apache/iceberg-cpp/blob/main/cmake_modules/IcebergThirdpartyToolchain.cmake#L186

stephanlachnit:
In principle it wouldn't be a problem to have a CMake target and pkg-config target with different names - this is also the case for zstd (pkg-config: libzstd). However then we should ensure that upstream actually generates a pkg-config file for the C++ version to be consistent. I opened a PR which will include one named `avro-cpp` (upper-case names are rather unusual for pkg-config files): https://github.com/apache/avro/pull/3479

eli-schwartz:
> In principle it wouldn't be a problem to have a CMake target and pkg-config target with different names - this is also the case for zstd (pkg-config: libzstd).

Let's stop right there.

Official meson policy is that it's a big problem, breaks assumptions, and requires what is officially termed as "a workaround" in projects that do `dependency('pkgconfig-name')` and consequently don't see any CMake package configs by other names. That means idiomatic projects cannot find the dependency at all, even if the CMake dependency finder is operational, if pkg-config isn't installed (and this scenario happens not infrequently on Windows).

I will add to this by pointing out that zstandard canonically, officially, doesn't provide CMake package config files at all. The sole *supported* build system is GNU Makefiles, all other build systems are unofficial. The unofficial meson support carefully matches the Makefiles, while the unofficial CMake and visual studio solution files do not. One of the ways the CMake unofficial files differ from the Makefiles is that they (alone) do provide unvetted CMake package config files that the zstandard developers have made no real effort to consider the correctness of.

You can't assume that those zstd CMake package config files exist at all -- e.g. there are some Linux distros that do provide them, while others do not provide them.

stephanlachnit:
Right, I have not considered the case where pkg-config is not installed, so point taken. I guess then if the CMake dependency is called `AvroCpp`, so should the pkg-config file. I will rename it in the PR then once a name is settled on in https://github.com/apache/avro/pull/3477.

WillAyd:
Thanks for all of the upstream work on this @wgtmac and @stephanlachnit !

stephanlachnit:
@WillAyd @wgtmac is there any ETA for 1.13? Would be nice so that we can publish this wrap :)

martin-g:
v1.13 of what ? 
If you talk about Apache Avro v1.13 then it will take a while!
There are talks about new releases of 1.12.x and 1.11.x for few months now. 1.13 has not be mentioned in any plans yet.

I'd recommend you to downport any changes from `main` (1.13) to `branch-1.12` if you need it sooner.

stephanlachnit:
Ah, I didn't expect that 1.12.1 included the changes which make boost optional, but indeed it is included as well as the package renaming we did. This should be enough to get the wrap working :+1:

wgtmac:
@martin-g Is there any blocker to release v1.13.0? It is worth doing at least for C++ since I have committed some codes that are dependent by Apache Iceberg C++.

martin-g:
> @martin-g Is there any blocker to release v1.13.0? It is worth doing at least for C++ since I have committed some codes that are dependent by Apache Iceberg C++.

Only the Rust SDK was extracted from the monorepo and can be released separately.
All other SDKs are released together and AFAIK it is a cumbersome process.

So far no one has mentioned releasing 1.13.
There are several regressions in 1.12 and releasing a new patch version for it is more important.

stephanlachnit:
Alright, this should be 95% there. The only thing missing is the `avrogencpp` tests which I will add later.

However there seems to be a problem with symbol visibility. In principle the annotations are there, but when I set it to `hidden` I get missing symbols when compiling the tests, and the msys2 CI seems to agree with that. Need to investigate and probably fix upstream.

Side note: the warning pragmas are incorrect since the assume MSVC on Windows, which is not always the case. Should be trivial to fix.

The test failure on Alpine armv7 and x86 I don't understand, there seems to be an actual issue in the code when compiling for 32-bit architectures with clang/libc++. First point to check might even be the compiler warnings generated there.

WillAyd:
> However there seems to be a problem with symbol visibility. In principle the annotations are there, but when I set it to `hidden` I get missing symbols when compiling the tests, and the msys2 CI seems to agree with that. Need to investigate and probably fix upstream.

From looking at the source, I believe you need to define `AVRO_DYN_LINK` and `AVRO_SOURCE` when building this project, and the former for downstream projects. So something like:

```python
avro_compile_args = []
if get_option('default_library') != 'static'
    avro_compile_args += ['-DAVRO_DYN_LINK']
endif

avro_lib = library(
    'avro',
    ...
    cpp_shared_args: ['-DAVRO_SOURCE'] + avro_compile_args,
)
avro_dep = declare_dependency(
    link_with: avro_lib,
    ...,
    compile_args: avro_compile_args,
)
```

It looks like upstream only tweaks symbol visibility when `_WIN32` is defined, so you'd have to upstream some changes to make `gnu_symbol_visibility: 'hidden'` work. But Windows seems like it can pass for now

stephanlachnit:
> From looking at the source, I believe you need to define AVRO_DYN_LINK and AVRO_SOURCE when building this project

This I already do:

```meson
# Set static and shared flags manually since Meson 1.3.0 is still too new
avro_flags += '-DAVRO_SOURCE'
if get_option('default_library') == 'shared'
  avro_flags += '-DAVRO_DYN_LINK'
elif get_option('default_library') == 'both'
  error('Building both static and shared libraries is not possible')
endif
```

> It looks like upstream only tweaks symbol visibility when _WIN32 is defined, so you'd have to upstream some changes to make gnu_symbol_visibility: 'hidden' work.

Ah, this is unfortunate, if there is already symbol visibility annotation, then they should be enabled at all times IMO since they don't give you any disadvantage.

WillAyd:
That looks good for the exporting library, but on Windows you also need the importing library (downstream project) to define `-DAVRO_DYN_LINK`

stephanlachnit:
> That looks good for the exporting library, but on Windows you also need the importing library (downstream project) to define `-DAVRO_DYN_LINK`

Ah right, I'll add this then.

stephanlachnit:
Ok, so I think I go the symbol visibility to work. There are some warning for certain inline methods but I don't think they are critical to fix. The only thing that is missing is the `armv7` and `x86` failure on Alpine, which seem to be actual 32-bit bugs. However, I don't think that those should be a blocker for WrapDB...?

In any case, we need to wait for 1.12.2 which includes the symbol visibility patch.

eli-schwartz:
> The only thing that is missing is the `armv7` and `x86` failure on Alpine, which seem to be actual 32-bit bugs. However, I don't think that those should be a blocker for WrapDB...?

You are correct -- they aren't blockers, just a useful heads up. Please report them upstream anyway, of course. :)

stephanlachnit:
@martin-g @WillAyd any news on a 1.12.2 release? I saw https://github.com/apache/avro/pull/3724 but not much activity otherwise

martin-g:
I don't have any other information too.

manuzhang:
@stephanlachnit [avro 1.12.2 has been released](https://github.com/apache/avro/releases/tag/release-1.12.20), can you revive this PR?

stephanlachnit:
> @stephanlachnit [avro 1.12.2 has been released](https://github.com/apache/avro/releases/tag/release-1.12.20), can you revive this PR?

Thanks for the headsup - I rebased and now it seems to work nicely (Alpine CI failures are unrelated).

stephanlachnit:
There seems to be an interesting failure on x86 and armv7:

```
../subprojects/avro-cpp-1.12.2/test/DataFileTests.cc(428): error: in "DataFile tests_ test9.df/DataFileTest__testReaderSyncSeek": check num == count has failed [0 != 1000]
../subprojects/avro-cpp-1.12.2/test/DataFileTests.cc(430): error: in "DataFile tests_ test9.df/DataFileTest__testReaderSyncSeek": check actual.size() == count has failed [0 != 1000]
../subprojects/avro-cpp-1.12.2/test/DataFileTests.cc(469): error: in "DataFile tests_ test10.df/DataFileTest__testReaderSyncDiscovery": check sync_points_syncing == sync_points_reading has failed
../subprojects/avro-cpp-1.12.2/test/DataFileTests.cc(472): error: in "DataFile tests_ test10.df/DataFileTest__testReaderSyncDiscovery": check sync_points_syncing.size() > 10 has failed [1 <= 10]
../subprojects/avro-cpp-1.12.2/test/DataFileTests.cc(497): error: in "DataFile tests_ test11.df/DataFileTest__testReaderSplits": check actual_count == count has failed [0 != 1000]
```

stephanlachnit:
@WillAyd if you want to take another at this PR, it is ready to be merged

WillAyd (approved):
lgtm

PR #2271, opened 2025-07-15
pcre2: improve function checks and reliability

carenas:
Eliminate the risk of accidental crashes that was created by enabling by default an experimental feature.

Reduce the number of checks that will be done for most cases, and improve their reliability.


eli-schwartz:
Please also bump the revision in releases.json in order to release an updated wrap that has these fixes.

Otherwise lgtm.

subprojects/packagefiles/pcre2/meson.build
```diff
@@ -133,20 +133,27 @@ foreach h : check_headers
...
 endforeach
 
-check_funs = [
-  '__builtin_unreachable',
-  'bcopy',
-  'memmove',
-  'realpath',
-  'strerror',
-]
+check_funs = ['bcopy', 'memmove', 'realpath', 'strerror']
```
eli-schwartz: Pfft, mandatory formatting that collapses arrays of check functions if it's possible to rewrite them to be all ugly and on one line. Remind me why people like mandatory autoformatters?

bgilbert: Because it's still better than the inconsistent mess we had before?

neheb: Personally I prefer

```
foreach f : ['bcopy', 'memmove', 'realpath', 'strerror']
```

since it's so short.

bgilbert:
What's the macro definition mangling that was corrected in a couple cases?  From looking at the code I'd expect the macros to be the same before and after the changes.

subprojects/packagefiles/pcre2/meson.build
eli-schwartz: Muon fmt at least has a quirk, trailing commas indicate you want arrays to be broken across lines.

@bruchar1

carenas:
> What's the macro definition mangling that was corrected in a couple cases?

Since b89abe5c (pcre2: update to 10.43 (#1411), 2024-03-01), all '_' characters from the function names were being strip()

> From looking at the code I'd expect the macros to be the same before and after the changes.

The new version doesn't use strip() when checking at function names, and has moved those `__builtin*` that started that logic out of any function array as well.

bgilbert:
`strip()` is supposed to remove the specified characters only from the beginning and end of the string, not from anywhere in it.  Are you seeing different behavior?

carenas:
> Are you seeing different behavior?

Just a misunderstanding then, should I squash all commits and fix that in the description?

subprojects/packagefiles/pcre2/meson.build
bruchar1: You are right. I didn't realized this. It shouldn't be difficult to implement.

carenas: the number of functions on each array is short enough that maybe using it as a literal as @neheb suggested will please the autoformatter and avoid the ugly looking array definition.

bgilbert:
The changes are nicely broken out into separate commits so I'd say they don't need to be squashed.  Please do update the commit messages though.

PR #2272, opened 2025-07-15
gdbm: update to 1.25

neheb:


bgilbert:
macOS failure is a regression that's since been fixed upstream: https://github.com/mesonbuild/wrapdb/pull/1989#issuecomment-2756900109

PR #2273, opened 2025-07-16
Update nanobind to 2.8.0

hpkfft:
Also, improve `meson.build` as follows:
+ Require robin-map version >=1.4.0
+ Only strip output objects if not debug build
+ Add variable 'stubgen' to nanobind_dep

@WillAyd, @oscargus: Comments (and testing) welcome!

Also, if there's a more idiomatic way to obtain the absolute path of the `stubgen.py` script, please let me know.


subprojects/packagefiles/nanobind/meson.build
```diff
@@ -42,19 +42,27 @@ if freethreaded
...
-  if py_soabi.startswith('cp')
-    resp_file = meson.project_source_root() / 'cmake/darwin-ld-cpython.sym'
-  else
-    resp_file = meson.project_source_root() / 'cmake/darwin-ld-pypy.sym'
+if not get_option('debug')
+  dep_compile_args += ['-DNB_COMPACT_ASSERTIONS']
+
+  # The following, from nanobind_strip(), applies both to the nanobind shared
+  # library (if it's built) and to the developer's extension:
+  if host_machine.system() == 'darwin'
```
WillAyd, line 50: Just curious - why is this moved in the `not get_option('debug')` branch? Wouldn't the linker flags still be required in debug builds?

hpkfft: I think you're right about `Wl,@ + resp_file`.  Thanks!
For the rest, no, these linker flags strip out symbols, which we want to leave in a debug build.

Note that the nanobind cmake function `nanobind_strip`, which is where one finds these flags, only applies if `${NB_OPT}` is true.  (It's done using generator expressions at line 315.  Cmake is confusing!)  At line 90, `NB_OPT` is set if the build configuration is `Release` or `MinSizeRel`.

I'll force-push the change for the response file.  (I believe that process is preferred by this project.)

It would be lovely if someone could test `--buildtype release` and `--buildtype debug` on MacOS; I cannot.

WillAyd: Makes sense. I also cannot verify on macOS - we might just have to trust what we have and leave it to someone that can to fix later, if there is an issue :-)

WillAyd (approved):

oscargus:
I have a mac at my office for testing purposes, but I do not have any plans to visit that for another two-three weeks...

Hence, I think this should be good to go and we can sort out any issues if they arise.

Thanks for working on this!

PR #2274, opened 2025-07-17
tools: add `meson format` wrapper script

bgilbert:
`meson format` needs a few arguments and it's annoying to copy the command from `sanity_checks.py` output.  Add a wrapper script that passes the proper args and only formats files in updated wraps.


tools/format.py
```diff
@@ -0,0 +1,59 @@
...
+
+def main() -> None:
+    with open('releases.json', 'r', encoding='utf-8') as f:
+        releases = json.load(f)
+
+    tags = set(
+        subprocess.check_output(
+            ['git', 'tag', '--merged'], text=True
+        ).splitlines()
+    )
```
eli-schwartz: This is only 91 characters long if you put all 5 lines on one line. I cannot imagine why a human would agree it's better to make it 5 lines...

tools/format.py
```diff
@@ -0,0 +1,59 @@
...
+                ]
+
+    if files:
+        cmd: list[str | Path] = [
+            "meson",
+            "format",
+            "--configuration",
+            "./meson.format",
+            "--inplace",
+        ] + files
```
eli-schwartz: Somewhat intrigued that this is the only part of the file using double quotes.

Also it is probably considerably simpler and avoids needing explicit annotations if you define cmd in one pass, then invoke `subprocess.run(cmd + files, check=True)`. It's even got an advantage in being easier to tell at a glance that there's two logical parts to the command: a static base command and a list of files.

tools/format.py
bgilbert: Because my terminal is 80 characters wide as tradition demands.  :slightly_smiling_face:

tools/format.py
bgilbert: Yup, both good catches.  The double quotes are because this was assembled from pieces of `sanity_checks.py`.

tools/format.py
bgilbert: Yeah, could have done some asymmetric thing.  Anyway it's unwrapped now.

PR #2281, opened 2025-07-21
protobuf: do not use -march=native

meator:
There is no native arch when cross compiling. This decision should be left to the wrap user, it should not be forced by default.

`protobuf` is one of the only wraps that set `-march` (to my knowledge). And it does so not only once, but twice:

https://github.com/mesonbuild/wrapdb/blob/575af61a1883376bb00d684ea4fa8b9dffd81181/subprojects/packagefiles/protobuf/src/meson.build#L44

I don't find this second `-march=native` useful either, but it at least doesn't break cross compilation.

This PR should fix CI failure for #2212


subprojects/packagefiles/protobuf/meson.build
```diff
@@ -66,13 +66,20 @@ libutf8_range_src = files(
   'third_party/utf8_range/naive.c',
   'third_party/utf8_range/range-avx2.c',
   'third_party/utf8_range/range-neon.c',
-  'third_party/utf8_range/range-sse.c',
   'third_party/utf8_range/utf8_validity.cc',
 )
+libutf8_range_args = []
+if build_machine.cpu_family() == 'x86_64'
```
eli-schwartz: Why build machine?

subprojects/packagefiles/protobuf/meson.build
```diff
@@ -66,13 +66,20 @@ libutf8_range_src = files(
...
+  # range-sse.c is wrapped in #ifdef __x86_64__
+  libutf8_range_src += files('third_party/utf8_range/range-sse.c')
+  # It requires -march=native (or some other more targetted way)
+  # to be able to use SSE.
```
eli-schwartz: I can surely think of a targeted way to be able to use SSE!

eli-schwartz (changes requested):

subprojects/packagefiles/protobuf/meson.build
meator: I tried to do that now. This complicates the logic, because compilers with gcc argument syntax, compilers with msvc argument syntax and strangely enough clang-cl all behave a bit different from each other.

I first wanted to leave `-march=native` as is because I assumed it worked everywhere since it was there before, and I wanted to make the least amount of changes to fix the issue.

meator:
If passing all CI checks is a requirement, this PR is currently blocked on #2282

meator:
Should I deal with the second `-march=native` (currently untouched by this PR) too?

subprojects/packagefiles/protobuf/meson.build
eli-schwartz: > I first wanted to leave `-march=native` as is because I assumed it worked everywhere since it was there before, and I wanted to make the least amount of changes to fix the issue.

Obviously it cannot possibly work everywhere, given that it may *compile* everywhere but it will SIGILL when you try to run the installed artifacts on any other computer.

Using `-march=native` could make sense for `native: true` programs only used as uninstalled code generators. It should be forbidden anywhere else...

meator:
`clang` reports that the minimum required SSE version is 4.1:

<details><summary>Details</summary>
<p>

```
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:121:33: error: always_inline function '_mm_shuffle_epi8' requires target feature 'ssse3', but would be inlined into function 'utf8_range' that is compiled without support for 'ssse3'
  121 |             __m128i first_len = _mm_shuffle_epi8(first_len_tbl, high_nibbles);
      |                                 ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:125:29: error: always_inline function '_mm_shuffle_epi8' requires target feature 'ssse3', but would be inlined into function 'utf8_range' that is compiled without support for 'ssse3'
  125 |             __m128i range = _mm_shuffle_epi8(first_range_tbl, high_nibbles);
      |                             ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:131:28: error: '__builtin_ia32_palignr128' needs target feature ssse3
  131 |                     range, _mm_alignr_epi8(first_len, prev_first_len, 15));
      |                            ^
/usr/lib/clang/18/include/tmmintrin.h:157:13: note: expanded from macro '_mm_alignr_epi8'
  157 |   ((__m128i)__builtin_ia32_palignr128((__v16qi)(__m128i)(a), \
      |             ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:137:19: error: '__builtin_ia32_palignr128' needs target feature ssse3
  137 |             tmp = _mm_alignr_epi8(first_len, prev_first_len, 14);
      |                   ^
/usr/lib/clang/18/include/tmmintrin.h:157:13: note: expanded from macro '_mm_alignr_epi8'
  157 |   ((__m128i)__builtin_ia32_palignr128((__v16qi)(__m128i)(a), \
      |             ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:146:19: error: '__builtin_ia32_palignr128' needs target feature ssse3
  146 |             tmp = _mm_alignr_epi8(first_len, prev_first_len, 13);
      |                   ^
/usr/lib/clang/18/include/tmmintrin.h:157:13: note: expanded from macro '_mm_alignr_epi8'
  157 |   ((__m128i)__builtin_ia32_palignr128((__v16qi)(__m128i)(a), \
      |             ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:170:22: error: '__builtin_ia32_palignr128' needs target feature ssse3
  170 |             shift1 = _mm_alignr_epi8(input, prev_input, 15);
      |                      ^
/usr/lib/clang/18/include/tmmintrin.h:157:13: note: expanded from macro '_mm_alignr_epi8'
  157 |   ((__m128i)__builtin_ia32_palignr128((__v16qi)(__m128i)(a), \
      |             ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:179:22: error: always_inline function '_mm_shuffle_epi8' requires target feature 'ssse3', but would be inlined into function 'utf8_range' that is compiled without support for 'ssse3'
  179 |             range2 = _mm_shuffle_epi8(df_ee_tbl, tmp);
      |                      ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:181:43: error: always_inline function '_mm_shuffle_epi8' requires target feature 'ssse3', but would be inlined into function 'utf8_range' that is compiled without support for 'ssse3'
  181 |             range2 = _mm_add_epi8(range2, _mm_shuffle_epi8(ef_fe_tbl, tmp));
      |                                           ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:186:28: error: always_inline function '_mm_shuffle_epi8' requires target feature 'ssse3', but would be inlined into function 'utf8_range' that is compiled without support for 'ssse3'
  186 |             __m128i minv = _mm_shuffle_epi8(range_min_tbl, range);
      |                            ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:187:28: error: always_inline function '_mm_shuffle_epi8' requires target feature 'ssse3', but would be inlined into function 'utf8_range' that is compiled without support for 'ssse3'
  187 |             __m128i maxv = _mm_shuffle_epi8(range_max_tbl, range);
      |                            ^
../subprojects/protobuf-25.2/third_party/utf8_range/range-sse.c:220:14: error: always_inline function '_mm_testz_si128' requires target feature 'sse4.1', but would be inlined into function 'utf8_range' that is compiled without support for 'sse4.1'
  220 |         if (!_mm_testz_si128(error, error))
      |              ^
```

</p>
</details> 

Here's what args everyone supports:

| gcc + clang | [msvc](https://learn.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-170) | clang-cl        |
| ----------- | ----            | --------        |
| `-msse`     | `/arch:SSE2`    | `-msse4.1`      |
| `-msse2`    | `/arch:SSE4.2`  | `-msse4.2`      |
| `-msse3`    | `/arch:AVX`     | `/arch:AVX`     |
| `-mssse3`   | `/arch:AVX2`    | `/arch:AVX2`    |
| `-msse4.1`  | `/arch:AVX512`  | `/arch:AVX512`  |
| `-msse4.2`  | `/arch:AVX10.1` | `/arch:AVX512F` |
| `-msse4`    |                 |                 |

Interestingly `clang-cl` doesn't support MSVC's `/arch:SSE*` flags, so I've resorted to using `clang` flags instead. It also doesn't support `-msse4`. I therefore use `-msse4.1` in gcc, clang and clang-cl, as everything seems to support that.

MSVC's SSE args aren't as granular as on Linux, so I chose `/arch:SSE4.2` (because there is no `/arch:SSE4.1`).

I determined this by examining compiler flag documentation and error messages, I do not actually know how these things work. If anyone here has more insight into this, please comment here.

Here's sample code which has the same SSE requirements as protobuf (useful for testing):

```c
#include <smmintrin.h>

int main()
{
	__m128i vec1 = _mm_set1_epi32(0);
	__m128i vec2 = _mm_set1_epi32(0);
	return _mm_testz_si128(vec1, vec2);
}
```

subprojects/packagefiles/protobuf/meson.build
meator: > Using `-march=native` could make sense for `native: true` programs only used as uninstalled code generators. It should be forbidden anywhere else...

I'll leave the second `-march=native` be then.

I have explained my reasoning for choosing the appropriate flags in https://github.com/mesonbuild/wrapdb/pull/2281#issuecomment-3101296549. I'll mark this as resolved.

PR #2291, opened 2025-07-26
libpng: Generate pkgconfig file via Meson module. Introduce Meson option for enable/disable tests

Zaburunier:


bgilbert:
Thanks for the PR.  Does this solve a problem you're having, or is it just a cleanup?

Zaburunier:
> Thanks for the PR. Does this solve a problem you're having, or is it just a cleanup?

@bgilbert Hey there! Yeah there was a problem with a libpng.pc file. Meson-generated file specifies `Libs: -lz` while autoconf-generated file would have `Libs: -lz -lm`.

This caught me unexpectedly while experimenting with static build pipelines. I have got compiler error that some math functions not found

subprojects/packagefiles/libpng/meson.build
```diff
@@ -109,32 +109,20 @@ install_headers(
...
-  input: 'libpng.pc.in',
-  output: 'libpng16.pc',
-  configuration: cdata,
-  install_dir: get_option('libdir') / 'pkgconfig',
-  install: true,
+pkgconfig = import('pkgconfig')
+
+pkgconfig.generate(
+  libpng,
+  name: 'libpng16',
```
bgilbert: Should be `libpng` to match upstream.

subprojects/packagefiles/libpng/meson.build
```diff
@@ -109,32 +109,20 @@ install_headers(
...
+  description: 'Loads and saves PNG files',
+  # -lz added just to match pkgconfig file generated via GNU autoconf
+  libraries_private: ['-lz'],
```
bgilbert: Actually I don't think we need this.  Upstream does indeed have it, but both upstream and the Meson-generated `.pc` file also have `Requires.private: zlib` which should do the same thing.

subprojects/packagefiles/libpng/meson.build
```diff
@@ -109,32 +109,20 @@ install_headers(
...
 )
 
-# FIXME: this should be a symlink to libpng16.pc
-configure_file(
-  input: 'libpng.pc.in',
-  output: 'libpng.pc',
-  configuration: cdata,
+install_symlink(
```
bgilbert: To avoid bumping the minimum Meson version, I'd suggest replacing this with a second `pkgconfig.generate()` (or rather, calling it in a loop) with a different `filebase`.

subprojects/packagefiles/libpng/meson.build
```diff
@@ -109,32 +109,20 @@ install_headers(
...
-
-configure_file(
-  input: 'libpng.pc.in',
-  output: 'libpng16.pc',
-  configuration: cdata,
-  install_dir: get_option('libdir') / 'pkgconfig',
-  install: true,
+pkgconfig = import('pkgconfig')
+
+pkgconfig.generate(
```
bgilbert: Needs a `subdirs` argument to correctly set the include path.

subprojects/packagefiles/libpng/meson.build
Zaburunier: Ah, I followed existing Meson version instead of upstream`s autoconf :slightly_frowning_face: 

`# FIXME: this should be a symlink to libpng16.pc` - confusing note...

subprojects/packagefiles/libpng/meson.build
neheb:
```
Requires.private: zlib
```

is more correct than -lz for what it's worth.

neheb:
@bgilbert you think it's worth it to remove all of these pc.in configurations?

```
git grep \\\.pc\\\.in
subprojects/packagefiles/argparse/meson.build:  input: 'packaging/pkgconfig.pc.in',
subprojects/packagefiles/asio/meson.build:  input: 'asio.pc.in',
subprojects/packagefiles/cairomm/data/meson.build:  input: 'cairomm.pc.in',
subprojects/packagefiles/cairomm/data/meson.build:    input: pc_file + '.pc.in',
subprojects/packagefiles/cppzmq/meson.build:  input: 'cppzmq.pc.in',
subprojects/packagefiles/libpng/meson.build:  input: 'libpng.pc.in',
subprojects/packagefiles/libpng/meson.build:  input: 'libpng.pc.in',
subprojects/packagefiles/libsigcplusplus-3/meson.build:  input: 'sigc++.pc.in',
subprojects/packagefiles/libsigcplusplus-3/meson.build:  input: 'sigc++-uninstalled.pc.in',
subprojects/packagefiles/xxhash/meson.build:  input: 'libxxhash.pc.in'
```

subprojects/packagefiles/libpng/meson.build
bgilbert: Well, that's frustrating.  In that case, I think the right fix is to install the symlink if Meson is new enough, and call `pkgconfig.generate()` a second time otherwise.  libpng is foundational enough that maintaining the broader Meson compatibility range seems worthwhile.

Zaburunier: Sorry I thought you did not see this :D Found workaround for this problem, sent another commit

bgilbert: Hmm.  That does indeed work around the issue, but seems a bit awkward.  There's a risk of the two `.pc` files diverging, and also the awkwardness executes on all Meson versions instead of only legacy ones.  So I'm still in favor of using the symlink when we can; that lets us match upstream more closely, and also we can eventually remove the workaround.

bgilbert: For the historical record, the deleted comment was:

> Working on this, tried to do following your suggestion:
>
>     libpng_pkgconf_names = ['libpng', 'libpng16']
>      
>     foreach libpng_pkgconf_name : libpng_pkgconf_names
>       pkgconfig.generate(
>         libpng,
>         name: libpng_pkgconf_name,
>         subdirs: pngincsubdir,
>         description: 'Loads and saves PNG files',
>       )
>     endforeach
>
> Got warning `libpng| WARNING: Already generated a pkg-config file for png16 `🙁 PR checks will fail due to this.
>
> According to documentation and sources, `pkgconfig.generate` does not return any public value which I could use with `configure_file()` to replicate `libpng.pc`

Zaburunier: Yeah I totally agree. I will rollback `install_symlink()` way, now with dangerous fallback for Meson < 0.61.0

bgilbert: To be clear, I was thinking we'd call `pkgconfig.generate()` again in the fallback path.  CI won't test that path, and the extra warning is harmless (if it's even emitted by Meson that old).

Zaburunier: I think you are right, but strangely two pkgconfigs differs which did not happen while generating with workaround :thinking:

bgilbert:
@neheb Yeah, I do.

subprojects/packagefiles/libpng/meson.build
bgilbert: Apparently for the second `.pc` file Meson depends on the first one with `Requires:` rather than replicating it.  Actually, that should work fine, and if we explicitly ask it to do that we can avoid the warning to console.  I've submitted a PR to your PR here: https://github.com/Zaburunier/meson-wrapdb/pull/1

Zaburunier: This is much better now, thank you! PR accepted.

Zaburunier:
@bgilbert @neheb I suggest we are waiting for final checks and then hopely for merge. Can you help with these? :slightly_smiling_face: 

UPD: I guess it should be done with https://github.com/mesonbuild/wrapdb/pull/2292 simultaneously

subprojects/packagefiles/libpng/meson.build
```diff
@@ -146,6 +118,32 @@ libpng_dep = declare_dependency(
...
+  libpng,
+  name: 'libpng',
+  description: 'Loads and saves PNG files',
+  filebase: 'lib' + png_libname,
+  subdirs: pngincsubdir,
+)
+
+if meson.version().version_compare('>=0.61')
+  install_symlink(
```
bgilbert, line 131: So, it seems Windows still doesn't allow non-administrators to create symlinks unless developer mode or a group policy is enabled on the machine, and I don't think we can assume that they are.  The conditional should be updated to exclude Windows, or perhaps we should remove the symlink case entirely for simplicity.  The pkg-config `Requires` indirection seems fine.

Sorry for all the churn on this one.

Zaburunier: Well, that is frustrating too! Won`t this be a problem on OSX? (zero experience with Apple MacBook)

bgilbert: As far as I know, OSX supports symlinks without restriction, on native filesystems at least.

eli-schwartz: Symlinks are quite standard for POSIX platforms. As a general rule I found that Windows is usually the sole edge case whenever some edge case comes up.

The notion that a symlink is such a trusted operation that it should be security-restricted from use by non admins, is a uniquely Microsoft form of utter insanity.

Zaburunier: Well, that is great news from some perspective.

In that case I would propose to stick with the `install_symlink()` call as this is the common way for most usecases, and keep two fallbacks (old Meson, Windows build OS) until they won`t be useful.

I will add the Windows branch tomorrow

Zaburunier: @bgilbert Done

Zaburunier:
@bgilbert sorry for pinging you again but I think we have to decide whether to accept PR or to close it because there are probably no unanswered questions left.

UPD: same for https://github.com/mesonbuild/wrapdb/pull/2292 because they can be merged into one wrap release I believe.

bgilbert:
Pings are okay if a PR really seems to have stalled, but note that there are no service time guarantees.  I'll get back to this on my next review pass through the repo.

What you can do in the meantime is squash your commits here, add the commit from #2292 as a separate commit, and close #2292.

bgilbert:
Thanks for all your work on this!

PR #2292, opened 2025-07-26
libpng: Introduce Meson option for enable/disable tests

Zaburunier:
This is just convenience change as many projects allow toggling tests off.

Note that this PR has same version with https://github.com/mesonbuild/wrapdb/pull/2291 . They have nothing in common and thus separated as two PRs, but release version increments are the same.


subprojects/packagefiles/libpng/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option(
+    'tests',
+    type: 'boolean',
+    value: true,
```
bgilbert, line 4: `true` is the default.

Zaburunier: > `true` is the default.

Yeah I know. This is not for changing default Meson value at all. I used to write default value always just to be explicit and have more clean style. Got this from some other projects, e. g. https://gitlab.gnome.org/GNOME/glib/-/blob/main/meson.options

By the way, we can see the same on current repo: https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/libffi/meson_options.txt

bgilbert: The current repo contains a lot of things.  :slightly_smiling_face:  But yeah, I'm okay with leaving it.

Zaburunier: Well, I don`t know personally anyone who has significant experience with Meson. I would happy to hear whether some writing style is common bad practice :slightly_smiling_face:

bgilbert: There's redundancy in repeating a default, but also the fact that booleans default to `true` is unusual and not necessarily obvious.  Personally I'd remove, but there are indeed arguments both ways.

Zaburunier:
Closing this PR as it is added into https://github.com/mesonbuild/wrapdb/pull/2291

PR #2293, opened 2025-07-27
xxhash: use pkgconfig generator

neheb:
Also remove GCC inline workaround as a fix was merged upstream.


neheb:
hrm extra fix is needed

```
./bin/libxxhash.dll
./include/llvm/Support/xxhash.h
./include/xxhash.h
./lib/libxxhash.a
./lib/libxxhash.dll.a
./lib/pkgconfig/libxxhash.pc
./share/licenses/xxhash
./share/licenses/xxhash/LICENSE
```

neheb:
ping @tristan957

PR #2303, opened 2025-07-30
libgrapheme: use find_program instead of relying on shebang

Gottox:
Apparently during testing it slipped through that meson does not apply file permissions from packagefiles. So we do not execute the chdir_wrapper directly now. Instead we search for a python executable and call the interpreter with the script instead.


Gottox:
libgrapheme is upstream-broken on windows. It seems to be fixed in master, so I would suggest to ignore the broken CI here.

subprojects/packagefiles/libgrapheme/gen/meson.build
```diff
@@ -24,6 +24,13 @@ gen_headers_tgt = {
 }
 
 chdir_wrapper = files('chdir_wrapper.py')
```
bgilbert, line 26: You should be able to undo all of these changes (including the removal of the `#!` line) and just change this to
```suggestion
chdir_wrapper = find_program('chdir_wrapper.py')
```

Meson has a special case for scripts starting with `#!/usr/bin/env python3`, so execute permission shouldn't be needed.  However, with `custom_target()` this works for `find_program()` but not for `files()`.

bgilbert (commented):
CI will require you to convert the wrap to the `dependency_names` provide syntax.  See the CI log for details.

PR #2308, opened 2025-08-06
openssl: update to 3.0.21

amyspark:
Hi all,

This is a PR to update OpenSSL 3.x to 3.0.21, and also to upstream the updates I'd made to my fork of the 1.x wrap.

The new logic enables:
- capturing and building Windows resource files
- support MSVC 32 and 64-bits x86
- support for MSVC with Nasm (bumps version to 0.64 to make it Meson managed)
- support for MinGW with GAS

I've also slightly modified the Makefile and generator script to be easier to re-run when debugging updates, and with support for running in platforms other than Linux (I daily drive either macOS or Windows).

All feedback is appreciated.

EDIT Aug 20, 2026: I bumped this to the latest 3.0 LTS version, which required an additional fix to Node's GYPI generator. Upstream's tag includes a half-baked update to 3.5, which needs to be reverted.
I also updated the patches to account for the latest MINGW64 Perl not including `Text::Template`, and its `system` command using `CreateProcess` instead of spawning a shell.


Jannik2099:
do you have any interest in pursuing this? All CI runs are failing with
`subprojects/openssl-3.0.17/meson.build:388:3: ERROR: Unknown variable "libcrypto_def"`

amyspark:
@Jannik2099 took me a while to catch up, I've fixed that. It should be back to the missing Perl packages error.

bgilbert:
#2554 updated to 3.0.10, so this needs to be rebased.

amyspark:
Working on it, Node's 22.19.0 LTS has a broken GYPI generator that needs to be addressed.

amyspark:
Update done, and PR description fixed.

PR #2314, opened 2025-08-13
zlib: bump to fix licenses

meator:
These two packages contain invalid license info (license names not conforming to SPDX) in their latest versions. The issue has been fixed in both of these wraps already, but it has not been released yet. This PR bumps these two.

My project uses Meson's dependency manifest as input to generate SBOMs. I want to get this fixed so that my scripts receive valid input.


meator:
`Alpine (riscv64)` and `Alpine (ppc64le)` CI failures should be fixed by https://github.com/mesonbuild/wrapdb/pull/2281 (which by the way I would really appreciate if someone could merge, since several of my PRs are potentially blocked on that).

The VisualStudio failures also aren't my fault. Other recent PRs are hit by this. Here's some debugging I've done:

```
Message: name = protoc
Message: prog.full_path() = D:/a/wrapdb/wrapdb/_build/subprojects/protobuf-25.2/src/protoc.exe
Message: meson.global_build_root() = D:\a\wrapdb\wrapdb\_build
Message: meson.global_source_root() = D:\a\wrapdb\wrapdb

meson.build:25:2: ERROR: Assert failed: prog.full_path().startswith(
meson.global_build_root()
) or prog.full_path().startswith(
meson.global_source_root()
)
```

(The `Message:`s are added by me.)

The issue seems to be with a mismatch between POSIX-style and DOS-style paths.

neheb:
feel free to rebase.

neheb:
For the VS failures, you could try .substring() and see if that works.

meator:
Thank you @neheb!

Merging https://github.com/mesonbuild/wrapdb/pull/2281 has bumped protobuf, so I no longer need to bump it here.

meator:
> For the VS failures, you could try .substring() and see if that works.

I don't think I understand.

Eli Schwartz has mentioned that this could have been caused by a breaking change in the newest release candidate of Meson (but it was a guess). If that is indeed the case, there is little mesonbuild/wrapdb could do to fix this.

PR #2315, opened 2025-08-13
sanity_checks: add tip about using old format

meator:
I spent a lot of time trying to follow this error message's advice to run `tools/format.py` without success. It took me a while to figure out that newer versions of Meson modified its formatting rules, so a file that was correctly formatted with Meson 1.6.1 (to give an example) will be considered unformatted by Meson 1.9.0rc2 (or whatever the runners use now). 1.6.1 isn't even _that_ old.

The wording may need to be changed, this PR is more of a proposal.

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -692,7 +692,7 @@ class TestReleases(unittest.TestCase):
             unformatted_str = ', '.join([str(f) for f in unformatted])
             self.fail(
                 f'''Not formatted files found: {unformatted_str}
-Run tools/format.py to format these files.''')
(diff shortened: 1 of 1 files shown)


bgilbert:
Perhaps `format.py` should install the latest Meson release into a virtualenv and run it from there, either always or only if the system's Meson is too old.

bgilbert:
https://github.com/mesonbuild/wrapdb/pull/2325 changes `format.py` to run Meson from a virtualenv it maintains with the latest prerelease.

meator:
Wow, thanks @bgilbert!

neheb:
No longer relevant.

PR #2318, opened 2025-08-14
libusb: add license

meator:


subprojects/packagefiles/libusb/meson.build
```diff
@@ -3,6 +3,7 @@ project(
   'c',
   version: '1.0.29',
   meson_version: '>=0.49.0',
+  license: 'LGPL-2.1-or-later',
   default_options: ['cpp_std=c++20'],
```
neheb: I think this warning can be avoided by using override options.

meator: What?

neheb: untested fix:

```
--- a/subprojects/packagefiles/libusb/meson.build
+++ b/subprojects/packagefiles/libusb/meson.build
@@ -3,7 +3,6 @@ project(
   'c',
   version: '1.0.29',
   meson_version: '>=0.49.0',
-  default_options: ['cpp_std=c++20'],
 )

 cc = meson.get_compiler('c')
@@ -166,6 +165,7 @@ config_inc = include_directories('.')
 libusb = library(
   'libusb-1.0',
   sources,
+  override_options: host_machine.system() == 'emscripten' ? 'cpp_std=c++20' : [],
   dependencies: [thread_dep, macos_dep, udev_dep],
   include_directories: incdirs,
   version: '0.5.0',
```

neheb:
```
GitHub Actions
/ Ubuntu (x86_64)

Minimum Meson version is 0.63.0

0.38.0: configuration_data.get()
0.46.0: pkgconfig.generate optional positional argument
0.47.0: User option "feature", dict
0.48.0: compiler.has_function_attribute, gnu_symbol_visibility arg in library
0.49.0: compiler.get_argument_syntax
0.63.0: cpp_std in subproject default_options
```

subprojects/packagefiles/libusb/meson.build
meator: What override options? The warnings indicate that `meson_version` needs to be increased, since there are quite a few features from newer features used.

neheb: Not really. It's already set to 0.49.0. The fix I posted is a fix to avoid bumping it to 0.63.0.

meator: I didn't reload, I see your reply now. How did you know that this is `emscripten` related?

neheb: I wrote the original meson.build file. This is a C library. C++ is needed for Emscripten.

meator: I didn't read the warning message that closely until now. Why are we being warned about "too new" features when they are all included in `>=0.49.0`? Also, your suggestion makes sense to me now.

neheb: It's a subtle issue that got fixed in 0.63.0. Basically even though options are specified in default_options, they don't get applied. Instead, options from the root project are the only ones used.

meator:
Thank you again @neheb!

Can we do anything about the riscv64 CI test or at least tell CI to skip it?

neheb:
I would dig too much into it. It's just typical QEMU bugs.

PR #2321, opened 2025-08-20
fdk-aac: Match the upstream build systems

mstorsjo:
Produce library files named "fdk-aac", not "fdk_aac". Include the right soversion and version, to produce shared libraries with the same names as the upstream build systems.


subprojects/packagefiles/fdk-aac/meson.build
```diff
@@ -391,14 +391,16 @@ if get_option('remove-date')
...
-  'fdk_aac',
+  'fdk-aac',
   fdk_aac_src,
   dependencies: m_dep,
   include_directories: fdk_aac_inc,
   install: true,
   cpp_args: cc.get_supported_arguments('-fno-exceptions', '-fno-rtti') + extra_args,
   c_args: extra_args,
   vs_module_defs: 'fdk-aac.def',
+  soversion: '2',
```
bgilbert: The value of `version` will cause `soversion` to default to `2`, so this is redundant.

mstorsjo:
> You'll also need to add a `license` argument to `project()`.

Ok, added that, and removed the superfluous `soversion:`.

PR #2325, opened 2025-08-20
Use latest Meson prerelease for `meson format`

bgilbert:
Formatting rules change over time, and CI uses the latest Meson prerelease, so we should match that version on developers' systems.  If not running in CI, install Meson into a virtualenv that we update at most once per day.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ __pycache__
 .idea
 .vscode
 *~
+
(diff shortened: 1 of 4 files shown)


bgilbert:
Actually, we need to use the same Meson for the format checks in `sanity_checks.py`.

bgilbert:
Any objections?

PR #2326, opened 2025-08-21
argparse: generate `.pc` file with pkg-config module

bgilbert:
It's less error-prone to use Meson's module rather than upstream's template.


neheb (approved):

neheb:
wonder how I missed this.

bgilbert:
There's a couple others too, not counting cairomm and libsigcplusplus which will be fixed by their next upstream releases.  I'm going through them now.

PR #2329, opened 2025-08-21
New Wrap: xdgmime

tesseract241:
I'm new at using wrap, so let me know if there's anything I should change.  
`meson setup -Dwrap=xdgmime; meson compile` works.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -1470,6 +1470,16 @@
       "liblzma-dev"
     ]
   },
+  "xdgmime": {
+    "_comment" : [
+      "Requires following XDG standards"
+    ],
+    "build_on" : [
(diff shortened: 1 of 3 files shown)


bgilbert:
`wrap-git` is not allowed in WrapDB.  You'd need to use `wrap-file`, which requires a source archive from an upstream release.  The archive can be autogenerated from a Git tag if needed, but there must be formal releases.  Upstream has no Git tags and the README says:

```
There are no formal releases planned for this module, and it is not
intended to be installed at this time.  Rather, it is meant to be used
by other libraries or applications to add support for the MIME system.
```

Unfortunately that means we won't be able to carry a wrap for this project.

PR #2332, opened 2025-08-23
Work around Meson bug ignoring `c_std`/`cpp_std` in subprojects; other 1.9.0rc updates

bgilbert:
On Meson versions affected (or expected to be affected) by mesonbuild/meson#14939, add C and C++ languages to the root project so subprojects will respect `c_std` and `cpp_std`.

We can adjust the version range if needed once the bug is fixed.  There's probably some value in continuing to carry this check for a while after the fix is released, since developers might have older Meson versions on their machines.

Also remove obsolete Strawberry Perl workaround and ignore `subprojects/.wraplock`.

### DIFF
--- a/.github/workflows/sanity_checks.yml
+++ b/.github/workflows/sanity_checks.yml
@@ -166,10 +166,6 @@ jobs:
           architecture: 'x86'
           python-version: '3.12'
 
-      # https://github.com/actions/runner-images/issues/5459#issuecomment-1532856844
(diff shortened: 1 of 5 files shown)


neheb:
It’d be nice to patch a bunch of wraps moving the assignment from default options to override options.

bgilbert:
Yeah, it's frustrating that the more robust approach requires more boilerplate.

PR #1527, opened 2024-05-23
[SFML] the wrap dont have the same name than pkg-config

nda-cunh:
close #1526

PR #2339, opened 2025-08-26
sfml: Update to SFML3.0.1

vid512:
Attempt to update to SFML version 3.  imgui-sfml will follow.

New stuff, compared to meson.build for SFML2:
- SFML3 uses miniaudio instead of openal.
- Added build instructions for most of examples.
  - Resource files needed for examples are now copied to build directory using `import('fs').copyfile()`, instead of `configure_file()`.
  - Running examples on Windows required manual copying of dependency DLLs into directory where example's .exe is. Not sure if/how this can be automated by Meson.
- Added license info (Zlib license)

Not changed:
- MacOS-specific example (cocoa) was not added.
- Build options / targets were left in the same basic state, as they in the old meson.build for SFML2. In particular, see: https://github.com/mesonbuild/wrapdb/pull/1527

Tested on Win32 and Linux with provided examples, and with my larger-than-just-a-demo pet project. Worked fine in all cases.

There were few gotchas (some extra macros had to be defined by the build system, eg. for endianness, using stdint.h instead of incorrect manual defaults). I wouldn't be too surprised if there were some minor problems left, that I have missed.


vid512:
Question: How should I submit PR for imgui-sfml v3.0, when it depends on this PR to work / test correctly? Should I wait until this is accepted, or...?

vid512:
The clang-cl build seems to fail due to a bug in build of libflac depenency, unrelated to sfml. I can replicate the same problem locally even on wrapdb/master by running  (in VS developer console) 
```
set CC=clang-cl
set CXX=clang-cl
meson setup build -Dwraps=flac
meson compile -C build
```

vid512:
Blocked by: https://github.com/mesonbuild/wrapdb/pull/2342

subprojects/packagefiles/sfml/meson.build
```diff
@@ -1,14 +1,15 @@
 project(
   'sfml',
   'cpp',
-  version: '2.6.2',
+  license: 'Zlib',
+  version: '3.0.1',
   default_options: ['cpp_std=c++17', 'wrap_mode=forcefallback'],
```
neheb: We've been replacing cpp_std in default options with override_options. Would it be possible here or is it too annoying?

vid512: Done.

PR #2342, opened 2025-08-27
flac: Fix clang-cl build

vid512:
Fixes build error with clang-cl.

getopt.c declares getenv() in a way that clang-cl considers incompatible to standard library version, unless macro "getenv" is defined. So just define getenv to getenv.


subprojects/packagefiles/flac/src/share/getopt/meson.build
```diff
@@ -9,10 +9,19 @@ else
     required: false,
   )
 endif
+
+# fix problem with conflicting getenv() declaration in getopt.c
+# when using clang-cl
+getopt_c_args = []
+if meson.get_compiler('c').get_id() == 'clang-cl'
```
neheb: I think it’s better to match on host system being windows instead

eli-schwartz: Is it an issue for MSVC too?

neheb: Not but it’s using the same libc. MSVC just happens to be more lenient, for now:

vid512: Done

PR #2345, opened 2025-08-27
Allow implicit `dependency_names = <wrap-name>` and forbid explicit form

bgilbert:
Meson implicitly adds a wrap's name to its provides, and always has; the project may or may not choose to add an override by that name.  However, `sanity_checks.py` treats a wrap file's explicit provides as canonical (when cross-checking with `releases.json` and with the overrides), making it impossible to use the implicit form.

Have `sanity_checks.py` treat `releases.json` as canonical instead, then special-case the wrap name so it need not be mentioned in `dependency_names`.  The behavior of transitional `<wrap-name> = <name>_dep` provides remains unchanged.  Wraps that want to declare a self-named provide should include it in their `releases.json` `dependency_names` as usual.

Then, for brevity/simplicity, forbid `dependency_names = <wrap-name>` in wrap files.


eli-schwartz:
I'm not convinced of this. "Explicit is better than implicit" is a popular adage for a reason.

The implicit form has value in particular because you can checkout a git submodule to subprojects/libfoo/ and have the submodule use override_dependency; you then don't need a wrap file at all. If you do have a wrap file anyway though (e.g. because you're using tarball fetches) what's the reason for leaving it to "guesswork" rather than including two lines that document intent?

What's the reason for making it an error for a project that provides two dependencies, to include one of them inside a line that has to exist anyway?

bgilbert:
This is meant as prep for converting ports over to `meson.override_dependency()`, where it'd be nice not to add a bunch of boilerplate that restates existing defaults.

I think in part the default feels weird because we aren't in the habit of relying on it.  The reason not to be explicit is the same as the reason e.g. not to specify `value: 'auto'` for every feature option: it obscures meaningful directives among routine ones.  While we don't strictly need to forbid the explicit case, I worry we'd get an inconsistent mess that'd be less legible than the status quo.

I'm not strongly attached to this though.

eli-schwartz:
- The thing is that it's a defensible choice to list it. So I don't see a good reason to punish people for doing so and telling them they "must not".
- Churning existing files to remove them has the same logical basis as churning existing files to remove `required: true` from everywhere you see it.
- if you rename a `foo.wrap` to `foo-1.2.3.wrap` to include the version number I'm fairly positive that the implicit form breaks into a dozen pieces and stops working. There are people who do this sort of thing for reasons I don't understand, and deleting the explicit form as a matter of WrapDB policy could then lead to non-obvious failure modes that aren't even detected immediately because the default configuration found a system dependency.
- It's not quite the same as an implicit `value: 'auto'` since the implicit default is computed from the state of the filesystem rather than having an intrinsic value. Same with `required: true`. There's no circumstances where `dependency('foo')` results in "required: false".

bgilbert:
I agree we shouldn't churn _releases_ to fix lints, but in general I do favor repo-wide updates where it's practical, to prevent each component from falling into a slightly different state of decay.  (Preferably alongside a CI test to prevent reoccurrence.)  For similar reasons I don't think "punishment" is the right way to think about it; there's a lot of value in enforcing legibility standards in a repo, even if other choices are defensible and other repos will make different decisions.

But I didn't realize wrap renaming was a thing, and yeah, that completely breaks the implicit provide.

eli-schwartz:
> I agree we shouldn't churn _releases_ to fix lints, but in general I do favor repo-wide updates where it's practical, to prevent each component from falling into a slightly different state of decay. (Preferably alongside a CI test to prevent reoccurrence.)

Yes and no. The trick there is to figure out what "matters", which is slightly subjective. My point wasn't really "we shouldn't do this because churn" but more "we should figure out whether the distinction matters enough to make it worth spending the churn budget to enforce a specific style". And there are ancillary changes you'd probably want to make at the same time, such as adding it to the refs that should be ignored by `git blame`.

> But I didn't realize wrap renaming was a thing, and yeah, that completely breaks the implicit provide.

Yes, unfortunately people will be people. :D

PR #2349, opened 2025-08-28
tools/update-packagefiles: Don't copy .meson-subproject-wrap-hash

vid512:
Update the `tools/update-packagefiles.py` script to ignore ` .meson-subproject-wrap-hash` files.


bgilbert:
Thanks!

PR #2350, opened 2025-08-29
abseil-cpp: update to 20250127.1-1

neheb:


subprojects/packagefiles/abseil-cpp/meson.build
```diff
@@ -1,9 +1,9 @@
 project(
   'abseil-cpp',
   'cpp',
-  version: '20240722.0',
+  meson_version: '>=0.49.0',
+  version: '20250127.1',
   license: 'Apache-2.0',
-  default_options: ['cpp_std=c++17'],
```
bgilbert, line 6: No `override_options`?

neheb: I'm trying to get CI to fail. The latest version probably will.

neheb: Hmm I don't think that'll happen unless GCC gets downgraded.

neheb: huh, GCC 9 compiles.

neheb:
OK so this is fascinating. There's C++14 support and only on macOS does this issue appear. Incredible.

bgilbert:
I don't love the duplication, but okay.  What would you think about defining a `cpp_std` or `override_options` variable?

PR #2351, opened 2025-08-29
github CI: add legacy GCC to Ubuntu

neheb:
Starting with GCC 11, C++17 becomes the default standard. GCC 9 will help find wraps that need it specified.


bgilbert:
This is just a copypaste of the existing Ubuntu job, so I'd be inclined to extend that one instead:

```yaml
    strategy:
      matrix:
        platform: ['x86_64', 'aarch64']
        include:
          - platform: x86_64
            legacy-compiler: GCC 9
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0

      - uses: egor-tensin/setup-gcc@v1
        if: matrix.legacy-compiler
        with:
          version: 9
```

Where the value of `legacy-compiler` is just there for legibility in the job list.

bgilbert:
Well, also, `Report Meson version dependencies` would need a conditional because it's only supposed to run in one job for the whole workflow.

bgilbert:
Also also, maybe we should downgrade `cpp_std` in the root `meson.build` instead?  Since that's there, I wouldn't think the change of compiler default would do anything.

neheb:
Oh that's a good point. Probably why it's not failing...

bgilbert:
The original PR was https://github.com/mesonbuild/wrapdb/pull/380 and the reasoning for it makes no sense.  But there's an argument for setting _older_ C/C++ standards in the root config (which would effectively become the minimums) rather than adding jobs with specific old compilers.

neheb:
Maybe we should just keep the last commit and remove the first one. GCC 9 defaults to C++14 and the macOS compiler either C++98 or C++11. That's probably good enough already.

bgilbert:
Works for me.

neheb:
yeah it's coming back to me now. This was before 0.63.0 was released. I think that PR was meant to be temporary. So much for that.

neheb:
some interesting errors have appeared...

```
error: invalid value 'gnu++17,c++17' in '-std=gnu++17,c++17'
```

bgilbert:
@neheb Where?

neheb:
Somewhere from the build all workflow for macOS.

PR #2352, opened 2025-08-29
Use `dependency_names` for ports that already call `override_dependency()`

bgilbert:


neheb:
I'd split off libxslt to a separate PR to fix MSVC.

bgilbert:
#2354

PR #2353, opened 2025-08-29
Use `dependency_names` for upstreams that call `override_dependency()`

bgilbert:


bgilbert:
Ubuntu tests pass (including the projects that aren't bumped here) on Meson 1.8.3.  Meson 1.9.0 is improperly passing `-std=c++14` when compiling libxmlpp, despite #2332, but someone else can track that one down.

bgilbert:
Okay, Meson Git master fixes it.

PR #2354, opened 2025-08-29
libxslt: fix dependency overrides

bgilbert:
Both overrides omitted the proper `lib` prefix and the libexslt provide pointed to the wrong library.

Also switch to `dependency_names`.


neheb:
```
..\subprojects\libxslt-1.1.43\libexslt/libexslt.h(13,10): fatal error: 'win32config.h' file not found
   13 | #include <win32config.h>
```

PR #2355, opened 2025-08-29
libccp4c: Set C standard to gnu89

taw10:
Recent GCC (15.2+) defaults to gnu23/C23 standard, which forbids a type of function declaration (without arguments) that libCCP4 relies on.  This causes a build failure with the following error when building on e.g. Fedora 41+:

```
./subprojects/libccp4-8.0.0/ccp4/library_utils.c: In function ‘ccp4_utils_setenv’:
../subprojects/libccp4-8.0.0/ccp4/library_utils.c:152:11: error: too many arguments to function ‘putenv’; expected 0, have 1
  152 |   return (putenv (param));
      |           ^~~~~~  ~~~~~
../subprojects/libccp4-8.0.0/ccp4/library_utils.c:144:7: note: declared here
  144 |   int putenv ();
      |       ^~~~~~
```

The code should, of course, use the standard C library's declaration for this routine, but that's an upstream problem.  It will build correctly up to gnu18, but I chose to pin it to an older standard since the code is rather ancient.


subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -3,8 +3,9 @@ project(
   'libccp4c',
   'c',
   version: '8.0.0',
-  meson_version: '>=0.49.0',
+  meson_version: '>=0.63.0',
   license: 'LGPL-3.0-or-later',
+  default_options: ['c_std=gnu89,c89'],
```
neheb: Use override_options instead.

taw10: Thanks, changed!

taw10:
This can almost be built on Windows.  Defining `__BYTE_ORDER` allows it to build on MSVC (see [this commit](https://github.com/mesonbuild/wrapdb/commit/c3c6ffacfdf7ba3570f2cd013cd8955265795b6a)).  However, this seems like the sort of thing that should be defined by the system and shouldn't be overridden like this.

Compiling on MSYS2 will need upstream work as mentioned previously (#687) .  Unfortunately the upstream changes are unlikely to ever get made (though I am trying!).

Given both of these things, I've disabled building on Windows completely.

neheb:
MINGW might be fixed by passing -include stdint.h

taw10:
> MINGW might be fixed by passing -include stdint.h

Indeed, that would probably work, though it also isn't very satisfying.  My suggestion would be to do this in two PRs.  First this one, to make it work for recent GCC, then add Windows build options in a new revision after that.

@neheb thanks for reviewing!  Any further suggestions, or would you consider this ready to merge?

PR #2356, opened 2025-08-30
yyjson: New wrap

lhearachel:
I have a few questions as part of this PR:

- This implementation is pretty basic and only targets building the library for use by a parent project. Does more of the original CMake configuration need to be ported for this to be accepted?
- `yyjson` stores both header and source in the same directory in the source tree. Should this wrap use a workaround directory to isolate the header file, similar to the `cJSON` wrap?


subprojects/packagefiles/yyjson/meson.options
```diff
@@ -0,0 +1,55 @@
...
+    type: 'boolean',
+    description: 'Disable JSON writer',
+    value: false,
+)
+
+option(
+    'YYJSON_DISABLE_INCR_READER',
+    type: 'boolean',
+    description: 'Disable incremental reader',
+    value: false,
```
bgilbert: These should not directly use the names of the upstream defines.  Uppercase is not traditionally used for Meson options and `YYJSON_` is redundant.  Probably `feature` options make sense, in case autodetection is desired in the future - something like:
```suggestion
    'incr-reader',
    type: 'feature',
    description: 'Enable incremental reader',
```

which will default to `auto`.

subprojects/yyjson.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = yyjson-0.12.0
+source_url = https://github.com/ibireme/yyjson/archive/refs/tags/0.12.0.tar.gz
+source_filename = yyjson-0.12.0.tar.gz
+source_hash = b16246f617b2a136c78d73e5e2647c6f1de1313e46678062985bdcf1f40bb75d
+patch_directory = yyjson
+
+[provide]
+yyjson = yyjson_dep
```
bgilbert:
```suggestion
dependency_names = yyjson
```

subprojects/packagefiles/yyjson/meson.build
```diff
@@ -0,0 +1,42 @@
+project(
+  'yyjson',
+  'c',
+  version: '0.12.0',
+  license: 'MIT',
+  meson_version: '>=1.1.0',
```
bgilbert: 1.1 is only needed for the name `meson.options`, which isn't really worth the tighter compatibility constraints.  Better to rename to `meson_options.txt` and drop this to `0.54.0`.

bgilbert:
> This implementation is pretty basic and only targets building the library for use by a parent project. Does more of the original CMake configuration need to be ported for this to be accepted?

No, that's fine.  The wrap should be useful to somebody but does not need to be complete.

> yyjson stores both .h and .c in the same directory in the source tree. Should this wrap use a workaround directory to isolate the header file, similar to the cJSON wrap?

cJSON documentation says that applications should `#include <cjson/cJSON.h>`, which is why the wrap has to force the header into a subdirectory.  yyjson doesn't really seem to document the include directive at all, but both Debian and Fedora put the header directly into `/usr/include` without a subdir.  So, no workaround should be necessary.

lhearachel:
> cJSON documentation says that applications should `#include <cjson/cJSON.h>`, which is why the wrap has to force the header into a subdirectory. yyjson doesn't really seem to document the include directive at all, but both `Debian` and `Fedora` put the header directly into `/usr/include` without a subdir. So, no workaround should be necessary.

I can confirm that this is true [for the package in Arch Linux's Extra repository, too](https://archlinux.org/packages/extra/x86_64/yyjson/):

```console
> pacman -Qs yyjson
local/yyjson 0.12.0-1
    A high performance JSON library written in ANSI C

> ls -l /usr/include | grep 'yyjson.h'
.rw-r--r--  322k root 18 Aug 14:10 yyjson.h
```

subprojects/packagefiles/yyjson/meson_options.txt
```diff
@@ -0,0 +1,47 @@
...
+option(
+    'writer',
+    type: 'feature',
+    description: 'Enable JSON writer',
+)
+
+option(
+    'incr-reader',
+    type: 'feature',
+    description: 'Enable incremental writer',
```
bgilbert:
```suggestion
    description: 'Enable incremental reader',
```

subprojects/packagefiles/yyjson/meson_options.txt
```diff
@@ -0,0 +1,47 @@
...
+)
+
+option(
+    'json-utils',
```
bgilbert: Optional: could rename to `utils` since `json` is arguably implied.

subprojects/packagefiles/yyjson/meson.build
```diff
@@ -0,0 +1,42 @@
...
+  if get_option(feature).disabled()
+    c_args += f'-D@flag@'
+  endif
+endforeach
+
+yyjson_inc = include_directories('src')
+
+yyjson_lib = library(
+  'yyjson',
+  sources: files('src/yyjson.c', 'src/yyjson.h'),
```
bgilbert: Headers don't need to be included in `sources`.

bgilbert:
Thanks!

PR #2361, opened 2025-08-31
gtest: pass C++17 for macOS

neheb:


bgilbert:
I'd consider at least adding a variable with the std rather than copying the logic everywhere.

neheb:
Done.

PR #2364, opened 2025-08-31
imgui-sfml: Update to SFML3

vid512:
Update imgui-sfml to version 3.0, which uses SFML 3.0


vid512:
Failure to build on MacOs is probably caused by: https://github.com/mesonbuild/wrapdb/pull/2366

PR #2365, opened 2025-08-31
ffmpegthumbnailer: fix compilation with macOS

neheb:


neheb:
FFS API broke again

neheb:
https://github.com/dirkvdb/ffmpegthumbnailer/pull/240

given how infrequent the releases are, this will take a while...

PR #2369, opened 2025-09-03
tools: add utility code to load/save wrap files and JSON metadata

bgilbert:
Consolidate I/O code for `releases.json`, `ci_config.json`, and wrap files.  For the JSON metadata, also ensure type hints are available in all the tools.

Also do some minor cleanups.

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -29,7 +29,7 @@ import sys
 import shutil
 
 from pathlib import Path
-from utils import Releases, Version, ci_group, is_ci, is_alpinelike, is_debianlike, is_macos, is_windows, is_msys, read_wrap, FormattingError, format_meson
+from utils import CIConfig, ProjectCIConfig, Releases, Version, ci_group, is_ci, is_alpinelike, is_debianlike, is_macos, is_windows, is_msys, read_wrap, FormattingError, format_meson
 
 PERMITTED_FILES = {'generator.sh', 'meson.build', 'meson_options.txt', 'meson.options', 'LICENSE.build'}
 PER_PROJECT_PERMITTED_FILES: dict[str, set[str]] = {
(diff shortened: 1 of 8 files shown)


tools/utils.py
```diff
@@ -16,14 +16,14 @@
...
 import os
-import sys
-import typing as T
 from pathlib import Path
 import platform
+import re
 import subprocess
+import sys
 import time
+import typing as T
```
eli-schwartz: While we're here, I think we usually leave `import typing as T` at the bottom. It's a bit "different" as an import because it's only used for annotations, not for "real" code.

tools/utils.py
```diff
@@ -104,6 +106,20 @@ def __cmp(self, other: 'Version', comparator: T.Callable[[T.Any, T.Any], bool])
...
+def read_wrap(name: str) -> configparser.ConfigParser:
+    config = configparser.ConfigParser(interpolation=None)
+    config.read(wrap_path(name), encoding='utf-8')
+    return config
+
+def write_wrap(path: Path, config: configparser.ConfigParser) -> None:
+    # configparser write() adds multiple trailing newlines, collapse them
+    buf = io.StringIO()
+    config.write(buf)
+    path.write_text(buf.getvalue().rstrip('\n') + '\n', encoding='utf-8')
```
eli-schwartz: pathlib concrete methods such as write_path() or resolve() tend to be rather scary and full of weird edge cases. The old version in the other file(s) used `with open(...) as f: f.write()` for this sort of thing, and I think that pattern is quite helpful.

It's also faster and even manages to be more readable as it avoids mixing encoding arguments with the string you're building for the contents.

tools/sanity_checks.py
```diff
@@ -211,14 +192,10 @@ def setUpClass(cls):
...
             cls.releases = Releases.load()
-            fn = 'ci_config.json'
-            with open(fn, 'r', encoding='utf-8') as f:
-                cls.ci_config = json.load(f)
+            cls.ci_config = CIConfig.load()
         except json.decoder.JSONDecodeError as ex:
             raise RuntimeError('metadata is malformed') from ex
 
-        system = platform.system().lower()
-        cls.skip = T.cast(T.List[str], cls.ci_config[f'broken_{system}'])
```
eli-schwartz, line 230: First argument to T.cast should always be quoted. Apparently it wasn't. Still, we can fix that while we're moving this around...

tools/utils.py
```diff
@@ -104,6 +108,72 @@ def __cmp(self, other: 'Version', comparator: T.Callable[[T.Any, T.Any], bool])
...
+def read_wrap(name: str) -> configparser.ConfigParser:
+    config = configparser.ConfigParser(interpolation=None)
+    config.read(wrap_path(name), encoding='utf-8')
+    return config
+
+def write_wrap(path: Path, config: configparser.ConfigParser) -> None:
+    # configparser write() adds multiple trailing newlines, collapse them
+    buf = io.StringIO()
+    config.write(buf)
+    with path.open('w', encoding='utf-8') as f:
```
eli-schwartz, line 174:
```suggestion
    with open(path, 'w', encoding='utf-8') as f:
```

works too. :)

path.open is just:

```python
    def open(self, mode='r', buffering=-1, encoding=None,
             errors=None, newline=None):
        """
        Open the file pointed to by this path and return a file object, as
        the built-in open() function does.
        """
        if "b" not in mode:
            encoding = io.text_encoding(encoding)
        return io.open(self, mode, buffering, encoding, errors, newline)
```

bgilbert: Oh indeed.  I find `path.open()` a bit cleaner when I have a `Path`.  Not strongly tied to it though.

PR #2370, opened 2025-09-03
cppzmq: use dependency names for pkg-config generation

stephanlachnit:
This reverts commit 149072b1a8af445827f44514eb6b4dea9b4fde87 from https://github.com/mesonbuild/wrapdb/pull/2327.

The reason is that unfortunately this effectively requires Meson 1.9.0 :/

When building with older Meson versions, you'll get this error when using libzmq as subproject (e.g. via #905):

```
Executing subproject cppzmq 
cppzmq| Project name: cppzmq
cppzmq| Project version: 4.10.0
cppzmq| C++ compiler for the host machine: ccache g++ (gcc 14.2.1 "g++ (GCC) 14.2.1 20250110 (Red Hat 14.2.1-7)")
cppzmq| C++ linker for the host machine: g++ ld.bfd 2.41-4
cppzmq| Dependency libzmq found: YES 4.3.5 (overridden)
subprojects/cppzmq-4.10.0/meson.build:18:20: ERROR: requires argument not a string, library with pkgconfig-generated file or pkgconfig-dependency object, got <InternalDependency null: True>
```

This issue comes from the pkg-config generator and was only fixed in https://github.com/mesonbuild/meson/pull/14750.

For cppzmq this is a problem in particular since the main reason to use it as a fallback is to also fallback libzmq with the draft API, which is typically not included in any stable distros. If libzmq is installed on a system level, then one can easily install cppzmq as well, since both are widely available.

\cc @bgilbert @neheb


bgilbert:
Good catch.  I've verified that no other recent pkgconfig conversions should be affected.

It should be possible to fix this without dropping the pkgconfig generator using something like (untested):

```meson
depnames = ['libzmq']
deps = []
foreach name : depnames
  deps += dependency(name)
endforeach

# [...]

import('pkgconfig').generate(
  name: meson.project_name(),
  description: 'C++ binding for libzmq',
  url: 'https://github.com/zeromq/cppzmq',
  # Meson < 1.9.0 fails on 'requires: deps' if any deps are sibling subprojects
  # https://github.com/mesonbuild/meson/pull/14750
  requires: depnames,
  install_dir: get_option('datadir') / 'pkgconfig',
)
```

bgilbert:
Docs update for the Meson behavior change in https://github.com/mesonbuild/meson/pull/14994.

stephanlachnit:
Probably better the other way around?

```meson
requires = []
foreach dep : deps
  requires += dep.name()
endforeach
```

This way you have more control over the dependencies (e.g. optional ones).

PR #2376, opened 2025-09-04
Add wrap files for adamyaxley-obfuscate

punytroll:


subprojects/packagefiles/adamyaxley-obfuscate/meson.build
```diff
@@ -0,0 +1,17 @@
+project(
+  'obfuscate',
+  'cpp',
+  version: '1.0.0',
+  license: 'Unlicense',
+  license_files: ['LICENSE'],
+  meson_version: '>=1.2.0',
```
bgilbert: This can be reduced to 0.54.0 if you make `default_options` a list and remove `license_files`.

punytroll:
> Shouldn't you also build https://github.com/adamyaxley/Obfuscate/blob/master/test_unit.cpp and run it as a `test()`?

I could. I thought about it, and decided against it, because I was unsure about the added dependency on gtest - whether it's really worth it, or not, given that it's based on a release by upstream. But let me try. :)

> This can be reduced to 0.54.0 if you make `default_options` a list and remove `license_files`.

Ah, so the wraps should support the oldest possible meson? I can make it '>=0.0.0' if I apply these changes but I see no benefit. What's special about 0.54.0 to pick that version? Updating meson should be encouraged, right? At least the dictionary has immediate readability benefits.

eli-schwartz:
> I could. I thought about it, and decided against it, because I was unsure about the added dependency on gtest - whether it's really worth it, or not, given that it's based on a release by upstream. But let me try. :)

In meson_options.txt, add a feature option "-Dtests=auto" so that you can search for gtest, but do nothing if it's not installed. That way people who use this as a wrap aren't forced to install gtest.

> Ah, so the wraps should support the oldest possible meson? I can make it '>=0.0.0' if I apply these changes but I see no benefit. What's special about 0.54.0 to pick that version? Updating meson should be encouraged, right? At least the dictionary has immediate readability benefits.

The purpose of the wrapdb is obviously to let people integrate it into their projects and their workflows. It's not unlikely that people are still using for example Debian Bookworm (oldstable) with its meson 1.0.1, or Ubuntu 22.04 LTS with its meson 0.61.2, or Ubuntu 24.04 LTS with its meson 1.3.2

One of those three is good enough to build the current version of this wrap. Two of those aren't good enough.

The question then becomes:
- do you care
- how much effort does it take to support LTS distros
- does supporting older meson releases just mean avoiding simple "style and neatness" changes, or are there genuine, meaningful bug fixes it would be painful to do without?

Those are all questions, not demands. As the wrap submitter, your opinion on the matter is one of the strongest influences on "what shall be done". But in general I'd encourage, for example, not assuming anything newer than meson 1.3 (if you can do without it) because "it doesn't work on the most recent Ubuntu LTS" will kind of cramp the style of most people. And you do support the most recent Ubuntu LTS, you just don't support the (still widely used) one before that.

The CI emits a non-blocking comment on the Files Changed panel, showing the current effective minimum version plus what features contributed to this effective minimum. The idea is just to make it trivial to double check that you're listing the minimum version you actually intended to support, and make it easy to tell whether it's worth the time investment to try lowering that. You're using:
- one feature from 1.2.0 (default_options dict)
- one feature from 1.1.0 (license_files)
- one feature from 0.54.0 (override_dependency)

If you're willing to support 1.0.0 or 0.61.2 then you get 0.54.0 support "for free" just by changing the declared minimum version. There's nothing special about 0.54.0, just that if you allow <=1.0.0 then it's the same thing as supporting 0.54.0

subprojects/packagefiles/adamyaxley-obfuscate/meson.build
```diff
@@ -0,0 +1,37 @@
...
+  meson_version: '>=0.59.0',
+  default_options: ['cpp_std=c++14'],
+)
+
+obfuscate_dep = declare_dependency(
+  include_directories: include_directories('.'),
+)
+
+if get_option('tests').allowed()
+  gtest_dep = dependency(
```
eli-schwartz: Feature options are designed to make this all extremely simple. You don't need 20 lines to find a dependency.

```meson
gtest_dep = dependency('gtest_main', version: '>=1.10.0, required: get_option('tests')

if gtest_dep.found()
    # make executable and test
endif
```
This intrinsically handles the case where enabled and not found, and it also logs a message if tests are disabled.

eli-schwartz (changes requested):

subprojects/packagefiles/adamyaxley-obfuscate/meson.build
punytroll: Ah, I was looking at spdlog, as I knew it was working there. I'll try again ...

eli-schwartz: That is, erm, it shouldn't be doing that. :upside_down_face:

punytroll: Got it. That IS simpler ... I wasn't sure about the allow_fallback but that is handled as well. :+1:

subprojects/packagefiles/adamyaxley-obfuscate/meson.build
```diff
@@ -0,0 +1,31 @@
...
+    'test_unit',
+    executable(
+      'test_unit',
+      ['test_unit.cpp'],
+      dependencies: [gtest_dep, obfuscate_dep],
+    ),
+  )
+endif
+
+meson.override_dependency('adamyaxley-obfuscate', obfuscate_dep)
```
bgilbert: The provide should match upstream's pkg-config name.  Upstream doesn't have one, but if it did, I imagine they would choose `obfuscate` (or `Obfuscate`?) instead of `adamyaxley-obfuscate`.  The origin prefix is a WrapDB wrap naming convention and doesn't apply to provides.

bgilbert (commented):
Looks good to me, just one naming thing.

subprojects/packagefiles/adamyaxley-obfuscate/meson.build
punytroll: > If the project name is too generic or ambiguous (e.g. benchmark), consider using organization-project naming format (e.g. google-benchmark).

That's from the documentation https://mesonbuild.com/Adding-new-projects-to-wrapdb.html

Since "obfuscate" is really generic, I added the "organisation". Same as with tl-expected and tl-optional.

Can we have a third opinion on that?

bgilbert: Yes, that's for the project name.  I'm talking about the provide name.

punytroll: Ah, OK. Thanks for the clarification - now I see, that you addressed that in your post already. But isn't that the same "namespace"? Again: tl-expected and tl-optional do the same thing.

punytroll: I've used Obfuscate, in line with their github project name.

bgilbert: It isn't the same namespace, no.  Upstreams don't always name pkg-config files after their project, and may ship multiple libraries in one project with a pkg-config file for each.  For examples, see the `google-*` wraps.  Or also: `apache-orc.wrap` provides `orc`, but [GStreamer's ORC](https://github.com/GStreamer/orc) provides `orc-0.4`.

(Meson makes the namespaces overlap a bit by implicitly adding a provide for the wrap name.  But the wrap doesn't _have_ to do an `override_dependency()` for that name, and WrapDB's CI forbids use of that feature in any case.)

eli-schwartz: Noting here that a provide is not the same thing as providing a dependency, it is just a "check here and see if it does end up providing the dependency". :)

For much the same reason, a `[provide]` entry can list something that is only available on a single operating system or only with certain build options, so if the subproject gets configured on the wrong system or with the wrong options it will just error out slightly *later* with "not actually provided".

PR #2379, opened 2025-09-05
fmt: properly provide header-only dependency

miklhh:
The logic for providing fmts header-only dependency is currently causing problems with building projects targeting WASM. This seems to be because, when the fmt meson option `header-only` is set, `fmt_lib` and `fmt_dep` are still provided in fmts meson.build file, even though they are not the header-only targets of fmt. The following error message is emitted when trying to build a project using the WASM32/Emscripten backend (even though `fmt_lib` and `fmt_dep` are unused throughout the project):

> ERROR: ld.wasm does not support shared libraries.

This PR fixes this issue by only providing `fmt_lib` and `fmt_dep` when `header-only` is **not** set. When `header-only` is set, this meson patch only provides `fmt_header_only_dep`.

### Extract from faulty build:

```
  apytypes 0.4.0.dev78+g45a1126c2
  
    Subprojects
      fmt         : YES
      highway     : YES
      nanobind    : YES
      robin-map   : YES (from nanobind)
  
    User defined options
      Cross files : /tmp/cibw-run-pz6qboh_/cp312-pyodide_wasm32/build/venv/lib/python3.12/site-packages/pyodide_build/tools/emscripten.meson.cross
      Native files: /home/runner/work/apytypes/apytypes/.mesonpy-3pdk3853/meson-python-native-file.ini
      b_ndebug    : if-release
      b_vscrt     : md
      buildtype   : release
  
  Found ninja-1.13.1 at /usr/local/bin/ninja
  Running compile:
  Working directory:  /tmp/tmp0n0pvf2j
  Code:
  int i = static_cast<int>(0);
  -----------
  Command line: `/tmp/tmpsmf8oq99/c++ /tmp/tmp0n0pvf2j/testfile.cpp -o /tmp/tmp0n0pvf2j/output.o -D_FILE_OFFSET_BITS=64 -c -std=c++17` -> 0
  Compiler accepts -std=c++17: YES
  Running compile:
  Working directory:  /home/runner/work/apytypes/apytypes/.mesonpy-3pdk3853/meson-private/tmpd3dggrpi
  Code:
  
  -----------
  Command line: `/tmp/tmpsmf8oq99/c++ /home/runner/work/apytypes/apytypes/.mesonpy-3pdk3853/meson-private/tmpd3dggrpi/testfile.cpp -o /home/runner/work/apytypes/apytypes/.mesonpy-3pdk3853/meson-private/tmpd3dggrpi/output.o -c -fexceptions -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -Werror=implicit-function-declaration -Werror=unknown-warning-option -Werror=unused-command-line-argument --print-search-dirs` -> 0
  stdout:
  programs: =/home/runner/.cache/cibuildwheel/emsdk-3.1.58/emsdk-3.1.58/upstream/bin
  libraries: =/home/runner/.cache/cibuildwheel/emsdk-3.1.58/emsdk-3.1.58/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten
  -----------
  Running compile:
  Working directory:  /tmp/tmpedgdo8rp
  Code:
  int i = static_cast<int>(0);
  -----------
  Command line: `/tmp/tmpsmf8oq99/c++ /tmp/tmpedgdo8rp/testfile.cpp -o /tmp/tmpedgdo8rp/output.o -D_FILE_OFFSET_BITS=64 -c -std=c++14` -> 0
  Compiler accepts -std=c++14: YES
  
  ERROR: ld.wasm does not support shared libraries.   # <------
```


neheb:
Have you tried defaultlibrary=static?

miklhh:
> Have you tried defaultlibrary=static?

Yes, we have tried `default_library=static,` which yields the same error whether it is set or not. The WASM compiler still complains that ld.wasm does not support shared libraries unless we apply the provided patch to fmt.

releases.json
```diff
@@ -985,6 +985,7 @@
       "fmt"
     ],
     "versions": [
+      "11.2.0-2",
```
oscargus, line 988: I think you should remove the line below (no need to provide two versions of 11.2.0-wrappers).

subprojects/packagefiles/fmt/meson.build
```diff
@@ -15,40 +15,34 @@ endif
...
+    compile_args: '-DFMT_HEADER_ONLY',
+  )
+else
+  fmt_lib = library(
+    'fmt',
+    'src/format.cc',
+    'src/os.cc',
+    cpp_args: fmt_private_cpp_args,
+    include_directories: 'include',
+    build_by_default: not header_only,
```
oscargus: This and below should be possible to set to `true` as we know that header_only is false here.

subprojects/packagefiles/fmt/meson.build
```diff
@@ -15,40 +15,34 @@ endif
...
-
-if not header_only
+  fmt_dep = declare_dependency(
+    include_directories: 'include',
+    compile_args: fmt_interface_cpp_args,
+    link_with: fmt_lib,
+  )
+  if meson.version().version_compare('>=0.54.0')
+    meson.override_dependency('fmt', fmt_dep)
+  endif
```
oscargus, line 46: As the linter claims that meson 0.63 is required (and probably should be defined somewhere?), this if-clause should always be true and therefore should always be used (possibly even for header only?).

oscargus: The linter should be fixed by adding `  meson_version: '>=0.63.0',` to the project definition at the top.

releases.json
bgilbert: Nope, existing releases must not be removed.

subprojects/packagefiles/fmt/meson.build
neheb: The 0.63 warning can be avoided by using override_options for the library instead of default_options.

miklhh: I opted to add `meson_version => 0.63.0` to `project` instead of using `override_options`. I don't quite understand how `override_options` works or how it helps here. Let me know if you prefer using `override_options` over adding the `meson_version => 0.63.0`.

neheb: It’s a bug with older meson versions. Even though they get specified they get ignored.

I’d like to have 0.61 compatibility as Ubuntu 22.04 ships with that

miklhh: Okey. I've now added `override_options: 'cpp_std=c++14'` to the `fmt_lib` library, and I completely removed `meson_version >= 0.63.0` from the `fmt` project. Did I understand your wish correctly @neheb?

I don't know how to `override_options` for the dependency `fmt_header_only_dep`, as it doesn't link against any `library`. Are there any other ways in which I can add the `cpp_std=c++14` option to `fmt_header_only_dep`?

neheb: No. That’s up to the users of the header.

neheb:
Add meson_version of 0.50.0 and this looks good to go.

miklhh:
I've added `meson_version >=0.50.0` to the fmt `project`. Should I also remove `default_options: 'cpp_std=c++14'` from the project, since the linter is complaining about it?

neheb:
Yes

miklhh:
All right. I removed the default option `cpp_std=c++14` now as well.

PR #2382, opened 2025-09-07
giflib: fix compilation under macOS

neheb:


bgilbert:
It seems better to be explicit about the standard, since future (compiler default) standards could conceivably break something.

bgilbert:
`util_lib` needs `override_options` too.

PR #2384, opened 2025-09-07
liburing: update from 2.5 to 2.12

apache-hb:


subprojects/.wraplock
bgilbert, line 1: This shouldn't be committed.

subprojects/packagefiles/liburing/meson.build
```diff
@@ -1,9 +1,9 @@
 project(
   'liburing',
   ['c', 'cpp'],
-  version: '2.5',
+  version: '2.12',
   license: '(MIT OR LGPL-2.1-only) AND (MIT OR GPL-2.0-only WITH Linux-syscall-note)',
-  meson_version: '>=0.57.0',
+  meson_version: '>=0.59.0',
```
bgilbert: 0.63.0 because of `c_std` and `cpp_std` in `default_options`, unless you'd prefer to move those to the `override_options` of every target.

subprojects/packagefiles/liburing/meson.build
```diff
@@ -117,19 +120,36 @@ has_cxx = cpp.compiles(
 )
 
 has_ucontext = (cc.has_type(
-  'ucontext_t',
+    'ucontext_t',
```
bgilbert: Optional: some line breaks can be removed here to make the formatter less weird about this.

subprojects/packagefiles/liburing/meson_options.txt
```diff
@@ -1,13 +1,13 @@
 option(
     'examples',
-    type: 'boolean',
-    value: false,
+    type: 'feature',
+    value: 'auto',
```
bgilbert: Nit: `auto` is the default.

subprojects/packagefiles/liburing/meson.build
```diff
@@ -117,19 +120,36 @@ has_cxx = cpp.compiles(
...
   prefix: '#include <ucontext.h>',
-))
+  ))
 
 has_nvme_uring_cmd = cc.has_type(
   'struct nvme_uring_cmd',
   prefix: '#include <linux/nvme_ioctl.h>',
 )
 
+has_futexv = cc.compiles(
```
bgilbert, line 135: Is this checking for `struct futex_waitv` or `FUTEX_32`?  It would be a bit clearer to use `cc.has_header_symbol()`.

subprojects/packagefiles/liburing/src/include/liburing/meson.build
```diff
@@ -35,10 +35,32 @@ struct open_how {
 '''
 endif
 
+has_block_uring_cmd_discard = cc.compiles(
```
bgilbert: Could be `cc.has_header_symbol()`.

subprojects/packagefiles/liburing/test/meson.build
```diff
@@ -7,18 +7,24 @@ all_tests = [
   '917257daa0fe.c',
   'a0908ae19763.c',
   'a4c0b3decb33.c',
-  'accept.c',
   'accept-link.c',
+  'accept-non-empty.c',
   'accept-reuse.c',
   'accept-test.c',
+  'accept.c',
```
bgilbert, line 14: Optional: if you make this a `files()` the formatter will keep the list sorted.

subprojects/packagefiles/liburing/test/meson.build
```diff
@@ -169,22 +226,34 @@ if has_cxx
...
-runtests_quiet_sh = find_program('runtests-quiet.sh')
 
 xcflags = ['-Wno-sign-compare']
 if has_stringop_overflow
   xcflags = xcflags + ['-Wstringop-overflow=0']
 endif
 if has_array_bounds
-  xcflags = xcflags + ['-Warray-bounds=0']
+  xcflags = xcflags + [
+    cc.get_id() == 'gcc' ? '-Warray-bounds=0' : '-Warray-bounds',
```
bgilbert: Could possibly be `cc.has_argument('-Warray-bounds=0') ?`

subprojects/packagefiles/liburing/test/meson.build
```diff
@@ -169,22 +226,34 @@ if has_cxx
...
+  ]
 endif
 
+has_io_uring_kernel_support = host_machine.cpu() in ['x86', 'x86_64', 'aarch64']
```
bgilbert, line 245: Is there a more future-proof way to test for this?

subprojects/packagefiles/liburing/test/meson.build
```diff
@@ -169,22 +226,34 @@ if has_cxx
...
+
 fs = import('fs')
 
 foreach test_source : all_tests
   test_name = fs.stem(test_source)
+  test_kwargs = {
```
bgilbert: This is only used once, so could just be arguments to that `test()` call.

subprojects/packagefiles/liburing/meson.build
apache-hb: This was a direct copy from the liburing configure script, It seems safest to do 1:1 translation

subprojects/packagefiles/liburing/test/meson.build
apache-hb: This list cannot be a `files` as its later iterated and `file` doesnt have `startswith`

subprojects/packagefiles/liburing/test/meson.build
apache-hb: I think the only real way to test this reliably is from parsing `/boot/config-$(uname -r)` which feels a bit too much for a configure script to do

subprojects/packagefiles/liburing/src/include/liburing/meson.build
```diff
@@ -35,10 +35,28 @@ struct open_how {
 '''
 endif
 
+has_block_uring_cmd_discard = cc.has_header_symbol(
+  '<linux/blkdev.h>',
```
bgilbert: This is always returning false.
```suggestion
  'linux/blkdev.h',
```

PR #2385, opened 2025-09-08
libccp4c: Windows compatability and #13713 fix

taw10:
Unfortunately, the previous update to the libccp4c wrap (https://github.com/mesonbuild/wrapdb/pull/2355) runs straight into https://github.com/mesonbuild/meson/issues/13713, breaking the build with Meson older than 1.3.0.  Since this is a scientific piece of software, it's compiled on old systems more often than it's compiled on very new systems, so this leaves us in a worse place than before the update.  This PR adds the workaround mentioned in https://github.com/mesonbuild/meson/issues/13713 for libccp4c.

As discussed in https://github.com/mesonbuild/wrapdb/pull/2355, this PR also adds the necessary flags to make it compile on Windows.  If we didn't care about Windows at all, another reasonable solution would be to simple change the standard from `gnu89,c89` to `gnu89`.


subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -13,8 +13,25 @@ mdep = cc.find_library(
...
+    )
+  elif host_machine.endian() == 'big'
+    add_project_arguments(
+      '-D__BYTE_ORDER=__BIG_ENDIAN',
+      language: 'c',
+    )
+  endif
+endif
+
+if host_machine.system() == 'windows' and cc.get_id() != 'msvc'
```
taw10: Is it better to do `cc.get_id() != 'msvc'`, or something like `(cc.get_id() == 'gcc' or cc.get_id() == 'clang')`?  I only know about the Windows compilers running in the CI pipeline here.

subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -64,7 +82,9 @@ libccp4c = library(
   'ccp4/cmtzlib.c',
   'ccp4/library_utils.c',
   dependencies: mdep,
-  override_options: ['c_std=gnu89,c89'],
+  override_options: [
+    meson.version().version_compare('>=1.3.0') ? 'c_std=gnu89,c89' : 'c_std=gnu89',
```
eli-schwartz: The whole moving override_options to individual targets thing seems kinda wild to me. At any rate, if it's being applied only to a specific target anyway then you can check it inside a *compiler* conditional instead of a *version* conditional, and do it up above, too.

subprojects/packagefiles/libccp4c/meson.build
eli-schwartz: How do you want to handle clang-cl?

That would be `cc.get_argument_syntax() == 'msvc'` compatible, but has its own unique ID.

subprojects/packagefiles/libccp4c/meson.build
neheb: I think @bgilbert wants this to be its own variable.

subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -13,8 +13,25 @@ mdep = cc.find_library(
   required: false,
 )
 
-if host_machine.system() == 'windows'
-  error('Building libccp4c is unsupported on Windows')
+if cc.get_id() == 'msvc'
```
neheb, line 16: Not sure how clang-cl handles this, but I would guess the same as MSVC. So use get_argument_syntax.

subprojects/packagefiles/libccp4c/meson.build
```diff
@@ -13,8 +13,25 @@ mdep = cc.find_library(
...
+  elif host_machine.endian() == 'big'
+    add_project_arguments(
+      '-D__BYTE_ORDER=__BIG_ENDIAN',
+      language: 'c',
+    )
+  endif
+endif
+
+if host_machine.system() == 'windows' and cc.get_id() != 'msvc'
+  add_project_arguments(
```
neheb: that's a bit heavy handed. Isn't it only needed for the library? If so, c_args there.

subprojects/packagefiles/libccp4c/meson.build
taw10: I switched it to `cc.get_argument_syntax() == 'gcc'`, because it turns out that the extra include is also needed for Windows+clang-cl.

subprojects/packagefiles/libccp4c/meson.build
taw10: I've changed this to `override_options: ['c_std=' + cstd]`, where `cstd` is a variable that gets set to either `c89` or `gnu89`, depending on the compiler, in the tests right a the top of the meson.build file.  This way, there's no need for a fallback and https://github.com/mesonbuild/meson/issues/13713 is side-stepped.  Did I understand your meaning correctly?

subprojects/packagefiles/libccp4c/meson.build
taw10: It seems to just be MSVC that needs the extra byte order declaration, so I left it as `cc.get_id() == 'msvc'`.

subprojects/packagefiles/libccp4c/meson.build
taw10: Done.  Now it uses `c_args: cargs`, where `cargs` is constructed at the top of the file, and contains the byte order declaration and/or forced header include (stdint.h) as necessary.

subprojects/packagefiles/libccp4c/meson.build
bgilbert: I think that's fine, sure.

(For the record I'm on the fence about `override_options`.  I don't love the extra boilerplate, and AIUI the forcing function right now is a short-term bug in Meson 1.9.0.  Having a variable helps reduce the boilerplate, but in a case like this where there's only one target, the variable isn't super-important.  Compiler conditional does seem better than version conditional here.)

taw10:
> Since you have later commits that apply review feedback to earlier commits, please either squash everything down to one commit, or rebase it down so we're back to one commit per logical change.

Thanks, now done.  The 13713 fix and Windows compatibility are two sides of the same coin, so I kept them together.  I separated out the extra warning suppression (`-Wno-array-parameter`) and the version number bump.

bgilbert:
Thanks!

Issue #2388, opened 2025-09-09
Google benchmark declares a benchmark-main dependency, but pkgconf name is benchmark_main


WillAyd:
Perhaps we should expose the dependency as `benchmark_main` to match the .pc file?

bgilbert:
Yes indeed.  Could you send a PR?

WillAyd:
Absolutely. Do we need to keep the current dependency name for backwards compatibility?

bgilbert:
I don't think so.  We've had a few PRs lately that haven't kept back-compat names for options or provides.  Anyone who runs `meson wrap update` will immediately break their fallback builds, at which point they can fix them.

PR #2389, opened 2025-09-10
Fix dependency name for google benchmark

WillAyd:
closes https://github.com/mesonbuild/wrapdb/issues/2388

### DIFF
--- a/subprojects/packagefiles/google-benchmark/meson.build
+++ b/subprojects/packagefiles/google-benchmark/meson.build
@@ -2,6 +2,7 @@ project(
   'benchmark',
   'cpp',
   version: '1.8.4',
+  meson_version: '>=0.49.0',
   default_options: [
     meson.version().version_compare('>=1.3.0') ? 'cpp_std=gnu++14,c++14' : 'cpp_std=gnu++14',
   ],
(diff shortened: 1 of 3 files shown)


bgilbert:
While we're here, we might as well do a couple other cleanups.  I've PRed your PR in https://github.com/WillAyd/wrapdb/pull/1.

WillAyd:
Great - thanks @bgilbert

subprojects/packagefiles/google-benchmark/meson.build
```diff
@@ -2,6 +2,7 @@ project(
   'benchmark',
   'cpp',
   version: '1.8.4',
+  meson_version: '>=0.48.0',
```
bgilbert: Argh, this should actually be 0.49.0.  The CI report missed this because it only matters on Windows.

WillAyd: OK no worries - will fix it up.

Do you need me to squash this here? Not sure how that works with retaining your contribution

bgilbert: Yeah, please do squash.  You can add a `Co-authored-by:` line to the commit message if you want, e.g. as in b89abe5c4f180b43144667e6e4dd37da3a19e825, but I'm not worried about it.

PR #2392, opened 2025-09-12
libcbor: add 0.13.0

apache-hb:
Adds version 0.13.0 of [libcbor](https://github.com/PJK/libcbor/tree/master)

### DIFF
--- a/subprojects/packagefiles/libcbor/meson.build
+++ b/subprojects/packagefiles/libcbor/meson.build
@@ -1,0 +1,162 @@
+project(
+  'libcbor',
+  'c',
+  version: '0.13.0',
+  meson_version: '>=0.59.0',
+  default_options: ['warning_level=3'],
+  license: 'MIT',
+)
+
+cc = meson.get_compiler('c')
+
+tests_opt = get_option('tests').disable_auto_if(meson.is_subproject())
+
+subdir('src/cbor')
+
+src = [
+  'src/cbor.c',
+  'src/allocators.c',
+  'src/cbor/streaming.c',
+  'src/cbor/internal/encoders.c',
+  'src/cbor/internal/builder_callbacks.c',
+  'src/cbor/internal/loaders.c',
+  'src/cbor/internal/memory_utils.c',
+  'src/cbor/internal/stack.c',
+  'src/cbor/internal/unicode.c',
+  'src/cbor/encoding.c',
+  'src/cbor/serialization.c',
+  'src/cbor/arrays.c',
+  'src/cbor/common.c',
+  'src/cbor/floats_ctrls.c',
+  'src/cbor/bytestrings.c',
+  'src/cbor/callbacks.c',
+  'src/cbor/strings.c',
+  'src/cbor/maps.c',
+  'src/cbor/tags.c',
+  'src/cbor/ints.c',
+]
+
+inc = include_directories('src')
+
+c_args = ['-DCBOR_BUILD=1']
+
+if get_option('default_library') == 'static'
+  c_args += ['-DCBOR_BUILD_STATIC=1']
+endif
+
+libcbor = library(
+  'cbor',
+  src,
+  config_file,
+  c_args: c_args,
+  include_directories: inc,
+  install: true,
+  version: meson.project_version(),
+  override_options: ['c_std=c11'],
+)
+
+libcbor_dep = declare_dependency(
+  link_with: libcbor,
+  include_directories: inc,
+)
+
+meson.override_dependency('libcbor', libcbor_dep)
+
+headers = [
+  config_file,
+  'src/cbor/streaming.h',
+  'src/cbor/cbor_export.h',
+  'src/cbor/data.h',
+  'src/cbor/encoding.h',
+  'src/cbor/serialization.h',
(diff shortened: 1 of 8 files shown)


subprojects/packagefiles/libcbor/meson.build
```diff
@@ -0,0 +1,161 @@
...
+  'c',
+  version: '0.13.0',
+  meson_version: '>=0.59.0',
+  default_options: ['warning_level=3'],
+  license: 'MIT',
+)
+
+cc = meson.get_compiler('c')
+
+tests_opt = get_option('tests').disable_auto_if(meson.is_subproject())
```
xclaesse, line 12: Why disable if it's a subproject?

subprojects/packagefiles/libcbor/src/cbor/cbor_export.h
```diff
@@ -0,0 +1,17 @@
+#ifndef CBOR_EXPORT
+# if (defined(__GNUC__) && __GNUC__ >= 4) || defined(__ICC)
+#  define CBOR_EXPORT __attribute__ ((visibility("default")))
+# elif (defined(__SUNPRO_C) || defined(__SUNPRO_CC))
+#  define CBOR_EXPORT __global
+/* Windows is special and you cannot just define entry points unconditionally. */
+# elif defined(_WIN32) && !defined(CBOR_BUILD_STATIC)
+#  ifdef CBOR_BUILD
+#   define CBOR_EXPORT __declspec(dllexport)
```
xclaesse, line 9: I don't think you ever include that file. Maybe open a PR upstream to include it?

xclaesse: Oh, upstream uses cmake generator, nevermind.

I wish Meson had it too... https://github.com/mesonbuild/meson/pull/10199.

subprojects/packagefiles/libcbor/meson.build
```diff
@@ -0,0 +1,161 @@
...
+cmocka = dependency(
+  'cmocka',
+  required: tests_opt,
+)
+m = cc.find_library(
+  'm',
+  required: false,
+)
+
+if cmocka.found() and tests_opt.allowed()
```
eli-schwartz: This is entirely incorrect.

cmocka.found() will only ever be true when tests_opt.allowed() is also true. Please do not manually check the value of feature options without a very good reason, it is very niche and intended for only highly advanced usage because the default is that feature options simply do the correct thing via dependency lookup.

Specifically, the primary reason feature options exist at all, is that cmocka will print a log message when disabled:
"Dependency cmocka **skipped**: feature **tests** disabled".

That means:
- pass feature options to the required kwarg of testsuite dependencies
- check if testsuite dependencies have been `.found()`, and leave it at that.

subprojects/packagefiles/libcbor/meson.build
```diff
@@ -0,0 +1,161 @@
...
+    'cbor_test',
+    ['test/assertions.c', 'test/stream_expectations.c', 'test/test_allocator.c'],
+    include_directories: inc,
+    dependencies: [libcbor_dep, cmocka, m],
+    c_args: c_args,
+  )
+
+  foreach path : tests
+    name = path.split('/')[1].replace('.c', '')
+    test_exe = executable(
```
eli-schwartz, line 146: Also use `build_by_default: false`.

This means you can also skip checking if the build is a subproject. The tests won't be built when including libcbor as a wrap, even if the testsuite is enabled -- they will only be built when running `meson test`.

subprojects/packagefiles/libcbor/meson.build
apache-hb: If the tests are left enabled in a subproject they'll all be built when running `ninja install`, which can be time consuming or at worst lead to build failures on stranger build configurations.

subprojects/packagefiles/libcbor/meson.build
apache-hb: Unless im mistaken targets with `build_by_default: false` will still be built during `ninja install` or equivalent, which can slow down builds a fair amount or lead to build failures. I really prefer to keep tests behind a feature for this reason.

xclaesse: Unless you add install: true, they won't be built. 

That being said, it's pretty common to have a yielding feature tests option. I just don't like it behave differently when it's a subproject.

eli-schwartz: Your information is very out of date: https://mesonbuild.com/Release-notes-for-1-7-0.html#test-targets-no-longer-built-by-default

> I really prefer to keep tests behind a feature for this reason.

They are ***already*** behind a feature, as you need a feature option to determine whether cmocka is a mandatory dependency. I do not see the urgency of also checking whether it's a subproject, that's all.

Irrespective of ANY other considerations though, I'm telling you to add `build_by_default: false` so that people building libcbor standalone with tests, will... have the tests only built when running `meson test`, which allows faster edit-compile-edit-compile development cycles, with testing deferred to later on.

apache-hb:
Sorry for the delay, busy week at work.

PR #2402, opened 2025-09-14
sanity_checks: require use of `dependency_names` where possible

bgilbert:
In wrap file `provide` sections, the `dependency_names = foo` syntax is preferred and works on every Meson release that supports the transitional `foo = foo_dep` syntax.  The transitional syntax is needed with projects that don't call `meson.override_dependency()`, but has also crept into a lot of wraps that don't require it.

For downstream ports, we control both the wrap file and the `meson.build`, so there's never a need for the transitional syntax.  Require ports to switch to `dependency_names` and `meson.override_dependency()` when they're updated.

For all other wraps, search the upstream `meson.build` files for `meson.override_dependency` and forbid the transitional syntax if found.  This will keep us in sync as upstreams add these calls, and in particular it'll prevent us from diverging from the provide names used upstream (see #2353).

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -350,6 +350,8 @@ class TestReleases(unittest.TestCase):
                                 self.check_meson_version(name, ver, patch_path)
                         if patch_path:
                             self.check_project_args(name, config)
+                        else:
+                            self.check_nonport_source(name, config)
                     else:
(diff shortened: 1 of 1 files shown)


bgilbert:
Will merge soon.

neheb (approved):

PR #2403, opened 2025-09-14
Update HTTP links to HTTPS; use ftp.gnu.org mirrors

bgilbert:


bgilbert:
The last commit doesn't fix all the problems with the wraps (mainly m4) but I think it's a net improvement.

subprojects/packagefiles/gdbm/meson.build
```diff
@@ -3,7 +3,7 @@ project(
   'c',
   version: '1.26',
   license: 'GPL-3.0-or-later',
-  meson_version: '>=0.49.0',
+  meson_version: '>=0.47.0',
```
neheb: This is suspicious.

I remember testing meson 0.49 on debian 9? a while back. One other bug that's tested neither by meson nor CI is that ternary operators (a ? b : c vs if a b else c) end up crashing python. IIRC @eli-schwartz blamed that on the version of python there.

Anyway, I don't know if 0.49 or older compatibility should be kept. It's ancient at this point.

PR #2406, opened 2025-09-17
ci: add IBM-hosted ppc64le and s390x runners

bgilbert:
They're only available in `mesonbuild/wrapdb`, not in forks.  As a result the emulated ppc64le job is potentially still useful, but it would likely bitrot since it wouldn't run in the main repo.  Drop that job.

Self-hosted runners don't count toward the concurrent job limit so this actually improves our parallelism.

setup-alpine doesn't understand that `$GITHUB_WORKSPACE` and `$RUNNER_TEMP` might not be under `$HOME/work`, so explicitly bind those into the chroot until this is fixed upstream.

GitHub doesn't have per-version Python builds for these arches, so just use the default Python.

In forks, the new jobs would sit indefinitely waiting for a runner to pick them up, blocking the workflow from completing.  We want to skip those jobs, but the job's `if:` directive can't read the `matrix` context.  Instead, preprocess the matrix from the prelude job.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -23,13 +23,15 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       cache-key: ${{ steps.cache-key.outputs.cache-key }}
+      matrix-alpine: ${{ steps.alpine.outputs.matrix }}
+      matrix-ubuntu: ${{ steps.ubuntu.outputs.matrix }}
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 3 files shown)


bgilbert:
Enablement is tracked in https://github.com/IBM/actionspz/issues/10.

bgilbert:
Prep commits broken out into #2422.

neheb:
Wonderful.

bgilbert:
Should be ready to go.

PR #2407, opened 2025-09-17
Bump croaring version

WillAyd:


subprojects/packagefiles/croaring/meson.build
```diff
@@ -3,12 +3,15 @@ project(
   'croaring',
   'c',
   'cpp',
-  version: '1.3.0',
+  version: '4.3.11',
+  meson_version: '>=0.49.0',
   license: 'MIT',
+  default_options: ['c_std=c11', 'cpp_std=c++11'],
```
neheb: Should be replaced with override_options

WillAyd: Here directly or are you asking to move these to the `library` call?

neheb: Move to library

subprojects/packagefiles/croaring/meson.build
```diff
@@ -36,9 +39,13 @@ croaring_lib = library(
   'croaring',
   croaring_sources,
   include_directories: croaring_headers,
+  override_options: ['c_std=c11', 'cpp_std=c++11'],
```
neheb: hrm I checked this project out.

So the compiled files are C only. The headers are what seem to have C++ support. Which means c++11 is not needed here.

WillAyd: Alright - I removed the cpp mentions from the project config and override_options

WillAyd:
Hey @neheb just wanted to toss a ping your way in case this got lost in the shuffle. I think all feedback has been addressed

PR #2408, opened 2025-09-18
ci: restrict `GITHUB_TOKEN` to read access where possible

bgilbert:
The `create_release` job needs write access, but nothing else does.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}
   # We do NOT want `cancel-in-progress` here since only one website job
@@ -13,6 +16,8 @@ concurrency:
 jobs:
   create_release:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read
```
eli-schwartz, line 9: I don't see the point of changing this workflow, which has a legitimate need for a token and runs only on merges to master. The other workflows make more sense to restrict.

bgilbert: You don't think it's worth restricting the `update_website` job?  It installs a bunch of packages from PyPI.

dcbaker: Is there any drawback to making it read only? I tend to think it's best to be as restrictive as possible with these sort of things

eli-schwartz: I guess I don't feel strongly one way or another. But I wonder why it's necessary to list permissions twice, once to take them away, and once to give them back again. Maybe just mark a single job as lower permissions, it's half as many lines to do it that way.

bgilbert: The idea is to set reasonable defaults in every workflow.  Better for a newly-added job to accidentally have too few permissions than too many.

PR #2410, opened 2025-09-21
JoltPhysics: Add wrap for 5.3.0

apache-hb:
Adds a wrap for the physics library [Jolt Physics](https://github.com/jrouwe/JoltPhysics/tree/master)

### DIFF
--- a/subprojects/packagefiles/joltphysics/meson.build
+++ b/subprojects/packagefiles/joltphysics/meson.build
@@ -1,0 +1,368 @@
+project(
+  'joltphysics',
+  'cpp',
+  license: 'MIT',
+  version: '5.3.0',
+  meson_version: '>=0.59.0',
+  default_options: ['cpp_rtti=false', 'cpp_eh=none'],
+)
+
+cpp = meson.get_compiler('cpp')
+
+# meson reports msys2 as windows, so we have to check for clang + windows
+# to catch clang on msys2. This doesn't catch clang-cl on windows so thats
+# fine.
+if cpp.get_id() == 'clang' and host_machine.system() == 'windows'
+  error(
+    'unsupported architecture: Clang on cygwin/msys2 cannot build jolt due to linking issues, sorry',
+  )
+endif
+
+if host_machine.cpu_family() in ['ppc64', 'loongarch64']
+  error(
+    'unsupported architecture: Jolt 5.3.0 cannot be built on ppc64 or loongarch64, 5.4.0+ is required',
+  )
+endif
+
+# Source files
+sources = files(
+  'Jolt/AABBTree/AABBTreeBuilder.cpp',
+  'Jolt/Core/Color.cpp',
+  'Jolt/Core/Factory.cpp',
+  'Jolt/Core/IssueReporting.cpp',
+  'Jolt/Core/JobSystemSingleThreaded.cpp',
+  'Jolt/Core/JobSystemThreadPool.cpp',
+  'Jolt/Core/JobSystemWithBarrier.cpp',
+  'Jolt/Core/LinearCurve.cpp',
+  'Jolt/Core/Memory.cpp',
+  'Jolt/Core/Profiler.cpp',
+  'Jolt/Core/RTTI.cpp',
+  'Jolt/Core/Semaphore.cpp',
+  'Jolt/Core/StringTools.cpp',
+  'Jolt/Core/TickCounter.cpp',
+  'Jolt/Geometry/ConvexHullBuilder.cpp',
+  'Jolt/Geometry/ConvexHullBuilder2D.cpp',
+  'Jolt/Geometry/Indexify.cpp',
+  'Jolt/Geometry/OrientedBox.cpp',
+  'Jolt/Math/Vec3.cpp',
+  'Jolt/Physics/Body/Body.cpp',
+  'Jolt/Physics/Body/BodyCreationSettings.cpp',
+  'Jolt/Physics/Body/BodyInterface.cpp',
+  'Jolt/Physics/Body/BodyManager.cpp',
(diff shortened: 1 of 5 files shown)


IogaMaster:
Can we get this pushed?

subprojects/packagefiles/joltphysics/meson.build
```diff
@@ -0,0 +1,684 @@
+project(
+  'jolt',
```
bgilbert:
```suggestion
  'joltphysics',
```

subprojects/packagefiles/joltphysics/meson.build
```diff
@@ -0,0 +1,684 @@
...
+)
+
+cpp = meson.get_compiler('cpp')
+
+# meson reports msys2 as windows, so we have to check for clang + windows
+# to catch clang on msys2. This doesn't catch clang-cl on windows so thats
+# fine.
+if cpp.get_id() == 'clang' and host_machine.system() == 'windows'
+  error(
+    'unsupported architecture: Clang on cygwin/msys2 cannot build jolt due to linking issues, sorry',
```
bgilbert, line 17: This is technically an abuse of `unsupported architecture`, but indeed we don't have a separate disable flag for this.

What are the linking issues?  Is it an upstream bug?  Wrap bug?

subprojects/packagefiles/joltphysics/meson.build
```diff
@@ -0,0 +1,684 @@
...
+
+if host_machine.cpu_family() == 'ppc64'
+  error(
+    'unsupported architecture: Jolt 5.3.0 cannot be built on ppc64, 5.4.0+ is required',
+  )
+endif
+
+# Source files
+all = [
+  'Jolt/ConfigurationString.h',
```
bgilbert: This works, but is a bit unconventional.  Usually we separate headers and source code into separate lists in the `meson.build`.

Please use `files()` so the formatter will automatically sort.

subprojects/packagefiles/joltphysics/meson.build
```diff
@@ -0,0 +1,684 @@
...
+args = []
+build_args = []
+
+if get_option('default_library') == 'shared'
+  args += ['-DJPH_SHARED_LIBRARY']
+  build_args += ['-DJPH_BUILD_SHARED_LIBRARY']
+endif
+
+buildtype = get_option('buildtype')
+build_debug = buildtype.contains('debug')
```
bgilbert: This won't work on all Meson versions.

```suggestion
build_debug = get_option('debug')
```

subprojects/packagefiles/joltphysics/meson.build
```diff
@@ -0,0 +1,684 @@
...
+  args += ['-DJPH_ENABLE_ASSERTS']
+endif
+
+args += cpp.get_supported_arguments(['-mfpmath=sse'])
+
+jolt = library(
+  'Jolt',
+  sources,
+  include_directories: '.',
+  extra_files: headers,
```
bgilbert: This isn't really necessary unless your IDE needs it.  You could potentially stop listing the headers in this file entirely.

subprojects/packagefiles/joltphysics/meson.build
```diff
@@ -0,0 +1,684 @@
...
+
+if profile.allowed()
+  args += ['-DJPH_PROFILE_ENABLED']
+endif
+
+if asserts_opt.allowed()
+  args += ['-DJPH_ENABLE_ASSERTS']
+endif
+
+args += cpp.get_supported_arguments(['-mfpmath=sse'])
```
bgilbert: We want to require the parent project to be compiled with this option too?

subprojects/packagefiles/joltphysics/meson.build
```diff
@@ -0,0 +1,684 @@
...
+      'UnitTests/Math/HalfFloatTests.cpp',
+      'UnitTests/Math/VectorTests.cpp',
+    ]
+    if object_stream
+      test_src += ['UnitTests/ObjectStream/ObjectStreamTest.cpp']
+    endif
+    test_exe = executable(
+      'UnitTests',
+      test_src,
+      include_directories: ['.', 'UnitTests'],
```
bgilbert: `.` is automatic.

subprojects/packagefiles/joltphysics/meson_options.txt
```diff
@@ -0,0 +1,56 @@
...
+)
+option(
+    'fpu_exceptions',
+    type: 'boolean',
+    value: false,
+)
+option(
+    'object_stream',
+    type: 'boolean',
+    value: true,
```
bgilbert: `true` is the default.

subprojects/packagefiles/joltphysics/meson.build
apache-hb: I honestly can't remember, I think it may have been a clang bug at the time

bgilbert: We don't require CI to be completely green.  If this isn't a known unfixable problem for some reason, it might be better to remove the check to avoid hiding the issue.

subprojects/packagefiles/joltphysics/meson_options.txt
```diff
@@ -0,0 +1,55 @@
...
+)
+option(
+    'disable_custom_allocator',
+    type: 'boolean',
+    value: false,
+)
+option(
+    'cross_platform_determinism',
+    type: 'boolean',
+    value: true,
```
bgilbert, line 40: (default)

apache-hb:
Oh damnit github why did it merge rather than fast forwarding, I'll force push

PR #2412, opened 2025-09-23
Update Abseil flags to match upstream recommendations and add missing source files.

smcallis:
Sources for these changes here:
  https://abseil.io/docs/cpp/platforms/compilerflags
  https://github.com/abseil/abseil-cpp/issues/727
  https://github.com/abseil/abseil-cpp/issues/1779

These better match what upstream recommends and suppresses spurious warnings about things like __int128. Compiles cleanly with gcc and clang now.

### DIFF
--- a/subprojects/packagefiles/abseil-cpp/meson.build
+++ b/subprojects/packagefiles/abseil-cpp/meson.build
@@ -21,7 +21,6 @@ flags = cpp.get_supported_arguments(
   '-Wno-float-promotion',
   '-Wno-format-literal',
   '-Wno-gcc-compat',
-  '-Wno-gcc-compat',
   '-Wno-old-style-cast',
   '-Wno-packed',
   '-Wno-padded',


subprojects/packagefiles/abseil-cpp/meson.build
```diff
@@ -12,9 +12,32 @@ cpp = meson.get_compiler('cpp')
...
-  '-Wno-sign-compare',
+  '-Wcast-qual',
+  '-Wconversion-null',
+  '-Wmissing-declarations',
+  '-Wno-comma',
+  '-Wno-conversion',
+  '-Wno-double-promotion',
+  '-Wno-float-promotion',
+  '-Wno-format-literal',
+  '-Wno-gcc-compat',
```
neheb, line 23: That's doubled.

smcallis: Can you elaborate?  I only see them defined once.

neheb: There are two no-GCC-compatible entries

smcallis: Ah yes indeed, thank you, fixed.

PR #2421, opened 2025-09-28
expat: update to 2.7.3

neheb:
ping @hartwork


neheb:
@bgilbert so this came out of CVE converns. Would it make sense to add CVE tracking to WrapDB?

bgilbert:
Library version should be 1.11.1.

Yeah, possibly.  There are notoriously a lot of bogus/unimportant CVEs but it'd be nice to get some degree of visibility.  Not sure how easy it is to query CVEs by upstream project.

neheb:
alright, added extra windows stuff. probably pointless.

eli-schwartz:
I am curious to hear, what exactly should a wrap do about a CVE.

neheb:
In this case, a version update is enough.

eli-schwartz:
So the idea would be to just discover that the wrap has a newer version available upstream? Doesn't really need a CVE integration. :)

bgilbert:
There are a lot of stale wraps here.  It might be useful to know that certain ones should be prioritized.

Issue #2224, opened 2025-06-28
`create_release.py` creates incomplete release if something goes wrong


bgilbert:
If `create_release.py` fails after `find_upload_url()`, it will have already created a GitHub release but won't upload the wrap file, so it produces a broken release like [this one](https://github.com/mesonbuild/wrapdb/releases/tag/libgpiod_1.6.5-3) without a source archive or wrap.  The workflow job will [fail](https://github.com/mesonbuild/wrapdb/actions/runs/15940439760/job/44967757393), but upon a rerun it'll find that the tag already exists and won't retry that release.

PR #2424, opened 2025-09-29
create_release: upload to draft release, then publish

bgilbert:
We've been creating a public release and then uploading files into it.  However, if we crash halfway through, we leave behind a Git tag and an unusable release without a wrap file, since that's the last thing to be uploaded.  Subsequent runs see the tag and don't retry.

(Also, if we enable [immutable releases](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/immutable-releases) in the future, GitHub won't allow us to upload to a published release.)

Instead, create the release in draft form, upload to it, and then publish the release.  GitHub doesn't create the tag until we publish, so newly-created tags will reflect only releases that have actually completed.

As a consequence, we now retry failed releases on subsequent runs.  However, if we locate an existing draft release and any assets have been uploaded to it, GitHub refuses to let us upload those again.  Instead of manually clearing out existing assets, just delete any existing drafts with the intended tag name.

Fixes: #2224

### DIFF
--- a/tools/create_release.py
+++ b/tools/create_release.py
@@ -121,10 +121,13 @@ class CreateRelease:
         response.raise_for_status()
         for r in response.json():
             if r['tag_name'] == self.tag:
-                self.release_id = r['id']
-                self.upload_url = r['upload_url'].replace('{?name,label}','')
-                print('Found release:', self.upload_url)
-                return
+                if r['draft']:
+                    delete_api = f'https://api.github.com/repos/{self.repo}/releases/{r["id"]}'
+                    response = requests.delete(delete_api, headers=headers)
+                    response.raise_for_status()
+                    print('Deleted stale release draft:', r['id'])
+                else:
+                    raise Exception('Refusing to recreate existing release')
 
         content = {
             'tag_name': self.tag,
--- a/tools/create_release.py
+++ b/tools/create_release.py
@@ -129,9 +129,12 @@ class CreateRelease:
                 else:
                     raise Exception('Refusing to recreate existing release')
 
+        cmd = ['git', 'rev-parse', 'HEAD']
+        commit = subprocess.check_output(cmd, text=True).strip()
         content = {
             'tag_name': self.tag,
             'name': self.tag,
+            'target_commitish': commit,
             'draft': True,
         }
         response = requests.post(api, headers=headers, json=content)
--- a/tools/create_release.py
+++ b/tools/create_release.py
@@ -42,6 +42,7 @@ class CreateRelease:
             self.create_source_fallback()
             self.create_patch_zip()
             self.create_wrap_file()
+            self.finalize()
 
     def warn(self, message: str) -> None:
         if is_ci():
@@ -120,6 +121,7 @@ class CreateRelease:
         response.raise_for_status()
         for r in response.json():
             if r['tag_name'] == self.tag:
+                self.release_id = r['id']
                 self.upload_url = r['upload_url'].replace('{?name,label}','')
                 print('Found release:', self.upload_url)
                 return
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> (Also, if we enable [immutable releases](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/immutable-releases) in the future, GitHub won't allow us to upload to a published release.)

It has been occasionally necessary to modify them long after the fact, which I think makes a good practical argument that it's a good thing we never had immutable releases.

As an extrapolation, we will probably not ever enable them in the future.

I agree that draft releases are good all by themselves solely for solving the crash-halfway issue. :)

bgilbert:
> It has been occasionally necessary to modify them long after the fact

That seems surprising.  When would we want to modify an existing release instead of creating a new one?

bgilbert:
Added one more fix:

`create_release` always tags the repo's current `HEAD` commit, not the commit checked out by the release job.  Thus, if two PRs that modify the same wrap land in quick succession, we could release the contents of the first PR under a tag attached to the second one.  If both PRs include identical release bumps (as seems likely), merge will elide the second one, so the second PR will not cause an additional release with the intended contents but a manual investigation will find the tag in the expected place.
    
Pass the commit hash of the current checkout during release creation, ensuring the correct commit is tagged.

eli-schwartz:
> That seems surprising. When would we want to modify an existing release instead of creating a new one?

It has been a while, so I can't remember anymore. :(

There are various possibilities why one might wish to, though. For example meson (not wrapdb) regularly releases binary installers, but not always by the same person as the uploader of the source releases. Or else there are simply issues doing the Windows / macOS signing and rather than delaying the release it is released source-only at first and later on the installers are added.

Three reasons I can think of, off the top of my head:
- ironically, fix a broken automated tag by manually creating and uploading the fixed asset. :)
- backdating brand new source fallback urls, since they are anyways optional bonuses, instead of adding a new release that many people will see and think they *have to* update to
- removing a release entirely for e.g. legal reasons. It seems from the GitHub documentation wording that you would not be allowed to "yank" a release, regardless of reason. Perhaps GitHub Support would be able to? No idea.

In general and on theoretical grounds I'm a believer in non-repudiation / auditable changelogs at the active expense of immutability. But my understanding is that GitHub only (starts to) offer the former as an accidental side effect of the latter.

tools/create_release.py
```diff
@@ -120,17 +121,28 @@ def find_upload_url(self) -> None:
...
+                    delete_api = f'https://api.github.com/repos/{self.repo}/releases/{r["id"]}'
+                    response = requests.delete(delete_api, headers=headers)
+                    response.raise_for_status()
+                    print('Deleted stale release draft:', r['id'])
+                else:
+                    raise Exception('Refusing to recreate existing release')
 
+        commit = subprocess.check_output(
+            ['git', 'rev-parse', 'HEAD'], text=True
+        ).strip()
```
eli-schwartz: Preferably function calls in python don't close on their own line at the same indentation level as they started. This is strictly avoided in meson itself. Unlike say dicts, it's not really idiomatic to do this in a language with significant indentation.

I'd usually try to fit it all on one line if possible. You could also define `cmd = []` on one line and reuse it on the next line and still have code that is strictly more compact and readable.

eli-schwartz (changes requested):
Overall this looks really solid (but I haven't tested). Thanks. :)

bgilbert:
> It seems from the GitHub documentation wording that you would not be allowed to "yank" a release, regardless of reason.

In my tests, it's possible to delete an immutable release and also an immutable tag, but they cannot be recreated afterward.  I don't know if that's the intended behavior when immutable releases come out of beta.

PR #2425, opened 2025-09-30
vo-aacenc.wrap: Fix building as a subproject

nirbheek:
Due to the header layout, this was previously only usable after installation into a prefix.

### DIFF
--- a/subprojects/packagefiles/vo-aacenc/meson.build
+++ b/subprojects/packagefiles/vo-aacenc/meson.build
@@ -8,7 +8,8 @@ project(
   ],
   license: 'Apache-2.0',
   version: '0.1.3',
-  meson_version: '>= 0.59.0',
+  # Need 1.2.0 for cc.compiles() with a file generated at setup time. This is only used on arm, so the CI will warn on x86_64.
+  meson_version: '>= 1.2.0',
 )
 
 armv7neon = false


nirbheek:
CI fails due to https://github.com/mesonbuild/meson/pull/15062.

Can be re-run when 1.9.2 or newer versions are released.

neheb:
```
In subprojects/vo-aacenc.wrap, replace `<name> = <name>_dep` entries with
`dependency_names = <name>`.
```

subprojects/packagefiles/vo-aacenc/LICENSE.build
```diff
@@ -0,0 +1,503 @@
+		  GNU LESSER GENERAL PUBLIC LICENSE
```
bgilbert: Wait, where did this come from?  This PR makes a fairly small change to the wrap, you aren't the original wrap author, and the upstream project is Apache 2.0, so it seems strange that the PR wants to relicense the wrap to LGPLv2.1.

nirbheek: You know what, that's a great question. I have no idea how that change got in. I don't recall adding it.

nirbheek: Ah, it got memory-holed because it was so long ago. LGPL-2.1 is the license of makedef.py. This change is of course wrong, I will fix it.

nirbheek:
Will merge in a week if there are no objections.

subprojects/packagefiles/vo-aacenc/LICENSE.build
bgilbert: Oh okay, I understand the situation now.  `LICENSE.build` should indeed have the MIT license, but also the LGPL license with some text saying that it applies to `makedef.py`.

bgilbert:
Could you add a comment above `meson_version` explaining why we need 1.2.0?  CI (which checks on x86_64) will continue complaining about that value, so it'd be helpful to have some documentation.

LGTM aside from those two comments.

subprojects/packagefiles/vo-aacenc/LICENSE.build
nirbheek: Done.

nirbheek:
> Could you add a comment above `meson_version` explaining why we need 1.2.0? CI (which checks on x86_64) will continue complaining about that value, so it'd be helpful to have some documentation.

Done. Will merge tomorrow.

PR #2255, opened 2025-07-08
yaml-cpp: New wrap v0.8.0

crvux:

PR #2428, opened 2025-10-01
yaml-cpp: clean-up meson build files

stephanlachnit:
This PR:

- Cleans up the build files, e.g. by removing unnecessary clang-format runs or adding compiler warnings (should be done via `warning_level` or manually via compile args, but really in a subproject this doesn't make a lot of sense)
- Reduces the required Meson version to 0.61.2, in order to be able to use this on Ubuntu 22.04 LTS.


stephanlachnit:
What I haven't checked if we can run unbundled gtest, and the tests don't run as part of the CI.

subprojects/packagefiles/yaml-cpp/meson.build
```diff
@@ -2,39 +2,31 @@ project(
...
 if cc.has_multi_arguments('-include', 'cstdint')
   lib_cpp_args += ['-include', 'cstdint']
 endif
-if not meson.is_subproject()
-  lib_cpp_args += cc.get_supported_arguments(
-    '-Wall',
-    '-Wextra',
-    '-Wshadow',
-    '-Weffc++',
-    '-Wno-long-long',
```
eli-schwartz, line 26: In general it makes perfect sense for a project to use -Wno-* even in a subproject, since you want a warning free build.

Arguments can of course be made in favor of "f*** C++" being an inappropriate option to use in a subproject (potentially to use anywhere, but I digress).

subprojects/packagefiles/yaml-cpp/meson.build
```diff
@@ -58,16 +50,3 @@ tests_feature = get_option('tests').disable_auto_if(meson.is_subproject())
...
-
-if not meson.is_subproject()
-  clang_format_prog = find_program(
-    'clang-format',
-    required: false,
-  )
-  if clang_format_prog.found()
-    run_target(
-      'format',
-      command: [clang_format_prog, '--style=file', '-i', src],
```
eli-schwartz, line 70: The more interesting question is why was this ever added given that meson automatically produces clang-format rules whenever clang-format is installed.

It does feel wasteful to invest time and energy into maintaining that inside an unofficial downstream patched-in meson.build, indeed.

subprojects/packagefiles/yaml-cpp/meson.build
stephanlachnit: Ah right, the `-Wno` can stay

subprojects/packagefiles/yaml-cpp/meson.build
stephanlachnit: I guess it slipped under the radar in https://github.com/mesonbuild/wrapdb/pull/2113 / https://github.com/mesonbuild/wrapdb/pull/2255 since I didn't see any comments on that part

eli-schwartz: The former doesn't appear to have had anything to comment on. :)

subprojects/packagefiles/yaml-cpp/meson.build
stephanlachnit: But on further inspection - I don't see those warnings ever popping up? They appear in none of the compilers in the CI, and there was no comment on why they were needed. I would keep them removed for now, and if they appear again in the future (I haven't tested old compilers), we can add them again. There is still the old version on WrapDB so this will not be a big issue if they pop up.

stephanlachnit:
> Can't we just configure this in ci_config.json? It should have always been enabled, as soon as the option was added, imo.

Tried but:
- doesn't compile due to warnings on arm (and those were not the ignored ones)
- does not compile on windows (I think due to bundled gtest)

I don't care enough about the test to take the time to fix the tests - if you don't mind, I would merge as is and leave this open for someone else to fix. The PR already improve the quality of the wrap.

subprojects/packagefiles/yaml-cpp/meson.build
eli-schwartz: Ha, interesting. Maybe blindly added from upstream?

eli-schwartz:
> The PR already improve the quality of the wrap.

"Cleaning up" is also not high priority to merge though, since it doesn't really affect things? But besides, I do think that the fact that part of this wrap doesn't successfully compile because it isn't enabled in CI is a good reason to not keep skipping a resolution.

simonspa:
One thing this would solve is that currently this wrap does not work under Ubuntu 22.04 LTS due to the meson version requirement:

```
Run-time dependency yaml-cpp found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency yaml-cpp
Downloading yaml-cpp source from https://github.com/jbeder/yaml-cpp/archive/refs/tags/0.8.0.zip
Download size: 1308318
Downloading: ..........
Downloading yaml-cpp patch from https://wrapdb.mesonbuild.com/v2/yaml-cpp_0.8.0-1/get_patch
Download size: 4073
Downloading: ..........
Executing subproject yaml-cpp 
subprojects/yaml-cpp-0.8.0/meson.build:1:0: ERROR: Meson version is 0.61.2 but project requires >=0.63.0
```

stephanlachnit:
@eli-schwartz how about:

- I removed the tests for now and we merge this
- I will open a new PR where I re-add the tests with wrapdb gtest and we go from there?

stephanlachnit:
@eli-schwartz another reason to drop the tests: yaml-cpp 0.8 is so old that gtest 1.17 does not work at all. When trying to build from wrapdb (`meson setup build -Dwraps=yaml-cpp -Dyaml-cpp:tests=enabled`), I was not able to get gtest 1.11 even with version restrictions in the call to `dependency('gmock', version: '1.11.0')`

stephanlachnit:
@eli-schwartz I managed to fix the tests by switching the default to C++17

neheb:
```
0.63.0: cpp_std in subproject default_options
```
this gets avoided with override_options.

stephanlachnit:
> ```
> 0.63.0: cpp_std in subproject default_options
> ```
> 
> this gets avoided with override_options.

Done!

eli-schwartz:
> @eli-schwartz I managed to fix the tests by switching the default to C++17

Do they run in CI?

stephanlachnit:
> Do they run in CI?

Sorry forgot, now they do!

stephanlachnit:
@eli-schwartz can we merge this now?

PR #2432, opened 2025-10-05
abseil-cpp: enable tests, add tracing.cc to base sources

apache-hb:
adds `tracing.cc` to `absl_base_sources`, and adds tests for base, numeric, hash, crc32, container, and logging to try and catch any other missing sources.


subprojects/packagefiles/abseil-cpp/meson.build
```diff
@@ -1,7 +1,7 @@
 project(
   'abseil-cpp',
-  'cpp',
-  meson_version: '>=0.49.0',
+  ['cpp', 'c'],
```
neheb: I wouldn't do this. Instead, call add_languages when tests are enabled.

subprojects/packagefiles/abseil-cpp/meson.build
```diff
@@ -10,8 +10,17 @@ override_cpp = 'cpp_std=c++17'
 
 cpp = meson.get_compiler('cpp')
 
+tests_opt = get_option('tests').disable_auto_if(meson.is_subproject())
```
neheb, line 13: @eli-schwartz

nirbheek: Why are tests auto-disabled when a subproject? Do they take a lot of time to build?

apache-hb: Yeah they can take a few extra minutes and have some extra dependencies that abseil doesnt have by default.

nirbheek:
Will merge in a week if there are no objections.

subprojects/packagefiles/abseil-cpp/meson.build
```diff
@@ -883,3 +914,805 @@ absl_status_dep = declare_dependency(
...
+meson.override_dependency('absl_debugging', absl_debugging_dep)
+meson.override_dependency('absl_random', absl_random_dep)
+meson.override_dependency('absl_crc', absl_crc_dep)
+meson.override_dependency('absl_time', absl_time_dep)
+meson.override_dependency('absl_types', absl_types_dep)
+meson.override_dependency('absl_synchronization', absl_synchronization_dep)
+meson.override_dependency('absl_container', absl_container_dep)
+meson.override_dependency('absl_flags', absl_flags_dep)
+meson.override_dependency('absl_log', absl_log_dep)
+meson.override_dependency('absl_status', absl_status_dep)
```
bonzini: You can put these in "names" as well.

bonzini:
Looks good, with just an aesthetic comment that you can disregard too.

nirbheek:
clang-cl arm64 job failure seems to be related.

MSYS2 CLANGARM64 job failure needs investigating

apache-hb:
Seems the msys2 failure is due to https://github.com/msys2/MSYS2-packages/issues/2058, I would guess the msys2/setup-msys2 action needs to updated to resolve that failure.

PR #2434, opened 2025-10-08
Fix Windows naming issue with grpc find_library call to protoc

WillAyd:


subprojects/packagefiles/grpc/meson.build
```diff
@@ -56,2079 +56,2081 @@ absl_type_traits_dep = dependency('absl_type_traits')
...
 libssl_dep = dependency('libssl')
 protobuf_dep = dependency('protobuf')
-protoc_dep = dependency(
-  'protoc',
-  required: false,
-)
+protoc_dep = dependency('protoc', required: false)
 if not protoc_dep.found()
-  protoc_dep = cpp.find_library('protoc')  # pkgconfig version does not provide protoc.pc
+    # pkgconfig version does not provide protoc.pc
```
WillAyd: This branch is the core of the change; everything else is done to appease CI (looks like this file hasn't been touched in a while)

WillAyd:
Lots of CI failures here but I am of the impression that they don't have anything to do with this change and are pre-existing failures

WillAyd:
Upon further inspection I believe the VS/macOS errors are due to an upstream bug - see https://github.com/grpc/grpc-dotnet/issues/2646

@neheb any thoughts on this PR?

neheb:
pass something like

```
cpp.get_supported_arguments('-Wno-missing-template-arg-list-after-template-kw')
```

to cppflags.

WillAyd:
That helped, although I'm guessing the remaining failures stem from this wrap version being a few years old. 

@klokik  I see from the comments that this file was generated by a script - do you know where that script is? Can we run it against the latest gRPC release?

WillAyd:
@neheb looks like we probably need to update this configuration to make it compatiable with newer abseil releases.

My guess is that may end up being a significant amount of work. In its current state, the configuration also requires a lot of formatting updates. Should we do the formtting updates as a separate pre-cursor, or is the preference to still just do all of that in one PR?

neheb:
you can have multiple commits. No issue with that.

PR #2436, opened 2025-10-09
cpr: install headers and library

stephanlachnit:
The shared library was not installed by default, this PR fixes this.


stephanlachnit:
Is this error on windows

```
ERROR: File 'subprojects/cpr-1.12.0/cpr.lib' could not be found
```

a new Meson bug? I don't think there should be a `.lib` file

eli-schwartz:
Why shouldn't there be one? It is kinda mandatory for shared libraries to have one if they expose an ABI. Just the linker doesn't actually let you say whether you expect one to exist.

The fact that one wasn't created by the linker as a secondary output when producing the DLL means the DLL is useless. The tests should fail to link, but if you take a look at the logs the tests don't actually run because GTest is optional and therefore isn't found if only available as a subproject. Oops. :)

stephanlachnit:
Ah, I guess this is the typical MSVC symbol visibility issue then?

eli-schwartz:
That is my most likely guess, yeah.

neheb:
And it's C++. Can't do a def file.

stephanlachnit:
Included in https://github.com/mesonbuild/wrapdb/pull/2499

PR #2438, opened 2025-10-09
cjson: update to 1.7.19

mmagnuskeys:


mmagnuskeys:
Test case failures can be explained here [https://github.com/mesonbuild/wrapdb/pull/1657] from the prior CJSON upgrade.

neheb:
```
(exit status 3221225725 or signal 3221225597 SIGinvalid)
```

IIRC this is an rpath issue. static linking would fix but there's probably a different solution.

PR #2439, opened 2025-10-10
ftxui: update from 5.0.0 to 6.1.9

na-trium-144:


na-trium-144:
It seems ftxui 6.0.0 and later requires c++20 for building examples.
https://github.com/ArthurSonzogni/FTXUI/commit/e5eb822d448c8c34ad8d4d769a124887f196dd09

Should I bump cpp_std default option to c++20, or remove examples from meson.build and keep cpp_std=c++17?

neheb:
It's recommended to use override_options instead of default_options for wraps because of meson bugs before 0.63

subprojects/packagefiles/ftxui/examples/meson.build
```diff
@@ -1,2 +1,6 @@
+if get_option('examples').disabled()
```
neheb: Not sure how I feel about a feature being used as a bool.

Off the top of my head, since the examples require C++20, you could probably use

compiler.has_header()

on a C++20 header with a requires for the feature option.

na-trium-144: It seems the only C++20 feature that ftxui uses is designated initializers, not any headers.
Should we use compiler.compiles(), or compiler.has_define() with feature testing macro `__cpp_designated_initializers`?
But they don't have override_options arg to specify c++20.

neheb: My header suggestion was for C++20 specifically, not that it actually uses any C++20 headers.

You’re spot on with no override_options. Hence my suggestion. C++20 headers should return as found even if they can’t be used. That’s check_header.

na-trium-144: I can check one of c++20 headers that is not actually used by ftxui, though it doesn't looks correct for me.

neheb: bit might be better.

Concepts gets introduced in GCC10 IIRC. GCC8 introduces C++20 support IIRC.

PR #2441, opened 2025-10-12
base64: add 0.5.2

apache-hb:
Adds version 0.5.2 of [aklomps base64 library](https://github.com/aklomp/base64)

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -72,6 +72,16 @@
       "binutils-dev"
     ]
   },
+  "aklomp-base64": {
+    "_comment": [
+      "The test program has a bug that causes it to fail with a SIGILL on x86 platforms without avx512 (all CI runners)",
+      "Fix pending at https://github.com/aklomp/base64/pull/153"
+    ],
+    "build_options": [
+      "aklomp-base64:tests=enabled"
+    ],
+    "skip_tests": true
+  },
   "blueprint-compiler": {
     "_comment": "Tests require pygobject and Gtk4 typelib, and also they crash",
     "skip_tests": true
--- a/releases.json
+++ b/releases.json
@@ -141,6 +141,14 @@
       "1.0.0-1"
     ]
   },
+  "aklomp-base64": {
+    "dependency_names": [
+      "base64"
+    ],
+    "versions": [
+      "0.5.2-1"
+    ]
+  },
   "apache-orc": {
     "dependency_names": [
       "orc"
--- a/subprojects/aklomp-base64.wrap
+++ b/subprojects/aklomp-base64.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+directory = base64-0.5.2
+source_url = https://github.com/aklomp/base64/archive/refs/tags/v0.5.2.tar.gz
+source_filename = base64-0.5.2.tar.gz
+source_hash = 723a0f9f4cf44cf79e97bcc315ec8f85e52eb104c8882942c3f2fba95acc080d
+patch_directory = aklomp-base64
+
+[provide]
+dependency_names = base64
--- a/subprojects/packagefiles/aklomp-base64/meson.build
+++ b/subprojects/packagefiles/aklomp-base64/meson.build
@@ -1,0 +1,182 @@
+project(
+  'aklomp-base64',
+  'c',
+  version: '0.5.2',
+  license: 'BSD-2-Clause',
+  meson_version: '>=0.59.0',
+  default_options: ['warning_level=3'],
+)
+
+pkg = import('pkgconfig')
+
+cc = meson.get_compiler('c')
+tests_opt = get_option('tests').disable_auto_if(meson.is_subproject())
+
+c_args = ['-DBASE64_EXPORTS']
+config = configuration_data()
+
+if get_option('default_library') == 'static'
+  c_args += ['-DBASE64_STATIC_DEFINE']
+endif
+
+is_x86 = host_machine.cpu_family() in ['x86', 'x86_64']
+is_arm = host_machine.cpu_family() in ['arm', 'aarch64']
+config.set10('BASE64_WITH_NEON64', host_machine.cpu_family() == 'aarch64')
+
+arch_flags = {}
+
+if cc.get_id() == 'msvc'
+  if is_x86
+    arch_flags += {
+      'SSSE3': [],
+      'SSE41': [],
+      'SSE42': [],
+      'AVX': ['/arch:AVX'],
+      'AVX2': ['/arch:AVX2'],
+      'AVX512': ['/arch:AVX512'],
+    }
+  elif is_arm
+    arch_flags += {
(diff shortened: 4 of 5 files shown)


subprojects/packagefiles/base64/meson.build
```diff
@@ -0,0 +1,90 @@
...
+  'base64',
+  'c',
+  version: '0.5.2',
+  license: 'BSD-2-Clause',
+  meson_version: '>=0.59.0',
+  default_options: ['warning_level=3'],
+)
+
+cc = meson.get_compiler('c')
+tests_opt = get_option('tests').disable_auto_if(meson.is_subproject())
```
neheb: ping @eli-schwartz

subprojects/aklomp-base64.wrap
bgilbert, line 1: This should probably be `aklomp-base64.wrap`.  `base64` is a pretty generic name.

subprojects/packagefiles/base64/meson.build
```diff
@@ -0,0 +1,90 @@
...
+    'SSSE3': 'ssse3',
+    'SSE41': 'sse4.1',
+    'SSE42': 'sse4.2',
+    'AVX': 'avx',
+    'AVX2': 'avx2',
+  }
+  foreach feature, flag : features
+    if cc.has_argument('-m' + flag)
+      config.set('HAVE_' + feature, 1)
+      c_args += ['-DHAVE_' + feature, '-m' + flag]
```
bgilbert: Upstream sets these on a per-file basis, not for the whole library.  The code here produces a library that requires every CPU feature supported by the _compiler_.

releases.json
```diff
@@ -141,6 +141,14 @@
       "1.0.0-1"
     ]
   },
+  "aklomp-base64": {
+    "dependency_names": [
+      "aklomp-base64"
```
bgilbert: The dependency name, on the other hand, should match upstream's pkg-config name.  Upstream doesn't have one, but the CMake config is `base64-config.cmake`.  So this should probably be `base64`.

subprojects/packagefiles/aklomp-base64/meson.build
```diff
@@ -0,0 +1,182 @@
...
+)
+
+is_x86 = host_machine.cpu_family() in ['x86', 'x86_64']
+config.set10('BASE64_WITH_NEON32', neon32_supported)
+config.set10('BASE64_WITH_NEON64', host_machine.cpu_family() == 'aarch64')
+
+present_flags = {}
+arch_flags = {}
+
+if cc.get_argument_syntax() == 'msvc' and cc.get_id() != 'clang-cl'
```
bgilbert: Could just be

```suggestion
if cc.get_id() == 'msvc'
```

subprojects/packagefiles/aklomp-base64/meson.build
```diff
@@ -0,0 +1,182 @@
...
+    'AVX': ['-mavx'],
+    'AVX2': ['-mavx2'],
+    'AVX512': ['-mavx512vl', '-mavx512vbmi'],
+  }
+endif
+
+foreach feature, flags : arch_flags
+  present = is_x86 and cc.has_multi_arguments(flags)
+  config.set10('BASE64_WITH_' + feature, present)
+  present_flags += {
```
bgilbert: You could simplify by removing `present_flags` entirely and doing `arch_flags[feature] = []` if not present.

subprojects/packagefiles/aklomp-base64/meson.build
```diff
@@ -0,0 +1,182 @@
...
+
+  present_flags += {
+    'SSSE3': true,
+    'SSE41': true,
+    'SSE42': true,
+  }
+
+  config.set10('BASE64_WITH_SSSE3', is_x86)
+  config.set10('BASE64_WITH_SSE41', is_x86)
+  config.set10('BASE64_WITH_SSE42', is_x86)
```
bgilbert: `cc.has_multi_arguments([])` evaluates to true, so these are redundant.

subprojects/packagefiles/aklomp-base64/meson.build
```diff
@@ -0,0 +1,182 @@
...
+tests_opt = get_option('tests').disable_auto_if(meson.is_subproject())
+
+c_args = ['-DBASE64_EXPORTS']
+config = configuration_data()
+
+if get_option('default_library') == 'static'
+  c_args += ['-DBASE64_STATIC_DEFINE']
+endif
+
+neon32_supported = host_machine.cpu_family() == 'arm' and cc.has_argument(
```
bgilbert: Couldn't arm and aarch64 use the `arch_flags` mechanism?  I think the `is_x86` check isn't actually necessary because the compiler should just fail to support the options that aren't relevant for this CPU.

ci_config.json
```diff
@@ -72,6 +72,13 @@
       "binutils-dev"
     ]
   },
+  "aklomp-base64": {
+    "_comment": "The test program has a bug that causes it to fail with a SIGILL on x86 platforms without avx512 (all CI runners)",
```
bgilbert: Has this been reported upstream?  At minimum we should link to the issue report so it's possible to follow up.  Is there any way to work around the bug from `meson.build` without patching the source?

apache-hb: I've published a fix on the projects github page and linked it in the comment

subprojects/packagefiles/aklomp-base64/meson.build
```diff
@@ -0,0 +1,193 @@
...
+)
+
+codecs = []
+
+codecs += [
+  static_library(
+    'aklomp-base64-generic',
+    'lib/arch/generic/codec.c',
+    'lib/tables/tables.c',
+    cfg,
```
bgilbert: `cfg` is a header built at setup time, so doesn't need to be explicitly included in any source lists.

subprojects/packagefiles/aklomp-base64/meson.build
```diff
@@ -0,0 +1,193 @@
...
+  link_with: libbase64,
+)
+
+install_headers('include/libbase64.h')
+
+meson.override_dependency('base64', base64_dep)
+
+pkg.generate(
+  libbase64,
+  name: 'aklomp-base64',
```
bgilbert: Must match the provide name.

```suggestion
  name: 'base64',
```

subprojects/packagefiles/aklomp-base64/meson.build
```diff
@@ -0,0 +1,193 @@
...
+  description: 'Fast Base64 stream encoder/decoder in C99, with SIMD acceleration',
+  url: 'https://github.com/aklomp/base64',
+  version: meson.project_version(),
```
bgilbert: This is the default.

bgilbert:
Thanks!

PR #2446, opened 2025-10-14
imgui-docking: add sdl3_gpu support

IogaMaster:
I forgot to add it to the last pr
I thought that the sdl3 renderer was the gpu api, my bad. 
This is now fixed, if we need to do something special then guide me


IogaMaster:
@neheb

IogaMaster:
Thanks, i don't know how that slipped my mind lol

IogaMaster:
This appears to not function correctly

PR #2449, opened 2025-10-15
Add iceberg-cpp wrap

WillAyd:
@wgtmac

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1574,6 +1574,14 @@
       "1.9-1"
     ]
   },
+  "iceberg-cpp": {
+    "dependency_names": [
+      "iceberg"
+    ],
+    "versions": [
+      "064d53bfe5f51d6a03fd14d15dbe95de1e2a0118-1"
+    ]
+  },
   "icu": {
     "dependency_names": [
       "icu-data",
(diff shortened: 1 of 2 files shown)


releases.json
```diff
@@ -1530,6 +1530,15 @@
       "1.9-1"
     ]
   },
+  "iceberg-cpp": {
+    "dependency_names": [
+      "iceberg",
+      "iceberg_rest",
```
WillAyd: Upstream it looks like this mistakenly calls `meson.override_dependency('iceberg-rest', ...)` so need to fix that before release

wgtmac: I think `iceberg_rest` is far from being finalized. Perhaps add it later?

WillAyd: No problem

WillAyd:
I can't reproduce the issue here nor did the meson issue tracker yield anything. The iceberg-cpp default arguments contains: 

```python
project(
    'iceberg',
    'cpp',
    version: '0.2.0',
    license: 'Apache-2.0',
    meson_version: '>=1.3.0',
    default_options: [
        'cpp_std=c++23,c++latest',
  ],
)
```

When I run locally it appears to correctly pass `-std=c++23` but in CI I see `-std=c++23,c++latest` - @eli-schwartz any idea what might be going awry here?

WillAyd:
I think the "multiple standard" issue is described upstream in https://github.com/mesonbuild/meson/issues/15202

bgilbert:
Please rebase to fix the CI failures.

PR #2455, opened 2025-10-18
crow: add 1.3.0 wrap

na-trium-144:
https://github.com/CrowCpp/Crow

- translated to meson everything except tests/fuzz (requires environment variable and I don't know about ossfuzz), crow_amalgamate (I think it's not needed when using as subproject), and installation.
- all options are disabled by default because it's disabled by default in original CMakeLists.txt.
- `crow:ssl=auto` in CI config because openssl is not available on VisualStudio x86 CI.
- `catch2:tests=false` in CI config because it doesn't compile on MinGW (upstream issue?)

### DIFF
--- a/subprojects/packagefiles/crow/examples/meson.build
+++ b/subprojects/packagefiles/crow/examples/meson.build
@@ -1,0 +1,149 @@
+if not get_option('examples')
+  subdir_done()
+endif
+
+executable(
+  'helloworld',
+  'helloworld.cpp',
+  dependencies: crow_dep,
+  override_options: crow_private_options,
+  cpp_args: crow_private_args,
+)
+
+if zlib_dep.found()
+  executable(
+    'example_compression',
+    'example_compression.cpp',
+    dependencies: crow_dep,
+    override_options: crow_private_options,
+    cpp_args: crow_private_args,
+  )
+endif
+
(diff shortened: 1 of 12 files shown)


subprojects/packagefiles/crow/examples/meson.build
```diff
@@ -0,0 +1,149 @@
...
+
+executable(
+  'example_unix_socket',
+  'example_unix_socket.cpp',
+  dependencies: crow_dep,
+  override_options: crow_private_options,
+  cpp_args: crow_private_args,
+)
+
+if meson.get_compiler('cpp').get_id() == 'msvc'
```
neheb: is this also relevant for clang-cl?

subprojects/packagefiles/crow/meson.build
```diff
@@ -0,0 +1,102 @@
...
+endif
+
+crow_inc = include_directories('include')
+
+crow_dep = declare_dependency(
+  include_directories: crow_inc,
+  dependencies: crow_deps,
+  compile_args: crow_public_args,
+)
+if meson.version().version_compare('>=0.54.0')
```
neheb: I'd bump the minimum version instead. 0.49 is really old.

PR #2464, opened 2025-10-24
Bump Apache Arrow

WillAyd:


WillAyd:
FYI @kou

subprojects/packagefiles/arrow-cpp/meson.build
```diff
@@ -26,10 +26,10 @@ project(
   'arrow',
   'cpp',
   'c',
-  version: '21.0.0',
+  version: '22.0.0',
   license: 'Apache-2.0',
   meson_version: '>=1.3.0',
-  default_options: ['c_std=gnu11,c11', 'warning_level=2', 'cpp_std=c++17'],
+  default_options: ['c_std=c11', 'warning_level=2', 'cpp_std=c++17'],
```
WillAyd, line 32: I'm not entirely sure why this is problematic now, but the CI reports this error:

```
cc: error: unrecognized command-line option ‘-std=gnu11,c11’; did you mean ‘-std=gnu++11’?
```

https://github.com/mesonbuild/wrapdb/actions/runs/18844109396/job/53763852587

I don't think its a huge deal to change for now; can take a closer look if it becomes a problem in future versions

WillAyd:
I think this is as good as we are going to get for CI with Arrow 22.0.0. Windows support is still not fully implemented upstream, and Alpine is not implemented.

The macOS failure is a regression from the 21.0.0 release, but that appears to be a result of an upstream issue (see https://github.com/apache/arrow/issues/47961)

subprojects/packagefiles/arrow-cpp/meson.build
kou: I think that we can use `-std=gnu11` OR `-std=c11`.

WillAyd: Thanks - I can't remember why we needed the gnu extensions upstream in the first place, and maybe we don't any more. Will investigate that separately from this

PR #2470, opened 2025-10-29
brotli: Add -D_POSIX_SOURCE

dcbaker:
The CMake definitions rely on the fact that GCC and Clang use a gnuXX standard instead of a cXX standard. If we have a cXX standard then brotli fails to compile because of a missing definition for `fdopen`. We can fix that by defining `_POSIX_SOURCE`

I ran into this using the brotli wrap with a project defining `c_std=c99`.


dcbaker:
I don't understand why this is failing?

neheb:
```
AssertionError: Items in the first set but not the second:
'libbrotlidec'
'libbrotlicommon'
'libbrotlienc'
```

did you see this in cygwin or something?

subprojects/packagefiles/google-brotli/meson.build
```diff
@@ -158,6 +158,12 @@ brotli_decoder_dep = declare_dependency(
   dependencies: brotli_common_dep,
 )
 
+if meson.version().version_compare('>= 0.54')
```
neheb: I'd bump the main version instead. 0.54 is pretty old.

subprojects/packagefiles/google-brotli/meson.build
```diff
@@ -28,6 +28,14 @@ project(
 
 cc = meson.get_compiler('c')
 
+# Brotli relies on the default standard being used, which is gnuXX for Clang
+# and GCC on Linux-like systems. If it gets set to a cXX standard it breaks.
+# We can fix this even for cXX standards by adding `-D_POSIX_SOURCE`
+add_project_arguments(
+  '-D_POSIX_SOURCE',
```
neheb: c_args might make more sense.

Could also set override_options to c_std=gnu99 instead.

dcbaker: I thought about that, but that would mean needing to bump the minimum meson version a lot to allow fallback to `c99` for MSVC, and still might miss compilers that support _POSIX_SOURCE but not gnu99

dcbaker:
I see that everywhere, but it doesn't make any sense, I didn't add or remove any dependencies and those dependencies are absolutely in the `[provide]` section *and* now set by `override_dependency()` as well.

neheb:
It’s complaining that they’re not alphabetic.

subprojects/packagefiles/google-brotli/meson.build
neheb: what do you mean? override_options doesn't need a recent meson version.

dcbaker: To build with MSVC we'd ned to set the project options as:
```
project('brotli', default_options={'c_std=[gnu99, c99])`.
```

If you mean to add `override_options : `c_std : condition : 'gnu99' ? 'c99'`, that seems a lot more fragile than just defining `_POSIX_SOURCE`

neheb: how about _POSIX_SOURCE to c_args?

dcbaker: Sure, we can do that instead.

dcbaker:
I noticed one more issue, that cmake has updated it's log2 handling, so I've updated to match that as well.

dcbaker:
I finally figured out what that failing test meant, and it was not at all clear from the error message that it wanted me to use the `dependency_names` field in the wrap file. Ie, it wanted:
```ini
[provide]
libbrotlienc = dep_bortli_enc
```

to be replaced with:
```ini
[provide]
dependency_names = libbrotlienc
```

I opened: https://github.com/mesonbuild/wrapdb/pull/2475 to try to add a more helpful error message.

subprojects/packagefiles/google-brotli/meson.build
```diff
@@ -26,18 +26,19 @@ project(
...
+c_args = []
+
 cc = meson.get_compiler('c')
 
 dep_m = cc.find_library(
   'm',
   required: false,
 )
-if not cc.has_function(
+has_log2 = cc.has_function(
```
neheb, line 40: I personally use

```
 if cc.links('#include <math.h>\nint main(){log(0);}')
   m_dep = declare_dependency()
 else
   m_dep = cc.find_library('m')
 endif
```

to avoid a pointless dependency on libm. It's a moot point though.

subprojects/packagefiles/google-brotli/meson.build
eli-schwartz: What is the point of avoiding add_project_arguments for its intended purpose and passing it to individual targets instead?

subprojects/packagefiles/google-brotli/meson.build
eli-schwartz: If the goal is to handle c89 systems that don't have log2 support then your code would be incorrect, since finding -lm when log2 doesn't link standalone doesn't mean that the found -lm has log2...

subprojects/packagefiles/google-brotli/meson.build
neheb: It's one line vs 3. I''m not a fan of the formatting that was introduced.

subprojects/packagefiles/google-brotli/meson.build
dcbaker: And that's the thing that the CMake build is trying to solve, not "where does log2 come from", but "do you have log2"?

I didn't check, but I presume that they have their own log2 implementation that they use if your C stdlib doesn't have one.

subprojects/packagefiles/google-brotli/meson.build
dcbaker: Well, I think that the format settings that change:
```meson
add_project_arguments('-D_POSIX_SOURCE', language : 'c')
```
to
```meson
add_project_arguments(
  '-D_POSIX_SOURCE',
  language : 'c',
)
```
is interesting, but I'm just running the formatting tool after writing the former.

subprojects/packagefiles/google-brotli/meson.build
neheb: I find it hard to believe a system supports meson but not log2.

dcbaker: Looking at the code, android is the main non-implementor of `log2` (plus old MSVC).

Which maybe isn't a big deal, but it seems odd to not match what CMake is doing unless what they're doing is wrong or covered by some other meson construct.

eli-schwartz: I find it very easy to believe that a system doesn't support meson natively, but does allow cross compiling TO that system from another system which does support meson.

And that the build machine running meson does have log2, but the host machine perhaps doesn't have log2.

dcbaker:
And, speaking of things that Meson does CMake doesn't, the `-DBROTLI_BUILD_PORTABLE` define was removed in Brotli 1.1.0

subprojects/packagefiles/google-brotli/meson.build
neheb: if android is the concern, log2 seems to have been supported since 4.4 from googling.

According to https://apilevels.com/ , google dropped support for less than 5 in 2023.

Btw is there CI for Android that can be implemented here? Would be good to have.

dcbaker: I'm not arguing that something is sensible, I'm arguing that for ports we should not derivative from the official build system except when:
  1. Meson provides different way to achieve the same thing (sanitizers, static vs shared vs both, etc)
  2. is for some platform we don't support
  3. cannot be done
  4. there is a bug in the upstream build system

Maybe this case is moot since Brotli 1.2 was released recently and it has dropped this option.

eli-schwartz: I agree that the right thing to do is, if upstream supports checking for some distinction in a given wrap version, that we faithfully reimplement it in a meson port for that version.

If instead of fixing this wrap we upgrade to a new upstream release that removes the option -- that is a valid solution too.

neheb: Which would make my solution applicable. Apparently even though glibc moved all pthread functions into libc, some math ones remain external.

dcbaker: If I drop the log2 patch are you happy to merge the rest, and then we can look at 1.2, whcih will require everything except the log2 stuff anyway?

dcbaker: @neheb: ping: ^

PR #2475, opened 2025-10-30
tools/sanity_check: provide better error message for old [provide] syntax

dcbaker:
When I ran into this error it wasn't obvious what it was checking for and what it expected me to do. This attempts to add a more helpful error message with clear instructions on what to fix.

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -328,13 +328,13 @@ class TestReleases(unittest.TestCase):
                 # FIXME: Not all wraps currently comply, only check for wraps we modify.
                 if extra_checks and patch_path:
                     # Intentional leading whitespace
-                    errmsg = textwrap.dedent('''
+                    errmsg = textwrap.dedent(f'''
+                        In the meson.build file use `meson.override_dependency('<name>', <name>_dep)`
+                        for each dependency. Ensure that the minimum `meson_version` in the `project()`
+                        call is at least `>= 0.54`.
 
-                        In the meson.build file use `meson.override_dependenc('foo', foo_dep)`
-                        for each dependency.
-
-                        In subprojects/foo.wrap, replace `name = name_dep` entries with
-                        `dependency_names = foo`
+                        In subprojects/{name}.wrap, replace `<name> = <name>_dep` entries with
+                        `dependency_names = <name>`.
 
                         See https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#overriding-dependencies-in-the-submitted-project
                         for more information.


tools/sanity_checks.py
```diff
@@ -321,8 +322,20 @@ def test_releases(self) -> None:
                 # Downstream ports shouldn't use transitional provides syntax
                 # FIXME: Not all wraps currently comply, only check for wraps we modify.
                 if extra_checks and patch_path:
-                    with self.subTest(step="Ports must not use 'foo = foo_dep' provide syntax; use meson.override_dependency('foo', foo_dep) and 'dependency_names = foo'.  https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#overriding-dependencies-in-the-submitted-project"):
-                        self.assertEqual(self.get_transitional_provides(config), set())
+                    # Intentional leading whitespace
+                    errmsg = textwrap.dedent('''
+
+                        In the meson.build file use `meson.override_dependenc('foo', foo_dep)`
```
bgilbert:
```suggestion
                        In the meson.build file use `meson.override_dependency('foo', foo_dep)`
```

tools/sanity_checks.py
```diff
@@ -321,8 +322,20 @@ def test_releases(self) -> None:
...
+                        for each dependency.
+
+                        In subprojects/foo.wrap, replace `name = name_dep` entries with
```
bgilbert: Let's use consistent variable names so it's clear what's being replaced:
```suggestion
                        In subprojects/{name}.wrap, replace `foo = foo_dep` entries with
```

(Of course that also requires this to be a format string.)

tools/sanity_checks.py
```diff
@@ -321,8 +322,20 @@ def test_releases(self) -> None:
                 # Downstream ports shouldn't use transitional provides syntax
                 # FIXME: Not all wraps currently comply, only check for wraps we modify.
                 if extra_checks and patch_path:
-                    with self.subTest(step="Ports must not use 'foo = foo_dep' provide syntax; use meson.override_dependency('foo', foo_dep) and 'dependency_names = foo'.  https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#overriding-dependencies-in-the-submitted-project"):
-                        self.assertEqual(self.get_transitional_provides(config), set())
+                    # Intentional leading whitespace
+                    errmsg = textwrap.dedent('''
+
+                        In the meson.build file use `meson.override_dependenc('foo', foo_dep)`
+                        for each dependency.
```
bgilbert: Optional: do we want to mention that it'll need to be conditionalized if the config supports Meson < 0.54?

bgilbert (commented):
Looks good to me.

tools/sanity_checks.py
neheb: It’s a good excuse to bump the minimum honestly. I don’t think wrapdb should be supporting real old versions.

neheb: I don’t know if I mentioned this earlier, but meson 0.49.0 under I believe Debian Jessie crashes when ternary operators are used. I asked @eli-schwartz and he said it’s a python bug, not necessarily meson.

eli-schwartz: @neheb I forget the context for this...

Anyway I think it's fine if people want to support old versions when it's as simple as putting this line inside a conditional. It's not something I would *insist* on.

We are now at the point where Debian and Ubuntu no longer provide LTS releases which still ship with meson 0.53 -- Debian buster is *gone* and bullseye has 0.56, Ubuntu 22.04 has 0.61 and any older than that needs a "pro" subscription (20.04 is still "kinda" supported if you really want meson 0.53).

If you had asked me a few months ago, I'd have said don't bother dropping support for meson 0.53 for wraps that currently allow it. But at this point maybe it's fine.

dcbaker: I've added that the minimum should be bumped to at least `>= 0.54`. 0.54 is from early 2020, I think it's okay to ask people to use that unconditionally.

neheb: Context was me saying a ? b : c was broken on Debian Jessie (meson 0.49) but

```
if a
  b
else
  c
endif
```

works. This may have been on IRC. I'd need to look at logs.

As far as older OSes are concerned, pip is a solution.

neheb: Found it: https://gist.github.com/neheb/b2c274a6b20ecc682fc7daf42c0f8414

tools/sanity_checks.py
```diff
@@ -321,8 +322,20 @@ def test_releases(self) -> None:
...
                 if extra_checks and patch_path:
-                    with self.subTest(step="Ports must not use 'foo = foo_dep' provide syntax; use meson.override_dependency('foo', foo_dep) and 'dependency_names = foo'.  https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#overriding-dependencies-in-the-submitted-project"):
-                        self.assertEqual(self.get_transitional_provides(config), set())
+                    # Intentional leading whitespace
+                    errmsg = textwrap.dedent(f'''
+                        In the meson.build file use `meson.override_dependency({name}, {name}_dep)`
+                        for each dependency. Ensure that the minimum `meson_version` in the `project()`
+                        call is at least `>= 0.54`.
+
+                        In subprojects/foo.wrap, replace `{name} ={name}_dep` entries with
```
bgilbert:
```suggestion
                        In subprojects/{name}.wrap, replace `{name} = {name}_dep` entries with
```

tools/sanity_checks.py
```diff
@@ -321,8 +322,20 @@ def test_releases(self) -> None:
                 # Downstream ports shouldn't use transitional provides syntax
                 # FIXME: Not all wraps currently comply, only check for wraps we modify.
                 if extra_checks and patch_path:
-                    with self.subTest(step="Ports must not use 'foo = foo_dep' provide syntax; use meson.override_dependency('foo', foo_dep) and 'dependency_names = foo'.  https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#overriding-dependencies-in-the-submitted-project"):
-                        self.assertEqual(self.get_transitional_provides(config), set())
+                    # Intentional leading whitespace
+                    errmsg = textwrap.dedent(f'''
+                        In the meson.build file use `meson.override_dependency({name}, {name}_dep)`
```
bgilbert: We know the wrap is called `{name}.wrap` but the dependency may not be called `{name}`, and it seems confusing to imply otherwise.  The provide names are returned from `self.get_transitional_provides()`, so we could use an actual name here, but if you don't want to go to that trouble I think `foo` is better than giving a name that's potentially misleading.

dcbaker: Okay, I've used `<name>` in those places.

We could do the `get_transitional_provides()` thing, but that seems like a lot of work for a fairly uncommon issue.

tools/sanity_checks.py
bgilbert: Neither of these were addressed.

tools/sanity_checks.py
```diff
@@ -321,8 +322,20 @@ def test_releases(self) -> None:
                 # Downstream ports shouldn't use transitional provides syntax
                 # FIXME: Not all wraps currently comply, only check for wraps we modify.
                 if extra_checks and patch_path:
-                    with self.subTest(step="Ports must not use 'foo = foo_dep' provide syntax; use meson.override_dependency('foo', foo_dep) and 'dependency_names = foo'.  https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#overriding-dependencies-in-the-submitted-project"):
-                        self.assertEqual(self.get_transitional_provides(config), set())
+                    # Intentional leading whitespace
+                    errmsg = textwrap.dedent(f'''
+                        In the meson.build file use `meson.override_dependencyl<name>, <name>_dep)`
```
bgilbert:
```suggestion
                        In the meson.build file use `meson.override_dependency(<name>, <name>_dep)`
```

tools/sanity_checks.py
```diff
@@ -321,8 +322,20 @@ def test_releases(self) -> None:
                 # Downstream ports shouldn't use transitional provides syntax
                 # FIXME: Not all wraps currently comply, only check for wraps we modify.
                 if extra_checks and patch_path:
-                    with self.subTest(step="Ports must not use 'foo = foo_dep' provide syntax; use meson.override_dependency('foo', foo_dep) and 'dependency_names = foo'.  https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#overriding-dependencies-in-the-submitted-project"):
-                        self.assertEqual(self.get_transitional_provides(config), set())
+                    # Intentional leading whitespace
+                    errmsg = textwrap.dedent(f'''
+                        In the meson.build file use `meson.override_dependency(<name>, <name>_dep)`
```
bgilbert: Oh whoops, just noticed this.

```suggestion
                        In the meson.build file use `meson.override_dependency('<name>', <name>_dep)`
```

bgilbert (approved):

tools/sanity_checks.py
dcbaker: Gah, getting all the formatting right in a string is a pain.

neheb:
Conflicts.

bgilbert:
An old version of this commit was merged as part of #2470.  Fixed conflicts.

PR #2479, opened 2025-11-01
libffi: Require Meson >= 1.9.2 for MSVC/ARM64 toolchain

### LINKED ISSUE #2469: libffi fails to configure in MSVC cross-build (vs2019, build x64, host ARM64)
See https://gitlab.gnome.org/GNOME/glib/-/jobs/5728547#L893

```
FAILED: subprojects/libffi-3.5.2/src/ffi-8.dll.p/meson-generated_.._preprocessor_0.p_win64_armasm.masm.obj 
"armasm64" "-isubprojects\libffi-3.5.2\src\ffi-8.dll.p" "-nologo" "-g" "-isubprojects\libffi-3.5.2\src" "-i..\subprojects\libffi-3.5.2\src\aarch64" "-i..\subprojects\libffi-3.5.2\include" "-isubprojects\libffi-3.5.2\include" "-i..\subprojects\libffi-3.5.2" "-isubprojects\libffi-3.5.2" "-i..\subprojects\libffi-3.5.2\src" "-isubprojects\libffi-3.5.2\src" "-isubprojects\libffi-3.5.2\src\ffi-8.dll.p" -o subprojects/libffi-3.5.2/src/ffi-8.dll.p/meson-generated_.._preprocessor_0.p_win64_armasm.masm.obj subprojects/libffi-3.5.2/src/preprocessor_0.p/win64_armasm.masm
edge with deps=gcc but no depfile makes no sense
```

I can try to debug this further in the next few days :)

> **Note** the same issue happens with latest Meson: 1.9.1

lb90:
Fixes #2469


lb90:
Note: this is based on https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi/-/blob/meson/src/meson.build?ref_type=heads but the licensing is unclear. Maybe we can ask @nirbheek?

lb90:
Ci failures look unrelated to the PR:

* **Alpine (riscv64)**: libffi test failures
* **MacOS (arm64)**: libffi test failures
* **VisualStudio-clang-cl**: libffi test timeout

xclaesse:
I marked the fix in Meson for backport in 1.9.2. TBH, I'm not sure we want to maintain workarounds like that, I would just `error('Windows aarch64 requires Meson >= 1.9.2')`. On Windows it's trivial to pip install latest Meson, unlike Linux distro where it can take the age of the universe to update a package (I'm still building products on Ubuntu 20.04...).

lb90:
Yeah, makes sense! Thanks a lot for backporting

Should I add the `error('Windows aarch64 requires Meson >= 1.9.2')`, then? I don't know if it's worth upgrading the wrap version number just for that.

xclaesse:
> Should I add the `error('Windows aarch64 requires Meson >= 1.9.2')`, then? I don't know if it's worth upgrading the wrap version number just for that.

I think that's simple enough that it's worth it, to avoid others hitting the same issue. We already do that kind of version checks.

PR #2481, opened 2025-11-02
github CI: add Windows ARM64 

neheb:
ping @bgilbert


neheb:
Meh seems ilammy/msvc-dev-cmd is broken

neheb:
hrm nvm.

PR #2482, opened 2025-11-03
spdlog: add missing override_dependency

neheb:
No idea why CI didn't catch it.


neheb:
I see what the issue was. Old spdlog directory.

PR #2486, opened 2025-11-08
s2n-bignum: add wrap

apache-hb:
Adds a wrap for [s2n-bignum](https://github.com/awslabs/s2n-bignum/tree/main). This package doesn't have any versioned releases so I picked the latest commit to their mainline branch.


apache-hb:
Wow piping everything through `tr ';' '\n'` actually worked, thats horrible D:

bgilbert:
By policy, WrapDB requires an upstream release and cannot ship a random Git commit.  Can you talk to upstream about tagging a release?

apache-hb:
I have raised this with the maintainers but havent gotten a response. Is there anything I can do like tag a release myself? I know s2n-bignum has no backwards compatibility garuantees so every new release is a breaking change. Would it be acceptable to create a github fork that just tags each commit as a new `x.0.0` release?

bgilbert:
If upstream isn't willing to create releases, there's nothing we can do unfortunately.

PR #2495, opened 2025-11-17
pcre2: updates for 10.47

bgilbert:
`memfd_create()` and `secure_getenv()` should always have been checked unconditionally, and now `mkostemp()` too.  Upstream removed conditions on sealloc enablement.

Link with version scripts if supported.

### DIFF
--- a/subprojects/packagefiles/pcre2/meson.build
+++ b/subprojects/packagefiles/pcre2/meson.build
@@ -208,12 +208,30 @@ if host_machine.system() == 'windows'
   general_c_args += ['-D_CRT_SECURE_NO_DEPRECATE', '-D_CRT_SECURE_NO_WARNINGS']
 endif
 
+have_version_scripts = c_compiler.has_link_argument(
+  '-Wl,--version-script,@0@/src/libpcre2-8.sym'.format(
+    meson.current_source_dir(),
+  ),
+)
+foreach lib : ['pcre2-8', 'pcre2-posix', 'pcre2-16', 'pcre2-32']
+  set_variable(
+    lib.underscorify() + '_link_args',
+    have_version_scripts ? [
+      '-Wl,--version-script,@0@/src/lib@1@.sym'.format(
+        meson.current_source_dir(),
+        lib,
+      ),
+    ] : [],
+  )
+endforeach
+
 pcre2_8_lib = library(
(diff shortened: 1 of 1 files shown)


subprojects/packagefiles/pcre2/meson.build
```diff
@@ -61,27 +59,24 @@ if not get_option('jit').disabled() and platform_is_unixlike
...
-    endif
+foreach e : [
+  ['mkostemp', 'stdlib.h'],
+  ['memfd_create', 'sys/mman.h'],
+  ['secure_getenv', 'stdlib.h'],
+]
+  if c_compiler.has_function(
+    e[0],
+    prefix: '#include <@0@>'.format(e[1]),
+    args: ['-D_GNU_SOURCE'],
```
neheb: I wonder if this should not be simplified to just has_function.

when adding prefix, I've noticed the behavior of has_function changes to be more similar to has_header_symbol. That is, it stops finding the function if adding headers and wrong args.

bgilbert: Yeah, good call.

neheb: I looked at musl and glibc just in case. Both have all these functions as normal ones and not macros.

subprojects/packagefiles/pcre2/meson.build
```diff
@@ -208,12 +208,30 @@ if host_machine.system() == 'windows'
...
 
+have_version_scripts = c_compiler.has_link_argument(
+  '-Wl,--version-script,@0@/src/libpcre2-8.sym'.format(
+    meson.current_source_dir(),
+  ),
+)
+foreach lib : ['pcre2-8', 'pcre2-posix', 'pcre2-16', 'pcre2-32']
+  set_variable(
+    lib.underscorify() + '_link_args',
+    have_version_scripts ? [
```
neheb, line 219: Could use https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerget_supported_link_arguments

Not sure if worth it.

bgilbert: I did that first, but of course it spewed all four linker script paths in the configure output.  This seemed a bit friendlier.

PR #2498, opened 2025-11-17
sqlite3: update from 3.50.4 to 3.51.0

bgilbert:
Add `carray` and `percentile` extensions, disabled by default per upstream.

Use the project version as the library version, matching upstream's current autoconf tarballs (which actually use autosetup now).

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -1454,6 +1454,11 @@
       "--timeout-multiplier=2"
     ]
   },
+  "sqlite3": {
+    "build_options": [
+      "sqlite3:all-extensions=enabled"
+    ]
+  },
   "sqlitecpp": {
     "_comment": "required to not use the default sqlite3 in MacOS to avoid compilation errors/tests failures, also add pkgconfig path to find sqlite3(keg-only)",
     "build_options": [


subprojects/packagefiles/sqlite3/meson.build
```diff
@@ -141,6 +143,9 @@ sqlite3_dep = declare_dependency(
   include_directories: depinc,
   link_with: libsqlite3,
 )
+if meson.version().version_compare('>=0.54.0')
```
neheb, line 153: I'd update the minimum version instead.

bgilbert: SQLite is widely used and pretty low in the dependency stack, so I figure it's worth a couple conditionals to avoid narrowing its Meson support.

neheb: Up to you. I'd update since meson is very easy to update through pip.

bgilbert: Oh indeed.  I do think there's some value in supporting distro-packaged Meson though.

PR #2499, opened 2025-11-17
cpr: update to 1.14.1

stephanlachnit:
Will fail to build on Windows due to https://github.com/mesonbuild/wrapdb/pull/2436#issuecomment-3397189932 (this problem already persists with current cpr). I pledge to merge this anyway since this doesn't change the status quo. Alternatively, I could disable Windows builds in CI and error in the Meson build files if this is preferred.


stephanlachnit:
Opened an issue about symbol visibility here: https://github.com/libcpr/cpr/issues/1273

subprojects/packagefiles/cpr/meson.build
```diff
@@ -1,31 +1,40 @@
 project(
   'cpr',
   'cpp',
-  version: '1.12.0',
+  version: '1.13.0',
   license: 'MIT',
   default_options: 'cpp_std=c++17',
 )
```
bgilbert, line 7: Let's also add `meson_version: '>=0.63.0'` while we're here.  (Alternatively `0.54.0` if you move `cpp_std` to `override_options` on the `library()` and `executable()` calls, preferably via an intermediate variable.)

subprojects/packagefiles/cpr/meson.build
```diff
@@ -1,31 +1,40 @@
...
   'libcurl',
-  version: '>=7.64.0',
+  version: '>=7.71.0',
 )
 thread_dep = dependency('threads')
 
 includes = include_directories('.', 'include')
 subdir('cpr')
 
+includes = include_directories('include')
```
bgilbert, line 21: I think `.` was included because `cpr/cpr.h` includes `cpr/cprver.h`, and the latter is generated in `cpr/` in the build directory and currently is not installed.  So this will break the includes.

You should install `cprver.h`, of course, but that won't help a superproject or sibling subproject find that header when cpr hasn't been installed.  I'd suggest generating `cprver.h` from a new `include/cpr/meson.build`, and also calling `install_headers()` from there.  (I'm open to other ideas, though.)

If you do that, and invoke `subdir('include/cpr')` before `subdir('cpr')`, `includes` shouldn't need to contain `'.'` at all.

bgilbert (commented):
Better to let Windows CI continue to fail, I think.

subprojects/packagefiles/cpr/meson.build
stephanlachnit: AH right, done!

stephanlachnit:
Updated to 1.14.1

neheb:
```
2025-12-22T15:18:08.3728429Z DEPRECATION: Variable substitution with boolean value 'CURL_CA_BUNDLE' is deprecated.
2025-12-22T15:18:08.3728437Z 
2025-12-22T15:18:08.3728636Z subprojects\curl-8.12.1\meson.build:1751:21: ERROR: Fatal warnings enabled, aborting
2025-12-22T15:18:08.3728689Z 
```

Interesting.

stephanlachnit:
> ```
> 2025-12-22T15:18:08.3728429Z DEPRECATION: Variable substitution with boolean value 'CURL_CA_BUNDLE' is deprecated.
> 2025-12-22T15:18:08.3728437Z 
> 2025-12-22T15:18:08.3728636Z subprojects\curl-8.12.1\meson.build:1751:21: ERROR: Fatal warnings enabled, aborting
> 2025-12-22T15:18:08.3728689Z 
> ```
> 
> Interesting.

@neheb this seems unrelated to the cpr wrap, can we merge this anyway?

stephanlachnit:
@bgilbert @neheb can we merge this? The 1.12 branch is unsupported upstream by now, I would really like to update to 1.14

PR #2500, opened 2025-11-17
pcre2: move string formatting out of `set_variable()` call

bgilbert:
Avoid exception in `meson rewrite kwargs info project /` on Meson 1.9.1.

https://github.com/mesonbuild/meson/issues/15261


bgilbert:
Also disable the license check on Python 3.8, which has an older version of `license-expression` that doesn't understand pcre2's license expression.

PR #2504, opened 2025-11-19
sol2-3.3.0: add wrap file

IogaMaster:


subprojects/sol2.wrap
```diff
@@ -0,0 +1,11 @@
...
+directory = sol2-3.3.0
+source_url = https://github.com/ThePhD/sol2/archive/refs/tags/v3.3.0.tar.gz
+source_filename = sol2-3.3.0.tar.gz
+source_hash = b82c5de030e18cb2bcbcefcd5f45afd526920c517a96413f0b59b4332d752a1e
+
+[provide]
+dependency_names = sol2
+
+[meson]
+options = lua=wrap, lua:lua_cpp=true
```
bgilbert: It isn't possible to declare these here.  See `ci_config.json`.

subprojects/sol2.wrap
```diff
@@ -0,0 +1,11 @@
...
+directory = sol2-3.3.0
+source_url = https://github.com/ThePhD/sol2/archive/refs/tags/v3.3.0.tar.gz
+source_filename = sol2-3.3.0.tar.gz
+source_hash = b82c5de030e18cb2bcbcefcd5f45afd526920c517a96413f0b59b4332d752a1e
+
+[provide]
+dependency_names = sol2
+
+[patch_file]
+filename = use_wrapdb_lua.patch
```
bgilbert: WrapDB does not allow patching upstream sources.  We already have a `lua` wrap that should be picked up automatically.

IogaMaster: Okay, what about this lua_cpp error?

bgilbert:
Ah, okay.  It expects to use the Lua Meson wrapper from https://github.com/OrfeasZ/lua-meson/ and pass it a `lua_cpp` option, which WrapDB's Lua Meson wrapper doesn't support.  sol2 also supports LuaJIT and doesn't expect it to support that option, so one possibility is to configure CI to build with LuaJIT instead.  Another possibility is to add `lua_cpp` support to WrapDB's wrapper.

bgilbert:
You could potentially convince the sol2 maintainers to switch to our Lua wrapper, but I don't know whether they'd be likely to agree.

bgilbert:
Again, if you want to override build options in CI, you need to modify `ci_config.json`.

bgilbert:
It turns out that sol2 supports the `use_luajit` option in Git but not in any released version, which means we can't yet use it in WrapDB.  Unfortunately, until sol2 puts out a new release, your only alternative is to add a `lua_cpp` option to WrapDB's lua wrap.

IogaMaster:
I also want to use luajit.
Can we just use the main branch

bgilbert:
Unfortunately no, WrapDB policy requires an upstream release.

IogaMaster:
Okay, I will just wrap it in my own project and handle it from there. Thanks for the help

PR #2505, opened 2025-11-19
uthash: bump for corrected license

bgilbert:
The last upstream release was in 2021 and the wrap doesn't really need maintenance.  Let's go ahead and bump the revision to finish fixing the license tag, rather than waiting for the next routine update.


subprojects/packagefiles/uthash/meson.build
```diff
@@ -3,8 +3,13 @@ project(
   'c',
   version: '2.3.0',
   license: 'BSD-1-Clause',
+  meson_version: '>=0',
```
dcbaker: You might as well set this to something reasonable, even if it's old like 0.46, which is the oldest version that really has version checking.

bgilbert: Hmm, I'd rather not imply there's a known minimum when there isn't.  (And the version check in CI wouldn't like it either.)  I can remove the line if it's controversial.

There's some value to declaring that a wrap is believed to support all versions, though, to highlight any later changes that break that.

neheb: don't wraps begin to work with 0.56 anyway?

bgilbert: Wraps existed at least as far back as the first release notes, 0.37.

neheb: yeah but they only really work since 0.56. Otherwise fallback: needs to be specified.

bgilbert: Sure, but if the parent project knows it needs `fallback`, that's available.  I wouldn't leave it in any existing wraps but this one doesn't have any dependencies.

eli-schwartz: The need for fallback: has no bearing on whether this wrap is capable of being used in such a meson version. It is quite strange to have it be mentioned at all.

(A more interesting point is that meson <0.56 is increasingly unlikely to be available in any LTS distro, so who is running such an old meson version anyway? For example even Debian Bullseye / oldoldstable has meson 0.56.2 -- and that has LTS support but not security support. The previous version, Buster, had meson 0.49 but is totally out of support since June 2024. And even that had meson 0.56.1 in buster-backports.)

eli-schwartz:
```
$ cat meson.build 
project(
  'uthash',
  'c',
  version: '2.3.0',
  license: 'BSD-1-Clause',
)

fs = import('fs')
fs.suffix('foo/bar/baz.dll')  # .dll

uthash_dep = declare_dependency(
  include_directories: include_directories('src'),
)
```

> There's some value to declaring that a wrap is believed to support all versions, though, to highlight any later changes that break that.

@bgilbert, Let's see if this warns.

```
$ meson setup b --wipe
The Meson build system
Version: 1.9.1
Source dir: /home/eschwartz/git/mesonbuild/wrapdb/subprojects/uthash-2.3.0
Build dir: /home/eschwartz/git/mesonbuild/wrapdb/subprojects/uthash-2.3.0/b
Build type: native build
Project name: uthash
Project version: 2.3.0
C compiler for the host machine: ccache cc (gcc 14.3.1 "cc (Gentoo 14.3.1_p20250801 p4) 14.3.1 20250801")
C linker for the host machine: cc ld.bfd 2.45
Host machine cpu: x86_64
meson.build:9: WARNING: Project does not target a minimum version but uses feature introduced in '1.9.0': fs.suffix.
Build targets in project: 0
WARNING: Project specifies no minimum version but uses features which were added in versions:
 * 1.9.0: {'fs.suffix'}

Found ninja-1.14.0.git at /usr/bin/ninja
```

eli-schwartz: https://mesonbuild.com/Release-notes-for-1-6-0.html#default-to-printing-deprecations-when-no-minimum-version-is-specified

(I added this BTW.)

eli-schwartz: (Note to be fair, this works based on semver, so it implies `meson_version: '>=1.0.0'` in a sense.)

bgilbert: Yeah, WrapDB has to care about pre-1.0 compatibility.  I wish Meson did too, but the line had to be drawn somewhere and semver seems like the obvious place.

But that's not the issue here.  Even when Meson doesn't warn, PR CI generates a report with all known version dependencies.  The question is then: has the PR added any new ones?  Without a `meson_version` it isn't obvious what the wrap intended to support.  (That applies to 105/276 ports, 38%.)  This makes code review harder because the wrap has to be manually checked against its previous baseline to avoid compatibility breaks.  We could teach CI how to do that, but for a drive-by contributor it's much clearer to have the minimum version right there in the config and for local builds to fail if it's breached.

As to minimum versions, I understand the current policy to be maximum possible back-compat consistent with maintainability.  There's been interest in dropping support for very old versions and I'm okay with that, but then we should really have a documented minimum so users don't have any maintenance surprises.

bgilbert: Re minimum versions, I've filed https://github.com/mesonbuild/wrapdb/issues/2511 for visibility.

PR #2506, opened 2025-11-20
libarchive: update from 3.8.1 to 3.8.3; fix function detection; add tests

bgilbert:
mesonbuild/wrapdb#1511 reported configure failures when checking for functions in internal dependencies.  #1877 worked around this by adding `partial_dependency()` wrappers that omit the library linkage.  This is okay for header checks (though for internal dependencies it needs Meson &ge; 0.57), but causes all function checks to return false (including for regular, non-internal dependencies) since the dependency object no longer points to the relevant library.  Fix this by assuming internal dependencies have the desired functions.


bgilbert:
- Alpine fails character set conversion tests: https://github.com/libarchive/libarchive/issues/1475
- MSYS2 fails on a missing test file: https://github.com/libarchive/libarchive/pull/2793
- MSVC times out, even when the timeout is increased to two hours.  The failures don't all happen in the same place and I haven't tracked them down.

neheb:
The disablers I added in order to have the generated config be as close as possible to the one generated by cmake. No idea if it matters.

bgilbert:
How so?  I didn't see any place where they seemed to affect the outcome.

neheb:
It was something like:

calling set() on a config option resulted in

```
#undef X
```

vs.

```
// #define X
```

Don't remember the details.

bgilbert:
At least with the current version of the wrap, the `config.h` from a test build (with some deps found, some not) is identical with and without the disablers, on each of Meson 0.58.0, 1.8.5, and 1.9.1.

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -727,10 +712,7 @@ functions = {
   'linkat': [],
   'listxattr': [],
   'llistxattr': [],
-  'locale_charset': iconv_dep.partial_dependency(
```
eli-schwartz, line 722: This function is actually really silly and technically comes from libunistring instead. And I don't really know that you can blindly assume it exists for all possibl implementations of a subproject wrap that exposes the *icon* API.

eli-schwartz:
In general if it's possible it's better to do version checks against the dependency version (for newly added APIs in the dependency) or via querying some pkgconfig file property (see for example how libcurl exposes its build options).

bgilbert:
Yeah, this problem keeps coming up and I don't know a clean way out of it.  Property checks only work for internal dependencies where the wrap intentionally supports them, and I don't know of any that do (including curl).  Version checks don't work for configured-out APIs, future API removals (libxml2), or semi-interoperable implementations (iconv).  My thinking is: if we assume an internal dependency has a function it doesn't, the build will fail in an obvious way, and since the parent project controls its own wraps it can switch to a more suitable one.

bgilbert:
I've switched to header symbol checks for functions in dependencies.

PR #2512, opened 2025-11-21
Update freeglut to 3.8.0

TheFakeMontyOnTheRun:
Tested by manually copying the meson files into the FreeGLUT source tree.
Also checked the changelog - no new relevant build option (but I'm open to suggestions).
Everything building fine.


bgilbert:
Thanks!

PR #2513, opened 2025-11-21
Add x265 4.1

amyspark:
Hi all,

Coming here from https://gitlab.freedesktop.org/gstreamer/cerbero/-/merge_requests/1962 to upstream the wrap I made for the x265 codec library.

The main objective of this port is to support SDR, HDR10(+) and HDR12 in a single build, instead of doing three separate builds + a libtool/ar merge as is done by us and e.g. Homebrew (https://github.com/Homebrew/homebrew-core/blob/de72d493fe5311696d04cc079405ca478e65848d/Formula/x/x265.rb#L37).

All feedback is appreciated!

cc @nirbheek

### DIFF
--- a/subprojects/packagefiles/x265/meson.build
+++ b/subprojects/packagefiles/x265/meson.build
@@ -1,0 +1,29 @@
+project(
+  'x265',
+  'c',
+  'cpp',
+  version: '4.1',
+  meson_version: '>= 1.3.0',
+  license: 'GPL-2.0-or-later',
+  license_files: 'COPYING',
+  default_options: [
+    'c_std=gnu11,c11',
+    'cpp_std=gnu++14,c++14',
+    'warning_level=2',
+  ],
+)
+
+if get_option('optimization') != '3'
+  message(
+    'This library expects to be built with -Doptimization=3. If you find any visual artifacts, consider adding this flag.',
+  )
+endif
+
+subdir('source')
+
+x265_dep = declare_dependency(
+  link_with: x265_lib,
+  include_directories: include_dirs,
+)
+
+meson.override_dependency('x265', x265_dep)
--- a/subprojects/packagefiles/x265/source/common/meson.build
+++ b/subprojects/packagefiles/x265/source/common/meson.build
@@ -1,0 +1,461 @@
+vflags = ['-DX265_VERSION=@0@'.format(x265_version)]
+
+if enable_assembly.allowed()
+  c_args += ['-DENABLE_ASSEMBLY=1']
+endif
+
+primitives_libs = []
+warndisable = []
+
+if enable_assembly.allowed() and host_arch.startswith('x86')
+  sse3 = files('vec/dct-sse3.cpp')
+  ssse3 = files('vec/dct-ssse3.cpp')
+  sse41 = files('vec/dct-sse41.cpp')
+
+  if cpp.get_argument_syntax() == 'msvc'
+    primitives = sse3 + ssse3 + sse41
+    warndisable = cpp.get_supported_arguments('/wd4100')
+    # x64 implies SSE4, so only add /arch:SSE2 if building for Win32
+    if cc.get_id() == 'msvc'
+      if host_arch == 'x86'
+        primitives_arch_args = {
+          'sse3': ['/arch:SSE2'],
+          'ssse3': ['/arch:SSE2'],
+          'sse41': ['/arch:SSE2'],
+        }
+      else
+        primitives_arch_args = {
+          'sse3': [],
+          'ssse3': [],
+          'sse41': [],
+        }
+      endif
+    else
+      primitives_arch_args = {
+        'sse3': ['-msse3'],
+        'ssse3': ['-mssse3'],
+        'sse41': ['-msse4.1'],
+      }
+    endif
+    foreach bit, depth : bit_depths
+      foreach k, arch_args : primitives_arch_args
+        primitives_args = c_args + get_variable(f'@depth@_args') + vflags + warndisable + arch_args
+        primitives_libs += static_library(
+          'primitives-@0@-@1@'.format(depth, k),
+          get_variable(k),
+          c_args: primitives_args,
+          cpp_args: primitives_args,
+          include_directories: include_dirs,
+          build_by_default: false,
+                  # gnu_symbol_visibility: 'inlineshidden',
+        )
+      endforeach
+    endforeach
+  endif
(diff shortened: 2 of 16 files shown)


subprojects/packagefiles/x265/meson.build
```diff
@@ -0,0 +1,29 @@
...
+  license: 'GPL-2.0-or-later',
+  license_files: 'COPYING',
+  default_options: [
+    'c_std=gnu11,c11',
+    'cpp_std=gnu++14,c++14',
+    'warning_level=2',
+  ],
+)
+
+if get_option('optimization') != '3'
```
nirbheek, line 16: Why not also put this in default_options?

subprojects/packagefiles/x265/source/common/meson.build
```diff
@@ -0,0 +1,461 @@
...
+      primitives_args = c_args + arm_args + get_variable(f'@depth@_args') + vflags + aarch64_neon_dotprod_flag
+      primitives_libs += static_library(
+        f'asm-primitives-neon-dotprod-@depth@',
+        c_srcs_neon_dotprod + arm_asms_neon_dotprod,
+        c_args: primitives_args + get_variable(f'@depth@_asm_args'),
+        cpp_args: primitives_args,
+        include_directories: include_dirs,
+        override_options: aarch64_override_opts,
+        build_by_default: false,
+              # gnu_symbol_visibility: 'inlineshidden',
```
nirbheek, line 288: Why is this commented out and at the wrong indentation, and repeated everywhere?

subprojects/packagefiles/x265/source/meson.build
```diff
@@ -0,0 +1,809 @@
...
+version_data = run_command(
+  version_py,
+  '--git',
+  git,
+  meson.project_source_root(),
+  capture: true,
+  check: true,
+)
+
+message(version_data.stderr().strip())
```
nirbheek: Why are we printing this?

subprojects/packagefiles/x265/source/meson/version.py
```diff
@@ -0,0 +1,98 @@
...
+    # Default Settings, for user to be vigilant about x265 version being reported during product build.
+    x265_version = 'unknown'
+    x265_latest_tag = '0.0'
+    x265_tag_distance = '0'
+    x265_revision_id = ''
+
+    archive_file = Path(args.src_dir) / 'x265Version.txt'
+    git_dir = Path(args.src_dir) / '.git'
+
+    # FIXME: the original cannot parse shallow clones --Amy
```
nirbheek: Why is this a FIXME? What's there to fix exactly?

subprojects/packagefiles/x265/source/test/meson.build
```diff
@@ -0,0 +1,75 @@
...
+  'ipfilterharness.h',
+  'mbdstharness.cpp',
+  'mbdstharness.h',
+  'pixelharness.cpp',
+  'pixelharness.h',
+  'testbench.cpp',
+  'testharness.h',
+)
+
+# FIXME: on CMake checkasm-arm uses NASM_FLAGS,
```
nirbheek: Also this FIXME, sounds like it should just be a normal comment?

subprojects/packagefiles/x265/source/common/meson.build
amyspark: x265, as well as other CMake based dependencies of GStreamer's, leaks all STL-generated inline C++ symbols. This causes duplicate symbol warnings when linking against the resulting static library (and, in the case of GStreamer's, when putting together the static framework for iOS deployment, causing GitLab CI to run out of job log space).

The fix, like I did for librsvg, is to:
- mark them as `inlineshidden` (the line above),
- and then export the right symbols with prelinking.

Meson supports this already, but when I tried this on macOS, the library breaks down because of Meson implicitly assuming ELF/GNU symbol semantics and ignoring the target architecture: https://github.com/mesonbuild/meson/issues/15234

As for the formatting itself, I do not know; earlier CI runs errored out with "missing formatting", and running `tools/format.py` adjusted lines like this.

subprojects/packagefiles/x265/source/test/meson.build
amyspark: I marked these as FIXMEs because they need to be addressed by the upstream maintainers; it's not a bug on our end, but here we have the advantage of being able to case them out.

The particular bug here is that this line mixes NASM and GAS flags thus breaking the build on Arm64.

subprojects/packagefiles/x265/source/meson/version.py
amyspark: As I explained in the other comment, the original script expects a full clone to describe the version tag; otherwise it just errors out because there's no tag at all from which to describe. 

For this wrap I added extra logic to ensure we fall back to the embedded manifest.

subprojects/packagefiles/x265/meson.build
amyspark: It is a forbidden option: https://github.com/mesonbuild/wrapdb/blob/eabe86cc7cc9692fb35bcccd9c778f849c21814e/tools/sanity_checks.py#L698

(However, it's compulsory in order to run tests, so I instead made sure to disable them if the right optimization level isn't set.)

subprojects/packagefiles/x265/source/test/meson.build
nirbheek: Yeah no, we should not put FIXMEs in our code for things that other people have to fix, it's very misleading. Please remove it :)

nirbheek: As in, mark it as `FIX UPSTREAM` or something. Remove "FIXME"

subprojects/packagefiles/x265/meson.build
nirbheek: > it's not clear that wraps should have an opinion about [optimization defaults]

@bgilbert this is not true for many projects, and x265 is one of them. Upstream specifically wants you to only build with -O3, and will probably ignore bug reports for any other optimization level. I think we should allow wraps to override this sanity check.

subprojects/packagefiles/x265/source/meson/version.py
amyspark: Labeled as `FIXME UPSTREAM` as per https://github.com/mesonbuild/wrapdb/pull/2513#discussion_r2655412872

Issue #2511, opened 2025-11-20
Consider dropping support for very old Meson versions


bgilbert:
WrapDB doesn't formally have a minimum supported Meson version for its ports.  Often - when it doesn't undermine the maintainability of the port - a subproject will conditionalize its use of a Meson feature so it can continue to support older versions.  That does impose some maintenance burden, though, and the oldest Meson versions may no longer be worth supporting in this way.

Meson can of course be updated with pip, but Linux distros already ship Meson and it's convenient for users if we support the version they already have.  Some relevant distro versions:

- Debian 11 (bullseye, oldoldstable, community LTS): 0.56.2
- Debian 12 (bookworm, oldstable): 1.0.1
- Ubuntu 22.04 LTS (jammy): 0.61.2
- RHEL 8: 0.58.2
- RHEL 9: 0.63.3

Should we formally drop support for some older versions?  In particular, CI now requires ports to add `meson.override_dependency()` when updated; this requires 0.54+ and we already have 22 `>=0.54` checks in 17 ports.

If so, we should document this, both for contributors and for downstream projects that might theoretically be affected.

neheb:
0.61.2 as a minimum is probably good.

older OS (including Ubuntu 18.04 and 20.04) can update to 0.61.5 using pip.

Even older is probably broken and untested. Debian 9 for example (0.49) breaks on ternary operations. There's no CI check for this.

Seamless wrap usage comes with 0.56 IIRC. Before then, fallback: had to be specified which doesn't really work right (always falls back).

eli-schwartz:
@neheb your goals IMO are consistent with assuming 0.56 as a minimum. Any particular reason why you want to cater to Ubuntu 22.04 LTS users but require Debian oldoldstable and RHEL 8 users to use pip as a generalization?

My personal gut feeling is that I'd tend to look to the oldest of the commonly-used LTS distros when deciding on assumptions. That usually tends to be (and in this case is) Debian. I've said before that I think it's pretty safe at this point to assume nobody still has workflows depending on < 0.56 at this point. If they are, their OS is usually no longer supported for security patches except in a kind of fig leaf sense if even that much.

neheb:
Let’s mandate a minimum of 0.56 then

bgilbert:
Proposed implementation in https://github.com/mesonbuild/wrapdb/pull/2516.

PR #2516, opened 2025-11-22
Impose minimum `meson_version`

bgilbert:
Pursuant to https://github.com/mesonbuild/wrapdb/issues/2511, define Meson 0.56 as the oldest supported version for ports.  Require updated ports to declare a `meson_version` consistent with that minimum.  This makes it clear by inspection which versions a wrap supports, and also makes it more obvious when a wrap modification reduces the supported version range.

Update the CI version report to explicitly list the repo's minimum, and to omit feature dependencies older than that.


bgilbert:
Lifting draft.

neheb (approved):

PR #2527, opened 2025-11-25
Serve `releases.json` from GitHub Pages to fix publishing race

bgilbert:
Meson fetches `releases.json` directly from the repo, via a redirect from `wrapdb.mesonbuild.com`.  When a PR lands, `releases.json` is updated immediately but the GitHub release (with the wrap file, source mirror, and patch zip) is updated after a delay, which can be lengthy if CI is heavily loaded or down.  During this time, `meson wrap install` and `meson wrap update` fail on the newly-updated wraps.

Maintain a second copy of `releases.json` that lags the one in Git, updating it only after all the releases it describes have been published.  Publish it via GitHub Pages: have `create_release.py` generate a static site, then publish that in the release workflow.  Because of how GitHub Pages works with custom domains, the new file will be available from https://mesonbuild.com/wrapdb/releases.json.  Update the `wrapdb.mesonbuild.com` nginx config to redirect there instead.

While we're here, minify the published `releases.json`.

cc @tp-m

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,11 +25,29 @@ jobs:
       - name: Upload release assets
         run: |
           ./tools/create_release.py ${{ github.repository }} ${{ secrets.GITHUB_TOKEN }}
+      - name: Upload releases.json
+        uses: actions/upload-pages-artifact@v4
+
+  update_index:
+    needs: create_release
+    runs-on: ubuntu-slim
+    permissions:
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deploy.outputs.page_url }}releases.json
+    steps:
+      - name: Deploy releases.json
+        id: deploy
+        uses: actions/deploy-pages@v4
 
   # Ideally we should trigger Meson's CI to update the website, but unfortunately
   # it requires a Personal Access Token. Instead clone meson and do it ourself.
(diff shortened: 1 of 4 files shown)


bgilbert:
This requires one repo setting to be changed: the Pages deployment source must be GitHub Actions.

image

Making this change should automatically create a `github-pages` environment that only allows deployment from `master`, which is what we want.

bgilbert:
This PR fixes a user-visible race condition in WrapDB infrastructure, but requires a repository admin to make the configuration change described in https://github.com/mesonbuild/wrapdb/pull/2527#issuecomment-3573653788.  Could an admin, e.g. @eli-schwartz, please take a look?

eli-schwartz:
Done for GitHub Actions. I can't update the nginx config.

bgilbert:
Yup, thanks.  Will merge when I'm at a computer, confirm that it's working, and then the nginx config can be updated afterward.

bgilbert:
[Deployed successfully](https://mesonbuild.com/wrapdb/releases.json).  @tp-m, could you please sync the nginx config when you get a chance?

bgilbert:
@tp-m, ping, please sync the nginx config.

tp-m:
Sorry, missed the pings here. Best to ping me on IRC if you don't get a quick response :)

Should be deployed now.

PR #2528, opened 2025-11-25
Add libtommath 1.3.0

cyanogilvie:
Add libtommath 1.3.0


subprojects/libtommath.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+directory = libtommath-1.3.0
+source_url = https://github.com/libtom/libtommath/releases/download/v1.3.0/ltm-1.3.0.tar.xz
+source_filename = libtommath-1.3.0.tar.xz
```
bgilbert: This should be `ltm-1.3.0.tar.xz` because that's what upstream calls it.  You can avoid the CI error by adding it to `SOURCE_FILENAME_PREFIXES` in `sanity_checks.py`.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+  )
+  benchmark(
+    'benchmarks',
+    timing_exe,
+    timeout: 30,
+  )
+endif
+
+# Documentation target - build PDF manual from LaTeX sources
+if get_option('docs')
```
bgilbert, line 377: WrapDB wraps don't necessarily reimplement all functionality from the upstream build system.  When a library is being used as a wrap, the parent project won't want the docs, and so we normally wouldn't maintain the additional complexity of `build-manual.py` etc.

On the other hand, if your use case requires this (e.g. because you want to add the Meson build config to the upstream repo), it isn't _forbidden_ in WrapDB.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
+project(
+  'libtommath',
+  'c',
+  version: '1.3.0',
+  license: 'CC-PDDC',
```
bgilbert: This is not correct.
```suggestion
  license: 'Unlicense',
```

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+  'timing',
+  'demo/timing.c',
+  link_with: libtommath,
+  c_args: ['-DTIMER'],
+  include_directories: inc,
+  install: false,
+  build_by_default: false,
+)
+
+if get_option('benchmarks')
```
bgilbert, line 339: Likewise, we probably don't need these unless there's a reason to carry them.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+  # Create logs directory in builddir for benchmark output (timings binary segfaults without it)
+  run_command(
+    'mkdir',
```
bgilbert: Unix-specific.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
+project(
+  'libtommath',
+  'c',
+  version: '1.3.0',
+  license: 'CC-PDDC',
+  default_options: [
+    'c_std=c99',
+    'warning_level=3',
+    'werror=false',
+    'default_library=both',
```
bgilbert: Why are these needed?

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+version = meson.project_version()
+version_arr = version.split('.')
+version_major = version_arr[0]
+version_minor = version_arr[1]
+version_patch = version_arr[2]
+
+cc = meson.get_compiler('c')
+
+if cc.get_id() in ['msvc', 'clang-cl']
+  error('MSVC builds are not supported at this time.')
```
bgilbert: Upstream seems to support Windows, so we should too.  If there's a Windows-specific problem you aren't interested in debugging, this should be removed and CI should be allowed to fail.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+    )
+  elif cc.get_id() == 'msvc'
+    add_project_arguments(
+      '/O2',
+      language: 'c',
+    )
+  endif
+endif
+
+if cc.get_id() in ['msvc', 'clang-cl']
```
bgilbert, line 27:
```suggestion
if host_machine.system() == 'windows'
```

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+
+# Build both static and shared libraries by default
+# Users can override with -Ddefault_library=static or -Ddefault_library=shared
+libtommath = library(
+  'tommath',
+  sources,
+  install: true,
+  include_directories: inc,
+  version: version,
+  soversion: version_major,
```
bgilbert: This is already the default when `version` is set.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+
+# Make the dependency available to parent projects
+meson.override_dependency(meson.project_name(), libtommath_dep)
+
+tommath_test_exe = executable(
+  'tommath_test',
+  ['demo/test.c', 'demo/shared.c'],
+  link_with: libtommath,
+  include_directories: inc,
+  install: false,
```
bgilbert: This is the default value.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+# This makes the library usable as a Meson subproject
+libtommath_dep = declare_dependency(
+  link_with: libtommath,
+  include_directories: inc,
+)
+
+# Make the dependency available to parent projects
+meson.override_dependency(meson.project_name(), libtommath_dep)
+
+tommath_test_exe = executable(
```
bgilbert: For clarity (and I think not really for any other reason), these could be moved inside `if get_option('tests')`.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+  'bn_s_mp_toom_mul.c',
+  'bn_s_mp_toom_sqr.c',
+)
+
+inc = include_directories('.')
+
+# Add extra optimization flags to match Makefile and CMake builds
+# Note: Meson already adds -O3 when optimization=3 (from buildtype=release)
+# We only need to add the flags Meson doesn't add by default
+if get_option('buildtype') != 'debug'
```
bgilbert: This will not work properly on all Meson versions.
```suggestion
if not get_option('debug')
```

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,380 @@
...
+cc = meson.get_compiler('c')
+
+if cc.get_id() in ['msvc', 'clang-cl']
+  error('MSVC builds are not supported at this time.')
+endif
+
+# Override optimization for debug builds to use -Og instead of -O0 to work
+# around an issue with clang not optimizing away references to some
+# conditionally defined functions with forward references
+if get_option('buildtype') == 'debug' and get_option('optimization') == '0'
```
bgilbert:
```suggestion
if get_option('debug') and get_option('optimization') == '0'
```

bgilbert (commented):
Thanks for the PR!

subprojects/packagefiles/libtommath/meson.build
cyanogilvie: I spent a few hours trying to get it to build with MSVC and clang-cl but I couldn't find any approaches that seemed to be acceptable (it does build and work with MSYS2). The core of the issue is that the upstream source uses a trick to avoid `#ifdef`ing platform-specific RNG interfaces, relying on the compiler to optimize away unreachable code.  MSVC (and clang at `-O0`) don't do that and so the link fails with undefined symbols.

Approaches I tried:
- Turning on minimal optimization on `debug`, `optimization=0` builds: `-Og` for clang (works), `/O2` for MSVC (doesn't).
- Applying a diff that adds the `#ifdef` guards around the undefined references, but I couldn't see any other wraps that use that facility so I backed it out.

The diff to do that is:
```
--- a/bn_s_mp_rand_platform.c
+++ b/bn_s_mp_rand_platform.c
@@ -128,20 +128,32 @@ static mp_err s_read_ltm_rng(void *p, size_t n)
 }
 #endif
 
+#if 0
 mp_err s_read_arc4random(void *p, size_t n);
 mp_err s_read_wincsp(void *p, size_t n);
 mp_err s_read_getrandom(void *p, size_t n);
 mp_err s_read_urandom(void *p, size_t n);
 mp_err s_read_ltm_rng(void *p, size_t n);
+#endif
 
 mp_err s_mp_rand_platform(void *p, size_t n)
 {
    mp_err err = MP_ERR;
+#ifdef BN_S_READ_ARC4RANDOM_C
    if ((err != MP_OKAY) && MP_HAS(S_READ_ARC4RANDOM)) err = s_read_arc4random(p, n);
+#endif
+#ifdef WIN32_LEAN_AND_MEAN
    if ((err != MP_OKAY) && MP_HAS(S_READ_WINCSP))     err = s_read_wincsp(p, n);
+#endif
+#ifdef BN_S_READ_GETRANDOM_C
    if ((err != MP_OKAY) && MP_HAS(S_READ_GETRANDOM))  err = s_read_getrandom(p, n);
+#endif
+#ifdef BN_S_READ_URANDOM_C
    if ((err != MP_OKAY) && MP_HAS(S_READ_URANDOM))    err = s_read_urandom(p, n);
+#endif
+#ifdef BN_S_READ_LTM_RNG
    if ((err != MP_OKAY) && MP_HAS(S_READ_LTM_RNG))    err = s_read_ltm_rng(p, n);
+#endif
    return err;
 }
```
Upstream's stance is given in a comment added in a later (`develop`) version of that file:
```c
/*
 * Note: libtommath relies on dead code elimination
 * for the configuration system, i.e., the MP_HAS macro.
 *
 * If you observe linking errors in this functions,
 * your compiler does not perform the dead code compilation
 * such that the unused functions are still referenced.
 *
 * This happens for example for MSVC if the /Od compilation
 * option is given. The option /Od instructs MSVC to
 * not perform any "optimizations", not even removal of
 * dead code wrapped in `if (0)` blocks.
 *
 * If you still insist on compiling with /Od, simply
 * comment out the lines which result in linking errors.
 *
 * We intentionally don't fix this issue in order
 * to have a single point of failure for misconfigured compilers.
 */
```
I wasn't able to find a combination of MSVC options to get it to work, but I don't have direct access to that platform and using the CI sanity check as an oracle is slow and feels wasteful.

The only other way I could think of to work around this is to add stubbed implementations of the missing symbols in an overlayed `.c` file (a new one, not one that exists in the upstream source), but that smelled bad,

Thanks for the review. I'm new to the PR workflow this project uses, should I continue to edit and force-push my local branch with the fixes from this review, or commit them directly from the review as new commits (read somewhere that it's preferred to merge PRs as single commits)?

subprojects/packagefiles/libtommath/meson.build
cyanogilvie: My preference would be to contribute the meson build system upstream, but that's quite challenging - upstream is historically reluctant to create new releases, and isn't planning on any new releases in the v1.x series AFAIK (and there are 6+ year old PRs blocked waiting on v2).

Their own README notes that "Building is straightforward for GNU Linux only", so a complete build system that Just Works for any platform would be nice to have. The wrap isn't really the right place for it, but I don't really have a better one.

subprojects/packagefiles/libtommath/meson.build
cyanogilvie: Since good performance is such a core deliverable of this library there may be an argument for exposing this for consumers of the subproject (although that would also imply implementing the things like the per-architecture Toom-Cook cut-off tuning that `make tune` does).

I used it to investigate whether things like LTO and PGO where worth it for my target platforms, and I could see consumers of the wrap using the benchmark to make their own decision on that.

subprojects/packagefiles/libtommath/meson.build
cyanogilvie: Would the idiomatic approach be a python script?
```meson
run_command(
  python,
  '-c',
  'import os, sys; os.makedirs(sys.argv[1]) if not os.path.exists(sys.argv[1]) else None'
  meson.project_build_root() / 'logs',
  check: true,
)
```

subprojects/packagefiles/libtommath/meson.build
cyanogilvie: I'll drop them. I started out with `warning_level=3`, `werror=true` aspirationally but had to back off that to get the CI to pass on all the platforms - the upstream source uses `#warning` which is c23 (could pacify with `-Wno-pedantic`) and some instances of unitary `-` applied to unsigned types.

cyanogilvie:
After extensive hacking the build and tests succeed on Windows MSVC and clang-cl, so I've enabled them in the CI

subprojects/packagefiles/libtommath/meson.build
bgilbert: Yes, either inline or in a separate script file.

subprojects/packagefiles/libtommath/meson.build
bgilbert: Feels like upstream is relying on an implementation detail here.  Awesome.

> Applying a diff that adds the `#ifdef` guards around the undefined references, but I couldn't see any other wraps that use that facility so I backed it out.

Yeah, we don't allow patches to upstream sources.  I think it's okay to use CI as an oracle if you want (you could disable the other jobs when doing so).

Force-pushes are preferred here; thanks for double-checking.  I know different projects have different opinions about this.

ci_config.json
```diff
@@ -778,6 +778,13 @@
       "windows": false
     }
   },
+  "libtommath": {
+    "_comment": "MSVC build fails because it doesn't optimize away unreachable code leaving undefined symbols.",
```
bgilbert: Stale.

tools/sanity_checks.py
```diff
@@ -97,6 +97,10 @@
     'libtiff': {
         'cmakedefine.py',
     },
+    'libtommath': {
+        'build-manual.py',
+        'libtommath.def'
```
bgilbert: `.def` files are automatically allowed, so this is redundant.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,386 @@
+project(
+  'libtommath',
+  'c',
+  version: '1.3.0',
+  license: 'Unlicense',
+  default_options: [
+    'c_std=c99',
+    'default_library=both',
```
bgilbert: This seems surprising and inconsistent as a default.  If it's needed because tests require the static library, it seems better to build the library with `both_libraries()` (or possibly conditionalize on the `tests` option using `build_target()`) instead.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,386 @@
...
+
+# Build both static and shared libraries by default
+# Users can override with -Ddefault_library=static or -Ddefault_library=shared
+libtommath = library(
+  'tommath',
+  sources,
+  install: true,
+  include_directories: inc,
+  version: meson.project_version(),
+  link_args: cc.get_id() in ['msvc', 'clang-cl'] ? [
```
bgilbert, line 235: Mild preference for building these up beforehand, which is idiomatic, rather than using ternaries here.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,386 @@
...
+
+# Create a dependency object for use by other projects
+# This makes the library usable as a Meson subproject
+libtommath_dep = declare_dependency(
+  link_with: libtommath,
+  include_directories: inc,
+)
+
+# Make the dependency available to parent projects
+meson.override_dependency(meson.project_name(), libtommath_dep)
```
bgilbert: The pkg-config name and the wrap name are in distinct namespaces and there's no requirement that they match.  So it'd be more correct to use a string literal:

```suggestion
meson.override_dependency('libtommath', libtommath_dep)
```

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,386 @@
...
+
+# Build and run unit tests if requested
+if get_option('tests')
+  tommath_test_exe = executable(
```
bgilbert, line 286: Nit: there's some value to `build_by_default: false` for these (it works correctly on newer Meson at least) because it's possible to build with tests enabled but never actually run tests.

subprojects/packagefiles/libtommath/meson.build
```diff
@@ -0,0 +1,386 @@
...
+    ] : [],
+  )
+
+  test('tommath_test', tommath_test_exe)
+
+  # Differential test: mtest generates test cases, mtest_opponent validates them
+  # Usage: mtest <iterations> | mtest_opponent
+  test(
+    'mtest_opponent',
+    find_program('sh'),
```
bgilbert: Happens to work on the Windows CI runners, but will not generally work on Windows.  A Python wrapper would be better.

subprojects/packagefiles/libtommath/doc/build-manual.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Replace bn.tex with deterministic version
+    shutil.move('bn-deterministic.tex', 'bn.tex')
+
+    # Restore original timestamp
+    shutil.copystat('bn.bak', 'bn.tex')
+
+    # Build the manual
+    with open('bn.ind', 'w') as f:
+        f.write('hello\n')
```
bgilbert, line 73: What's happening here?

subprojects/packagefiles/libtommath/doc/build-manual.py
```diff
@@ -0,0 +1,110 @@
...
+
+    subprocess.run([latex, 'bn'], stdout=subprocess.DEVNULL, check=True)
+    subprocess.run([latex, 'bn'], stdout=subprocess.DEVNULL, check=True)
```
bgilbert, line 76: Is this intentionally doubled?

subprojects/packagefiles/libtommath/doc/build-manual.py
```diff
@@ -0,0 +1,110 @@
...
+    output_file = os.path.abspath(output_file)
+
+    # Change to output directory
+    os.chdir(outdir)
+
+    # Copy input to bn.tex
+    shutil.copy2(input_file, 'bn.tex')
+
+    # Create backup with same timestamp
+    shutil.copy2('bn.tex', 'bn.bak')
```
bgilbert, line 43: Seems like there's a lot of unnecessary code to move files around.  Why not just write the deterministic header directly into `bn.tex` (or into `tempfile.NamedTemporaryFile()`), copy `input_file` into it afterward, and set its timestamp from `input_file`?

subprojects/packagefiles/libtommath/doc/build-manual.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Cleanup
+    shutil.move('bn.bak', 'bn.tex')
+
+    cleanup_files = [
+        'bn.aux', 'bn.dvi', 'bn.log', 'bn.idx',
+        'bn.lof', 'bn.out', 'bn.toc', 'bn.ilg',
+        'bn.ind', 'bn.tex'
+    ]
+    for f in cleanup_files:
```
bgilbert, line 102: These are all written to the build directory.  Is it important to clean them up?

subprojects/packagefiles/libtommath/meson.build
bgilbert: ...in which case, you should probably also do something like this:

https://github.com/mesonbuild/wrapdb/blob/22e22114e021cf8be0c98ed52c5a5bd12ac4db95/subprojects/packagefiles/libarchive/libarchive/meson.build#L176-L181

cyanogilvie: Can we go with the alternative: conditionalize building and running the tests and set `default_library=both` in the CI? Hard-coding `both_libraries` internally and then re-implementing the intent of the `default_library` standard option seems brittle (the above code would defeat a consumer's intent if they explicitly specify `default_library=both` - handing them a `lib` rather than the `both_libs` they would expect).  This way the CI works and runs the tests on all platforms and the consumer gets what they expect with the least surprising internal implementation.

```meson
  libtommath_for_tests = libtommath
  if cc.get_id() in ['msvc', 'clang-cl']
    if get_option('default_library') == 'shared'
      error('Static library build is required to run tests with MSVC/clang-cl')
    endif
    libtommath_for_tests = libtommath.get_static_lib()
  endif
```
I would have liked to use `libtommath.get_static_lib().found()` to more directly test the relevant requirement, but it seems that `libraries()` doesn't return `both_libs` if only one is selected (and so the `get_*_library()` methods are missing, or only the corresponding method exists).

subprojects/packagefiles/libtommath/meson.build
cyanogilvie: I expressed these as a ternaries to make reasoning about the scope of their effect easier - like that it cannot affect anything other than the `link_args` of this target, whereas if it is coupled through global state by constructing it as a conditionally defined variable earlier it would require searching for other uses of that variable to understand its scope, but I'll rework it through a variable if the ternary clashes with the idiomatic style of this project?

subprojects/packagefiles/libtommath/meson.build
cyanogilvie: Replaced with a Python wrapper

subprojects/packagefiles/libtommath/doc/build-manual.py
cyanogilvie: Re this and the other review comments on this script: this is a python port of upstream's doc build rules:
```makefile
docs: manual

#LTM user manual
mandvi: bn.tex
    cp bn.tex bn.bak
    touch --reference=bn.tex bn.bak
    (printf "%s" "\def\fixedpdfdate{"; date +'D:%Y%m%d%H%M%S%:z' -d @$$(stat --format=%Y bn.tex) | sed "s/:\([0-9][0-9]\)$$/'\1'}/g") > bn-deterministic.tex
    printf "%s\n" "\pdfinfo{" >> bn-deterministic.tex
    printf "%s\n" "  /CreationDate (\fixedpdfdate)" >> bn-deterministic.tex
    printf "%s\n}\n" "  /ModDate (\fixedpdfdate)" >> bn-deterministic.tex
    cat bn.tex >> bn-deterministic.tex
    mv bn-deterministic.tex bn.tex
    touch --reference=bn.bak bn.tex
    echo "hello" > bn.ind
    latex bn ${silent_stdout}
    latex bn ${silent_stdout}
    makeindex bn
    latex bn ${silent_stdout}

#LTM user manual [pdf]
manual: mandvi
    pdflatex bn >/dev/null
    sed -b -i 's,^/ID \[.*\]$$,/ID [<0> <0>],g' bn.pdf
    mv bn.bak bn.tex
    rm -f bn.aux bn.dvi bn.log bn.idx bn.lof bn.out bn.toc
```
Which seem to contain a lot of voodoo in pursuit of deterministic builds, but I honestly don't understand the intent of each bit well enough to reason about which side effects are being relied upon (and the commit history didn't shed any light on it), so I instead aimed to replicate the process and side-effects as exactly as possible.

My python is pretty rusty (I last did any substantial work in python over 5 years ago, and even then not much), so I used AI for the initial port and then checked it as best I can, which probably hasn't helped matters.

For my use case I don't care about deterministic builds (especially for an optional documentation target). That seems like a core focus of other build systems like bazel, and I've seen it mentioned for meson but I don't know what the official stance is towards deterministic builds here. It's curious that libtommath seemed to work so hard on making this output deterministic 23 years go, long before I'd heard anything about deterministic builds.

subprojects/packagefiles/libtommath/doc/build-manual.py
cyanogilvie: I have no idea. It dates back to a commit in 2003 with the message `added libtommath-0.17` which I take to mean that this libtommath repo was extracted from some other project that imported libtommath from elsewhere (libtomcrypt?), or that back then libtommath development didn't use VC, so wasn't able to trace the intent.

subprojects/packagefiles/libtommath/doc/build-manual.py
cyanogilvie: Only insofar as it replicates what upstream's makefile does. My best guess is that there is some build stability hysteresis issue that produces different results if the output artifacts exist already

subprojects/packagefiles/libtommath/doc/build-manual.py
cyanogilvie: I would certainly leave that to the normal build dir handling, but since upstream's rules explicitly cleaned up these files separately from its `clean` target (which includes a superset of them), I interpreted it as meaning that their presence during subsequent builds would affect the determinism so I cargo culted it.

PR #2534, opened 2025-12-01
google-benchmark: do not build tests by default

mochaaP:
This reduces compilation time in downstream projects.


bgilbert:
Meson [older than 1.7](https://mesonbuild.com/Release-notes-for-1-7-0.html#test-targets-no-longer-built-by-default) doesn't respect `build_by_default: false` on tests, so we shouldn't deprecate `-Dtests=disabled` yet.

eli-schwartz:
I know -- and I qualify that as a simple bugfix in meson. For something that doesn't actually affect the build at all, just makes it a bit *faster*.

I question how important it is to keep hackish manual options around, that require the user to fiddle with the settings, just for a simple build time speed issue. It seems like a textbook case for "upgrade if you want it to be faster, use an old version if it's more important that it does build at all on some ancient deployment system".

bgilbert:
We should absolutely be tagging test executables as `build_by_default: false`, and I've started doing that.  But we shouldn't start logging complaints when parent projects want to skip building subproject tests on Meson versions more than _ten months old_.

Downstream projects need to run CI on the Meson versions they actually support, and even Ubuntu 24.04 only has Meson 1.3.2.  If we're serious about getting projects to use `--fatal-meson-warnings` in CI, we need to stop adding warnings for problems they can't actually fix.

PR #2536, opened 2025-12-01
stc: update to 6.0_rc3

mochaaP:


bgilbert:
We normally don't ship release candidates here.  In this case, there isn't even a tag to point to, and the wrap's commit hash points to a commit seven commits after RC3.  Arguably that's better than the current situation, where the wrap points to a commit that isn't even on the main branch, but I'm inclined to just wait for the 6.0 tag.

PR #2539, opened 2025-12-02
sanity_checks: ports cannot have upstream `meson.build`

bgilbert:
When a port is updated, check its upstream source archive for a `meson.build` file and fail if present.  WrapDB wraps should not maintain a downstream Meson config that overrides an upstream one.

As a special exception, allow overriding the config in the first upstream release added to WrapDB.  This has been allowed in the past and is a pragmatic way to add new wraps while fixes propagate to slow-moving upstreams.

Switch `cmocka` to its upstream Meson config.  There are other wraps that override upstream but they're all waiting on their next releases for fixes.

### DIFF
--- a/tools/sanity_checks.py
+++ b/tools/sanity_checks.py
@@ -27,7 +27,9 @@ import platform
 import io
 import sys
 import shutil
+import tarfile
 import textwrap
+import zipfile
 
 from pathlib import Path
 from utils import CIConfig, ProjectCIConfig, Releases, Version, ci_group, is_ci, is_alpinelike, is_debianlike, is_macos, is_windows, is_msys, read_wrap, FormattingError, format_meson, format_wrap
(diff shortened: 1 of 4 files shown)


bgilbert:
Upstream cmocka has build regressions.  Deferring that change for now.

tools/sanity_checks.py
```diff
@@ -377,6 +379,12 @@ def test_releases(self) -> None:
                                 self.check_meson_version(name, ver, patch_path)
                         if patch_path:
                             self.check_project_args(name, config)
+                            # allow a port to overwrite the upstream
+                            # Meson config when the wrap is first added to
+                            # WrapDB, as a development aid, but forbid it
+                            # on subsequent releases
```
eli-schwartz: What happens if:
- upstream doesn't have meson
- wrap is added
- upstream adds independently created meson, but it has bugs
- new release, added to wrapdb and CI fails, after investigation the bugs are fixed and PR'ed upstream

What should this sanity check do?

bgilbert: Yeah.  We could support a versioned override in `ci_config.json`; will take a look.

bgilbert:
Switched to a versioned override mechanism in `ci_config.json` instead of special-casing the first upstream version.

neheb (approved):

PR #2541, opened 2025-12-03
efsw-1.5.0: add wrap file

IogaMaster:


IogaMaster:
Im not making a meson build for this when cmake works just fine

PR #2542, opened 2025-12-04
arduinocore-avr: update from 1.8.2 to 1.8.6

cjmayo:
Releases are listed in:
https://downloads.arduino.cc/packages/package_index.json

Repository:
https://github.com/arduino/ArduinoCore-avr/


bgilbert:
For the provides, see docs [here](https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#overriding-dependencies-in-the-submitted-project).

cjmayo:
Thanks for the pointer. These dependency_names are usable, but they would never match system names because the compiled libraries are specific to the board variant they were compiled for.

Initial checks now pass - won't let me break the long dependency_names line (that seems known as utils.format_wrap() says "Does not understand continuation lines").

The compile checks are failing " subprojects/avr-1.8.6/meson.build:15:0: ERROR: Assert failed: Arduino projects can only be built in a cross build environment.".
I can see `debian_packages` for ci_config.json but it needs to be able to pass `--cross-file` to meson setup, not clear how to do that.

bgilbert:
> These dependency_names are usable, but they would never match system names because the compiled libraries are specific to the board variant they were compiled for.

That's fine.  It's still useful for the parent project to be able to write `dependency('arduinocore')` instead of `dependency('arduinocore', fallback: ['arduinocore-avr', 'arduinocore_dep'])`.

The cross-build requirement means we won't be able to do test builds in CI.  We have mechanisms to force CI to green if a platform or an architecture is unsupported, but this case isn't quite either of those, so I think we should probably just leave it red and accept this as a special case.

bgilbert (approved):

bgilbert:
Thanks!

PR #2544, opened 2025-12-05
README: dependency_names and program_names are no longer optional

cjmayo:


bgilbert:
They're optional in the sense that they only need to be there if the wrap provides libraries or programs, respectively.  Maybe we could say ```It must match information from the wrap's `[provide]` section, if any.```?

cjmayo:
Yes, I wrote that a bit quick. Another idea "may be omitted".

Documentation is the hardest part, I won't keep coming back on this. Just saying it wasn't immediately obvious to me - adding some error messages #2553 might even have been more valuable.

bgilbert:
Thanks!

PR #2545, opened 2025-12-06
libarchive: update from 3.8.3 to 3.8.4

bgilbert:


bgilbert:
MSVC timeouts and Alpine locale issues, as expected.

PR #2546, opened 2025-12-06
icu: update from 77.1 to 78.1

bgilbert:


bgilbert:
clang-cl failure is a [regression](https://groups.google.com/a/unicode.org/g/icu-support/c/61hEq3foH7o/m/kjkL8hs8BgAJ).

neheb:
wonder why MSVC even allows exporting deleted functions...

PR #2547, opened 2025-12-06
libxslt: update from 1.1.43 to 1.1.45

bgilbert:


neheb:
might be a bit pedantic but...

incdir includes config.h. Not sure if that should be exposed as the dependency.

neheb:
Hrm looking at this it needs include-workaround-meson

bgilbert:
Actually there are several problems with the headers, thanks for the pointer.  Fixing.

neheb (approved):

PR #1535, opened 2024-05-27
Added an option for range-v3 to build tests

adembudak:
Added range_v3_tests build option and set it off by default.

PR #2625, opened 2026-02-05
range-v3: add option to disable tests

bryceberger:
Previously #1535, not given an explicit reason for closing.

Use case is the same: I would like to not be _forced_ to compile 240 tests when including `range-v3`.

Took the specific syntax / style from the wrap for `pegtl`.


neheb:
Need new entry in releases.json.

bryceberger:
CI failure doesn't appear to be my fault.

Went ahead and fixed the formatting / applied suggestions from the script, but I'm unsure what to do about the license field. It seems to have been like that since the wrap was added, and contains (multiple?) non-spdx licenses.

bryceberger:
> Need new entry in releases.json.

There's one in the first commit.

neheb:
isn't the license supposed to be BSL-1.0 ?

subprojects/packagefiles/range-v3/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option(
+    'tests',
+    type: 'feature',
+    value: 'auto',
```
neheb: Not needed. This is default.

bryceberger:
According to https://github.com/ericniebler/range-v3/blob/ca1388fb9da8e69314dda222dc7b139ca84e092f/LICENSE.txt, it's BSL-1.0 and `lic++` and `Stepanov and McJones, "Elements of Programming"` and `SGI STL`

The section for `libc++` links to the old svn host. Currently, libc++ is licensed under Apache with LLVM exception. Edit: the old `libc++` license appears to be `NCSA and MIT`, according to https://github.com/llvm/llvm-project/commit/469bdefd448b76c5adcdd67256e9a44fabf7e027

Unsure about the other two licenses. Neither of them are on the [spdx list](https://spdx.org/licenses/)

neheb:
try replacing

```
,
```

with

```
 AND 
```

bryceberger:
Just replacing the commas isn't enough, since the parser (previously) rejected any unknown license text.

I can pull the sanity_checks change into a separate PR if needed, but I believe it (or similar) is required to have this wrap file pass CI.

bryceberger:
Most recent CI failure appears to be ephemeral, but I don't (think that I) have permission to restart it.

```

ERROR: test_releases (__main__.TestReleases.test_releases) (step='check_new_release', name='range-v3')
  File "/home/runner/work/wrapdb/wrapdb/./tools/sanity_checks.py", line 393, in test_releases
  File "/home/runner/work/wrapdb/wrapdb/./tools/sanity_checks.py", line 537, in check_new_release
    res, logs = do_setup(builddir, options, meson_env)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/wrapdb/wrapdb/./tools/sanity_checks.py", line 532, in do_setup
    logs = log_file.read_text(encoding='utf-8')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/pathlib.py", line 1027, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/pathlib.py", line 1013, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '_build/meson-logs/meson-log.txt'
```

Edit: looks like it was my fault, I had `range-v3:cpp_std:c++17` in `ci_config.json` instead of `range-v3:cpp_std=c++17`

bgilbert (changes requested):
If the SPDX expression is invalid, we shouldn't hide problems in warnings no one will see; that defeats the purpose of the check.  SPDX seems to assume (without explicitly specifying, argh) that license IDs contain no spaces, so this wrap's license expression is improper in any case.

As a practical matter, we may want to allow `LicenseRef-<string>` custom licenses.  I've already encountered a wrap where I had to add an additional license as a comment in the `project()` call, which is unhelpful.  `license_expression` rejects these because they don't unambiguously identify a license (https://github.com/aboutcode-org/license-expression/issues/109) but that ambiguity seems better than outright omission.

PR #2637, opened 2026-02-08
aws-c-cal: update from 0.9.10 to 0.9.13

apache-hb:
* Updates aws-c-cal from 0.9.10 to 0.9.13
* Adds emscripten support, its not in CI but I've successfully used it in a wasm project
* Adds support for the bring your own crypto option, which disables the default crypto provider and lets uses provide their own dynamically at runtime


stephanlachnit (approved):

subprojects/packagefiles/aws-c-cal/meson_options.txt
```diff
@@ -6,7 +6,13 @@ option(
...
-    type: 'boolean',
-    value: false,
+    type: 'feature',
     description: 'Experimental feature to support ED25519 keygen on platforms that do not support it in os libs (i.e. win/mac)',
 )
+
+option(
+    'byo_crypto',
+    type: 'feature',
+    value: 'disabled',
```
bgilbert, line 16: Personally I'm okay with this, given how it's used, but it is surprising that a feature doesn't default to `auto`.

subprojects/packagefiles/aws-c-cal/meson.build
```diff
@@ -11,34 +11,61 @@ fs = import('fs')
...
+  not may_have_openssl,
+)
+
 c_args = ['-DAWS_CAL_EXPORTS=1']
 
+public_c_args = []
+if (
+  get_option('default_library') == 'shared'
+  and host_machine.system() == 'windows'
+) or host_machine.system() != 'windows'
```
bgilbert, line 32:
```suggestion
if host_machine.system() != 'windows' or get_option('default_library') == 'shared'
```

subprojects/packagefiles/aws-c-cal/meson.build
```diff
@@ -11,34 +11,61 @@ fs = import('fs')
...
   'aarch64',
 ]
-  error('unsupported architecture: only x86_64 Windows is supported')
+  error('unsupported architecture: only x86_64 is supported on Windows')
 endif
 
 aws_c_common_dep = dependency('aws-c-common')
 libcrypto_dep = dependency(
-  'libcrypto',
+  'openssl',
```
bgilbert, line 45: Is this change correct?  openssl implies both libcrypto and libssl, the actual TLS library.

subprojects/packagefiles/aws-c-cal/meson.build
```diff
@@ -11,34 +11,61 @@ fs = import('fs')
...
   version: '>=1.1',
+  required: not byo_crypto.allowed(),
 )
+
 ncrypt = cc.find_library(
   'ncrypt',
-  required: host_machine.system() == 'windows',
+  required: false,
```
bgilbert, line 52: This, and the corresponding `required: false` for `foundation`, could theoretically produce a situation where `byo_crypto` is disabled and we also don't find the needed crypto library.

subprojects/packagefiles/aws-c-cal/meson.build
```diff
@@ -11,34 +11,61 @@ fs = import('fs')
...
-  error('unsupported architecture: only x86_64 Windows is supported')
+  error('unsupported architecture: only x86_64 is supported on Windows')
 endif
 
 aws_c_common_dep = dependency('aws-c-common')
 libcrypto_dep = dependency(
-  'libcrypto',
+  'openssl',
   version: '>=1.1',
+  required: not byo_crypto.allowed(),
```
bgilbert, line 47: This requires (and, further down the file, enables) OpenSSL on Windows and Darwin, contrary to the earlier comment.

subprojects/packagefiles/aws-c-cal/meson.build
```diff
@@ -52,45 +79,49 @@ src = files(
...
-  src += files(
-    'source/unix/openssl_aes.c',
-    'source/unix/openssl_platform_init.c',
-    'source/unix/openssl_rsa.c',
-    'source/unix/opensslcrypto_ecc.c',
-    'source/unix/opensslcrypto_hash.c',
-    'source/unix/opensslcrypto_hmac.c',
-  )
-else
-  error('Unsupported platform: ' + host_machine.system())
```
bgilbert, line 93: Shouldn't we still error out in this case?

bgilbert (commented):
I'm not 100% certain I've understood the intended behavior of the new logic.  I've flagged some inconsistencies but please double-check the result to make sure it properly handles all the cases.

PR #2638, opened 2026-02-08
aws-c-common: update from 0.12.4 to 0.12.6

apache-hb:
* Updates aws-c-common from 0.12.4 to 0.12.6
* Fixes a link error on AL2 and AL2023 due to not linking `-ldl`
* Adds emscripten support
* Fixes a few of the feature tests which I copy pasted wrong when making the wrap the first time


bgilbert:
@stephanlachnit FYI, we usually use rebase merges in WrapDB, and never merge commits.

@eli-schwartz, to avoid this mistake in future, it might be a good idea to turn off merge commits in the repo settings.

stephanlachnit:
Ah sorry for this - I agree rebasing is much nicer and thus have merge commits also disabled in all most of my GitHub repos, but I did not properly check before I merged this - sorry.

PR #2661, opened 2026-02-19
bzip2: make more robust to downstream warning flags

SoapGentoo:
* disable tests when run as subproject
* require GNU C, since downstream consumers maybe just request ISO C
* silence more pedantic warnings like `-Wunused-parameter` and `-Wimplicit-fallthrough`


subprojects/packagefiles/bzip2/meson.build
```diff
@@ -3,6 +3,8 @@ project(
   'c',
   version: '1.0.8',
   meson_version: '>=1.0.0',
+  default_options: ['c_std=gnu11,c11', 'warning_level=0'],
```
bgilbert:
```meson
meson.version().version_compare('>=1.3.0') ? 'c_std=gnu11,c11' : 'c_std=gnu11'
```

to avoid a dependency on 1.3.0+.

subprojects/packagefiles/bzip2/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option(
+    'tests',
+    type: 'feature',
+    value: 'auto',
```
bgilbert: Redundant.

subprojects/packagefiles/bzip2/meson.build
```diff
@@ -18,13 +20,18 @@ libbz2_lib = static_library(
...
+
 bzip2_exe = executable(
   'bzip2',
   'bzip2.c',
+  c_args: c_args,
   link_with: libbz2_lib,
 )
 
-if meson.can_run_host_binaries()
+if get_option('tests').disable_auto_if(meson.is_subproject()).allowed() and meson.can_run_host_binaries()
```
bgilbert: We've generally been discouraging this kind of automatic disablement.  The parent project can use the option if it wants.

SoapGentoo: I still don't see a realistic use case where downstream consumers want to run bzip2 tests **by default**

bgilbert: If I'm building a compression library I certainly want to make sure it isn't generating invalid output because of some compiler flag or library problem.

SoapGentoo:
@bgilbert updated

bgilbert (approved):

subprojects/packagefiles/bzip2/meson_options.txt
```diff
@@ -0,0 +1,5 @@
+option(
+    'tests',
+    type: 'feature',
+    description: 'build and run tests',
```
eli-schwartz, line 4: In general I'm not sure there's a huge advantage to this. If you want to disable tests in subprojects, it's probably a better idea to pass --suite as an option to `meson test`, which allows you to guarantee only the parent project tests run.

At any rate, this is an incorrect implementation. Using a feature option implies a tristate possibility, usually because auto means automagic dependency detection (very useful for random developers and users building from a git clone) whereas enabled / disabled is good for e.g. distro packaging.

But in this case, there's no dependency to detect and you're simply treating auto as equivalent to enabled. That means that choosing between those two possible values is a false choice that does nothing. Why not use a boolean option here instead?

eli-schwartz (changes requested):

subprojects/packagefiles/bzip2/meson_options.txt
SoapGentoo: made it a boolean

PR #2689, opened 2026-03-09
microsoft-gsl: Update to 2.4.1

mochaaP:
- Added pkgconfig file generation
- Install headers
- Use `meson.override_dependency('msft_gsl')` !! Breaking `dependency('microsoft_gsl')`, but codesearch shows most existing users use the subproject variable instead of the dependency

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -3073,9 +3073,10 @@
   },
   "microsoft-gsl": {
     "dependency_names": [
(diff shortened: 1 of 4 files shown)


mochaaP:
or rename the dependency to msft_gsl, since that's what wil and proxy does

PR #2705, opened 2026-03-21
xsimd: add 14.2.0

Alvaro-Kothe:


subprojects/packagefiles/xsimd/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option(
+    'xtl_complex',
+    type: 'boolean',
```
bgilbert: Should probably be a `feature` variable, and then `meson.build` should autodetect the library by default.

subprojects/packagefiles/xsimd/meson.build
```diff
@@ -0,0 +1,28 @@
...
+
+xsimd_inc = include_directories('include')
+
+deps = []
+if get_option('xtl_complex')
+  deps += dependency(
+    'xtl',
+    version: '^0.7.0',
+  )
+  add_project_arguments(
```
bgilbert: This won't do anything because the wrap itself compiles no code.  The define should be added to `compile_args` in `declare_dependency()` instead.

subprojects/packagefiles/xsimd/meson.build
```diff
@@ -0,0 +1,28 @@
...
+  version: '14.1.0',
+)
+
+xsimd_inc = include_directories('include')
+
+deps = []
+if get_option('xtl_complex')
+  deps += dependency(
+    'xtl',
+    version: '^0.7.0',
```
bgilbert: Meson will treat this as a requirement for exactly 0.7.0, which is probably not what you intended.

subprojects/packagefiles/xsimd/meson_options.txt
Alvaro-Kothe: Thanks I didn't know this option. Changed to be a `feature`.

subprojects/packagefiles/xsimd/meson.build
Alvaro-Kothe: Changed to conditionally add this definition if xtl if found.

subprojects/packagefiles/xsimd/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option(
+    'xtl_complex',
+    type: 'feature',
+    value: 'disabled',
```
bgilbert: Presumably this should default to `auto`?  In which case, you can just remove this line.

subprojects/packagefiles/xsimd/meson.build
Alvaro-Kothe: I copied this from their README. Changed to `[>=0.7.0, <1]`

subprojects/packagefiles/xsimd/meson.build
```diff
@@ -0,0 +1,28 @@
...
+  'cpp',
+  license: 'BSD-3-Clause',
+  meson_version: '>=0.56.0',
+  version: '14.1.0',
+)
+
+xsimd_inc = include_directories('include')
+xtl_dep = dependency(
+  'xtl',
+  version: ['>=0.7.0', '<1.0.0'],
```
bgilbert: Is an upcoming 1.0 known to break xsimd?  If not, we usually don't declare the upper bound.

subprojects/packagefiles/xsimd/meson_options.txt
Alvaro-Kothe: I don't have enough knowledge on `xsimd` to make this decision. I mainly adapted their cmake config to meson.

https://github.com/xtensor-stack/xsimd/blob/4050d947536137280ee9c01f2ec73fb3071e63f4/xsimdConfig.cmake.in#L18-L29

subprojects/packagefiles/xsimd/meson.build
Alvaro-Kothe: I don't think so. removed.

subprojects/packagefiles/xsimd/meson_options.txt
Alvaro-Kothe: Removed this line to maintain consistency with `xtensor`, which uses auto with `xsimd`. 

I also added the cpp_std=c++14 that I forgot.

PR #2735, opened 2026-04-07
TCLAP 1.2.5

ManuelAtWork:
`v1.2.5` fixes a problem with `C++20` (among other things)


subprojects/packagefiles/tclap/meson.build
```diff
@@ -13,8 +14,8 @@ project(
 # Library (Main Dependency)
 # --
 
-tclap_dep = declare_dependency(
-  version: '1.2.4',
+tclap = declare_dependency(
```
bgilbert: The name `tclap_dep` is technically part of the wrap's API and shouldn't be changed without a reason.  What CI wants you to do is:

```meson
meson.override_dependency('tclap', tclap_dep)
```

subprojects/packagefiles/tclap/meson.build
```diff
@@ -13,8 +14,8 @@ project(
...
+  version: '1.2.5',
```
bgilbert: `version` is the same as the one declared in `project()`, so is redundant and can be dropped.

bgilbert (commented):
Some quick observations, haven't fully reviewed.

subprojects/packagefiles/tclap/meson.build
ManuelAtWork: 👍

subprojects/packagefiles/tclap/meson.build
ManuelAtWork: Fixed, thanks for the info.

ManuelAtWork:
CI used to fail for Win32 platforms. Explanation:
- Tests in the original TCLAP rely on `bash`-scripts, and were never meant to run on Windows.
- Moreover, TCLAP is known to cause problems on Windows.
- Wraps are not meant to fix deficits in the original code.
- Therefore, I disabled tests on Windows.

subprojects/packagefiles/tclap/meson.build
bgilbert: I meant that both the dependency name `tclap` and the variable name `tclap_dep` are part of the API.  The `override_dependency` call I wrote is correct, and `tclap.wrap` and `releases.json` should be updated to match.

bgilbert:
What is the nature of the "problems on Windows"?  If it isn't expected to work at all there, we should disable the project entirely on Windows (with `build_on` in `ci_config.json`), not just the tests.

ManuelAtWork:
> What is the nature of the "problems on Windows"? If it isn't expected to work at all there, we should disable the project entirely on Windows (with `build_on` in `ci_config.json`), not just the tests.

There are 2 areas of concern:
- Occasional incompatilibty of the TCLAP implementation. (keywords: `std:sstream`, config.h, include paths, ...) It depends on the MSVC/MSYS version. Many MSVC users will be lucky, so I would not generally disable builds entirely.
- Automated tests: They used to rely on `bash` (and probably line endings). So they never ran on Windows.

The present wrap mirrors the policy of the original TCLAP developers: "_On Win32, use at your own risk. No tests._"

subprojects/packagefiles/tclap/meson.build
ManuelAtWork: Get it. 😅
Just force-pushed the changes.

PR #2748, opened 2026-04-16
libarchive: update from 3.8.6 to 3.8.7

bgilbert:


bgilbert:
Usual Alpine encoding issues and MSVC failures.  Also a `test_write_filter_gzip` failure on Ubuntu s390x, which also occurs with the previous release.

PR #2749, opened 2026-04-18
libarchive: fix link error with headers but no pkg-config file

bgilbert:
Stop enabling bzip2 if we only find `bzlib.h`, and subsequently failing to link.

Also test with `LRZIP=NOCONFIG`, matching upstream.


bgilbert:
Upstream PR in https://github.com/libarchive/libarchive/pull/2963.

bgilbert:
Upstream prefers a source change for the s390x test failure, so I've dropped that portion.

bgilbert:
Same failures as https://github.com/mesonbuild/wrapdb/pull/2748.

PR #2752, opened 2026-04-22
cpp-httplib: update to 0.43.1

jfsimoneau:


bgilbert:
Thanks for the PR.  FYI, version updates for this wrap (and some of the others) are automatically PRed every Wednesday and Saturday, so manual updates aren't usually needed unless there's an URL change or similar.

PR #2755, opened 2026-04-23
Glfw wayland fixes

Oxyd:
Previously the Wayland backend wouldn't build at all. I pretty much copied what the upstream CMake does in regards to the Wayland backend.

Second, GLFW can be built with multiple backends enabled, and the upstream CMake supports that. So I added support for that here as well.


luigifcruz:
This looks good to me. The multiple backend support is very useful.

luigifcruz:
@bgilbert Any chance we could prioritize this? Wayland is now the default compositor in many distros and `auto` support is paramount for a good user experience.

PR #2757, opened 2026-04-24
added vk-bootstrap to warp

yajatkaul:
Vk bootstrap is a very used library for vulkan users as it eliminates a lot of boilerplate code saving a lot of time, this is a must have imo


yajatkaul:
The reason why it first checks for vulkan and falls back to vulkan headers is because people would prefer using the sdk version since it gives addition headers unlike the vulkan headers such as `vk_enum_string_helper`

stephanlachnit (approved):
LGTM - You could add the tests (catch2 and glfw are both available on WrapDB) but not required.

The only question I have, but this is likely limited by my lack of knowledge about Vulkan, is why the dependency to `VulkanHeaders` is not enough - users of the wrap could still depend on `vulkan` themselves in case they need special headers, no?

Or does the SDK not provide the `VulkanHeaders` dependency? In that case maybe a build option is better to select which dependency should be used?

subprojects/packagefiles/vk-bootstrap/meson.build
```diff
@@ -0,0 +1,31 @@
...
+
+vk_bootstrap_inc = include_directories('src')
+
+vulkan_dep = dependency(
+  'vulkan',
+  required: true,
+  fallback: ['vulkan-headers', 'vulkan_headers_dep'],
+)
+
+vk_bootstrap_lib = static_library(
```
bgilbert, line 18: Why not just `library()`?

subprojects/packagefiles/vk-bootstrap/meson.build
```diff
@@ -0,0 +1,31 @@
...
+  'vk-bootstrap',
+  ['src/VkBootstrap.cpp'],
+  include_directories: vk_bootstrap_inc,
+  dependencies: [vulkan_dep],
+)
+
+vk_bootstrap_dep = declare_dependency(
+  include_directories: vk_bootstrap_inc,
+  link_with: vk_bootstrap_lib,
+  dependencies: [vulkan_dep],
```
bgilbert, line 28: I believe this is redundant, since `vk_bootstrap_lib` also depends on it.

subprojects/packagefiles/vk-bootstrap/meson.build
```diff
@@ -0,0 +1,31 @@
...
+  license: 'MIT',
+  meson_version: '>=0.63.0',
+  default_options: ['cpp_std=c++17'],
+)
+
+vk_bootstrap_inc = include_directories('src')
+
+vulkan_dep = dependency(
+  'vulkan',
+  required: true,
```
bgilbert, line 14: Redundant.

subprojects/packagefiles/vk-bootstrap/meson.build
```diff
@@ -0,0 +1,31 @@
...
+  fallback: ['vulkan-headers', 'vulkan_headers_dep'],
```
bgilbert, line 15: Any Meson new enough to meet WrapDB's global minimum version will support reading provides from the wrap file, so there's no longer a reason to include a fallback.

bgilbert (commented):
Please squash your commits.

I concur with the request for tests if possible.  Historically WrapDB wraps haven't always included tests but it'd be good for us to get better about this.

PR #2767, opened 2026-05-01
lua: support building as C++

v-for-vandal:
Add option to build lua as C++.


dcbaker:
Lua upstream doesn't support this, and we don't add features that are not available upstream.

I'm also confused on why you'd even want to compile these files as c++, they're C files and can be linked with C++ trivially.

v-for-vandal:
@dcbaker 
Lua most definetly supports C++ 
1. https://github.com/lua/lua/blob/c037162a1a657088d722f550e287015525bb2259/manual/manual.of#L2816
    >> (Lua will use exceptions if you compile it as C++; search for @id{LUAI_THROW} in the source code for details.)
2. https://github.com/lua/lua/blob/c037162a1a657088d722f550e287015525bb2259/luaconf.h#L327  defined(__cplusplus)
3. throw itself: https://github.com/lua/lua/blob/c037162a1a657088d722f550e287015525bb2259/ldo.c#L69

I actually found no working way to build lua as c++ using existing wrap. You can't just link it - you need to compile it (see links above). You can't use '-xc++' as option to compiler - even if you know how to set up compiler options only for subproject (  ```[lua:built-in options] c_args=['-xc++']``` in native file didn't work for me), it will not be enough - option 'c_std=c99' is incompatible with '-xc++'. So one need to also find a way to influence variable override_options (see my changes to meson.build) from within native file.  I dont think patching meson.build from native file is supported feature of Meson?

It is possible to redirect all C code to be compiled as C++ by overriding C  compiler and C linker in native file in global  [binaries] section, but then you can't use e.g. wrap for protobuf nor any other library that has real C code.

P.S.
Sorry, whats the rule here  - should I '@' mention person I am replying to, or not ?

dcbaker:
No need to at me.

This is annoying because the code exists, but lua itself injects -std=c99 into their makefiles.

The only really feasable way to do this is to do a copy of the files from `.c` to `.cpp` (or some other valid c++ extension).

Either with a generator or a loop of `fs.copyfile`, something like:
```meson
sources = [...]
if get_option('build_as_cpp')
  add_language('cpp', required : true)
  _sources = []
  foreach s : sources
    _sources += fs.copyfile(s, fs@s@.cpp')
  endforeach
  sources = _sources
endif
```

You'd then just compile the `sources`, and everything should just magically work.

dcbaker:
Whether that is something we want to support in our wrap is a question I can't answer though, let's see if one of the other devs has an opinion.

v-for-vandal:
Hi @bgilbert !
I am tentatively requesting a second opinion as "one of the other devs" regarding this PR :)

bgilbert:
Since Lua is meant to be embedded, I'm not surprised that there are build options not supported by the build system, and I do think it's useful to support this feature.  So I'm okay with something like https://github.com/mesonbuild/wrapdb/pull/2767#issuecomment-4380992681.

PR #2799, opened 2026-06-17
Adds freetype compilation unit when toggled

jUnG3:
Adds the `misc/freetype/imgui_freetype.cpp` compilation unit with its dependencies to the `imgui` lib when the option `freetype` is set.

that enables this kind of configuration in your project

`dependency('imgui', default_options: ['freetype=true'])`

### DIFF
--- a/subprojects/packagefiles/imgui/meson.build
+++ b/subprojects/packagefiles/imgui/meson.build
@@ -26,6 +26,12 @@ sources = files(
 cpp = meson.get_compiler('cpp')
 dependencies = []
 
+if get_option('freetype')
+    sources += files('misc/freetype/imgui_freetype.cpp')
+    add_project_arguments('-DIMGUI_ENABLE_FREETYPE', language: 'cpp')
+    dependencies += dependency('freetype2')
+endif
+
 # renderer backends
 dependencies += cpp.find_library(
   'd3dcompiler',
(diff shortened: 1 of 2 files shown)


subprojects/packagefiles/imgui/meson.build
```diff
@@ -26,6 +26,12 @@ sources = files(
 cpp = meson.get_compiler('cpp')
 dependencies = []
 
+if get_option('freetype')
+    sources += files('misc/freetype/imgui_freetype.cpp')
+    add_project_arguments('-DIMGUI_ENABLE_FREETYPE', language: 'cpp')
+    dependencies += dependency('freetype2')
+endif
```
stephanlachnit, line 33: I think a features makes more sense than a boolean option here:


```suggestion
freetype_dep = dependency('freetype2', required: get_option('freetype'))
if freetype_dep.found()
  sources += files('misc/freetype/imgui_freetype.cpp')
  add_project_arguments('-DIMGUI_ENABLE_FREETYPE', language: 'cpp')
  dependencies += freetype_dep
endif
```

Also instead of `add_project_arguments` create a separate variables for private compile args.

stephanlachnit (changes requested):

subprojects/packagefiles/imgui/meson_options.txt
```diff
@@ -88,3 +88,10 @@ option(
     type: 'feature',
     value: 'auto',
 )
+
+# freetype
+option(
+    'freetype',
+    type: 'boolean',
```
stephanlachnit: Still a bool, change to feature and remove value (uses auto as default).

PR #2811, opened 2026-06-27
sfml: update from 3.0.2 to 3.1.0

vSEK1RO:
meson.build files were carefully updated for `examples` and `src` directories, except for `src/Network` and `src/Graphics`.

- issue with Graphics is that `harfbuzz` and `freetype` have a circular dependency as mentioned [here](https://github.com/SFML/SFML/pull/3543). `sheenbidi` doesn't have wrap, but the project itself uses meson. also, sheenbidi caused a lot of `unused-function` warnings.

- issue with Network is that `mbedtls` library doesn't have build scripts for meson.

i have tested with harfbuzz, freetype, mbedtls installed from the system (arch) and sheenbidi added using simple `wrap-git`, it worked.


jpakkane:
As per the failing log:

> Unformatted files found.  Run tools/format.py to format these files.

When you do that, please also squash the commits into one at the same time, it makes things simpler.

PR #2818, opened 2026-07-01
libjpeg-turbo: update from 3.1.4.1 to 3.2.0

bgilbert:
Support for disabling full NEON intrinsics was removed upstream.

TurboJPEG now requires libspng, which already has a wrap in WrapDB.

Add RISC-V and Loongson SIMD.


bgilbert:
For loongarch64, upstream's compiler flags aren't supported in CI, so the SIMD build hasn't been tested.

clang-cl arm64 has the usual float ABI problem:

```
   28 | #error "NEON intrinsics not available with the soft-float ABI. Please use -mfloat-abi=softfp or -mfloat-abi=hard"
```

Alpine ppc64le is a regression:

```
 1/52 libjpeg-turbo:tjunittest-yuv              FAIL            16.48s   killed by signal 11 SIGSEGV
 2/52 libjpeg-turbo:tjunittest-yuv-alloc        FAIL            16.94s   killed by signal 11 SIGSEGV
 4/52 libjpeg-turbo:tjunittest-yuv-nopad        FAIL            15.97s   killed by signal 11 SIGSEGV
```

bgilbert:
Actually, no, the upstream CMake build has those segfaults in emulated Alpine ppc64le on both 3.1.4.1 and 3.2.0.

PR #2820, opened 2026-07-02
Add geo-utils-cpp 1.2.2

gistrec:
Adds [geo-utils-cpp](https://github.com/gistrec/geo-utils-cpp) 1.2.0 — a header-only C++17 library for spherical lat/lng geometry (distance, bearing, polygon area, point-in-polygon, snap-to-route, encoded polylines). Apache-2.0, no dependencies.

The port is a minimal header-only wrap (`declare_dependency` + `override_dependency` + header install); upstream's self-verifying `examples/gps_track.cpp` is built and registered as a smoke test so the wrap exercises the headers on CI.

`tools/sanity_checks.py` passes locally (macOS arm64, AppleClang 17): the wrap downloads, the dependency resolves as `geo-utils-cpp`, the smoke test compiles and runs.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1308,6 +1308,14 @@
       "0.20.6-1"
     ]
   },
+  "geo-utils-cpp": {
+    "dependency_names": [
+      "geo-utils-cpp"
+    ],
+    "versions": [
+      "1.2.2-1"
+    ]
+  },
   "gflags": {
     "dependency_names": [
       "gflags"
--- a/subprojects/geo-utils-cpp.wrap
+++ b/subprojects/geo-utils-cpp.wrap
@@ -1,0 +1,8 @@
+[wrap-file]
+directory = geo-utils-cpp-1.2.2
+source_url = https://github.com/gistrec/geo-utils-cpp/archive/refs/tags/v1.2.2.tar.gz
+source_filename = geo-utils-cpp-1.2.2.tar.gz
+source_hash = 2ecc7726c73d7565e7820feafd944fc3d1800f586ca79e8b8b7fdb67a7ae2274
+
+[provide]
+dependency_names = geo-utils-cpp


subprojects/packagefiles/geo-utils-cpp/meson.build
```diff
@@ -0,0 +1,27 @@
+project(
+  'geo-utils-cpp',
+  'cpp',
+  version: '1.2.0',
+  license: 'Apache-2.0',
+  meson_version: '>=0.56.0',
+  default_options: ['cpp_std=c++17'],
```
bgilbert: This doesn't do anything for the actual library because it's header-only, and it bumps the minimum Meson to 0.63.  I'd suggest moving it to `override_options` on the executable.

subprojects/packagefiles/geo-utils-cpp/meson.build
```diff
@@ -0,0 +1,27 @@
...
+install_subdir(
+  'include/geo',
+  install_dir: get_option('includedir'),
+)
+
+# Upstream's self-verifying end-to-end example doubles as a smoke test.
+geo_utils_cpp_smoke = executable(
+  'geo_utils_cpp_smoke',
+  'examples/gps_track.cpp',
+  dependencies: geo_utils_cpp_dep,
```
bgilbert:
```
build_by_default: false
```

bgilbert (commented):
LGTM generally.

gistrec:
Thanks! Both addressed in f7bd7b7 (cpp_std moved to override_options, build_by_default: false added).

bgilbert:
Looks good, go ahead and squash your commits and I'll merge.

gistrec:
The two `Visual Studio (clang-cl, …)` checks are failing on an internal Meson **1.12.0rc1** bug, not on anything in this wrap. `meson --pre` started resolving to 1.12.0rc1 (released a couple of hours before this run), and `ClangClCPPCompiler.get_options()` registers `vc++23` as a `cpp_std` value that is missing from Meson's `ALL_STDS`, so `UserStdOption.set_versions()` hits `assert all(std in self.all_stds …)` the moment a clang-cl C++ compiler is detected — for any C++ project:

```
cpp.py:958 get_options → cpp.py:823 _get_options_impl → options.py:599 set_versions
AssertionError  ("Unhandled python exception")
```

Regular MSVC (`cl`) and the other 21 checks pass; only clang-cl trips because it's the only compiler that advertises `vc++23`.

Upstream fix: mesonbuild/meson#15986 (one line — add `vc++23` to `ALL_STDS`). clang-cl should go green once that lands and `--pre` picks it up. Nothing to change here.

bgilbert:
Sorry, I was a little slow on the uptake here and just noticed that you're the upstream maintainer for this package.  A couple questions:

- Would it make sense to keep the Meson files as an additional build system directly in the upstream repo?
- Are there existing users for this wrap?  There's a lot of software on the Internet, and WrapDB should probably limit itself to packages that someone actually needs to consume as a wrap.  I myself maintain a widely-packaged C library that isn't in WrapDB because no one consumes it as a source-level dependency of a Meson build.

gistrec:
Thanks! Both questions are fair:
* **Meson upstream: done.**
  I just added a native meson.build upstream in v1.2.2 (https://github.com/gistrec/geo-utils-cpp/releases/tag/v1.2.2), and updated this PR, so WrapDB no longer carries my build files.
 * **Consumers: none yet**
  This wrap would be the first, and there's no Meson entry point to consume it through until it lands. The library is already on vcpkg, Conan, xrepo, and build2/cppget, so it's actively maintained; WrapDB is the one ecosystem it's missing. Your call on whether that's enough.

bgilbert:
I think it's better to wait until there's a downstream consumer, rather than speculatively adding a package here.  Now that the build files are upstream, anyone can add a wrap file directly to their project, so no one will be blocked on the package not being in WrapDB.

If a downstream consumer appears, though, it makes sense to add a WrapDB package so wrapfile maintenance can be shared between downstreams.

gistrec:
> I think it's better to wait until there's a downstream consumer

Makes sense. If someone starts using it as a Meson dependency, I'll open a new PR.

Thanks a lot for your time!

PR #2821, opened 2026-07-02
sdl3: Update to 3.4.12

res2k:
Latest SDL3 release


res2k:
s390x, ppc64le fail in "install Debian packages":
```
  The following packages have unmet dependencies:
   liboss4-salsa-asound2 : Conflicts: libasound2
  E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
```

Unfortunately, I have no idea how to address this...

res2k:
> s390x, ppc64le fail in "install Debian packages":

I tried to reproduce that by failure by locally running an image that is reported in the run's "Runner Image" information: https://github.com/IBM/action-runner-image-pz/releases/tag/1.2-release

Unfortunately, I could not reproduce the conflict above.
However, I also noticed that the commands, eg `sudo apt-get -y install build-essential ninja-build`, behave differently in my "fresh" container, when compared to the output of the workflow run.
Which makes me wonder: is it possible there are already some packages installed in the runner image(s), and could that cause the conflict?

res2k:
> Which makes me wonder: is it possible there are already some packages installed in the runner image(s), and could that cause the conflict?

Actually, I checked that by dropping an `apt list --installed` into the "Sanity checks" workflow Ubuntu builds.
And yeah ... a lot of packages are installed on the runners, including those causing the conflict.

res2k:
> And yeah ... a lot of packages are installed on the runners, including those causing the conflict.

Some more workflow abuse and I found out that the `temurin-*` packages seem to be the offenders, as they have `liboss4-salsa-asound2` as a dependency.

res2k:
I eventually found an approach that relies on `ci_config.json` only: 9da4b6edb401889257a14b4e6fe4669a5f1c84b2

The remaining sanity check failure is a Meson bug and, afaics, should go away once a new release containing https://github.com/mesonbuild/meson/commit/2b4f5dcd1b37a4a8915ba507b9da250a2a64aa4c is made.

ci_config.json
```diff
@@ -1460,7 +1460,8 @@
       "libwayland-bin",
       "libxkbcommon-dev",
       "libegl-dev",
-      "libgl-dev"
+      "libgl-dev",
+      "liboss4-salsa-asound2-"
```
jpakkane, line 1464: The dash at the end of this seems suspicious.

res2k: It's necessary, to _uninstall_ that package on ppc64le & s390x, otherwise it causes conflicts there.
(See https://github.com/mesonbuild/wrapdb/pull/2821#issuecomment-4869160063 & my other comments above.)

PR #2823, opened 2026-07-06
sparkcli: Update to 0.0.7

antoniovazquezblanco:


antoniovazquezblanco:
An optional depencency warning is thrown and that is causing builds to fail. How should that be managed?

```
 sparkcli| subprojects/sparkcli-0.0.7/assets/meson.build:6: WARNING: vhs not found: skipping documentation asset generation (install https://github.com/charmbracelet/vhs to enable it).
 ```

bgilbert:
You can set `fatal_warnings` to false in `ci_config.json`, along with a `_comment` explaining why.

antoniovazquezblanco:
Done!

Just for future releases, is it a good practice to use warnings for this or should that be an informational note?

Thanks!

bgilbert:
For something like this, where nothing will really go wrong, I'd suggest `message()` instead.

PR #2833, opened 2026-07-15
zxc: new wrap for 0.13.2

hellobertrand:
Adds a wrap for [zxc](https://github.com/hellobertrand/zxc), a high-performance asymmetric lossless compression library written in C (BSD-3-Clause).

Key points
- Native Meson build: meson.build is maintained upstream, so this wrap ships no patch_directory / packagefiles.
- Provides the libzxc dependency (via meson.override_dependency).
- Runtime SIMD dispatch (SSE2 / AVX2 / AVX-512 on x86-64, NEON on ARM) with a scalar fallback.
- Requires Meson >= 0.63.0 (uses c_std in default_options).
- Builds and tests clean on Linux, macOS and Windows

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -5167,6 +5167,14 @@
       "1.3.3-1"
     ]
   },
+  "zxc": {
+    "dependency_names": [
+      "libzxc"
+    ],
+    "versions": [
+      "0.13.2-1"
+    ]
+  },
   "zycore": {
     "dependency_names": [
       "zycore"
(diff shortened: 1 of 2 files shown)


bgilbert:
Thanks for the PR.  Are there any users yet who need this as a Meson wrap?  If so, the PR looks fine (though please squash your commits).  If not, it's probably premature to add it to WrapDB; we aren't in a position to properly maintain wraps for all available libraries that build with Meson.

hellobertrand:
@bgilbert thanks for taking the time to review.

For now, no one has explicitly requested a wrap for zxc just yet. I was submitting this proactively to make the library more accessible.

I completely understand your policy regarding the maintenance burden. I'll go ahead and close this PR for now to keep your queue clean, and I'll reopen it (with squashed commits) once there is concrete user demand for it.

PR #2838, opened 2026-07-16
Install fluidsynth library.

jpakkane:


jpakkane:
The failures are in no way related to the changes done. The Windows thing sucks, but VS fails if you don't add special flags and clang-cl fails if it does.

PR #2839, opened 2026-07-18
id3tag: add 0.16.4

0x0003:
Adds libid3tag 0.16.4, an MP3 id3 tag manipulation library.


subprojects/packagefiles/id3tag/meson.build
```diff
@@ -0,0 +1,135 @@
...
+id3tag_version_conf.set('PROJECT_VERSION_MINOR', version_split[1])
+id3tag_version_conf.set('PROJECT_VERSION_PATCH', version_split[2])
+
+id3tag_h = configure_file(
+  input: 'include/id3tag.h.in',
+  output: 'id3tag.h',
+  configuration: id3tag_version_conf,
+)
+
+libid3tag = static_library(
```
stephanlachnit:
```suggestion
libid3tag = library(
```

Any reason this is explicitly static?

subprojects/packagefiles/id3tag/meson.build
```diff
@@ -0,0 +1,135 @@
...
+
+gperf = find_program('gperf')
+
+id3tag_c_args = ['-DHAVE_ASSERT_H']
+# gperf <3.1 generates "unsigned int" length instead of "size_t"
+if not find_program(
+  'gperf',
+  version: '>=3.1',
+  required: false,
+).found()
```
bgilbert:
```suggestion
if not gperf.version().version_compare('>=3.1')
```

subprojects/packagefiles/id3tag/meson.build
```diff
@@ -0,0 +1,135 @@
...
+  id3tag_c_args += '-DHAVE_FTRUNCATE'
+endif
+if compiler.has_function('fseeko')
+  id3tag_c_args += '-DHAVE_FSEEKO'
+endif
+if compiler.has_function('ftello')
+  id3tag_c_args += '-DHAVE_FTELLO'
+endif
+
+cmake_prog = find_program('cmake')
```
bgilbert: Nope, we don't do that here.  You'll need to replicate the CMake logic directly in the Meson build code.

subprojects/packagefiles/id3tag/meson.build
```diff
@@ -0,0 +1,135 @@
...
+)
+
+libid3tag = static_library(
+  'id3tag',
+  [
+    id3tag_h,
+    compat_c,
+    frametype_c,
+    genre_dat,
+    'src/crc.c',
```
bgilbert: Nit: better to use `files()` for lists of filenames so `meson format` will keep the list sorted.

subprojects/packagefiles/id3tag/meson.build
```diff
@@ -0,0 +1,135 @@
...
+    'src/render.c',
+    'src/tag.c',
+    'src/ucs4.c',
+    'src/utf16.c',
+    'src/utf8.c',
+    'src/util.c',
+    'src/version.c',
+  ],
+  include_directories: include_directories('src'),
+  dependencies: [dependency('zlib')],
```
bgilbert, line 112: Please pull out the `dependency('zlib')` to a variable rather than invoking it more than once.

bgilbert (commented):
Does upstream have any test cases you can add to the Meson config?

subprojects/packagefiles/id3tag/meson.build
0x0003: Fixed.

No reason other than that I took [existing build file from MPD repo](https://github.com/MusicPlayerDaemon/MPD/blob/ff22bd411bedbda508eb9406e9fe77f1116a2d11/subprojects/packagefiles/id3tag/meson.build) to use as a base and it had it as static, I didn't pay any mind to it.

subprojects/packagefiles/id3tag/meson.build
0x0003: Applied and bumped minimum meson version to 0.62

subprojects/packagefiles/id3tag/meson.build
0x0003: I tried replicating it in this build file directly, but it looked so incredibly ugly. I ended up with a python script, is that adequate?

0x0003:
> Does upstream have any test cases you can add to the Meson config?

No, upstream does not have any tests.

subprojects/packagefiles/id3tag/meson.build
```diff
@@ -0,0 +1,128 @@
...
+  c_args: id3tag_c_args,
+  install: true,
+  vs_module_defs: 'id3tag.def',
+)
+
+install_headers(id3tag_h)
+
+libid3tag_dep = declare_dependency(
+  link_with: libid3tag,
+  dependencies: [zlib],
```
bgilbert: To be clear, a libid3tag user needs to link directly with zlib?  Usually the dependencies inherited from `link_with` are enough and an explicit `dependencies:` is not required.

subprojects/packagefiles/id3tag/meson.build
```diff
@@ -0,0 +1,128 @@
...
+  include_directories: [include_directories('src'), include_directories('.')],
```
bgilbert: `src` appears to only contain internal headers, which probably shouldn't be exported to library users.

subprojects/packagefiles/id3tag/meson.build
```diff
@@ -0,0 +1,128 @@
...
+      'src/util.c',
+      'src/version.c',
+    ),
+  ],
+  include_directories: include_directories('src'),
+  dependencies: [zlib],
+  c_args: id3tag_c_args,
+  install: true,
+  vs_module_defs: 'id3tag.def',
+)
```
bgilbert, line 117: Upstream uses a library version that matches the package version, so `version: meson.project_version()`.

subprojects/packagefiles/id3tag/meson.build
bgilbert: Yup, Python script is perfectly fine.

bgilbert (commented):
LGTM generally.

subprojects/packagefiles/id3tag/meson.build
0x0003: Right, link_with should cover it.

subprojects/packagefiles/id3tag/meson.build
0x0003: Added.

subprojects/packagefiles/id3tag/meson.build
0x0003: Dropped src/

bgilbert (approved):

subprojects/packagefiles/id3tag/meson.build
0x0003: Dropped the zlib variable too since it's only invoked once now.

subprojects/packagefiles/id3tag/meson.build
stephanlachnit: @0x0003 pulling the dependency out was reverted, can you change it back?

bgilbert: xref https://github.com/mesonbuild/wrapdb/pull/2839#discussion_r3705308647.  I'm okay with either approach since there's no duplication now.

stephanlachnit: Ah sorry, missed that

PR #2841, opened 2026-07-18
imgui: update to v1.92.8

Kanashimo:
Latest version of imgui from 12 May 2026, previous one was from 20 November 2025


subprojects/packagefiles/imgui/meson.build
```diff
@@ -2,8 +2,8 @@ project(
   'imgui',
   'cpp',
   license: 'MIT',
-  version: '1.92.5',
-  meson_version: '>=0.56.0',
+  version: '1.92.8',
+  meson_version: '>=1.0.0',
```
stephanlachnit: Why was the required Meson version bumped?

Kanashimo: I've looked at the other packages and most of them requires >= 1.0.0 right now, there is not technical reason for that. I just wanted to update it, but I don't think that supporting that old version make any sense. 1.0 is from December 2022 and it has been 4 years since its release, so it's well accommodated

stephanlachnit: If there is not technical reason then just revert it - the idea of the minimum version is to actually reflect the features used. Some poeple might still be on Ubuntu 22.04 with an old Meson, so let's leave it at `0.56.0`.

Kanashimo: Okay, I'll change that in an hour

Kanashimo:
@stephanlachnit, fixed. I would be greatful if you could review the #2848 too. Thanks.

PR #2842, opened 2026-07-18
libarchive: Fix bcrypt logic on Windows

meator:
See added comment for explanation.

Before, the user was allowed to set `cng` to `disabled` to completely disable `bcrypt`, which caused compilation issues if bcrypt is unconditionally needed:

```
libarchive.a(archive_util.c.obj) : error LNK2019: unresolved external symbol BCryptOpenAlgorithmProvider referenced in function __archive_mktempx
libarchive.a(archive_util.c.obj) : error LNK2019: unresolved external symbol BCryptCloseAlgorithmProvider referenced in function __archive_mktempx
libarchive.a(archive_util.c.obj) : error LNK2019: unresolved external symbol BCryptGenRandom referenced in function __archive_mktempx
```

This PR makes the build system recognize the scenario where bcrypt must be present.


subprojects/packagefiles/libarchive/meson.build
```diff
@@ -388,6 +403,8 @@ if bcrypt_dep.found() and cc.has_header(
     cdata.set('ARCHIVE_CRYPTO_@0@_WIN'.format(f.to_upper()), 1)
   endforeach
   deps += bcrypt_dep
+elif bcrypt_unconditionally_needed
```
bgilbert, line 406: The `__archive_mktempx` case also requires `bcrypt.h` (but not `HAVE_BCRYPT_H`) so this seems misleading.  I think it's better to drop `bcrypt_unconditionally_needed` and also replace the second `get_option('cng')` call with `bcrypt_required`.  We can then skip setting `ARCHIVE_CRYPTO_*_WIN` if `get_option('cng').disabled()`.

meator: Sorry, but I do not understand your feedback, could you explain it more please? Are you talking about a stylystic or a functional change?

bgilbert: Currently, if `bcrypt_unconditionally_needed` is true, this PR skips checking for `cc.has_header('bcrypt.h')`.  However, libarchive's code still needs `bcrypt.h` in this case, even though the PR could be read to imply otherwise.  I think it's clearer to drop the `elif` branch here and change the check immediately above it to `if bcrypt_dep.found() and cc.has_header('bcrypt.h', required: bcrypt_required)`.

However, we shouldn't set the `ARCHIVE_CRYPTO_*_WIN` defines if the user has explicitly disabled the `cng` option, so you'd need to add a check before setting those: `if not get_option('cng').disabled()`.

meator: My original version made more sense to me (I percieve it to be more readable), but OK.

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -375,18 +375,35 @@ if lzo2_dep.found()
...
+# bcrypt is needed in two places: for crypto stuff (optional) and
+# for __archive_mktempx implementation (unconditional dependency).
+#
+# bcrypt is needed for __archive_mktempx if
+# defined(_WIN32) && !defined(__CYGWIN__) is true.
+#
+# See
+# https://github.com/libarchive/libarchive/blob/7219b0134d771dc4b51bf86b4d01761b87398b1b/libarchive/archive_util.c#L44-L51
+bcrypt_unconditionally_needed = host_machine.system() == 'windows'
+if bcrypt_unconditionally_needed
```
bgilbert:
```suggestion
if host_machine.system() == 'windows'
```

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -375,18 +375,35 @@ if lzo2_dep.found()
...
   'bcrypt.h',
-  required: get_option('cng'),
+  required: bcrypt_required,
 )
-  cdata.set('HAVE_BCRYPT_H', 1)
-  foreach f : ['md5', 'sha1', 'sha256', 'sha384', 'sha512']
-    cdata.set('ARCHIVE_CRYPTO_@0@_WIN'.format(f.to_upper()), 1)
-  endforeach
+  if not get_option('cng').disabled()
+    cdata.set('HAVE_BCRYPT_H', 1)
```
bgilbert:
```suggestion
  cdata.set('HAVE_BCRYPT_H', 1)
  if not get_option('cng').disabled()
```

bgilbert (commented):
Couple small things, LGTM otherwise!

subprojects/packagefiles/libarchive/meson.build
meator: I mentally treat it as two different kinds of dependencies. When writing this, I consider the scenario where bcrypt is required because `host_machine.system() == 'windows'` but the user doesn't want it otherwise, as that's the most complicated state here.

My original PR reflected this viewpoint, but you want to merge them more. Should `HAVE_BCRYPT_H` be defined even when the user does not want it (when `get_option('cng').disabled() == true`)? If we went a little further with the merging logic, I could make Meson ignore `get_option('cng')` if `host_machine.system() == 'windows'` and just make it fully required.

I have tested it localy and the library compiles on Windows without `HAVE_BCRYPT_H` set.

I guess reading the code more to understand this, inspecting upstream CMake build system or possibly asking upstream could help.

meator: Update: Upstream's CMake doesn't care, it gladly let's you disable `ENABLE_CNG`, which will lead to linker errors on `__archive_mktempx`. I will make an issue/PR upstream. I'll let them decide when the macros should be defined.

bgilbert: Ultimately these sorts of questions usually have to be resolved by searching the upstream source.  With some upstreams we're forced to use configuration tricks, for example by disabling a `HAVE_X` macro when we have X but want to disable a feature that's tied to it.  Here that isn't necessary because we can set the `ARCHIVE_CRYPTO_*_WIN` macros if we want to use CNG, and skip them otherwise.

And so, there's a [Gricean](https://en.wikipedia.org/wiki/Cooperative_principle#Maxim_of_relation_(relevance)) thing going on here: if we decline to set `HAVE_BCRYPT_H` when we do indeed have `bcrypt.h`, we're implying that we _need_ to make that distinction for some reason, which needlessly makes the code more difficult to understand.

You're right that this doesn't make a functional difference; it's just a cleanliness thing.  If you feel strongly about it, I'm okay leaving it as is.

meator: The upstream has answered in https://github.com/libarchive/libarchive/pull/3486. I modified this PR to reflect their changes as best as I could.

To anybody interested, please review my PR and cross reference it with https://github.com/libarchive/libarchive/commit/e1dcb2947920ee7a5161beb011808f388ecec185.

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -375,18 +375,28 @@ if lzo2_dep.found()
...
   'bcrypt.h',
-  required: get_option('cng'),
+  required: true,
+  dependencies: bcrypt_dep,
 )
   cdata.set('HAVE_BCRYPT_H', 1)
-  foreach f : ['md5', 'sha1', 'sha256', 'sha384', 'sha512']
-    cdata.set('ARCHIVE_CRYPTO_@0@_WIN'.format(f.to_upper()), 1)
-  endforeach
+  if not get_option('cng').disabled()
```
kleisauke: FWIW, to match PR https://github.com/libarchive/libarchive/pull/3500, this will also be needed in a future release of libarchive:
```suggestion
  if not get_option('cng').disabled()
    cdata.set('ARCHIVE_CRYPTO_WIN', 1)
```

meator: I've added a comment to the wrap to indicate that the Meson build system should be rechecked when a new version of libarchive is released[^1]. I think it's too soon to implement such change, even if defining a "useless" macro should do no harm, for a similar reason to the one [bgilbert pointed out](https://github.com/mesonbuild/wrapdb/pull/2842#discussion_r3764787359).

If I remember to do it when the time comes, I'd be willing to handle the update to the wrap when new libarchive is released.

I briefly looked through the release procedure scripts and the comments I added to `subprojects/libarchive.wrap` should be stripped before the release wrap is generated, so the comment should be present only in the git repository (which is correct).

[^1]: ~~CI doesn't seem to like that for some reason. I'll test this further.~~ Resolved.

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -375,18 +375,28 @@ if lzo2_dep.found()
   deps += lzo2_dep
 endif
 
+if host_machine.system() == 'windows'
+  bcrypt_required = true
+else
+  bcrypt_required = get_option('cng')
```
kleisauke: IIUC, the bcrypt library and corresponding CNG feature is Windows-only. So, perhaps this should be wrapped in `host_machine.system() == 'windows'`, for example:
```diff
@@ -376,28 +376,24 @@ if lzo2_dep.found()
 endif
 
 if host_machine.system() == 'windows'
-  bcrypt_required = true
-else
-  bcrypt_required = get_option('cng')
-endif
-
-bcrypt_dep = cc.find_library(
-  'bcrypt',
-  required: bcrypt_required,
-)
+  bcrypt_dep = cc.find_library(
+    'bcrypt',
+    required: true,
+  )
 
-if bcrypt_dep.found() and cc.has_header(
-  'bcrypt.h',
-  required: true,
-  dependencies: bcrypt_dep,
-)
-  cdata.set('HAVE_BCRYPT_H', 1)
-  if not get_option('cng').disabled()
-    foreach f : ['md5', 'sha1', 'sha256', 'sha384', 'sha512']
-      cdata.set('ARCHIVE_CRYPTO_@0@_WIN'.format(f.to_upper()), 1)
-    endforeach
+  if bcrypt_dep.found() and cc.has_header(
+    'bcrypt.h',
+    required: true,
+    dependencies: bcrypt_dep,
+  )
+    cdata.set('HAVE_BCRYPT_H', 1)
+    if not get_option('cng').disabled()
+      foreach f : ['md5', 'sha1', 'sha256', 'sha384', 'sha512']
+        cdata.set('ARCHIVE_CRYPTO_@0@_WIN'.format(f.to_upper()), 1)
+      endforeach
+    endif
+    deps += bcrypt_dep
   endif
-  deps += bcrypt_dep
 endif
 
 crypto_dep = dependency(
```
(untested)

subprojects/packagefiles/libarchive/meson.build
kleisauke: Great! Please feel free to resolve this conversation, as I seem unable to do so for some reason.

subprojects/packagefiles/libarchive/meson.build
meator: I applied your proposed change in the latest commit.

@kleisauke Would you please be able to take a look at it? I assume that Cygwin/MSYS2 users[^1] could still use cng/bcrypt. Will test soon.

[^1]: Meson has two names for Windows: `windows` for "plain" Windows and `cygwin` for Cygwin/MSYS2, see [reference table](https://mesonbuild.com/Reference-tables.html#operating-system-names).

kleisauke: I took a look at the libarchive sources and noticed that they use `#if defined(_WIN32) && !defined(__CYGWIN__)` guards around cng/bcrypt usage, see e.g.:
https://github.com/libarchive/libarchive/blob/v3.8.9/libarchive/archive_cryptor_private.h#L65
https://github.com/libarchive/libarchive/blob/v3.8.9/libarchive/archive_hmac_private.h#L55
https://github.com/libarchive/libarchive/blob/v3.8.9/libarchive/archive_random.c#L60
https://github.com/libarchive/libarchive/blob/v3.8.9/libarchive/archive_random.c#L81

So, I think(?) that `host_machine.system() == 'windows'` would be sufficient in this case.

subprojects/packagefiles/libarchive/meson.build
```diff
@@ -375,19 +375,42 @@ if lzo2_dep.found()
...
+  endif
+else
+  if cng_feature.auto()
+    message('cng will not be used on Windows')
+  elif cng_feature.enabled()
+    error(
+      'cng is a Windows-only dependency! If you are cross compiling,',
+      'please set correct Windows values in the cross file.',
+    )
+  endif
```
kleisauke, line 413: FWIW, we could use [`feature.require()`](https://mesonbuild.com/Reference-manual_returned_feature.html#featurerequire) for this, e.g.:
```meson
cng_enabled = get_option('cng').require(
    host_machine.system() == 'windows',
    error_message: 'CNG is a Windows-only dependency, please reconfigure with -Dcng=disabled'
).allowed()
```

Though, that requires Meson 0.59.0. So, let's leave this as-is.

subprojects/packagefiles/libarchive/meson.build
kleisauke: I forgot to mention, I'm just an external contributor, so take my comments with a grain of salt. :)

meator: I know about `#if defined(_WIN32) && !defined(__CYGWIN__)`. This means that bcrypt isn't mandatory to build libarchive on Cygwin/MSYS2 Windows, but it is usable (correct me if I'm wrong). This is handled by

```meson
  bcrypt_required = true
else
  bcrypt_required = cng_feature
endif
```

kleisauke: Ah, I see what you mean. The `archive_digest` struct could _potentially_ use bcrypt on Cygwin, since there's no `!defined(__CYGWIN__)` guard here:
https://github.com/libarchive/libarchive/blob/v3.8.9/libarchive/archive_digest_private.h#L162-L177
(although, I'm not sure whether bcrypt is actually supported/usable on Cygwin, someone brave will have to verify that)

It's a bit odd, because libarchive doesn't seem to depend on bcrypt elsewhere when compiled on Cygwin. For example, on Cygwin, `archive_random()` uses OpenBSD's random number generator:
https://github.com/libarchive/libarchive/blob/v3.8.9/libarchive/archive_random.c#L95
https://github.com/libarchive/libarchive/blob/v3.8.9/libarchive/archive_random.c#L103
https://github.com/libarchive/libarchive/blob/v3.8.9/libarchive/archive_random.c#L261

It also relies on external libraries (i.e. Mbed TLS, Nettle, or OpenSSL) for the `archive_cryptor` / `archive_hmac` structs (see links above).

Given this, I think the current state of this PR is OK. We could do:
```diff
@@ -377,16 +377,10 @@ endif
 
 cng_feature = get_option('cng')
 
-if host_machine.system() in ['windows', 'cygwin']
-  if host_machine.system() == 'windows'
-    bcrypt_required = true
-  else
-    bcrypt_required = cng_feature
-  endif
-
+if host_machine.system() == 'windows'
   bcrypt_dep = cc.find_library(
     'bcrypt',
-    required: bcrypt_required,
+    required: true,
   )
 
   if bcrypt_dep.found() and cc.has_header(
```
if we are certain that bcrypt is not supported/usable on Cygwin.

meator: Thank you for the research you have done, it is highly relevant to the discussion.

I tested this PR on MSYS2 UCRT64 with `-Dcng=enabled` and it builds, tests build, tests pass and both the test executable and the produced .dll depend on bcrypt (I checked using Dependency Walker). I'm not an expert on Windows dynamic linking, but I believe that the dependency would not be recorded if it was not used.

Dependency Walker resolved the following functions: `BCryptCloseAlgorithmProvider`, `BCryptCreateHash`, `BCryptDeriveKeyPBKDF2`, `BCryptDestroyHash`, `BCryptDestroyKey`, `BCryptEncrypt`, `BCryptFinishHash`, `BCryptGenRandom`, `BCryptGenerateSymmetricKey`, `BCryptGetProperty`, `BCryptHashData`, `BCryptOpenAlgorithmProvider`, `BCryptSetProperty`.

Because of this, I believe that my PR is ready to go in its current state.

kleisauke: I think the MSYS environment is the only one that provides the Cygwin runtime library. See for example:
https://www.msys2.org/docs/environments/#overview

I can also confirm that it compiles and links correctly against bcrypt in that environment.

<details>
  <summary>Details</summary>

```console
$ pacman -S gcc git meson
$ git clone -b pr/fix-libarchive-bcrypt-windows https://github.com/meator/wrapdb.git
$ cd wrapdb/
$ cat > cygwin-x86_64.ini << EOF
[binaries]
c = '/usr/bin/x86_64-pc-cygwin-gcc'
cpp = '/usr/bin/x86_64-pc-cygwin-g++'
ar = '/usr/bin/x86_64-pc-cygwin-gcc-ar'
pkg-config = 'x86_64-pc-cygwin-pkg-config'
strip = '/usr/bin/strip'

[host_machine]
system = 'cygwin'
cpu_family = 'x86_64'
cpu = 'x86_64'
endian = 'little'
EOF
$ meson setup build --cross-file cygwin-x86_64.ini -Dwraps=libarchive -Dlibarchive:cng=enabled
[...]
libarchive| C compiler for the host machine: /usr/bin/x86_64-pc-cygwin-gcc (gcc 15.2.0 "x86_64-pc-cygwin-gcc (GCC) 15.2.0")
libarchive| C linker for the host machine: /usr/bin/x86_64-pc-cygwin-gcc ld.bfd 2.46
[...]
libarchive| Library bcrypt found: YES
libarchive| Has header "bcrypt.h" with dependency -lbcrypt: YES
$ meson compile -C build
$ pacman -S llvm
$ llvm-readobj --coff-imports build/subprojects/libarchive-3.8.9/libarchive/libarchive-13.dll
[...]
Import {
  Name: bcrypt.dll
  ImportLookupTableRVA: 0xE7480
  ImportAddressTableRVA: 0xE78F8
  Symbol: BCryptCloseAlgorithmProvider (2)
  Symbol: BCryptCreateHash (6)
  Symbol: BCryptDestroyHash (13)
  Symbol: BCryptFinishHash (27)
  Symbol: BCryptHashData (35)
  Symbol: BCryptOpenAlgorithmProvider (39)
}
```
</details>

(note that it links against only a subset of the bcrypt functions you mentioned, which confirms my previous comment)

> Because of this, I believe that my PR is ready to go in its current state.

Indeed, it looks good to me as well. 👍

kleisauke: Forgot to mention, `-Dlibarchive:cng=disabled` compiles also correctly on Cygwin. In that case, I see:
```console
libarchive| Library bcrypt skipped: feature cng disabled
```

meator: > I think the MSYS environment is the only one that provides the Cygwin runtime library. See for example:
> https://www.msys2.org/docs/environments/#overview

You're right, my mistake. My experience with Windows development is limited.

I redid the thest with MSYS MSYS2 environment and I came to the same conclusion as you, that `BCryptOpenAlgorithmProvider`, `BCryptCloseAlgorithmProvider`, `BCryptCreateHash`, `BCryptDestroyHash`, `BCryptFinishHash`, `BCryptHashData` are used.

I can also confirm that `bcrypt` is unused when `cng` is disabled.

PR #2851, opened 2026-07-21
Add libvmaf wrap (Netflix VMAF)

diegonieto:
Add libvmaf wrap (Netflix VMAF 3.2.0)

VMAF (Video Multimethod Assessment Fusion) is a perceptual video
quality assessment algorithm developed by Netflix.

**Dependency:** `libvmaf`
**Upstream:** https://github.com/Netflix/vmaf
**Version:** 3.2.0 (tag v3.2.0)

**Notes:**

- Upstream already ships a Meson build system inside `libvmaf/`.
  A root `meson.build` is provided in `packagefiles/` to serve as
  the entry point when used as a subproject.

- `libvmaf_dep` and `meson.override_dependency()` are declared in
  the root `meson.build` to enable `dependency('libvmaf')`.

- Build requires `nasm` (on x86) and `xxd`.

- **MSVC is not supported.** MSVC lacks support for
  upstream's GCC-specific AVX2 intrinsics.

- Docs and tools are disabled in CI; tests are enabled and pass
  on all other platforms.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -929,6 +929,29 @@
       "libudev-dev"
     ]
   },
+  "libvmaf": {
+    "build_options": [
+      "libvmaf:enable_docs=false"
+    ],
+    "test_options": [
+      "--timeout-multiplier=3"
+    ],
+    "debian_packages": [
+      "nasm",
+      "xxd"
+    ],
+    "alpine_packages": [
+      "nasm",
+      "xxd"
+    ],
+    "brew_packages": [
+      "nasm"
+    ],
+    "msys_packages": [
+      "nasm",
+      "neovim"
+    ]
+  },
   "libxcursor": {
     "_comment": "X11 needs a POSIX system",
     "build_on": {
--- a/subprojects/packagefiles/libvmaf/libvmaf/src/meson.build
+++ b/subprojects/packagefiles/libvmaf/libvmaf/src/meson.build
@@ -1,0 +1,737 @@
+vmaf_soversion = vmaf_api_version_major
+
+# Build libvmaf
+feature_src_dir = './feature/'
+src_dir = './'
+model_dir = '../../model/'
+cuda_dir = './cuda/'
+
+vmaf_base_include = include_directories(
+  './',
+  './feature',
+  './feature/common',
+  '../include',
+)
+
+if cc.get_id() != 'msvc'
+  vmaf_cflags_common = [
+    '-pedantic',
+    '-DOC_NEW_STYLE_INCLUDES',
+  ]
+else
+  vmaf_cflags_common = [
+    '-wd4028',  # parameter different from declaration
+    '-wd4996',  # use of POSIX functions
+    '-DOC_NEW_STYLE_INCLUDES',
+  ]
+endif
+
+
+cdata = configuration_data()
+cdata_asm = configuration_data()
+
+if cc.symbols_have_underscore_prefix()
+  cdata.set10('PREFIX', true)
+  cdata_asm.set10('PREFIX', true)
+endif
+
+is_asm_enabled = get_option('enable_asm') == true
+is_cuda_enabled = get_option('enable_cuda') == true
+is_avx512_enabled = get_option('enable_avx512') == true
+is_nvtx_enabled = get_option('enable_nvtx') == true
+
+if is_nvtx_enabled
+  cdata.set10('HAVE_NVTX', is_nvtx_enabled)
+endif
+if is_cuda_enabled
+  cdata.set10('HAVE_CUDA', is_cuda_enabled)
+endif
+
+if is_asm_enabled
+  if host_machine.cpu_family().startswith('arm64') or host_machine.cpu_family().startswith(
+    'aarch64',
+  )
+    cdata.set10('HAVE_ASM', is_asm_enabled)
+    cdata.set10(
+      'ARCH_AARCH64',
+      host_machine.cpu_family() == 'aarch64' or host_machine.cpu() == 'arm64',
+    )
+    cdata.set10(
+      'ARCH_ARM',
+      host_machine.cpu_family().startswith('arm') and host_machine.cpu() != 'arm64',
+    )
+
+    if host_machine.system() == 'darwin'
+      asm_format = 'macho'
+    else
+      asm_format = 'elf'
+    endif
+  endif
+
+  if host_machine.cpu_family().startswith('x86')
+    cdata.set10('HAVE_ASM', is_asm_enabled)
+    cdata.set10('ARCH_X86', host_machine.cpu_family().startswith('x86'))
+    cdata.set10('ARCH_X86_64', host_machine.cpu_family() == 'x86_64')
+    cdata.set10('ARCH_X86_32', host_machine.cpu_family() == 'x86')
+
+    # check NASM version
+    nasm = find_program('nasm')
+    if nasm.found()
+      nasm_r = run_command(
+        nasm,
+        '-v',
+        check: true,
+      )
+
+      if nasm_r.returncode() != 0
+        error('failed running nasm to obtain its version')
+      endif
+
+      out = nasm_r.stdout().strip().split()
+      if out[1].to_lower() == 'version'
+        if out[2].version_compare('<2.13.02')
+          error(
+            'nasm 2.13.02 or later is required, found nasm @0@'.format(out[2]),
+          )
+        elif out[2].version_compare('<2.14') and get_option('enable_avx512')
+          warning(
+            'nasm 2.14 or later is required for AVX-512 asm. Disabling AVX-512.',
+          )
+          is_avx512_supported = false
+        else
+          is_avx512_supported = true
+        endif
+        cdata.set10(
+          'HAVE_AVX512',
+          get_option('enable_avx512') and is_avx512_supported,
+        )
+        cdata_asm.set10(
+          'HAVE_AVX512',
+          get_option('enable_avx512') and is_avx512_supported,
+        )
+      else
+        error('unexpected nasm version string: @0@'.format(nasm_r.stdout()))
+      endif
+    endif
+
+    # Generate config.asm
+    cdata_asm.set10('ARCH_X86_64', host_machine.cpu_family() == 'x86_64')
+    cdata_asm.set10('ARCH_X86_32', host_machine.cpu_family() == 'x86')
+    cdata_asm.set10('PIC', true)
+    config_asm_target = configure_file(
+      output: 'config.asm',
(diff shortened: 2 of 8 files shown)


ci_config.json
```diff
@@ -929,6 +929,21 @@
       "libudev-dev"
     ]
   },
+  "libvmaf": {
+    "build_options": [
+      "libvmaf:enable_tests=false",
```
bgilbert: CI should build and run tests if feasible.

ci_config.json
```diff
@@ -929,6 +929,21 @@
...
+      "libvmaf:enable_docs=false",
+      "libvmaf:enable_tools=false"
```
bgilbert: Likewise, it's good to verify that the tools successfully build.

subprojects/packagefiles/libvmaf/meson.build
```diff
@@ -0,0 +1,53 @@
+project('libvmaf', ['c', 'cpp'],
```
bgilbert: Please add a comment indicating that these two files are synchronized with upstream, and specify the differences from the upstream version.

bgilbert:
```
AssertionError: Found files subprojects/packagefiles/libvmaf/meson.build with license headers in a project without a LICENSE.build.  The LICENSE.build file in the patch ZIP defaults to MIT unless the patch directory has its own LICENSE.build, which should state the license for the wrap's build files.
```

False positive; you'll need to reword the header comment.

ci_config.json
diegonieto: fixed

ci_config.json
diegonieto: fixed

diegonieto:
@bgilbert  I've update the PR to pass the CI. Both `msvc` and `msys2` have been disabled since it fails in either building or passing some of the tests. See https://github.com/Netflix/vmaf/pull/1477.

ylatuya:
> @bgilbert I've update the PR to pass the CI. Both `msvc` and `msys2` have been disabled since it fails in either building or passing some of the tests. See [Netflix/vmaf#1477](https://github.com/Netflix/vmaf/pull/1477).

`vmaf`  supports `mingw64`, their CI runs compilation and tests successfully ->  https://github.com/Netflix/vmaf/actions/runs/29288566688/job/86946759491

bgilbert:
Okay, what's the situation here?  Is the intention to temporarily carry a downstream fork of the Meson build scripts while PRing changes back to the upstream project?

`build_on` is meant for projects that categorically don't support an OS.  If support is expected in an upcoming release, just let the CI fail.

diegonieto:
> Okay, what's the situation here? Is the intention to temporarily carry a downstream fork of the Meson build scripts and PR changes back to the upstream project?

Yes, the idea is to make `libvmaf` part of `wrapdb` and upstream all the portable changes.

> 
> `build_on` is meant for projects that categorically don't support an OS. If support is expected in an upcoming release, just let the CI fail.

The "windows": false (Visual Studio/MSVC) is permanent, not temporary. Upstream libvmaf does not support the MSVC compiler — it uses GCC-specific SIMD intrinsics (__m256i type casts, __builtin_clz, etc.) and requires POSIX headers like pthread.h. The wrap's `meson.build` explicitly errors out for MSVC/clang-cl.

This is consistent with upstream, which also doesn't test MSVC in their CI. MSYS2 (GCC-based Windows builds) is the supported path upstream, and the one we should support here as well.

Should I leave it building without passing the tests for msys2?

bgilbert:
> Yes, the idea is to make `libvmaf` part of `wrapdb` and upstream all the portable changes.

Making sure I understand the "portable" qualification: will the wrap eventually just be the root `meson.build` and `meson_options.txt` with the paths changed, as in early drafts of this PR?

> The "windows": false (Visual Studio/MSVC) is permanent, not temporary. Upstream libvmaf does not support the MSVC compiler — it uses GCC-specific SIMD intrinsics (__m256i type casts, __builtin_clz, etc.) and requires POSIX headers like pthread.h. The wrap's `meson.build` explicitly errors out for MSVC/clang-cl.

You previously linked to https://github.com/Netflix/vmaf/pull/1477, which has a comment saying the PR will be reviewed before the next release.

> Should I leave it building without passing the tests for msys2?

Yes, if they can't be made to pass.  But the current problem is MSYS2 failing to install `vim`, which is a bug in the PR.

tools/sanity_checks.py
```diff
@@ -113,6 +113,9 @@
     'libuv': {
         'link_file_in_build_dir.py',
     },
+    'libvmaf': {
+        'attributes.h',
```
bgilbert: The PR doesn't have an `attributes.h` so this shouldn't be here.

ci_config.json
bgilbert: Not fixed; the override is still here.

subprojects/packagefiles/libvmaf/LICENSE.build
```diff
@@ -0,0 +1 @@
+SPDX-License-Identifier: BSD-2-Clause-Patent
```
bgilbert: In cases where `LICENSE.build` is needed, it should include a complete copy of the license text.  But it isn't needed here and should be removed, as previously discussed.  CI is objecting because of the word `license` in the header block of `meson.build`; reword that comment instead.

subprojects/packagefiles/libvmaf/meson.build
```diff
@@ -0,0 +1,85 @@
...
+  ['c', 'cpp'],
+  version: '3.2.0',
+  license: 'BSD-2-Clause-Patent',
+  default_options: [
+    'c_std=c11',
+    'cpp_std=c++11',
+    'warning_level=2',
+    'default_library=both',
+  ],
+  meson_version: '>= 0.56.1',
```
bgilbert: This is a tough one.  `c_std` and `cpp_std` is not respected in the `default_options` of subprojects until 0.63.0 and CI is warning about it.  However, upstream shouldn't bump its minimum Meson version because those options _are_ respected in root projects.  It might make sense to bump this only in WrapDB and add it to the list of changes that need to be made downstream.

diegonieto:
> > Yes, the idea is to make `libvmaf` part of `wrapdb` and upstream all the portable changes.
> 
> Making sure I understand the "portable" qualification: will the wrap eventually just be the root `meson.build` and `meson_options.txt` with the paths changed, as in early drafts of this PR?

Once those land upstream, the wrap reduces to just the root meson.build (with libvmaf/ path prefixes since we're one level up from upstream's entry point) and meson_options.txt. That's the goal, yes.

> 
> > The "windows": false (Visual Studio/MSVC) is permanent, not temporary. Upstream libvmaf does not support the MSVC compiler — it uses GCC-specific SIMD intrinsics (__m256i type casts, __builtin_clz, etc.) and requires POSIX headers like pthread.h. The wrap's `meson.build` explicitly errors out for MSVC/clang-cl.
> 
> You previously linked to [Netflix/vmaf#1477](https://github.com/Netflix/vmaf/pull/1477), which has a comment saying the PR will be reviewed before the next release.

Sorry, what I meant  by "permanent" is that we don't really know when `libvmaf` will actually merge and release those changes supporting `msvc`. 
The lack of supporting `msys2` in `wrapdb` (this PR) might be temporary since the `libvmaf` CI already supports it. However, after reviewing @ylatuya comment, I will try to make it work in this PR as well.

> 
> > Should I leave it building without passing the tests for msys2?
> 
> Yes, if they can't be made to pass. But the current problem is MSYS2 failing to install `vim`, which is a bug in the PR.

To avoid the CI back and forth I will try to make the tests pass in local for `msys2`. After that I'll let you know. Let me know whether you have any additional comments about this.

To give some context, this PR comes from: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/12101

subprojects/packagefiles/libvmaf/meson.build
diegonieto: Done, thanks

subprojects/packagefiles/libvmaf/LICENSE.build
diegonieto: Done, thanks.

tools/sanity_checks.py
diegonieto: Removed, thanks.

diegonieto:
@bgilbert I think I've addressed your feedback. It should be also  supporting now `msys2`, as `libvmaf` CI does. Let me know whether you have any additional comments.

ci_config.json
bgilbert: It's still here.  Disabling docs is fine but we should not disable tools unless there's a good reason.

ci_config.json
```diff
@@ -929,6 +929,33 @@
       "libudev-dev"
     ]
   },
+  "libvmaf": {
+    "build_options": [
+      "libvmaf:enable_docs=false",
+      "libvmaf:enable_tools=false"
+    ],
+    "build_on": {
+      "windows": false
```
bgilbert: No, just let the CI fail, please.  This flag is intended for cases where e.g. a library only exists to support POSIX-specific functionality.  We shouldn't set up a situation where we could accidentally leave Windows support disabled after upstream adds it.

subprojects/packagefiles/libvmaf/libvmaf/src/meson.build
```diff
@@ -0,0 +1,737 @@
...
+  endif
+
+  if host_machine.cpu_family().startswith('x86')
+    cdata.set10('HAVE_ASM', is_asm_enabled)
+    cdata.set10('ARCH_X86', host_machine.cpu_family().startswith('x86'))
+    cdata.set10('ARCH_X86_64', host_machine.cpu_family() == 'x86_64')
+    cdata.set10('ARCH_X86_32', host_machine.cpu_family() == 'x86')
+
+    # check NASM version
+    nasm = find_program('nasm')
```
bgilbert, line 78: For the record, upstream would ideally switch to the built-in nasm support in Meson 0.64+.

diegonieto: Ok, noted.

ci_config.json
diegonieto: Understood, updated.

bgilbert: I won't insist on it, but I'd tend to favor removing `error('MSVC and clang-cl are not supported. Use MSYS2 with GCC instead.')` as part of this.  While that check is indeed potentially helpful for users, upstream doesn't have it, and it likewise could unnecessarily break future builds if rotely carried forward.

bgilbert:
Our Alpine ppc64le CI is broken.

Thanks for the PR!

PR #2751, opened 2026-04-22
mimalloc: update to 3.3.2

stephanlachnit:

PR #2857, opened 2026-07-27
mimalloc: update to 3.5.1

stephanlachnit:


bgilbert:
Reopened to bump CI.

bgilbert:
Didn't work.  Could you force-push?

stephanlachnit:
Alpine (ppc64le) seems to be broken on all tests currently. The clang-cl arm64 failure seems to be a missing arm64 MSVC intrinsic in clang-cl (`__stlr64` and `__ldar64`).

bgilbert:
clang-cl arm64 succeeded in https://github.com/mesonbuild/wrapdb/pull/2751.  It's worth checking whether the previous release still builds successfully there (in which case it's an upstream regression and I'd hesitate to merge this) or doesn't (in which case it's a regression in the CI environment).

stephanlachnit:
My feeling is that they now a MSVC arm64 intrinsic which is not implemented in clang-cl yet (which IMO would not be a blocker). But I can double check.

bgilbert:
I generally consider breakage on previously-working platforms to be a blocker (which is why there are so many pending wrapdb-bot PRs).  Otherwise Meson projects could update their wraps, not have CI that notices the platform breakage, and then break their downstream users.  It probably doesn't matter if something like loongarch64 breaks, but I'm not sure whether clang-cl arm64 is sufficiently niche to ignore.

Are there any #defines we could use to work around this?

stephanlachnit:
> Are there any #defines we could use to work around this?

Unfortunately not, they are simply behind a

```c
#if defined(_M_ARM) || defined(_M_ARM64)
```

The corresponding clang-cl issue is https://github.com/llvm/llvm-project/issues/121689. Given the status of this issue my hopes are small that this is fixed any time soon and I guess there is nothing else we can do expect:

- Stop updating upstream
- Try to get a path upstream not to use those intrinsic on clang-cl (which I'm not willing to invest the time in)
- Accept that clang-cl is not supported by mimalloc (yet)

I am in favor of the last option. However 3.4.4 / 3.4.5 seemed to introduce a regression on x86 (https://github.com/microsoft/mimalloc/issues/1361), so let's see how that goes.

bgilbert:
Thanks for filing the x86 issue.  Agreed that we can't stop updating indefinitely if upstream isn't interested in fixing something.  If you're up for it, I'd favor at least reporting the arm64 issue to mimalloc, and if they don't opt to land a workaround in the next release, proceeding anyway.

stephanlachnit:
So - new update to 3.5.0, everything except clang-cl looks good now - since clang-cl is now also broken on x64, I'd say it's not supported upstream. Alpine CI failures are unrelated.

res2k:
v3.5.1 contains some fixes for clang-cl, try that version.

stephanlachnit:
Thanks @res2k - everything seems to work now

bgilbert (approved):
🎉

PR #2868, opened 2026-08-02
giflib: update to 6.1.3

neheb:


subprojects/packagefiles/giflib/meson.build
```diff
@@ -85,10 +86,11 @@ install_headers('gif_lib.h')
 
 if enable_progs or enable_tests
 
-  m_dep = cc.find_library(
-    'm',
-    required: false,
-  )
+  if cc.links('#include <math.h>\nint main(){log(0);}')
```
bgilbert, line 89: I tend to favor adding `name` so the check is visible, but nbd.

bgilbert (commented):
Does the library version need bumping?

neheb:
```
lrwxrwxrwx 1 root root    11 Jul 21 14:28 /usr/lib/libgif.so -> libgif.so.7*
lrwxrwxrwx 1 root root    15 Jul 21 14:28 /usr/lib/libgif.so.7 -> libgif.so.7.2.0*
-rwxr-xr-x 1 root root 42872 Jul 21 14:28 /usr/lib/libgif.so.7.2.0*
```

```
extra/giflib 6.1.3-2 [installed]
    Library for reading and writing gif images
```

subprojects/packagefiles/giflib/meson.build
neheb: It's a moot point. libm is only really needed with older glibc.

bgilbert (approved):

PR #2874, opened 2026-08-04
Update nanoarrow to 0.9.0

paleolimbot:
Updates nanoarrow to the latest release!


bgilbert:
Nope, that job is currently failing in all PRs.

FYI, nanoarrow is one of the wraps that's automatically updated by the bot twice a week, e.g. #2879, so routine manual updates usually shouldn't be necessary.

PR #2883, opened 2026-08-08
enet: add installation support

ichliebedich777:
There is support in CMakeLists, unlike in the Meson wrap.
```
include(GNUInstallDirs)
install(TARGETS enet
    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
)
install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/enet
    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
)
```


subprojects/packagefiles/enet/meson.build
```diff
@@ -77,6 +77,22 @@ src = [
   'win32.c',
 ]
 
+is_can_be_installed = cc.get_id() != 'msvc'
```
bgilbert: If we're installing, we should also do it on MSVC.  If upstream doesn't have a macro for injecting a dllexport declspec, you can include a `vs_module_defs` file with the wrap.

ELITE08: 👌

subprojects/packagefiles/enet/msvc_export.def
```diff
@@ -0,0 +1,77 @@
...
+	enet_socket_get_option   
+	enet_socket_listen   
+	enet_socket_receive   
+	enet_socket_send   
+	enet_socket_set_option   
+	enet_socket_shutdown   
+	enet_socket_wait   
+	enet_socketset_select   
+	enet_time_get   
+	enet_time_set
```
bgilbert: Missing final newline.

bgilbert (commented):
LGTM overall.  Please run `tools/format.py` and squash your commits.

bgilbert (approved):
Commit message will need a cleanup.  Feel free to do that yourself, or we can do it with a squash merge.

bgilbert:
Thanks!

PR #2884, opened 2026-08-08
robin-map: update to 1.4.1

stephanlachnit:


bgilbert (approved):

stephanlachnit:
Something I was wondering: it seems like the actual dependency name is `tsl-robin-map` (see https://packages.debian.org/forky/all/robin-map-dev/filelist or https://packages.fedoraproject.org/pkgs/robin-map/robin-map-devel/fedora-rawhide.html#files).

Upstream nanobind [`find_dependency(tsl-robin-map)`](https://github.com/wjakob/nanobind/blob/8e6c4f620888f24c0acc7c273dd88e2a45822d4c/cmake/nanobind-config.cmake#L291) - should be mirror this dependency name in the Wrap as well? As far as I can see only nanobind uses this dependency.

bgilbert:
Oh, good catch, [it's also like that upstream](https://github.com/Tessil/robin-map/blob/91362aab8f2c63ef90bf7c21fb2c3283ded5ae48/cmake/tsl-robin-mapConfig.cmake.in).  (Worth checking since, at least with pkg-config, distros sometimes invent their own downstream files.)

Yes, we should change the provide name here and bump the nanobind wrap as well.  Historically when changing provide names I've removed the old name immediately, since there's no way to flag an old name for a deprecation period before removing it.

stephanlachnit:
Done, I also added the unit tests for good measure. nanobind PR is https://github.com/mesonbuild/wrapdb/pull/2891

subprojects/packagefiles/robin-map/tests/meson.build
```diff
@@ -0,0 +1,23 @@
...
+  sources: files(
+    'custom_allocator_tests.cpp',
+    'main.cpp',
+    'policy_tests.cpp',
+    'robin_map_tests.cpp',
+    'robin_set_tests.cpp',
+  ),
+  dependencies: [robin_map_dep, boost_unit_test_framework_dep],
+  override_options: 'cpp_std=c++17',
+)
```
bgilbert, line 23: Good general practice: `build_by_default: false` for test executables.  Meson 1.7+ respects this, and it saves some build time even when users don't explicitly set `-Dtests=disabled`.

(Wraps that require 1.7+ probably shouldn't even have a tests option, but I don't think we're there yet.)

subprojects/packagefiles/robin-map/meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option(
+    'tests',
+    type: 'feature',
+    value: 'auto',
```
bgilbert: Redundant.

bgilbert (approved):

bgilbert (approved):

PR #2887, opened 2026-08-09
ci: define a "needs fixes" PR label and automatically remove it on push

bgilbert:
Add a privileged Actions job that removes any `needs fixes` label on a PR whenever that PR's branch is pushed.  This allows reviewers to set the label after giving review feedback, making it clear in the PR list which PRs don't need to be looked at again.


bgilbert:
Let's try this.  We can always roll it back if not useful.

PR #2897, opened 2026-08-15
c-capnproto: new wrap for 0.4.0

HaoZeke:
Pure-C Cap'n Proto runtime and the `capnpc-c` schema plugin, a maintained
fork of `opensourcerouting/c-capnproto`.

calls `meson.override_dependency('c-capnproto', libcapnp_dep)`.

The plugin is not declared in `program_names`: 0.4.0 does not call
`meson.override_find_program`, so declaring it would not match. I will
add that upstream and declare it in a later revision.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -407,6 +407,14 @@
       "1.20.1-1"
     ]
   },
+  "c-capnproto": {
+    "dependency_names": [
+      "c-capnproto"
+    ],
+    "versions": [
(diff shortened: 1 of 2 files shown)


bgilbert:
Thanks for the PR.  I'm hesitant to include a fork of an existing package, where the fork doesn't appear to have users yet.

HaoZeke:

PR #2898, opened 2026-08-15
capnp-fortran: new wrap for 0.3.0

HaoZeke:
A native modern-Fortran Cap'n Proto implementation: wire format, packed
and canonical codecs, RPC, and the `capnpc-fortran` schema plugin.

calls `meson.override_dependency('capnp-fortran', capnp_fortran_dep)`.
Meson tracks the Fortran module directory of a linked target itself, so
a consumer reaches `use capnp` without naming an include path.

The `ci_config.json` entry follows openblas, the other Fortran wrap:
gfortran for Alpine, Homebrew and Debian, `mingw-w64-ucrt-x86_64-fc` for
MSYS, and `build_on.windows = false` since there is no native Windows
Fortran compiler. This is the project's first exposure to the macOS and
MSYS toolchains, so I am expecting to iterate if CI disagrees.

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -160,6 +160,21 @@
   "cairomm": {
     "ignore_upstream_meson": "1.18.0"
   },
+  "capnp-fortran": {
+    "_comment": "Fortran; gfortran is what the runners need",
+    "alpine_packages": [
+      "gfortran"
+    ],
+    "brew_packages": [
+      "gfortran"
+    ],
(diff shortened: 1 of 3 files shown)


bgilbert:
Same comment as https://github.com/mesonbuild/wrapdb/pull/2899#issuecomment-5302461971: it's likely too early to include this wrap here.

HaoZeke:

PR #2899, opened 2026-08-15
capnp-janet: new wrap for 0.2.5

HaoZeke:
A Cap'n Proto implementation in C with a Janet native module on top. The
wrap provides the C runtime; the Janet binding is built by the same
project but is not what a Meson consumer depends on.

calls `meson.override_dependency('capnp-janet', capnp_janet_dep)`.

No `ci_config.json` entry: the runtime is C99 with no external
dependencies, and the optional live-peer interop targets are skipped
when capnp-C++ is absent.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -439,6 +439,14 @@
       "1.18.0-1"
     ]
   },
+  "capnp-janet": {
+    "dependency_names": [
+      "capnp-janet"
+    ],
+    "versions": [
+      "0.2.5-1"
+    ]
+  },
   "catch": {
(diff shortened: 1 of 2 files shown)


bgilbert:
Thanks for the PRs!  To limit our maintenance burden, WrapDB doesn't carry wraps for every project that uses Meson upstream; we try to stick to those that are actually needed by another Meson project.  It looks like this project is pretty new, so I'd recommend waiting until such a use case arises.

HaoZeke:

PR #2100, opened 2025-05-09
ENH: Update OpenBLAS to v0.3.29

HaoZeke:
Needs https://github.com/mesonbuild/wrapdb/pull/2005

PR #2005, opened 2025-03-31
ENH: Add LAPACK build to OpenBLAS

HaoZeke:
~Draft until `scipy` is tested a bit more.~ Tested in https://github.com/HaoZeke/OpenBLAS/pull/33#issuecomment-2813739350

The next update should be for the new 0.39 release.

Same builds / checks as in earlier PRs [1,2]

[1] https://github.com/mesonbuild/wrapdb/pull/1815
[2] https://github.com/mesonbuild/wrapdb/pull/1734

PR #2902, opened 2026-08-16
ENH: Update OpenBLAS to v0.3.34 and add LAPACK

HaoZeke:
Supersedes #2005 and #2100.

- wrap 0.3.28 -> 0.3.34 (`0.3.34-1`)
- fold OpenBLAS `lapack/` into `libopenblas` (`link_whole`), gated on `build_without_lapack` / `no_lapack`
- `F_INTERFACE_*` from OpenBLAS `cmake/fc.cmake` (LLVM Flang -> `GFORT`, classic Flang -> `FLANG`)
- `libtinfo6`, drop `build_on.windows`, `unsupported architecture` skip string (bgilbert on #2005)
- `ignore_upstream_meson` for 0.3.34

LAPACK fold was SciPy-tested on the 0.3.28 tree in https://github.com/HaoZeke/OpenBLAS/pull/33#issuecomment-2813739350. Same wrapdb checks as #1734 / #1815.

MSVC is still `c_std=gnu11`; VS jobs may stay red. `laed3` still uses the netlib Fortran sources.


bgilbert:
Upstream appears to support Windows, so we shouldn't make Windows CI green when it doesn't actually work.  It's okay to just let the CI fail.

PR #2913, opened 2026-08-31
pcre2: update from 10.47 to 10.48

bgilbert:


bgilbert:
Locale issues for some reason only on Alpine loongarch64, which wasn't tested in CI at the time of the previous release.

PR #2915, opened 2026-09-01
sfml_vala: add 3.0.0-1

nda-cunh:
Add Sfml_Vala 

https://gitlab.com/nda-cunh/sfml-vala-binding

### DIFF
--- a/ci_config.json
+++ b/ci_config.json
@@ -220,6 +220,17 @@
       "openssl"
     ]
   },
+  "sfml_vala": {
+    "debian_packages": [
+      "valac"
+    ],
+    "alpine_packages": [
+      "vala"
+    ],
+    "msys_packages": [
+      "vala"
+    ]
+  },
   "csfml" : {
     "alpine_packages": [
       "libudev-zero-dev",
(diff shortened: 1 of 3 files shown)


bgilbert:
Does this library have users that want to consume it via a Meson wrap?  WrapDB can't publish a wrap for every library that builds with Meson, so we try to stick to those where we can help avoid duplication of maintenance effort by downstream projects.

nda-cunh:
> Does this library have users that want to consume it via a Meson wrap? WrapDB can't publish a wrap for every library that builds with Meson, so we try to stick to those where we can help avoid duplication of maintenance effort by downstream projects.

It's my library, but I get it! Besides, I would have maintained it anyway since they're just vapis. I would have loved to know if there's a better way to provide Vala vapis with Meson, though! Do as you wish, don't worry, it's fine by me.

https://www.sfml-dev.org/download/bindings/

bgilbert:
Okay, I'll close for now.  Good luck with the library!  If a downstream project shows up that wants to consume sfml_vala via wrap, let's revisit this.

PR #2918, opened 2026-09-03
picohttpparser: add 1.2-1

### LINKED ISSUE #2919: Add picohttpparser
https://github.com/h2o/picohttpparser

nda-cunh:
close #2919

### DIFF
--- a/subprojects/packagefiles/picohttpparser/meson.build
+++ b/subprojects/packagefiles/picohttpparser/meson.build
@@ -1,0 +1,21 @@
+project(
+  'picohttpparser',
+  'c',
+  meson_version: '>=0.56.0',
+  version: '1.2',
+  license: 'MIT OR Artistic-1.0-Perl OR GPL-1.0-or-later',
+)
+
+lib_pico = library(
+  'picohttpparser',
+  'picohttpparser.c',
+  vs_module_defs: 'picohttpparser.def',
+  install: true,
+)
+
+dep_pico = declare_dependency(
+  include_directories: include_directories('.'),
+  link_with: lib_pico,
+)
+
+meson.override_dependency('picohttpparser', dep_pico)
--- a/releases.json
+++ b/releases.json
@@ -3801,6 +3801,14 @@
       "3.2.0-1"
     ]
   },
+  "picohttpparser": {
+    "dependency_names": [
+      "picohttpparser"
+    ],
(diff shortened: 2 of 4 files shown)


subprojects/packagefiles/picohttpparser/meson.build
```diff
@@ -0,0 +1,21 @@
+project(
+  'picohttpparser',
+  'c',
+  meson_version: '>= 0.58.0',
+  version: '1.2',
+  license: 'MIT',
```
bgilbert: The README says it's dual-licensed under MIT or "the Perl license", presumably the combination described [here](https://dev.perl.org/licenses/).

subprojects/packagefiles/picohttpparser/meson.build
```diff
@@ -0,0 +1,21 @@
+project(
+  'picohttpparser',
+  'c',
+  meson_version: '>= 0.58.0',
```
bgilbert: Why is 0.58 needed?

bgilbert (commented):
This should also build the test program (with `build_by_default: false`) and invoke it as a test.

subprojects/packagefiles/picohttpparser/meson.build
nda-cunh: I think that’s what my `meson init` gave me
Do you have a version you prefer?

bgilbert: WrapDB supports >= 0.56.0 unless there's a reason to be more restrictive.

nda-cunh: Okay :)

subprojects/packagefiles/picohttpparser/meson.build
nda-cunh: done

nda-cunh:
> This should also build the test program (with `build_by_default: false`) and invoke it as a test.

I looked into this and unfortunately test.c isn't buildable from the release tarball.

test.c starts with #include "picotest/picotest.h", and picotest/ is a git submodule pointing at h2o/picotest (see .gitmodules). GitHub's generated archive for tag v1.2 contains picotest/ as an empty directory, so the test's only test-framework dependency is missing

bgilbert (approved):

PR #2925, opened 2026-09-07
glew: update to v2.3.1

Kanashimo:


subprojects/glew.wrap
```diff
@@ -1,9 +1,9 @@
 [wrap-file]
-directory = glew-2.2.0
-source_url = https://downloads.sourceforge.net/glew/glew-2.2.0.tgz
-source_filename = glew-2.2.0.tgz
-source_hash = d4fc82893cfb00109578d0a1a2337fb8ca335b3ceccf97b97e5cc7f08e4353e1
+directory = glew-2.3.1
+source_url = https://github.com/nigels-com/glew/releases/download/glew-2.3.1/glew-2.3.1.tgz
```
bgilbert, line 3: For the record, this is indeed the repo linked from the SourceForge page.  :+1:

subprojects/packagefiles/glew/meson.build
```diff
@@ -121,3 +121,5 @@ glew_dep = declare_dependency(
   include_directories: glew_include,
   dependencies: [dep_gl, dep_egl, dep_glu],
```
bgilbert: This line is redundant because of the identical line in `library()`, so it could be removed while you're here.

bgilbert (commented):
Thanks!

bgilbert:
The arm64 failure can be fixed by implementing https://github.com/nigels-com/glew/pull/384 here.

Kanashimo:
Thank you for your fast review, I'll fix both in an hour or two

subprojects/packagefiles/glew/meson.build
```diff
@@ -76,7 +76,10 @@ if host_system == 'windows'
   glew_sources += windows.compile_resources('build/glew.rc')
   name = 'glew32'
 
-  if cc.get_argument_syntax() == 'msvc'
+  if cc.get_argument_syntax() == 'msvc' and host_machine.cpu_family() != 'aarch64'
+    c_args += ['-DVC_EXTRALEAN', '-GS-']
+    ld_args += ['-nodefaultlib', '-noentry']
+  elif cc.get_argument_syntax() == 'msvc'
     c_args += ['-DVC_EXTRALEAN', '-GS-']
```
Kanashimo, line 83: This is a little bit redundant, but I think it's more readable when done like this

Kanashimo:
Done 🫡

subprojects/packagefiles/glew/meson.build
```diff
@@ -76,7 +76,10 @@ if host_system == 'windows'
   glew_sources += windows.compile_resources('build/glew.rc')
   name = 'glew32'
 
-  if cc.get_argument_syntax() == 'msvc'
+  if cc.get_argument_syntax() == 'msvc' and host_machine.cpu_family() == 'aarch64'
```
bgilbert: I'd prefer we mirror upstream's logic by only adding the base address on `x86`.

Kanashimo: What about `x86_64`?

bgilbert (approved):
Thanks!

PR #2934, opened 2026-09-18
bearssl: add 0.6-1

### LINKED ISSUE #2935: add lib bearssl

nda-cunh:
close #2935

This MergeRequest add : 'https://bearssl.org/'


nda-cunh:
Can i disable clang-cl, and for .def there is a better method ?

